PR #271, opened 2013-02-11
[BIG] Consolidated ddoc files

andralex:
Overhauled ddoc files, factoring together macros that are specific to our website apart from ddoc macros that generate generic html files. Now ddoc/html.ddoc may be used to generate HTML documents with rendering 100% controlled from stylesheets. The file css/html-skeleton.sty contains all of the appropriate entries, while otherwise being neutral.

LaTeX ddoc files are also being pushed with this. As of now the LaTeX generic macros are not yet separated from the D-specific macros.

This overhaul allowed for a simple improvement on the website's "jump to" navigation system. This is a temporary step before generating one page per entity.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -28,11 +28,6 @@ body
 	hyphens: manual;
 }
 
-.jumpto
-{
-	text-decoration: none;	/* Eliminate underlines from jumpto anchors */
-}
-
 .param
 {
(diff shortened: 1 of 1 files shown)


andralex:
The essentials of this have been committed - yay!

PR #320, opened 2013-05-14
Moved DLangSpec files into their own subdirectory.

bordenr:
This is to help with my future efforts in allowing compilation to EPUB. I have also updated the win32.mak and posix.mak files to accommodate these changes. I've done some simple testing with the posix.mak file but almost none on the win32.mak file. I can't think how I've broken anything (unless, of course, win32 make is substantially different than *nix make).


andralex:
looks like this won't happen - sorry.

quickfur:
Why not? Isn't it just a matter of updating the makefiles so that they install the generated html files in the right place? The html directory structure doesn't necessarily have to correspond with the source file directory structure.

PR #690, opened 2014-11-04
Add a explanation for function templates that is missing

mingodad:
Thanks for the expalanation given by H. S. Teoh at http://forum.dlang.org/post/mailman.1476.1415061701.9932.digitalmars-d-learn@puremagic.com

### DIFF
--- a/function.dd
+++ b/function.dd
@@ -954,7 +954,7 @@ $(H4 $(LNAME2 parameters, Function Parameters))
 
         $(P Parameter storage classes are $(D in), $(D out),
         $(D ref), $(D lazy), $(D const), $(D immutable), $(D shared),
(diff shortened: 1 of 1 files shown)


andralex:
@quickfur could you please take this over?

quickfur:
Whoa, one thing at a time! I'm still wading through std.algorithm's docs...

Speaking of which, (when) are we gonna split up std.algorithm into more manageable chunks? I already have to manually delete half the file in order to run its unittests when locally testing PR's, because otherwise it soaks up all RAM and dies. Or maybe I should start doing that as part of the ddoc improvements...

ntrel:
I've now implemented the suggestions here in #751.

PR #725, opened 2014-12-24
Issue 3652 - Allow explicit and implicit casting of dynamic array slices of known size to static array

9rnsr:
https://issues.dlang.org/show_bug.cgi?id=3652

Even if a poor D compiler does not support any const-folding, `arr[1 .. 3]` should be convertible to `T[2]`. This PR will add the minimum requirement for the feature.

### DIFF
--- a/expression.dd
+++ b/expression.dd
@@ -1056,6 +1056,46 @@ $(GNAME SliceExpression):
 
     $(P A $(I SliceExpression) is not a modifiable lvalue.)
 
+    $(P If the slice bounds can be known at compile time, the slice expression
+        is implicitly convertible to an lvalue of static array. For example:
+
+        -------------
+        arr[a .. b]     // typed T[]
+        -------------
+
+        If both $(CODE a) and $(CODE b) are integers (may be constant-folded),
+        the slice expression can be convered to a static array type
+        $(D T[b - a]).
+
+        -------------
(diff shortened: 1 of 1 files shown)


9rnsr:
@AndrejMitrovic Thanks, fixed three wordings.

expression.dd
```diff
@@ -1207,6 +1247,31 @@ $(GNAME StringLiterals):
         $(TROW $(D immutable(dchar)[]))
     )
 
+    $(P By default, a string literal is typed as a dynamic array, but the element
+        count is known at compile time. So all string literals can be
+        implicitly connvered to static array types.
+
```
AndrejMitrovic: Ah small typo, converted should be with one "n" :)

expression.dd
```diff
@@ -1235,6 +1300,39 @@ $(GNAME ArrayLiteral):
         ---
     )
 
+    $(P By default, an array literal is typed as a dynamic array, but the element
+        count is known at compile time. So all array literals can be
+        implicitly connvered to static array types.
+
```
AndrejMitrovic: ditto

expression.dd
9rnsr: Sorry... fixed.

AndrejMitrovic:
Thanks.

PR #726, opened 2014-12-24
Use hierarchical Jump-to links

ntrel:
https://issues.dlang.org/show_bug.cgi?id=5136

Show only top-level symbols in the main module jump-to links, with remaining symbols shown in separate jump-to lists before their respective parents.

Show non-top-level jump-to links even when `inhibitQuickIndex` is set.
Show all jump-to links with same background color as symbol declarations.

Note: std.typecons, std.stdio and std.container.rbtree are nice modules to test this with.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -735,6 +735,11 @@ TABLE.book CAPTION
     text-align: left;
 }
 
+.quickindex {
+	font-weight: normal;
+	background-color: #E4E9EF;
+}
+
 .question
 {
     display: inline;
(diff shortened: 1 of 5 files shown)


ntrel:
Rebased on top of #740. I still need to show anchor overload links, will do tomorrow.

> I'd love to have a small arrow next to each struct/class

Should be doable, I'll investigate. Alternatively we could just show the first line of nested jump to links until expanded.

andralex:
This good to go?

MartinNowak:
Needs a rebase since #740 was merged.

andralex:
rebase pliz

ntrel:
@MartinNowak I've made the `getElementsByClassName('quickindex')` change, and also used `''` as the parent key name.

@andralex Should be good now, sorry for the delay.

andralex:
... and it's up!

ntrel:
Great!

PR #733, opened 2014-12-29
Factor all dlang.org-specific stuff into dlang.org.ddoc

andralex:
This eliminates a long-standing redundancy across doc.ddoc and std.ddoc.


ntrel:
> I think we need to also update win32.mak

Done in #741. Phobos is now also updated for this and #732.

PR #734, opened 2014-12-30
Much nicer jump-to anchors

andralex:
Example: http://erdani.com/d/phobos-prerelease/std_range_package.html


quickfur:
This PR has broken Phobos docs builds. If you clone a clean copy of dlang.org and phobos and build dlang.org then `cd phobos; make -f posix.mak html`, it will produce documentation pages with no content in `phobos-prerelease`. I'd revert this PR, but the github reverter seems to have trouble with this, as it appears to be conflicting with subsequent changes in the repo.

kuettler:
It seems DIVID is missing in std.ddoc.

ntrel:
@andralex Now done in #740.

@quickfur Hopefully fixed with https://github.com/D-Programming-Language/phobos/pull/2835, please test.

PR #737, opened 2015-01-01
Generation of verbatim and plaintext output from ddoc

andralex:


AndrejMitrovic:
If we merge your *undefined macro substitution* pull then we could get rid of most of these `= $0` macros. :)

andralex:
@AndrejMitrovic: Those are needed because they're predefined.

AndrejMitrovic:
Oh good point!

PR #740, opened 2015-01-02
Remove nestingLevel JS code

ntrel:
We can use `DDOC_ANCHOR`'s argument instead, which is fully qualified.


std.ddoc
```diff
@@ -37,13 +44,14 @@ function listanchors()
     values.sort();
 
     var newText = "";
-    for(var i = 0; i < values.length; i++) {
+    for (var i in values) {
```
MartinNowak: Rather keep the `for(var i = 0; i < values.length; ++i)` loop.

std.ddoc
```diff
@@ -15,16 +15,23 @@ DDOC = <!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN"
...
 //
 function defineSymbol(sym) {
-  if (nestingLevel > 0) return;
+  // ignore ditto overloads
   if (sym == lastSymbolDefined) return;
-  if (true) lastSymbolDefined = sym;
+  // ignore scoped symbols, but show module-level overloads
+  var pos = sym.indexOf('.');
+  if (pos >= 0 && (sym.lastIndexOf('.') != pos || isNaN(lastName(sym)))) return;
+  // Note: empty comment avoids macro def
```
MartinNowak: What does that mean?

std.ddoc
```diff
@@ -15,16 +15,23 @@ DDOC = <!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN"
...
+}
 //
 function defineSymbol(sym) {
-  if (nestingLevel > 0) return;
+  // ignore ditto overloads
   if (sym == lastSymbolDefined) return;
-  if (true) lastSymbolDefined = sym;
+  // ignore scoped symbols, but show module-level overloads
+  var pos = sym.indexOf('.');
+  if (pos >= 0 && (sym.lastIndexOf('.') != pos || isNaN(lastName(sym)))) return;
```
MartinNowak: How about a comment with an example of what you're filtering out here?

std.ddoc
ntrel: Why?

std.ddoc
ntrel: The next line is `/**/ lastSymbolDefined = sym;`. Without the `/**/`, ddoc would end the current macro definition and start a new macro called `lastSymbolDefined`. Andrei used `if (true)` instead, but I thought an empty comment was slightly better. IMO ddoc should not allow macro definitions when there is indentation.

std.ddoc
MartinNowak: [javascript - Why is using "for...in" with array iteration such a bad idea? - Stack Overflow](http://stackoverflow.com/questions/500504/why-is-using-for-in-with-array-iteration-such-a-bad-idea)

std.ddoc
MartinNowak: I think we should rather put it into a listanchors.js file.

std.ddoc
ntrel: Hmm, so it's slower, and iterates in order of element creation rather than index order.

std.ddoc
ntrel: Nice idea. I've now done that for my other branch #726, which includes this pull. I'll update this pull soon (ran out of time again today).

MartinNowak:
Let's go on with this.

PR #741, opened 2015-01-02
Use needed dlang.org.ddoc for HTML generation on Windows

ntrel:
Note: `posix.mak` already uses it - see https://github.com/D-Programming-Language/dlang.org/pull/733/files#diff-8.


AndrejMitrovic:
Thanks.

PR #745, opened 2015-01-04
Document byKeyValue in changelog.

quickfur:


AndrejMitrovic:
I haven't yet checked, does the generated changelog look ok? (Personally I always mess up closing the parens in the wrong place before I get it right (: )

quickfur:
That's where vim's jump-to-matching-parenthesis feature comes in very handy. :-)

You can see a preview here: http://eusebeia.dyndns.org/~hsteoh/tmp/web/changelog

PR #746, opened 2015-01-05
fix order of pattern rules

MartinNowak:
- pre-3.8.2 make uses first matching pattern rule
- since 3.8.2. best match is used


andralex:
Thanks! Let's see how this works.

PR #750, opened 2015-01-09
Issue 6574 - Erroneous recursive call in template instantiation

9rnsr:
https://issues.dlang.org/show_bug.cgi?id=6574

Improve mangling scheme to distinguish instantiations with specialized template parameters.

If a template argument matches to specialized template parameter, the argument is mangled with prefix 'H'.


9rnsr:
Corresponding compiler fix: https://github.com/D-Programming-Language/dmd/pull/4270

andralex:
Is there a way in github to specify that one pull is dependent on another?

PR #758, opened 2015-01-11
Add -conf= to dmd invocations

andralex:
This does help ddoc generation, but not the ddox stuff. How to I pass a dmd flag to dub? Thanks! cc @MartinNowak

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -160,13 +160,13 @@ ALL_FILES = $(ALL_FILES_BUT_SITEMAP) $(DOC_OUTPUT_DIR)/sitemap.html
 # Pattern rulez
 
 $(DOC_OUTPUT_DIR)/%.html : %.dd $(DDOC) $(DMD)
-	$(DMD) -c -o- -Df$@ $(DDOC) $<
+	$(DMD) -conf= -c -o- -Df$@ $(DDOC) $<
 
 $(DOC_OUTPUT_DIR)/%.verbatim : %.dd verbatim.ddoc $(DMD)
 	$(DMD) -c -o- -Df$@ verbatim.ddoc $<
 
 $(DOC_OUTPUT_DIR)/%.php : %.php.dd $(DDOC) $(DMD)
-	$(DMD) -c -o- -Df$@ $(DDOC) $<
+	$(DMD) -conf= -c -o- -Df$@ $(DDOC) $<
 
 $(DOC_OUTPUT_DIR)/css/% : css/%
 	@mkdir -p $(dir $@)
@@ -184,7 +184,7 @@ $(DOC_OUTPUT_DIR)/% : %
 	cp $< $@
 
 $(DOC_OUTPUT_DIR)/dmd-%.html : %.ddoc dcompiler.dd $(DDOC) $(DMD)
-	$(DMD) -c -o- -Df$@ $(DDOC) dcompiler.dd $<
+	$(DMD) -conf= -c -o- -Df$@ $(DDOC) dcompiler.dd $<
 
 $(DOC_OUTPUT_DIR)/dmd-%.verbatim : %.ddoc dcompiler.dd verbatim.ddoc $(DMD)
 	$(DMD) -c -o- -Df$@ verbatim.ddoc dcompiler.dd $<
@@ -212,7 +212,7 @@ $(DOC_OUTPUT_DIR)/sitemap.html : $(ALL_FILES_BUT_SITEMAP) $(DMD)
 	  && echo \
 	    "$F	`sed -n 's/<title>\(.*\) - D Programming Language.*<\/title>/\1/'p $(DOC_OUTPUT_DIR)/$F`")) \
 	  | sort --ignore-case --key=2 | sed 's/^\([^	]*\)	\(.*\)/<a href="\1">\2<\/a><p>/' >> sitemap.dd
-	$(DMD) -c -o- -Df$@ $(DDOC) sitemap.dd
+	$(DMD) -conf= -c -o- -Df$@ $(DDOC) sitemap.dd
 	rm sitemap.dd
 
 ${LATEST}.ddoc :
@@ -245,7 +245,7 @@ dlangspec.d : $(addsuffix .dd,$(SPEC_ROOT))
 	$(RDMD) ../tools/catdoc.d -o$@ $^
 
 dlangspec.html : $(DDOC) ebook.ddoc dlangspec.d $(DMD)
-	$(DMD) $(DDOC) ebook.ddoc dlangspec.d
+	$(DMD) -conf= $(DDOC) ebook.ddoc dlangspec.d
 
 dlangspec.zip : dlangspec.html ebook.css
 	rm -f $@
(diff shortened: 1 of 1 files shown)


andralex:
rebased, ping

andralex:
ping @WalterBright @MartinNowak

MartinNowak:
> How to I pass a dmd flag to dub

Using DFLAGS env, but you can't disable the compiler conf for dpl-docs because it depends on druntime/phobos. Also I hardly see how this helps for the other targets, because you still need a functional config to generate phobos docs.

andralex:
@MartinNowak: Well the long and short of it is I do have a functioning config but dub can't build with it. Advice?

MartinNowak:
The stable compiler comes with it's own working dmd.conf. Apparently you have a dmd.conf in your HOME or dlang.org dir that's used instead and turns the stable compiler nonfunctional.
I tried to convince you guys how useless that behavior is.
https://github.com/D-Programming-Language/dmd/pull/4256#issuecomment-68908317

We can't set an explicit config for the stable compiler.
My current advice would be to move your dmd.conf next to the compiler you're using, so it doesn't interfere with the stable compiler.

andralex:
I don't think that's useless, and at any rate it's not productive to count on user education. Robust is what it is. I think we must fix this. Setting DFLAGS in the environment won't help? If not, this is a bug in dub.

andralex:
@MartinNowak Sadly now it seems after all your work with -conf= we're still not where we need to be!

More context on ~/dmd.conf: it points to my dev dmd most of the time, and I don't want it to appear in my git status output. Occasionally I run git clean -dfx etc. We can't rely on the user NOT having a dmd.conf in their $HOME.

MartinNowak:
> More context on ~/dmd.conf: it points to my dev dmd most of the time, and I don't want it to appear in my git status output. Occasionally I run git clean -dfx etc. We can't rely on the user NOT having a dmd.conf in their $HOME.

It's in [.gitignore](https://github.com/D-Programming-Language/dmd/blob/74430a4869b97052c8dbfd86ec74e2edc45eb7a0/.gitignore#L6), so if you run `git clean -df` it will survive and it won't show up in git status either.

> More context on ~/dmd.conf: it points to my dev dmd most of the time, and I don't want it to appear in my git status output. Occasionally I run git clean -dfx etc. We can't rely on the user NOT having a dmd.conf in their $HOME.

Sure, and one solution to this is to change the configuration precedence, because the current one is responsible for our issues.
https://github.com/D-Programming-Language/dmd/pull/4256#issuecomment-68908317

What we can do for now, is to explicitly select a dmd.conf when using the stable compiler.
And that's already non-trivial.

andralex:
OK we seem to have reached a stalemate here.

Please allow me to explain the matter again. My ~/dmd.conf is as follows:

    DFLAGS=-I/Users/aalexandre/code/d/druntime/import -I/Users/aalexandre/code/d/phobos -L-L/Users/aalexandre/code/d/phobos/generated/osx/release/64

It is working for my usual development setup. I can build and run programs perfectly find using that setup. It doesn't seem comely to ask me and others to change a working setup to make up for an insufficiency in our tooling.

With that working setup, if I run make in dlang.org I get:

```
../dub-0.9.22/bin/dub build --root=dpl-docs --compiler=/tmp/.stable_dmd-2.066.1/dmd2/osx/bin/dmd
Error executing command build: Failed to invoke the compiler /tmp/.stable_dmd-2.066.1/dmd2/osx/bin/dmd to determine the build platform: {
  "compiler": "dmd",
  "frontendVersion": 2066,
  "compilerVendor": "Digital Mars D",
  "platform": [
/tmp/dub_platform_probe.d(43): Error: template std.array.join cannot deduce function from argument types !()(string[], string), candidates are:
/Users/aalexandre/code/d/phobos/std/array.d(1559):        std.array.join(RoR, R)(RoR ror, R sep) if (isInputRange!RoR && isInputRange!(Unqual!(ElementType!RoR)) && isInputRange!R && is(Unqual!(ElementType!(ElementType!RoR)) == Unqual!(ElementType!R)))
/Users/aalexandre/code/d/phobos/std/array.d(1624):        std.array.join(RoR, E)(RoR ror, E sep) if (isInputRange!RoR && isInputRange!(Unqual!(ElementType!RoR)) && is(E : ElementType!(ElementType!RoR)))
/Users/aalexandre/code/d/phobos/std/array.d(1683):        std.array.join(RoR)(RoR ror) if (isInputRange!RoR && isInputRange!(Unqual!(ElementType!RoR)))
/tmp/dub_platform_probe.d(11):        called from here: determinePlatform()
/tmp/dub_platform_probe.d(11):        while evaluating pragma(msg, "    " ~ determinePlatform())
  ],
  "architecture": [
/tmp/dub_platform_probe.d(87): Error: template std.array.join cannot deduce function from argument types !()(string[], string), candidates are:
/Users/aalexandre/code/d/phobos/std/array.d(1559):        std.array.join(RoR, R)(RoR ror, R sep) if (isInputRange!RoR && isInputRange!(Unqual!(ElementType!RoR)) && isInputRange!R && is(Unqual!(ElementType!(ElementType!RoR)) == Unqual!(ElementType!R)))
/Users/aalexandre/code/d/phobos/std/array.d(1624):        std.array.join(RoR, E)(RoR ror, E sep) if (isInputRange!RoR && isInputRange!(Unqual!(ElementType!RoR)) && is(E : ElementType!(ElementType!RoR)))
/Users/aalexandre/code/d/phobos/std/array.d(1683):        std.array.join(RoR)(RoR ror) if (isInputRange!RoR && isInputRange!(Unqual!(ElementType!RoR)))
/tmp/dub_platform_probe.d(14):        called from here: determineArchitecture()
/tmp/dub_platform_probe.d(14):        while evaluating pragma(msg, "    " ~ determineArchitecture())
   ],
}
/Users/aalexandre/code/d/phobos/std/range/primitives.d(325): Error: static assert  "Cannot put a string into a Appender!string."
/Users/aalexandre/code/d/phobos/std/format.d(3310):        instantiated from here: put!(Appender!string, string)
/Users/aalexandre/code/d/phobos/std/conv.d(113):        instantiated from here: formatValue!(Appender!string, void*, char)
/Users/aalexandre/code/d/phobos/std/conv.d(912):        instantiated from here: toStr!(string, void*)
/Users/aalexandre/code/d/phobos/std/conv.d(287):        ... (4 instantiations, -v to show) ...
/Users/aalexandre/code/d/phobos/std/uni.d(2329):        instantiated from here: FormatSpec!char
/Users/aalexandre/code/d/phobos/std/uni.d(1853):        instantiated from here: InversionList!(GcPolicy)


make: *** [dpl-docs] Error 2
```

We seem to be unable to make a tool run that doesn't depend on a [Liberum Veto](https://en.wikipedia.org/wiki/Liberum_veto) among everything on the user's computer. You'd have a hard time convincing me this is not an issue with our tooling. 

Please advise.

andralex:
ping @MartinNowak 

Would be great if at least (a) this were acknowledged as a tooling issue; (b) this request was pulled as a step forward toward fixing the tooling issue.

Again: the simple-stated case here is I have a setup on my machine that works perfectly well, yet dub insists that it interferes with it.

MartinNowak:
> Would be great if at least (a) this were acknowledged as a tooling issue; (b) this request was pulled as a step forward toward fixing the tooling issue.

Sure, it's a problem that dmd couldn't be configured, we partly solved that, but that doesn't fix 2.066.1.
Only this pull doesn't fix your issue.

> Again: the simple-stated case here is I have a setup on my machine that works perfectly well, yet dub insists that it interferes with it.

Once again, this is a dmd _not_ a dub problem. The conf in your home dir takes preference over anything else.
Even with the recently added `-conf=` switch, we'd still have to explicitly override any dmd.conf in your environment. To do so, we'd need to add a bunch of dmd.conf files to dlang.org (every OS times every dmd version, 6 x 3 = 18), not a good plan IMO.
I tried to convince you guys how useless the current precedence is, https://github.com/D-Programming-Language/dmd/pull/4256#issuecomment-68908317.

PR #759, opened 2015-01-12
Fix dlang.org generation following backquoted escapes

andralex:
https://github.com/D-Programming-Language/dmd/pull/4228 is a bit extreme in handling backquotes, breaking the website build.

This fix gets things going for now but in the future I think we should do what github does - if a backquote is not paired on the same line, just don't consider it an escape and output it verbatim. cc @adamdruppe


adamdruppe:
OOoh, I didn't even think about the `linker quoted stuff'. I'll do a small fix PR to the doc.c with the same line thing and I'll probably write the updated ddoc stuff too.... maybe tonight, I have a very busy week ahead of me starting in 15 minutes!

adamdruppe:
On Sun, Jan 11, 2015 at 10:31:00PM -0800, Andrei Alexandrescu wrote:
> https://github.com/D-Programming-Language/dmd/pull/4228 is a bit extreme in handling backquotes, breaking the website build.

https://github.com/D-Programming-Language/dmd/pull/4290 loosens it.

PR #761, opened 2015-01-12
Use code font for module name

Geod24:
http://forum.dlang.org/thread/m8rn6c$1ntq$1@digitalmars.com

Ping @andralex

### DIFF
--- a/dpl-docs/views/layout.dt
+++ b/dpl-docs/views/layout.dt
@@ -111,8 +111,10 @@ html(lang='en-US')
       - auto hidx = std.string.indexOf(title, ' ');
       - if (title.endsWith(" - multiple declarations"))
         h1 <code><a href="#{info.linkTo(info.mod)}">#{info.mod.qualifiedName}</a>.#{title[0 .. $-24]}</code> #{title[$-24 ..$]}
-      - else if (hidx > 0 && title[0 .. hidx] != "Module" && info.mod)
+      - else if (hidx > 0 && !title.startsWith("Module ") && info.mod)
         h1 #{title[0 .. hidx]} <code><a href="#{info.linkTo(info.mod)}">#{info.mod.qualifiedName}</a>.#{title[hidx+1 .. $]}</code>
+      - else if (hidx > 0 && title.startsWith("Module ") && info.mod)
+        h1 Module <code>#{info.mod.qualifiedName}</code>
       - else
         h1= title
       block body
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,9 @@ dlangspec.d
 /web
 /2.*.ddoc
 Makefile
+
+deleteme.*
+docs-prerelease.json
+docs.json
+.dub/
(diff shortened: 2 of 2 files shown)


dpl-docs/views/layout.dt
```diff
@@ -113,6 +113,8 @@ html(lang='en-US')
         h1 <code><a href="#{info.linkTo(info.mod)}">#{info.mod.qualifiedName}</a>.#{title[0 .. $-24]}</code> #{title[$-24 ..$]}
       - else if (hidx > 0 && title[0 .. hidx] != "Module" && info.mod)
         h1 #{title[0 .. hidx]} <code><a href="#{info.linkTo(info.mod)}">#{info.mod.qualifiedName}</a>.#{title[hidx+1 .. $]}</code>
+      - else if (hidx > 0 && title[0 .. hidx] == "Module" && info.mod)
+        h1 Module <code>#{info.mod.qualifiedName}</code>
```
andralex: Thanks for looking into this! Could we expand ddoc macros here instead of hardcoding tags? I'm thinking e.g. $(D #{info.mod.qualifiedName}).

dpl-docs/views/layout.dt
```diff
@@ -113,6 +113,8 @@ html(lang='en-US')
         h1 <code><a href="#{info.linkTo(info.mod)}">#{info.mod.qualifiedName}</a>.#{title[0 .. $-24]}</code> #{title[$-24 ..$]}
       - else if (hidx > 0 && title[0 .. hidx] != "Module" && info.mod)
         h1 #{title[0 .. hidx]} <code><a href="#{info.linkTo(info.mod)}">#{info.mod.qualifiedName}</a>.#{title[hidx+1 .. $]}</code>
+      - else if (hidx > 0 && title[0 .. hidx] == "Module" && info.mod)
```
andralex: I might be a lil pedantic here: `else if (hidx > 0 && title[0 .. hidx + 1] == "Module " && info.mod)`

i.e. space should be part of the matching...better yet, can't we use `title.startsWith("Module ")`? What the limit of stuff usable in these .dt documents?

dpl-docs/views/layout.dt
Geod24: Nope, diet templates are pure HTML templates, so they would not understand DDOC's macros.
But what do you want to achieve ? Linking ?

dpl-docs/views/layout.dt
Geod24: Updated.

> What the limit of stuff usable in these .dt documents?
There shouldn't be any.

andralex:
I hope you tried this :)

dpl-docs/views/layout.dt
MartinNowak: > What the limit of stuff usable in these .dt documents?

With limits you mean D code? It's all converted and compiled as normal D code.
The template engine is very similar to jade http://jade-lang.com/reference/.

PR #762, opened 2015-01-13
Document the new backticked inline code feature in ddoc

adamdruppe:
This is current with both of my recent dmd pull requests.


Hackerpilot:
:+1:

PR #763, opened 2015-01-13
Add template for library index

kuettler:
This is not exactly pretty, rather a proof of concept. I think there is more discussion needed.

Also, in order to build it, I need to do
`rm -rf ~/.dub/packages/ddox-0.10.0/.dub; rm web/library-prerelease/sitemap.xml; make -f posix.mak apidocs-prerelease`, which does not seem appropriate either.

### DIFF
--- a/dpl-docs/views/ddox.overview.dt
+++ b/dpl-docs/views/ddox.overview.dt
@@ -1,0 +1,85 @@
+extends ddox.layout
+
+block ddox.defs
+	- import ddox.api;
+
+block ddox.title
+	- auto title = "API documentation";
+
+block ddox.members
+
+	- void moduleInfoRec(Package pack)
+		- foreach( p; pack.packages )
+			- moduleInfoRec(p);
+		- foreach( m; pack.modules )
+			- if (!m.qualifiedName.startsWith("std.c.") && !m.qualifiedName.startsWith("core.stdc.") && !m.qualifiedName.startsWith("core.sys."))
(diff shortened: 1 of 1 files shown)


andralex:
Do you have the result uploaded somewhere?

kuettler:
You can find this version here: http://kuettler.orgfree.com/library-prerelease/index.html

As said I have mixed feelings about this:

* This patch adds phobos package names to the `ddox.overview.dt` file. I guess the `*.dt` files are meant to be generic and the data (which package to show in which list) should come from D source files. To do that some kind of documentation metadata would be needed. I do not know if something of that sort exists.
* The `dpl-docs` executable on my x86_64 Linux is 41M huge. The phobos-specific templates are compiled into a library in `~/.dub/packages/ddox-0.10.0/.dub`. Maybe my setup is wrong. (I am in favor of all changes that make the required setup more obvious.)
* I do not know whether to create patches for `ddoc` or `ddox` right now. I know too little about either system, however, from what I have seen I prefer the simplicity of `ddoc`.

andralex:
I think I'll cc @MartinNowak and @s-ludwig for more review of this.

MartinNowak:
Looks pretty good.
There isn't anything wrong with your approach. You override ddox' overview template with a dpl-docs specific one.
Dub doesn't currently track dependencies on overridden templates, see https://github.com/D-Programming-Language/dub/issues/331.

PR #767, opened 2015-01-15
Add DDOC_UNDEFINED_MACRO as a comment in the generated file

andralex:
This helps us weed undefined macros by just grepping through the generated content.


andralex:
I've already built the site with this patch.

JakobOvrum:
LGTM. Pulling it reduces the hassle of helping weed out undefined macros, so we should merge ASAP.

quickfur:
LGTM

andralex:
FYI the current html files at dlang.org have been generated with this macro. Happy hunting!

PR #770, opened 2015-01-16
Css minify

andralex:
This will allow nicely documented css files without passing the overhead to the users of dlang.org.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -22,6 +22,7 @@ GIT_HOME=https://github.com/D-Programming-Language
 DPL_DOCS_PATH=dpl-docs
 DPL_DOCS=$(DPL_DOCS_PATH)/dpl-docs
 REMOTE_DIR=d-programming@digitalmars.com:data
+
 # stable dub and dmd versions used to build dpl-docs
 DUB_VER=0.9.22
 STABLE_DMD_VER=2.066.1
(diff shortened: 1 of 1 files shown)


andralex:
Well? With opt-in this seems to be reasonable. Pull?

posix.mak
```diff
@@ -43,6 +44,9 @@ REBASE = MYBRANCH=`git rev-parse --abbrev-ref HEAD` &&\
 CHANGE_SUFFIX = \
  for f in `find "$3" -iname '*.$1'`; do\
   mv $$f `dirname $$f`/`basename $$f .$1`.$2; done
+ 
+ # Set to 1 in the command line to minify css files
+ CSS_MINIFY=
```
MartinNowak: Why is it off by default?

posix.mak
```diff
@@ -161,6 +165,14 @@ $(DOC_OUTPUT_DIR)/%.verbatim : %.dd verbatim.ddoc $(DMD)
 $(DOC_OUTPUT_DIR)/%.php : %.php.dd $(DDOC) $(DMD)
 	$(DMD) -c -o- -Df$@ $(DDOC) $<
 
+$(DOC_OUTPUT_DIR)/css/% : css/%
+	@mkdir -p $(dir $@)
+ifeq (1,$(CSS_MINIFY))
+	curl -X POST -s --data-urlencode 'input@$<' http://cssminifier.com/raw >$@
```
MartinNowak: Is there some local program we could settle on, e.g. yuglify.
Make sure to include -f and -S (-fsS) as curl arguments.

posix.mak
andralex: Just for documentation. Recall that in makefiles setting a variable in the command line overrides manual setting with `=`.

posix.mak
andralex: >Is there some local program we could settle on, e.g. yuglify?

I don't know.

>Please make sure to include -f and -S (-fsS) as curl arguments.

Done.

PR #772, opened 2015-01-16
Issue 13991 - Invariant grammar does not document that parenthesis are o...

Hackerpilot:
...ptional

https://issues.dlang.org/show_bug.cgi?id=13991


Hackerpilot:
Don't merge yet, I found another change that I need to make.

andralex:
tanx!

PR #773, opened 2015-01-16
Issue 13979 - ForeachType grammar does not allow 'ref' to appear after t...

Hackerpilot:
...ype constructors

https://issues.dlang.org/show_bug.cgi?id=13979

### DIFF
--- a/grammar.dd
+++ b/grammar.dd
@@ -657,10 +657,16 @@ $(GNAME ForeachTypeList):
     $(GLINK ForeachType) , $(I ForeachTypeList)
 
 $(GNAME ForeachType):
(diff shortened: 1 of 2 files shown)


9rnsr:
This change does not allow:
```d
shared int[1] arr;
foreach (shared ref const a; arr) {}
```

So, I'd suggest to add ForeachTypeAttrs that provides a list with arbitrary order of `TypeCtors` and `ref`.

Hackerpilot:
I've updated the pull to allow arbitrary ordering of the foreach type attributes.

9rnsr:
LGTM.

PR #777, opened 2015-01-17
$(D ) for code in changelog

MartinNowak:


andralex:
Would you want to use backticks instead?

9rnsr:
Good.

MartinNowak:
> Would you want to use backticks instead?

Next time

PR #778, opened 2015-01-17
Redo comparison document, remove a bunch of hardcoded html tags

andralex:
This started as a replacement of raw HTML tags with ddoc macros, but ended up being for the most part a restyling of comparison.html from a meaningless one-column comparison to a nested itemization.

It was a ton of tedious work... but I seem to think the page might be eliminated entirely. I don't think a laundry list of features is the best way to introduce D. Thoughts?


andralex:
See new page: http://erdani.com/d/comparison.html

changelog.dd
```diff
@@ -802,7 +802,7 @@ $(BUGSTITLE DMD Compiler bugs,
 
 $(P
 $(LI $(BUGZILLA 648): DDoc: unable to document mixin statement)
-$(LI $(BUGZILLA 846): Error 42: Symbol Undefined "<mangle_of_class_template>__arrayZ")
+$(LI $(BUGZILLA 846): Error 42: Symbol Undefined "$(LESS)mangle_of_class_template$(GREATER)__arrayZ")
```
andralex: A bunch of this stuff was simply invisible!

CyberShadow:
:+1: 

Maybe move Garbage Collection somewhere below, since it's not something to be very proud of today ;)

PR #779, opened 2015-01-18
mobile tweaks - hide sidebars if small screen, use mobile nav to compensate

adamdruppe:
My comment here:
https://issues.dlang.org/show_bug.cgi?id=13997#c1

The sidebars make the content too narrow on a small screen, so this removes them. However, the sidebar serves a useful nav purpose, so I added a new file which you can see by clicking the logo on the small thing. idk if that's any good, but it seems better than nothing.

Also, I was not able to actually test these changes since I can't get the website to build on my box. Trying to do that wasted too much time and put me overbudget on this, I still have a lot to do tonight! But it isn't a complicated change, so I'm reasonably confident it will do what I want it to do.

whether that sucks or not is another question though.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -959,3 +959,44 @@ input.runButton:active, input.resetButton:active, input.argsButton:active, input
 { position: relative; top: 1px; }
 input.resetButton{display: none}
 /* Runnable-examples css -end */
+
+/* Mobile related tweaks */
+
+#twitter ~ #content {
+	/* Makes room for the twitter thing only if it is actually there */
+	margin-right: 20em;
+}
+
+/* Only show twitter sidebar on an especially wide screen */
+@media screen and (max-width: 960px) {
+	#twitter {
+		display: none;
+	}
+	#twitter ~ #content {
+		margin-right: 0.5em;
+	}
+}
+
+#header-tinyscreen {
+	display: none;
+}
+
+/* Also hide the navigation sidebar on a smaller screen */
+@media screen and (max-width: 640px) {
+	#header {
+		display: none;
+	}
+
+	#header-tinyscreen {
+		display: block;
+	}
+		
+	div#navigation {
+		display: none;
+	}
+	div#content {
+		margin-left: 0.5em;
+	}
+}
+
+/* End mobile tweaks */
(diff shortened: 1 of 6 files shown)


andralex:
Do you have a mirror somewhere?

adamdruppe:
You can see the one page I hacked up here:

http://arsdnet.net/dlang.org/

resize your browser to see the differences. The mobilenav.html though
is completely blind, I couldn't even generate it.


> Do you have a mirror somewhere?

Not really, I couldn't get the site to build locally (dmd -D produced
blank pages and make -f posix.mak spewed about 200 errors after
surprising me by modifying my dmd pull request tree - extremely
rude btw, who expects a call to make to start messing with files
in a ../ directory?!)

I ran out of time messing with that nonsense, so I just made the edits
based on a quick look at the live site and one private copy of it.

The mobilenav.dd page is done on blind faith entirely.

jacob-carlborg:
I does not work on a mobile device. I test both on an iPhone and in the iOS simulator.

jacob-carlborg:
Is it possible to make the transition smooth? Just before the twitter side bar is removed the content is very narrow, then when the side bar is removed the width of the content expands.

jacob-carlborg:
Oh, BTW, Bootstrap has already solved these problems, why don't use that instead.

adamdruppe:
> I does not work on a mobile device. I test both on an iPhone and in the iOS simulator.

I might have just picked a wrong number for the size, I guess the screen is wider than I thought. (I don't actually have a smartphone so doing that blind too.)

Refresh and see how it looks now:
http://arsdnet.net/dlang.org/

adamdruppe:
> Is it possible to make the transition smooth? 

Sort of, but not really - Bootstrap has size jumps too since there's a cut-off point where the media queries take effect (it uses this same style of implementation, just with a LOT more code).

jacob-carlborg:
> Refresh and see how it looks now:

No difference. Don't you need to add something like this to make it work?

```html
<meta name="viewport" content="width=device-width, initial-scale=1">
```

jacob-carlborg:
> Sort of, but not really - Bootstrap has size jumps too since there's a cut-off point where the media queries take effect (it uses this same style of implementation, just with a LOT more code).

Take a look at [Bootstrap site](http://getbootstrap.com/getting-started/). When the right side menu disappears it looks like they're adding more margin to the main content to compensate.

adamdruppe:
Closing this since other people have stepped up with better implementations!

PR #781, opened 2015-01-18
Issue 13941 - NewExpression grammar is insufficient

9rnsr:
https://issues.dlang.org/show_bug.cgi?id=13941

- Fix `Type` rule by using `BasicType2`
  It should be consistent with parseType() in dmd code.

- `BasicType2` should handle repetition
  It should be consistent with parseBasicType2() in dmd code.


Hackerpilot:
:white_check_mark:

9rnsr:
Thanks.

PR #782, opened 2015-01-18
fix issue 13325 - Supplemental change for the pull 774

9rnsr:
https://issues.dlang.org/show_bug.cgi?id=13325

1. Rename `Vector` rule to `TypeVector`
2. Fix `BasicType` rule in declaration.dd to be consistent with grammar.dd


9rnsr:
@Hackerpilot How about this?

Hackerpilot:
This looks reasonable.

9rnsr:
Thanks.

PR #785, opened 2015-01-19
Kill TOCENTRIES macro and use ITEMIZE instead

Geod24:
These two macros appear to do the (almost) same job.

The only difference is that `TOCENTRIES` expansion is much more dangerous (some parser might break on it).


Hackerpilot:
I don't see any obvious problems with this.

andralex:
thanks, this kind of cleanup is needed

doc.ddoc
```diff
@@ -69,14 +69,14 @@ _=
...
-    $(UL $(TOCENTRIES
-        comparison,   D Feature List,
-        builtin,      Rationale for Builtins,
-        ctod,         Converting C to D,
-        cpptod,       Converting C++ to D,
-        pretod,       The C Preprocessor vs D,
-        cppcontracts, D Contract Programming vs C++
-    ))
+    $(ITEMIZE
+        $(A comparison,   D Feature List),
```
andralex: There's an unrelated issue here - this should use e.g. comparison.html so we don't futz with two possible names for pages. I'll fix that.

PR #787, opened 2015-01-20
Removed CONTRIBUTING.md because information is now in README.md

JinShil:


andralex:
Thanks! I was to do it but slipped my mind.

PR #791, opened 2015-01-20
Made menus narrower to save screen real estate

JinShil:
narrow_menu

### DIFF
--- a/css/cssmenu.css.dd
+++ b/css/cssmenu.css.dd
@@ -96,6 +96,8 @@ Ddoc
   -webkit-border-radius: 4px;
   -moz-border-radius: 4px;
   border-radius: 4px;
+  padding-top: 4px;
+  padding-bottom: 4px;
 }
 #cssmenu > ul > li > a:hover {
   text-decoration: none;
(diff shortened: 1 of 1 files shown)


andralex:
did you try with the stdlib menus? Some names there get pretty long

CyberShadow:
Not sure about this, narrower = smaller click target, more difficult to navigate. We have literally nothing else in that vertical space, so all space under the menu is wasted. How about a dynamic CSS rule based on the display height?

JinShil:
> did you try with the stdlib menus? Some names there get pretty long

Yes, it looks fine for me.

andralex:
@CyberShadow: I think I was wrong, that doesn't reduce the click target, only the padding.

JinShil:
> Not sure about this, narrower = smaller click target, more difficult to navigate. We have literally nothing else in that vertical space, so all space under the menu is wasted. How about a dynamic CSS rule based on the display height?

The library reference requires too much scrolling, so I though the narrow menus would be a good idea.

CyberShadow:
> The library reference requires too much scrolling, so I though the narrow menus would be a good idea.

That's true. But why not just apply this to the library reference?

JinShil:
> That's true. But why not just apply this to the library reference?

I think that would look weird because the menus would be inconsistent with the rest of the site.

CyberShadow:
There is already some inconsistency, the module list uses a monotype font.

andralex:
Oh oh... s/narrower/shorter/ and I understood everything

PR #793, opened 2015-01-20
Word wrap long menu names

JinShil:
word_wrap


MartinNowak:
Yikes, how about a variable sized menu instead?

CyberShadow:
Can we insert a zero-width space after each `.` so it doesn't break in the middle of each word?

JinShil:
> Can we insert a zero-width space after each . so it doesn't break in the middle of each word?

That's a good idea.  Please give me 24 hours to modify.

MartinNowak:
> Can we insert a zero-width space after each . so it doesn't break in the middle of each word?

How about omitting the `std` prefix?
Also `std.container` should be another submenu IMO, but that's a bit more work.

CyberShadow:
> Also std.container should be another submenu IMO, but that's a bit more work.

We should avoid nesting things too deeply, as that's more clicks and complexity.

JinShil:
Added zerowidth breaking space.

wrap2

I will submit another pull request to experiment with some of @MartinNowak 's ideas.

JinShil:
> How about omitting the std prefix?

Sorry @MartinNowak, I tried to do that, but it turned out to be a little more difficult than I anticipated due to the way the DDoc macros were written... and I'm not that invested.

andralex:
I'll pull this because it's useful regardless of later improvements.

PR #797, opened 2015-01-20
update wc.dd

Panke:
More modern version of the word count example

### DIFF
--- a/wc.dd
+++ b/wc.dd
@@ -6,123 +6,62 @@ $(D_S Example: wc,
     program.
     It serves to demonstrate how to read files, slice arrays,
     and simple symbol table management with associative arrays.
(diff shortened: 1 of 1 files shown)


wc.dd
```diff
@@ -6,123 +6,62 @@ $(D_S Example: wc,
...
-                {
-                    auto word = input[wstart .. j];
-
-                    dictionary[word]++;
-                    inword = false;
-                }
-            }
-            if (inword)
+            lCharCount += line.length;
+            foreach(char[] word; line.split)
```
andralex: splitter

andralex:
Please try splitter and then let's merge. Thanks for improving this!!

andralex:
will merge as is leaving splitter for future work

PR #798, opened 2015-01-20
Adjust padding and menu width to accomodate long names in library reference

JinShil:
This PR provides an alternative to PR #793.  It increases the menu width, and tries to make things a little more compact to accommodate long names in the library reference like `std.container.binaryheap`.  However, this and #793 are compatible and can be pulled together.

wider_menus

### DIFF
--- a/css/cssmenu.css.dd
+++ b/css/cssmenu.css.dd
@@ -44,7 +44,7 @@ Ddoc
   border-radius: 3px 3px 0 0;
 }
 #cssmenu > ul > li:first-child > a {
-  padding: 15px 10px;
+  padding: 4px;
   background: url(../images/pattern.png) top left repeat;
   border: none;
   border-top: 1px solid $(top_border);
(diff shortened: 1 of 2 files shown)


CyberShadow:
Perhaps we should only do this on library pages. In `std.ddoc` you could add a CSS class to `<body>` (`<body id="$(TITLE)" class='hyphenate std'>`), then add CSS rules such as `body.std #cssmenu ... { ... }`

andralex:
Alternative: remove the 'std.' prefix. Thoughts?

JinShil:
> Alternative: remove the 'std.' prefix. Thoughts?

Sounds good to me.  The same pattern should also be used for `etc`, and `core`.

andralex:
@JinShil could you please try that and see what it looks like?

JinShil:
> @JinShil could you please try that and see what it looks like?

@andralex :  I already did, but I couldn't get the recursice Ddoc macros to work the way I thought they should.  I'm sorry.

JinShil:
Here's what I attempted:

```
LIBITEM=$(LI $(A $1$(LIBITEM_RECURSE1 $+).html, $(D $(LIBITEM_RECURSE2 $+))))
LIBITEM_RECURSE1=_$1$(LIBITEM_RECURSE1 $+)
LIBITEM_RECURSE2=$1$(LIBITEM_RECURSE3 $+)
LIBITEM_RECURSE3=.&#x200B;$1$(LIBITEM_RECURSE3 $+)
_=
```

trailingdots

But, as you can see, it leaves a trailing dot.  If you see my error, let me know and I'll fix it and submit the pull request.  I'm sorry, but I don't understand why this is happening.

JinShil:
I cheated.  See #827

JinShil:
Closing due to #827.

PR #799, opened 2015-01-20
Add info about DDoc to readme

JinShil:


ntrel:
LGTM

CyberShadow:
We should not be relying on extension-less links, as they are an artifact of the server's current configuration.

PR #800, opened 2015-01-20
Remove collapse of already-open menu.

schveiguy:
When you expand one submenu and it collapses another, it can have particularly annoying results. For example, when in the standard library docs, and you want to expand multiple sections (std and core), you cannot. In addition, the following sequence is quite annoying:

1. expand std.
2. expand core.

So when you expand std, it has many entries, scrolling core off the screen. You must scroll down to reach the core menu. As soon as you open core, it collapses std. core has very few entries in it, so the overall size shrinks tremendously. This makes the menu suck itself up off the screen, and now you have to scroll up to get at the submenus.

This fix simply avoids collapsing the already-open menu.

### DIFF
--- a/js/cssmenu.js
+++ b/js/cssmenu.js
@@ -24,7 +24,7 @@ $('#cssmenu > ul > li > a').click(function() {
     checkElement.slideUp('normal');
   }
   if((checkElement.is('ul')) && (!checkElement.is(':visible'))) {
-    $('#cssmenu ul ul:visible').slideUp('normal');
+    /* $('#cssmenu ul ul:visible').slideUp('normal'); */
     checkElement.slideDown('normal');
   }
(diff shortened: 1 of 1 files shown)


MartinNowak:
:+1:

CyberShadow:
FWIW, this turns the accordion menu into a tree-style menu, but I agree that in our case it's a good change.

andralex:
Conspiracy! :) OK, will push the new site soon.

CyberShadow:
There is now a visual bug: when you expand a second submenu, the previously open one's "minus" sign changes to a "plus" sign, even though it's still expanded.

schveiguy:
> There is now a visual bug:

I'll fix it.

MartinNowak:
> this turns the accordion menu into a tree-style menu

Which might even be a better match for displaying packages.

PR #801, opened 2015-01-20
add links to other std.container.* modules

MartinNowak:


CyberShadow:
LGTM but we need to think about the horizontal overflow (#798 or #793).

Will we need to manually update this list every time the Phobos module set changes? Perhaps there could be a POSIX Make rule which updates a versioned file (as a workaround to win32.mak being underpowered)..

andralex:
how did I miss'em?

PR #805, opened 2015-01-20
a few JS optimizations

MartinNowak:


Hackerpilot:
Seems reasonable.

CyberShadow:
Did you test these changes thoroughly? This will certainly break any inline code which tries to access jQuery.

CyberShadow:
You can definitely see the menu flashing with a fresh cache, please fix/revert.

MartinNowak:
> You can definitely see the menu flashing with a fresh cache, please fix/revert.

The body onload code runs jquery ready, so the menu is already hidden. Will recheck.

> This will certainly break any inline code which tries to access jQuery.

Wasn't aware of that usage.

CyberShadow:
> The body onload code runs jquery ready, so the menu is already hidden. Will recheck.

The problem is the moment during which jQuery.js is downloaded/parsed/executed, the HTML and CSS is already available to the browser, which goes ahead and renders them. Since the CSS says that the menus are visible, the browser renders them as visible. Only after jQuery runs are the menus hidden.

> Wasn't aware of that usage.

I don't know if we use this right now (that's why I said testing is needed). Even if we don't, moving scripts to the bottom means we won't be able to add inline scripts that use jQuery in the future.

MartinNowak:
> Only after jQuery runs are the menus hidden.

Now fixed by using `<noscript><style>`. That in turn required to switch to HTML5 which seems a sane choice anyhow.
I still left the scripts at the bottom, couldn't find any jQuery usage in druntime/phobos/dlang.org.
And actually it's good that using more heavy-weight JS in docs would require and explicit change here.

CyberShadow:
Please don't use `<noscript>` since it may not always indicate the absence of JavaScript. Opera/Presto ignores `<noscript>` tags when JS is disabled.

CyberShadow:
OK, so now you got rid of `$(SCRIPT document.body.className += ' have-javascript';)` but instead added a hard-coded CSS rule in the HTML document. I might seem extremely nitpicky but I don't see how this is a good change :)

MartinNowak:
OK, removed the no-js stuff. What about HTML5?

css/style.css
```diff
@@ -24,6 +24,27 @@ a.btn:hover
...
 }
 
+/* replaces <big> tag */
+em.big
+{
+    font-style: normal;
+    font-size: larger;
+}
+
+/* replaces <tt> tag */
```
CyberShadow: Why not use `<code>`?

css/style.css
```diff
@@ -24,6 +24,27 @@ a.btn:hover
...
+em.tt
+{
+    font-style: normal;
+    font-family: monospace;
+}
+
+/* replaces <u> tag */
+em.u
```
CyberShadow: I don't think we need to replace `<u>`, at least according to http://www.w3schools.com/tags/tag_u.asp

css/style.css
MartinNowak: I was a bit worried about interference with other styling. And it would be odd if the TT macro would be a `<code>` tag, while the CODE macros is some `<code class="">` tag.

CyberShadow: I don't think we use the `code` tag currently. It does appear a few times in `contracts.dd` and once in `deprecate.dd` but I think that is not intentional.

CyberShadow: > And it would be odd if the TT macro would be a <code> tag

That one seems to be used in just two files, too (`iasm.dd` and `latex.ddoc`).

css/style.css
MartinNowak: It's semantic was changed and it should no longer be used to underline text.
http://www.w3.org/TR/html5-diff/#changed-elements

CyberShadow: I see. This macro is only used 6 times though, its usage should probably be replaced.

MartinNowak: Sure might do that in another pull.

CyberShadow:
LGTM, ready to merge if you are.

PR #808, opened 2015-01-20
Directly replace GLINK2 with GLINK in grammar page

9rnsr:
Currently some links in grammar page jump out from grammar page.

e.g. The `Type` under [EnumBaseType](http://dlang.org/grammar.html#EnumBaseType)

They should be internal links instead.

### DIFF
--- a/grammar.dd
+++ b/grammar.dd
@@ -1810,3 +1810,4 @@ Macros:
         PSCURLYSCOPE=$(GLINK NonEmptyOrScopeBlockStatement)
         TRAITS_LINK2=$(LINK2 traits.html#$1, $(D $1))
         GLINK_LEX=$(GLINK2 lex, $1)
+        GLINK2=$(GLINK $2)


MartinNowak:
Wouldn't it be cleaner to replace the usages of GLINK2 with GLINK?

9rnsr:
Thisis better way to maintain grammar page, because we can just copy-paste rulrs from other pages.

andralex:
I'm fine with this. FWIW we need to overhaul the various ways of defining links because of the mobi and pdf targets. I'm thinking there are three categories:

1. Link to an anchor on the same page
2. Link to another document but within the same directory (conceptual unit), e.g. dlang.org/xxx.html -> dlang.org/yyy.html;
3. Link to a document on dlang.org but in a different directory, i.e. cross from phobos to druntime or phobos to language definition;
4. Link to external site.

(A detail would be link to another document with an anchor, e.g. xxx.html#pos.)

These would be handled differently in the mobi and pdf documentations because they'd be either cross-references within the same document or clickable URL mentions to online documents.

andralex:
I'll merge this as is. Thanks @9rnsr !

PR #815, opened 2015-01-21
Integrate forum activity summary into dlang.org front page

CyberShadow:

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -963,6 +963,13 @@ input.runButton:active, input.resetButton:active, input.argsButton:active, input
 input.resetButton{display: none}
 /* Runnable-examples css -end */
 
+iframe#forum-summary {
+	width: 100%;
+	height: 254px;
+	border: 0;
+	outline: 2px solid #D0D0D0;
+}
+
 div#mobile-top, div#mobile-search {
 	display: none;
 }
--- a/index.dd
+++ b/index.dd
@@ -10,6 +10,9 @@ $(UL
 $(LI $(LINK2 http://dconf.org/2015, $(B $(RED DConf 2015))))
 $(LI $(LINK2 https://archive.org/search.php?query=title%3A%28DConf%202014%29&sort=+publicdate, $(B $(RED DConf 2014))))
 $(LI $(LINK2 http://dconf.org/2013/schedule/index.html, $(B $(RED DConf 2013)))))
+
+<iframe id="forum-summary" src="http://forum.dlang.org/frame"></iframe>
+$(BR)
 $(BR)
 
 <div style="text-align: right; font-size: 11px; margin-bottom: -35px; margin-right: 5px; position: relative; z-index: 10000">


MartinNowak:
Maybe a little more towards the bottom?

CyberShadow:
Below the example? It would be below the fold then, I think @andralex wants it to be front and center.

MartinNowak:
My first thought was, quite a lot of information. But then again, it's quite a good idea to make the forum more present.

CyberShadow:
@andralex Awaiting your feedback.

andralex:
My first impression: looks nice but I think everything should be smaller and "thinner". Essentially a la Facebook, which crams a bunch of good stuff in small spaces by using small but highly readable fonts, pictograms, and images. How about trying a few other (and smaller) fonts?

quickfur:
Agree that the fonts for the forum topics are a smidgen too large. Also, I think the code example should come first, since a newcomer would be more interested primarily in the language itself (that the code sample shows off) than the discussions surrounding the language. First you want to draw the reader to the language itself, then when some interest has registered, present the forum as a place to find out more of the latest-and-greatest on the language.

CyberShadow:
Now?

p0nce:
I think the forum activity should be in a separate panel much like the twitter feed. As of now it breaks the language presentation in parts.

CyberShadow:
Should I just stack the two columns on top of each other and put them above the Twitter feed?

> pictograms, and images

I could add a small avatar of the thread author.

michaelc37:
yea, it seems a bit awkward to this information in the center while the twitter feed is on the right. 

i think they might look better as separate columns above/below the twitter feed.

JinShil:
> Should I just stack the two columns on top of each other and put them above the Twitter feed?

I like this idea, if you make each feed collapsable like the sidenav.

CyberShadow:
Why would you ever want to collapse them, though? They'll only appear on the front page.

How about I will make them scrollable instead (like right now), and we can play with the number of visible rows if it ends up too tall.

JinShil:
> Why would you ever want to collapse them, though? They'll only appear on the front page.

So I don't have to scroll down, read the posts, then scroll back up to select something else towards the top.  

But do as you wish.

CyberShadow:
Surely scrolling is much easier than clicking a relatively tiny target???

I can understand that with navigation, having everything expanded is unwieldy, makes finding what you want harder, and can increase the page length for short pages.

JinShil:
> Surely scrolling is much easier than clicking a relatively tiny target???

If your prefer scrolling, you don't have to click to collapse.

CyberShadow:
!()[http://dump.thecybershadow.net/2df91276e145cd544b4b2fead2283421/00000142.png]

JinShil:
Looks good!

quickfur:
:+1:

andralex:
love it!!!

andralex:
@CyberShadow wait, I just pushed and it doesn't look like your latest screenshot. Was that just a mockup?

CyberShadow:
Sorry, yeah, since I have to keep things in sync with the forum I have local URLs changed to localhost, which is why the pull was out of date. And I need to update the "production" forum instance.

CyberShadow:
Real pull: https://github.com/D-Programming-Language/dlang.org/pull/823

jacob-carlborg:
@andralex @CyberShadow there's a big 404 on the main page because `http://forum.dlang.org/frame` doesn't exist.

PR #812, opened 2015-01-21
align title "D Programming Language" with content text

aG0aep6G:
little-improvements

See commit messages for what changed exactly.

PR #816, opened 2015-01-21
minor css and js fixes updates

dgallets:
This probably should have been multiple commits, I didn't plan on changing so much.
1. Styled and positioned search box.
2. Search box always shows on mobile (it now only takes one line).
3. Fixed issue where search box would disappear if you displayed it in "mobile" mode and re-sized back to "desktop" size.
4. Changed px based fonts to % so it would scale.
5. Fixed mobile menu being truncated when content is taller than screen (now it scrolls).
Note: I used font-awesome for icons because they scale unlike static images.


dgallets:
desktop
mobile

MartinNowak:
Looks pretty nice.
- [ ] delete the search image
- [ ] use $(DIVID) instead of <div id="">
- [ ] you could use http://fontello.com/ to select only needed icons
  [fa-search](http://fontello.com/d88e0ae88edd29343b6188e6cba16cef), [makefile integration](https://gist.github.com/puzrin/5537065)

CyberShadow:
Doesn't look very nice on my phone.

CyberShadow:
> 2. Search box always shows on mobile

I'm not sure this is a good idea - seems like a big waste of vertical space for people who don't use search. I don't think any major websites do this. Wikipedia does something clever and combines header and search, but I don't think this would fit dlang.org.

> (it now only takes one line).

I don't know what you mean by this.

CyberShadow:
BTW, there's a lot of whitespace noise in your diff.

CyberShadow:
> makefile integration

I thought we agreed to not do this :) 

(this = call 3rd-party web services over HTTP during our build)

dgallets:
>delete the search image

Done.

>use $(DIVID) instead of 

Instead of what, html? Is it possible to add additional attributes (like classes) when using $(DIVID)?

>Doesn't look very nice on my phone.

Please try it now, it should be fixed. I also put search toggle back.

>BTW, there's a lot of whitespace noise in your diff.

Are you talking about spaces instead of tabs? The use of 4 spaces vs tabs is inconsistent, what "standard" should we be using?

MartinNowak:
> (this = call 3rd-party web services over HTTP during our build)

You'd only need to use their service to update the font files.

CyberShadow:
> Instead of what, html?

Instead of `<div id="">` (GitHub ate Martin's HTML). For cases when you only want to specify the id.

> Are you talking about spaces instead of tabs? 

Yes, look at the diff here on GitHub. You could either use the same sort of whitespace around the code that you've modified, or convert tabs to spaces in a separate commit.

CyberShadow:
> Please try it now, it should be fixed.

Still a little crooked: http://dump.thecybershadow.net/61a9248675539683d768ece69be61af4/Screenshot_2015-01-21-19-11-05.png

BTW, why the rounded corners on the content? It doesn't really fit the "black bar" gradient skeumorphism.

MartinNowak:
> (this = call 3rd-party web services over HTTP during our build)

You'd only need to use their service to update the font files, not for building the site.
But I generally agree, do you happen to know a nice minify tool. Maybe we can talk Andrei into installing yuglify?

CyberShadow:
If we really need to, we can self-host the necessary tools somewhere on a server we already trust. We can also use SSL by including the server certificate with the repo.

dgallets:
>you could use http://fontello.com/ 

Do you think this is necessary? The files are around 300kb (not tiny, but not massive). If we use the CDN it will already be cached if any other site visited uses it.

>You could either use the same sort of whitespace around the code that you've modified

When you pointed it out I realized most of the lines near my edits were tabs so I replaced my spaces. I try to match whatever the file is using, but it's tricky when the file uses both.

> why the rounded corners on the content?

In desktop mode the other elements had rounded corners and I was trying to be consistent. Do you think they should never be rounded? or keep them round on desktop and not on mobile?

MartinNowak:
> Do you think this is necessary?

Not really necessary.

CyberShadow:
I am not a fan of rounding the content's corners on desktop either, but it is a matter of opinion.

aG0aep6G:
> 5. Fixed mobile menu being truncated when content is taller than screen (now it scrolls).

I think my PR #819 aims for the same, but the solution is a little different. You let menu and content scroll independently. I let them scroll together.

dgallets:
>I think my PR #819 aims for the same, but the solution is a little different. You let menu and content scroll independently. I let them scroll together.

Your solution is probably better. `position: fixed` does not work well in some mobile browsers.

aG0aep6G:
#819 has been merged since.

andralex:
I'm okay with this after @CyberShadow and @MartinNowak ensure all kinks get worked out.

css/style.css
```diff
@@ -405,44 +406,82 @@ img#github-ribbon
...
+div#search-dropdown select
 {
-	outline: none;
+	position: relative;
+	width: 12em;
+	z-index: 1;
+}
+
+div#search-dropdown:after{
+	content: "\f107";
```
aG0aep6G: Maybe add a comment about what that displays, like `/* angle down */`.

css/style.css
```diff
@@ -1035,22 +1090,10 @@ div#mobile-top, div#mobile-search {
...
-
-	div#search-dropdown {
-		display: inline;
-	}
-	div#search-box select {
-		width: 30%;
+		position: static;
+		text-align: center;
+		margin-top: 10px;
+		z-index: 1000;
```
aG0aep6G: `z-index` doesn't work with `position: static`.

css/style.css
```diff
@@ -1068,6 +1111,14 @@ div#mobile-top, div#mobile-search {
 		-webkit-box-shadow: 10px 0px 20px black;
 		box-shadow: 10px 0px 20px black;
 		z-index: 10001;
+		overflow: auto; /*needs to scroll if content bigger than screen*/
```
aG0aep6G: With #819 in, this isn't needed anymore.

doc.ddoc
```diff
@@ -44,16 +45,15 @@ $(SCRIPT document.body.className += ' have-javascript';)
 $(DIVID top,
     $(DIVID search-box,
         <form method="get" action="http://google.com/search">
-            <img src="http://dlang.org/images/search-left.gif" width="11" height="22" /><input id="q" name="q" placeholder="Search" /><input type="image" id="search-submit" name="submit" src="http://dlang.org/images/search-button.gif" />
             <input type="hidden" id="domains" name="domains" value="dlang.org" />
             <input type="hidden" id="sourceid" name="sourceid" value="google-search" />
-            $(DIVID search-dropdown,
+            $(DIVID search-query, <input id="q" name="q" />)$(DIVID search-dropdown,
```
aG0aep6G: Please add `placeholder="Search"`.

aG0aep6G:
Just like #812, the search box clashes with the GitHub ribbon on download.html

dgallets:
I made recommended changes including clash with github fork link (which is currently floating 0.875ems from the top in master due to header margin).

CyberShadow:
> Still a little crooked

Still not fixed.

Regression on Chrome on Android: for some reason when you tap the search box, typing the first letter closes the keyboard. Sometimes the keyboard is hidden immediately.

dgallets:
Mobile search should be fixed.

I thought the thick border was unnecessary and out of place. I added it back, but thinner, perhaps that is good compromise?

CyberShadow:
Great, just the crookedness left :)

Here's my take on it, plus fixes for very small screens:

```diff
diff --git a/css/style.css b/css/style.css
index 4fb7b48..3092b82 100644
--- a/css/style.css
+++ b/css/style.css
@@ -443,6 +443,7 @@ div#search-query
 	-webkit-border-radius: 5px 0 0 5px;
 	-moz-border-radius: 5px 0 0 5px;
 	border-radius: 5px 0 0 5px;
+	width: 40%;
 }
 
 div#search-dropdown
@@ -450,7 +451,8 @@ div#search-dropdown
 	position: relative;
 	border-right: none;
 	border-left: none;
-	width: 8em;
+	max-width: 8em;
+	width: 40%;
 	overflow: hidden;
 }
 
@@ -474,7 +476,6 @@ div#search-dropdown:after{
 
 div#search-submit
 {
-	line-height: 1.5em;
 	-moz-border-radius: 0 5px 5px 0;
 	-webkit-border-radius: 0 5px 5px 0;
 	border-radius: 0 5px 5px 0;
@@ -1054,8 +1055,7 @@ div#mobile-top, div#mobile-search {
 
 	div#mobile-top {
 		display: block;
-		padding: 5px 0;
-		padding-bottom: 10px;
+		padding: 7px 0;
 
 		background: -moz-linear-gradient(#555 0%, #333 100%);
 		background: -webkit-gradient(linear, left top, left bottom, color-stop(0%, #555), color-stop(100%, #333));
@@ -1065,18 +1065,19 @@ div#mobile-top, div#mobile-search {
 
 	div#mobile-search-button {
 		float: right;
-		padding-right: 10px;
+		padding-right: 15px;
 	}
 
 	div#mobile-hamburger {
 		float: left;
-		padding-left: 10px;
+		padding-left: 15px;
 	}
 
 	div#mobile-search-button, div#mobile-hamburger {
 		color: #aaa;
 		font-size: 1.7em;
 		cursor: pointer;
+		padding-top: 6px;
 	}
 
 	div#mobile-title {
@@ -1092,6 +1093,7 @@ div#mobile-top, div#mobile-search {
 		height: 38px;
 		max-width: 50%;
 		vertical-align: middle;
+		line-height: 1.0;
 	}
 
 	div#mobile-d-language a {
diff --git a/doc.ddoc b/doc.ddoc
index f3a18ea..92802cf 100644
--- a/doc.ddoc
+++ b/doc.ddoc
@@ -69,7 +69,7 @@ $(DIVID mobile-top,
         $(DIVID mobile-title,
             <img id='mobile-logo' width="50" height="38" border="0" alt="D Logo" src="http://dlang.org/images/dlogo.png">
             $(DIVID mobile-d-language,
-                <a href="http://dlang.org">D Programming Language</a>
+                <a href="http://dlang.org">D$(NBSP)Programming Language</a>
             )
         )
     )
diff --git a/std.ddoc b/std.ddoc
index 3c37841..187058b 100644
--- a/std.ddoc
+++ b/std.ddoc
@@ -48,7 +48,7 @@ $(DIVID mobile-top,
         $(DIVID mobile-title,
             <img id='mobile-logo' width="50" height="38" border="0" alt="D Logo" src="http://dlang.org/images/dlogo.png">
             $(DIVID mobile-d-language,
-                <a href="http://dlang.org">D Programming Language</a>
+                <a href="http://dlang.org">D$(NBSP)Programming Language</a>
             )
         )
     )
```

CyberShadow:
:+1:

Just needs a rebase now.

dgallets:
Just did rebase, never did it before, hopefully I didn't screw it up.

CyberShadow:
Aw, it ate my commit. Oh well. Generally the idea is to clean up the development history and merge commits, not necessarily squish everything into one commit.

dgallets:
Apologies, people told me to squash commits into one before rebase to get a "clean" pull request, so I assumed that's what you meant. Just a rebase would have been much simpler since I've been merging with upstream all along so I wouldn't get out of date. 

Do you recommend merging while working on a local branch or should I always just rebase?

CyberShadow:
> Do you recommend merging while working on a local branch or should I always just rebase?

Unless you need them personally, or the clutter gets in the way of reviewing, don't bother with such things until your pull request is approved. It might end up wasted effort.

CyberShadow:
Oh, forgot one thing. We need to update the makefiles when files are added/deleted, otherwise it'll complain during uploading, I think.

dgallets:
Update in #825 I'm using win32.mak which doesn't seem to reference them.

PR #817, opened 2015-01-21
align D logo with text

MartinNowak:
- drop shadow of logo isn't visible on dark background
  so shift the logo down by that amount so it's optically
  aligned with the text next to it

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -381,6 +381,8 @@ div#header
 img#logo
 {
 	vertical-align: text-bottom;
+	/* drop shadow isn't visible => align the logo */
+	margin-bottom: -3px;
 }
 
 /*div#d-language a*/


MartinNowak:
Alternative would be to use a different logo without shadow.

CyberShadow:
What does this fix? The logo is already aligned for me:

MartinNowak:
I seems to me, that the text should be a little higher, it sits so low because of the shadow.
Decide for yourself https://dlang.dawg.eu/.
align

CyberShadow:
Yeah, I applied the patch to compare, but I like the current version better. Right now the bottom of the text is aligned to the bottom of the image, with your patch I don't understand what's aligned to what and it just looks misaligned.

MartinNowak:
Seem like I see a different baseline in the logo.
Feel free to close this if you disagree.

CyberShadow:
Well, I'm not the gatekeeper of this repo, so far it's your opinion vs. mine. We should get a third opinion.

andralex:
What's with the green line?

andralex:
oh it was explanatory

andralex:
Yah I like the proposed version better... guess I'll merge.

PR #820, opened 2015-01-21
Replace plus/minus menu PNG icons with SVG icons

CyberShadow:


andralex:
@CyberShadow you forgot to update the makefiles, could you please? Thanks!

PR #822, opened 2015-01-22
Add missing navbar links to std.csv.

quickfur:
This appears to have been missing since the previous release, so update both navbars.


CyberShadow:
How many more pull requests with module list inconsistencies are we going to get?

Did you notice that std.csv was missing, or did you compare the displayed and actual module list?

quickfur:
Somebody mentioned it on the forum.

I do know that there's at least one broken link, namely
std.windows.charset. AFAIK that one has never worked, since posix.mak
never built docs for it, and even if it did, it wouldn't help 'cos it's
all `version(Windows)` anyway. I don't know how to solve this in a way
that isn't horribly ugly and unmaintainable.

PR #828, opened 2015-01-24
use user preference for font-size

aG0aep6G:
font-size currently is at 87.5% of the user preference. So it would go up a bit.

Other than myself, [H. S. Teoh has expressed a desire for this](http://forum.dlang.org/post/mailman.5189.1422081821.9932.digitalmars-d@puremagic.com).

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -231,7 +231,6 @@ pre
 	padding: 1ex;
 	margin: 1em 0 1em 0em;
 	font-family: monospace;
-	font-size: 1.2em;
 	line-height: normal;
 	border: 1px solid #ccc;
 	width: auto;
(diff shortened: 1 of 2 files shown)


CyberShadow:
> Because it's the right thing to do (TM).

Is it really, though?

It seems to me that user-configurable font sizes as a browser setting are an artifact of the times when browsers could not do full-page zoom.

aG0aep6G:
Well, I'm convinced it's the right thing to do. But I can understand if you value a neat presentation on factory defaults higher than respecting power-user settings.

CyberShadow:
> But I can understand if you value a neat presentation on factory defaults

No, that's not what I'm arguing for - I'm saying that I'm not sure if spending resources on settings used by a very small % of visitors is worth it. For example, making sure the website works without JavaScript does not just serve end users who choose to disable JavaScript, but also helps with accessibility and user agents who can't execute JavaScript (e.g. search engines). Even if the design work is a one-time effort, some changes can complicate the CSS files for little gain (increasing the maintenance costs), and still are very likely bitrot away in future changes to the website, as very few contributors are going to test at non-standard font sizes.

aG0aep6G:
If it's about maintenance cost, I don't think this adds to it. I'm mainly dropping font-size declarations. The rather complex comments are just spelling out what had been done before, too, just with a different font-size factor. And the body#Home thing is a drive-by change that could be separated, but it touches the same lines.

CyberShadow:
OK, but we are expending relatively much effort on something that can essentially be fixed on the end-user's side by Ctrl+= or Ctrl + mouse wheel up.

Looking at the new style, I'm not sure about the increase to the font size. There is now a striking difference between the size of the main content font, and that of the code examples, navigation, news widgets, etc.

A font size of 16px for the content is also unusual today. For example:

- Facebook: 12px
- Google: 13px
- Wikipedia: 14px

Thus, as odd as it sounds, I don't think that using the browser's default font size as-is is a good idea.

andralex:
can we see comparative screenshot for each of these?

CyberShadow:
Front page: [before](http://dump.thecybershadow.net/d7ae94a63370328b16985cd614d9f8e4/00000147.png), [after](http://dump.thecybershadow.net/495e7bad3448b9149115679c5e2ba574/00000144.png)
Module page: [before](http://dump.thecybershadow.net/5337f99af294078b98e30dc35e030499/00000146.png), [after](http://dump.thecybershadow.net/afdcc88b5b0b94f39de0e1e477fa2ced/00000145.png)

andralex:
>Thus, as odd as it sounds, I don't think that using the browser's default font size as-is is a good idea.

Agreed. This is not going in, sorry.

PR #830, opened 2015-01-24
fancy tables

aG0aep6G:
Mostly copied from Sebastiaan Koppe.

iconscreenshot1

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -249,21 +249,42 @@ dd, .d_decl_dd
 	margin-bottom: 1em;
 }
 
+table td
+{
+	padding: 0.5em;
+}
+table th
+{
+	padding: 0.5em;
+	background: rgb(139, 158, 169);
+	font-weight: bold;
+	color: white;
+}
+table tr:not(:last-of-type) td
+{
+	border-bottom: 1px dashed rgb(211, 211, 211);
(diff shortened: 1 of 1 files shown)


CyberShadow:
Can you reuse an existing color or its hue? I don't think we use that shade of gray anywhere now.

aG0aep6G:
Changed colors. Updated screenshot above.

andralex:
Nice. I want to keep booktable because it looks nice for wide entries. Also vertical lines are unrecommended in classical typography, see e.g. http://www.inf.ethz.ch/personal/markusp/teaching/guides/guide-tables.pdf.

So: @aG0aep6G could you please keep the booktable class BUT keep your proposed style for default tables? We'll use the default style for short tables.

Also please eliminate all dotted lines. Vertical ones are not nice anyway, and the horizontal ones are rendered unnecessary by the stripes in the background. Thanks!

aG0aep6G:
I'd rather not go with two different styles. If you like the 'book' style, I suggest making that the one table style:  #842

andralex:
@aG0aep6G Well we already have a different kind of table for "Parameters". That looks nice enough, but if you feel like improving it please revive this. Thanks!

PR #831, opened 2015-01-24
remove underlines from links

aG0aep6G:
Before/after:
iconscreenshot1

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -271,9 +271,16 @@ hr
 	margin: 2em 0;
 }
 
-a
+a, .question
 {
-	color: #006;
+	text-decoration: none;
+	color: #00D;
+	cursor:pointer;		/* Cursor is like a hand when someone rolls the mouse over the question */
+}
+
+a:hover, .question:hover
+{
+	text-decoration: underline;
 }
 
 a:visited
@@ -945,9 +952,6 @@ TABLE.book CAPTION
 {
     display: inline;
 	/* font-weight:bold;	/* Bold font */
-	color: #006;
-    text-decoration: underline;
-	cursor:pointer;		/* Cursor is like a hand when someone rolls the mouse over the question */
 }
 
 .answer {


CyberShadow:
:-1: Links should stay underlined, IMHO. This becomes even more ambiguous for links inside text that has other colors, such as symbol listings, or cross references in Ddox. This also clashes with D keywords in syntax-highlighted code.

aG0aep6G:
> This becomes even more ambiguous for links inside text that has other colors, such as symbol listings, or cross references in Ddox.

I'd say they should be blue, too.

> This also clashes with D keywords in syntax-highlighted code.

And those not blue.

If it's acceptable then, I'd try and change those accordingly.

CyberShadow:
You're welcome to try, although I can't imagine a variation of your proposal which I would consider a positive change. IMHO, non-underlined links are a gimmick which harms usability more than it improves style. But if you post your results to the forum and people will like it, it'll likely go in.

andralex:
I personally like the un-underlined jump links. Too many underlined words together - underlining is just noise. Also, it's rather obvious that they're supposed to be clicked on, otherwise why clump them together.

@aG0aep6G: Discuss on the forum. I'll fight for ya.

CyberShadow:
I don't mind as much removing underlining from indexes and other places where almost everything is a link (as we already do with navigation), but it should not be done everywhere. Doing so has accessibility consequences (e.g. it makes it difficult for colorblind users to discern links).

andralex:
Oh wait I missed that. Nonono, underlining should be disabled ONLY in the table shown in the screenshot. @aG0aep6G that can be done with a CSS class, could you please do that?

aG0aep6G:
> Doing so has accessibility consequences (e.g. it makes it difficult for colorblind users to discern links).

That's a really good point.

> I personally like the un-underlined jump links.
[...]
> could you please do that?

I'm not sure it's a good idea. Like CyberShadow says, "it would be inconsistent to do it selectively in the main content". But here it goes: #843

andralex:
nice work in #843 - closing this now

PR #834, opened 2015-01-24
px to em for #tool's position

aG0aep6G:
Should be pretty much invisible. Avoids weirdness with extraordinary user-set font-sizes.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -651,9 +651,10 @@ div#tools
 {
 	display: inline;
 	float: right;
-	margin: 0;
-	margin-top: -15px;
-	margin-right: -18px;
+	margin:
+		-1.1em /* = -(1.3em [#content's padding-top] - 0.2em) */
+		-1.4em /* = -(1.6em [#content's padding-right] - 0.2em) */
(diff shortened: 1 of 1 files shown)


css/style.css
```diff
@@ -651,9 +651,10 @@ div#tools
 {
 	display: inline;
 	float: right;
-	margin: 0;
-	margin-top: -15px;
-	margin-right: -18px;
+	margin:
+		-1.1em /* = -(1.3em [#content's padding-top] - 0.2em) */
+		-1.4em /* = -(1.6em [#content's padding-right] - 0.2em) */
+		1em 1em;
```
CyberShadow: Are comments allowed inside values? GitHub certainly doesn't like them.

aG0aep6G: Hm, <http://www.w3.org/TR/CSS21/syndata.html#tokenization> says:
> COMMENT tokens do not occur in the grammar (to keep it readable), but any number of these tokens may appear anywhere outside other tokens.

So, I think it's valid and GitHub gets it wrong.

But I could split it up, of course.

PR #835, opened 2015-01-24
remove margin top from the home page heading

aG0aep6G:
It slipped through CSS, because it's not an &lt;h1&gt;.

Before/after:
iconscreenshot1

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -162,7 +162,6 @@ h1, h2, h3, h4, h5, h6
 
 h1
 {
-	margin-top: 0;
 	font-size: 2.0em;
 }
 
@@ -538,6 +537,11 @@ div#content
 	-moz-border-radius: 5px;
 	-webkit-border-radius: 5px;
 	border-radius: 5px;
+}
+
+div#content *:first-child + * /* conceptually the first child, skipping #tools */
+{
+	margin-top: 0;
 }
 
 body.chm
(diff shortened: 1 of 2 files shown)


CyberShadow:
What about the "Improve this page" buttons?

aG0aep6G:
Yeah, they're clashing with the slogan now, forcing it to break:

iconscreenshot1

It's the same as on every other page now. But the home page is exceptionally long, of course, so that it happens easier.

I don't think it's too problematic. The slogan is overly long anyway. It has to be expected that it's broken in smaller windows.

CyberShadow:
OK, then how about making the slogan only break if necessary, so it prefers a break after "The D programming language."?

```diff
diff --git a/index.dd b/index.dd
index fcb176d..42c51bd 100644
--- a/index.dd
+++ b/index.dd
@@ -2,8 +2,7 @@ Ddoc
 
 $(D_S D Programming Language,
 
-$(SECTION3 The D programming language. Modern
-convenience. Modeling power. Native efficiency.,
+$(SECTION3 The D programming language. $(SPAN style="display: inline-block", Modern convenience. Modeling power. Native efficiency.),
 
 Conference websites, videos and slides:
 $(UL
```

aG0aep6G:
Sure thing. Updated.

PR #837, opened 2015-01-24
turn list items of Convenience/Power/Efficiency to paragraphs

aG0aep6G:
Before/after:
iconscreenshot1


andralex:
noice... thanks!

PR #827, opened 2015-01-24
Remove first prefix from submenus in library reference

JinShil:
no_std

PR #840, opened 2015-01-25
Redo #827: Remove first prefix from submenus in library reference

JinShil:
Had trouble rebasing.  This is a redo of #827.


andralex:
Close that. Merging this. Thanks!

PR #841, opened 2015-01-25
 Intent modules in packages 

CyberShadow:


JinShil:
Does the `container` expand/collapse?  If not, what is displayed when selecting `container`?

quickfur:
:+1: This is what we need.

Are the submodules collapsible? I suppose it's not a big deal if they aren't, though it would be nice. The primary thing is that the package names no longer bleed out of the navigation bar.

andralex:
OK this is good progress. I assume submodules are not collapsible. One question is whether they should. Three possibilities I see (out of many):

1. Uncollapsible, i.e. do nothing
2. Collapsible transitively
3. Floating submenu

Not sure which is best. I'd say staying with (1) until it hurts sounds like the best line of action.

quickfur:
I think (2) is a much better option. But since this PR already implements (1), I think we should merge this first, and then implement transitive collapsing in a followup PR.

I think (3) is a bit of an overkill.

andralex:
@quickfur roger that

CyberShadow:
Submodules are **not** collapsible.

I don't think we should make them collapsible. Navigating nested accordion menus is somewhat  frustrating since it requires multiple clicks on relatively small targets. Scrolling is much easier, and we don't win that much vertical space by collapsing them.

> Does the container expand/collapse? If not, what is displayed when selecting container?

For packages which declare a `package.d`, that file's documentation. Otherwise the package name is not a clickable link. Sorry, I couldn't figure out a  way to style links and non-links differently here in a way that looked nice.

CyberShadow:
It might make sense to collapse packages which have a `package.d`, i.e. can work transparently as a single module (mainly this applies to modules that have been split into packages, such as `std.algorithm`).

quickfur:
I'm OK with leaving them non-collapsible. They already look quite nice.

PR #842, opened 2015-01-25
use book style for all tables; also increase row spacing a little

aG0aep6G:
Properties before/after:
iconscreenshot1
std.range before/after:
iconscreenshot1

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -251,19 +251,39 @@ dd, .d_decl_dd
 
 table
 {
-	border-spacing: 0px;
+    border: solid black;
+    border-width: 2px 0;
+    border-collapse: collapse;
 }
 
-table, tr, td, th
+table tr
 {
-	border: 1px solid black;
-	/*border-style: solid;*/
+	border: none;
 }
 
-td
+table td, table th, table caption
 {
+    text-align: left;
+    vertical-align: top;
+    padding: 0.3em;
+}
+
+table td
+{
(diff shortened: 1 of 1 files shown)


andralex:
I think this is an improvement, but booktables look bad for very short and narrow tables, such as parameters. They look nice if the table is wide enough to occupy the entire page width. 

Therefore I'll pull this but I suggest @aG0aep6G you follow up with your previous work on tables and make it look sweet for small tables (2-4 entries, small width) and then apply that style appropriately. Thanks!

CyberShadow:
I noticed something wrong here, not sure if it's related:

http://dlang.org/phobos-prerelease/std_algorithm_mutation.html

aG0aep6G:
> I noticed something wrong here, not sure if it's related:
> 
> http://dlang.org/phobos-prerelease/std_algorithm_mutation.html

I don't think this PR has anything to do with it. The HTML is broken there:
```html
<div class="description"><table class="book"><caption>Cheat Sheet</caption></table><p></p>

<tr><th scope="col">Function Name</th> <th scope="col">Description</th></tr>
<p></p>
```

Looks like dmd inserts `<p></p>` on blank lines.

CyberShadow:
I see, thanks. Probably another victim of DDOC_PARAGRAPH.

https://github.com/D-Programming-Language/dmd/pull/4342

PR #843, opened 2015-01-25
un-underline jump links

aG0aep6G:
Before/after:

iconscreenshot1

As @CyberShadow said in #831, it's somewhat "inconsistent to do it selectively in the main content". So I'm not sure about this.

Phobos modules with manually maintained jump link tables will need to add a class to be affected by this. PR is on the way.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -933,6 +933,14 @@ td.compYes  /* comparison with D: "YES" */
 	background-color: #E4E9EF;
 }
 
+.quickindex a {
+	text-decoration: none;
+}
+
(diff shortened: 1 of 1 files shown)


aG0aep6G:
Phobos PR for the manually maintained tables: https://github.com/D-Programming-Language/phobos/pull/2907

CyberShadow:
Doesn't look too bad, really - the block is already styled differently (darker background), so the inconsistency is not very striking.

andralex:
that does look nice - thanks!

PR #847, opened 2015-01-26
Fall back to a self-hosted jQuery if Google CDN fails to load

CyberShadow:
Fixes https://issues.dlang.org/show_bug.cgi?id=14004


doc.ddoc
```diff
@@ -58,6 +58,7 @@ $(DIVID mobile-top,
 $(BODY)
 $(FOOTER)
 $(SCRIPTLOAD https://ajax.googleapis.com/ajax/libs/jquery/1.7.2/jquery.min.js)
+$(SCRIPT window.jQuery || document.write('<script src="js/jquery-1.7.2.min.js">\x3C/script>'))
```
andralex: clever

lplus: clever ,  very very very very clever !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
you win

PR #849, opened 2015-01-26
posix.mak fixes

CyberShadow:
- Don't use \t zsh-ism when building sitemap
- Add --chmod to rsync

CC @andralex

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -210,10 +210,10 @@ $(DOC_OUTPUT_DIR)/sitemap.html : $(ALL_FILES_BUT_SITEMAP) $(DMD)
 	cp -f sitemap-template.dd sitemap.dd
 	(true $(foreach F, $(TARGETS), \
 	  && echo \
-        "$F\t`sed -n 's/<title>\(.*\) - D Programming Language.*<\/title>/\1/'p $(DOC_OUTPUT_DIR)/$F`")) \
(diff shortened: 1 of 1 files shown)


posix.mak
```diff
@@ -210,10 +210,10 @@ $(DOC_OUTPUT_DIR)/sitemap.html : $(ALL_FILES_BUT_SITEMAP) $(DMD)
 	cp -f sitemap-template.dd sitemap.dd
 	(true $(foreach F, $(TARGETS), \
 	  && echo \
-        "$F\t`sed -n 's/<title>\(.*\) - D Programming Language.*<\/title>/\1/'p $(DOC_OUTPUT_DIR)/$F`")) \
+	    "$F	`sed -n 's/<title>\(.*\) - D Programming Language.*<\/title>/\1/'p $(DOC_OUTPUT_DIR)/$F`")) \
```
Orvid: You've got a tab here.

CyberShadow: Yes, just like every indented line in the entire Makefile. The spaces were an outlier.

Orvid: It's a tab followed by 4 spaces, shouldn't it be 2 tabs instead?

CyberShadow: > It's a tab followed by 4 spaces, shouldn't it be 2 tabs instead?

No, because the tab is part of the Makefile syntax, and the spaces are for indentation - this rule is indented with an indentation size of 2. Regardless, I didn't touch that part of the line :P

andralex:
noice

andralex:
yah I too format the makefile with 2-col indent

PR #851, opened 2015-01-26
Clarify scope of application of invariants

Dicebot:
Specifically, explain why it is not OK to refer to
other GC managed objects from invariant.

### DIFF
--- a/dbc.dd
+++ b/dbc.dd
@@ -181,6 +181,13 @@ class Date
 	$(LI postconditions)
 	)
 
+    $(P Because at the start of class/struct destructor object instance may
+    point to other object instances already collected by the GC trying to
+    check theirs state can lead to undefined behaviour. Invariants are
+    designed to ensure internal correctness of the object and shouldn't check
+    the global state or state of other GC objects (former can be done with
+    a lot of care but generally is discouraged).)
+
 	$(P The code in the invariant may not call any public non-static members
 	of the class or struct, either directly or indirectly.
 	Doing so will result in a stack overflow, as the invariant will wind
--- a/class.dd
+++ b/class.dd
@@ -778,8 +778,10 @@ $(GNAME Invariant):
 )
 
     Class invariants are used to specify characteristics of a class that always
-    must be true (except while executing a member function).
-    They are described in $(GLINK2 dbc, Invariants).
+    must be true (except while executing a member function). However, it only
+    applies to object internal consistency. Using invariant to check for global
+    program state can lead to subtle issues and is generally not recommended.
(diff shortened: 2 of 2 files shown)


CyberShadow:
I think this is a crappy reason which will massively restrict invariant usefulness. The definition of when an object is valid is, in practice, almost always in relation to other heap-allocated objects. For example, this is supposedly invalid:

---
class Account
{
    Currency balance;
    Transaction[] transactions;
    invariant()
    {
        assert(balance == transactions.map!(t => t.debit).sum);
    }
}
---

You say that the core reason is that the GC might destroy objects in arbitrary order. But if those objects don't themselves have destructors, then their state is actually valid, as the GC does not clobber freed objects' memory until it has finished running all destructors.

I've heard the argument that the above is just how the current implementation works, and should not be solidified into the language, as it might constrain future GC implementations. I'm not sure I agree, since with the current implementation, things are still memory-safe, no matter what you do in the destructor. Sure, you are allowed to access an object whose destructor has run and is now logically invalid, but that doesn't jeopardize memory safety.

I think we should go into the opposite direction and guarantee memory safety inside a destructor (with the remaining caveat that finalization order is arbitrary).

And as for the invariant getting called *after* the destructor... well, that's a bug. I checked TDPL, and even though it doesn't say much about invariants accessing other objects, it is explicit in that the invariant does not run after the destructor.

@andralex What do you think?

dbc.dd
```diff
@@ -181,6 +181,13 @@ class Date
 	$(LI postconditions)
 	)
 
+    $(P Because at the start of class/struct destructor object instance may
+    point to other object instances already collected by the GC trying to
```
1fish2: This needs a comma after "GC" so it doesn't parse as "... instances collected by [the GC trying to check...]".

dbc.dd
```diff
@@ -181,6 +181,13 @@ class Date
...
+    check theirs state can lead to undefined behaviour. Invariants are
```
1fish2: theirs -> their

1fish2:
It's really good to document the situation!

> However, it only applies to object internal consistency.

I'd want it to be plainly obvious what counts as "internal consistency". The [spec on destructors](http://dlang.org/class.html#destructors) says:

> This means that when the garbage collector calls a destructor for an object of a class that has members that are references to garbage collected objects, those references may no longer be valid.

That means an `invariant()` had better not check, say, that a string holds a suitable value (suitable for this object, as distinct from just being a well-formed string) or even that it's not a null reference.

[Separately, the situation is a nasty surprise. Should the collector skip calling the `invariant()` before the destuctor? At least, don't call it after invalidating any of the object's references?]

PR #852, opened 2015-01-26
use same search box on phobos pages as elsewhere

aG0aep6G:
Also DRY out other minor differences between doc.ddoc and std.ddoc.

### DIFF
--- a/doc.ddoc
+++ b/doc.ddoc
@@ -11,6 +11,7 @@ $(COMMON_SCRIPTS)
 _=
 
 ROOT = .
+SEARCHDEFAULT_PHOBOS =
 PROJECT = dlang.org
 _=
 
--- a/doc.ddoc
+++ b/doc.ddoc
@@ -1,59 +1,12 @@
 DDOC=
-<!DOCTYPE html>
-<html lang="en-US">
-<!--
-    Copyright (c) 1999-$(YEAR) by Digital Mars
-    All Rights Reserved Written by Walter Bright
-    http://digitalmars.com
-  -->
-<head>
-<meta charset="utf-8" />
-<meta name="keywords" content="$(META_KEYWORDS)" />
-<meta name="description" content="$(META_DESCRIPTION)" />
-$(T title, $(TITLE) - D Programming Language)
-<link rel="stylesheet" href="css/codemirror.css" />
-<link rel="stylesheet" href="css/style.css" />
-<link rel="stylesheet" href="css/print.css" media="print" />
-<link rel="stylesheet" href="css/cssmenu.css">
-<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.2.0/css/font-awesome.min.css">
-<link rel="shortcut icon" href="favicon.ico" />
-<meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=0.1, maximum-scale=10.0" />
-</head>
+$(DOCTYPE_OPENHTML_COPYRIGHT_HEAD)
 <body id='$(TITLE)'>
 $(SCRIPT document.body.className += ' have-javascript';)
 $(DIVID top,
-    $(DIVID search-box,
-        <form method="get" action="http://google.com/search">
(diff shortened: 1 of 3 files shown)


std.ddoc
```diff
@@ -1,55 +1,13 @@
...
-            <img id='mobile-logo' width="50" height="38" alt="D Logo" src="http://dlang.org/images/dlogo.png">
-            $(DIVID mobile-d-language,
-                <a href="http://dlang.org">D$(NBSP)Programming Language</a>
-            )
-        )
-    )
-)
-$(DIVID navigation, $(TOP)
-$(NAVIGATION_PHOBOS)
+    $(TOP_SEARCHBOX ,selected,)
```
aG0aep6G: This isn't particularly beautiful, but I couldn't think of a nicer way to pre-select an &lt;option&gt;.

CyberShadow: How about:
```
Macros:
    SEARCH_DEFAULT_STD=selected
```

doc.ddoc
```diff
@@ -67,6 +19,7 @@ $(SCRIPTLOAD js/cssmenu.js)
 </html>
 _=
 
+ROOT = .
```
CyberShadow: Why not empty string here, and use `$(ROOT)js/...` ?

doc.ddoc
```diff
@@ -1,60 +1,12 @@
...
-<link rel="stylesheet" href="css/print.css" media="print" />
-<link rel="stylesheet" href="css/cssmenu.css">
-<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.2.0/css/font-awesome.min.css">
-<link rel="shortcut icon" href="favicon.ico" />
-<meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=0.1, maximum-scale=10.0" />
-</head>
+$(DOCTYPE_OPENHTML_COPYRIGHT_HEAD)
 <body id='$(TITLE)'>
 $(SCRIPT document.body.className += ' have-javascript';)
 $(DIVID top,
```
CyberShadow: Why have two macros when you can have one? This bit is almost identical too.

std.ddoc
```diff
@@ -68,18 +26,7 @@ $(SCRIPT jQuery(document).ready(listanchors);)
 </html>
```
CyberShadow: More DRY potential here

doc.ddoc
aG0aep6G: I think

    ROOT = .
    ROOT = ..
    $(ROOT)/foo

are clearer and easier to get right than

    ROOT =
    ROOT = ../
    $(ROOT)foo

But if there's an upside to doing it the other way, I can do that of course.

CyberShadow:
Thanks, I was going to do this at some point as well. But I think more refactoring can be done :)

You can move stuff out of the inline HTML code by specifying it as a macro with different value in `std.ddoc` and `doc.ddoc`, then using it in `dlang.org.ddoc`.

doc.ddoc
CyberShadow: The upside is that HTTP requests won't have a `./` prefix.

aG0aep6G: Hm, shouldn't the browser usually swallow them?

CyberShadow: Maybe?

Anyway, this is just a nitpick. But a prefix with a trailing slash is a common pattern, e.g. the `--prefix` parameter to `./configure` scripts.

aG0aep6G:
Pushed a commit that consolidates the multiple macros into one TOP.

aG0aep6G:
and one to DRY the scripts

CyberShadow:
OK, not exactly what I had in mind but this LGTM as is. I'll have a go at my own suggestions myself

doc.ddoc
CyberShadow: The current approach has the advantage that `<a href="$(ROOT)">Index</a>` works properly, which allows linking to the main page without hard-coding the `http://dlang.org/` URL.

PR #853, opened 2015-01-26
New logo proposal

p0nce:
I thought I would give it a try and make something more into the current hype.
Included SVG derived from the original logo.


p0nce:
Before:
Before

After:
After

AndrejMitrovic:
A little too sharp on the edges for my taste. Could you make it a bit rounded and see what that looks like? :)

CyberShadow:
#812 moved the title to the right. Another tweak you could try is to make the logo as wide as the navigation.

p0nce:
@AndrejMitrovic Erm, let me see another day. Spent hours already. I don't really like rounded edges everywhere like on the current site, but it makes sense to have subtle ones like on the menu. Can't guarantee it will look better.

@CyberShadow Makes sense. Maybe we can display the SVG directly then (the PNG is exactly the SVG render). But going to be now.

WalterBright:
I appreciate the work you've put in, but I'd like to keep the current logo:

1. logo consistency is very important for branding. We've done a good job associating D with the current logo
2. I like the enameled look of the current one
3. I like the current rounded edges, the subtle variation in color, the soft edges of the D part, the D fills the logo, etc.

CyberShadow:
> logo consistency is very important for branding

A lot of companies (possibly all major ones) tweak their logo over time while keeping the core theme.

See for yourself: https://www.google.com/search?tbm=isch&q=logo+evolution

WalterBright:
I know some do. But not that quickly, and they launch new logos with marketing campaigns.

CyberShadow:
BTW, please use PNG for screenshots.

andralex:
BDFL-induced closing.

CyberShadow:
The comments of this week's This Week in D seem to centralize on the need of a new logo:

https://www.reddit.com/r/programming/comments/2tozgb/this_week_in_d_website_makeover_c_interop_import/

@WalterBright I suggest that you reconsider.

WalterBright:
There's no rationale given, either here or on reddit or on the other PR for a logo change. Changing an established logo is not something to be taken lightly. Let's not spend resources on an illusion of progress.

CyberShadow:
What's the rationale of all the work we've been doing this week on changing the website design?

p0nce:
Wow I didn't want to launch such a heated debate, it was really just a proposal.

I'm fully OK with the logo staying as it is: branding is indeed important and I like BDFL decisions.

I just wanted to try another thing and eventually make more material for D-related stuff like "d-idioms" or "This Week In D" where the current logo looks odd.

I will use this logo for "d-idioms" not because I think it looks better (such a statement I can't make, for example the original logo has lovely reflections) but because my logo will integrate better in the surroundings on this site.

WalterBright:
I do appreciate the effort you put into helping out! Thank you

AndrejMitrovic:
We should make a couple of proposal and the post to reddit and ask on opinions whether the logo should change. Then it's practically automatically accepted as the new logo (just like the recent *@changes*).

aldanor:
While I don't personally like the redesign suggested above (or maybe it just lacks a few finishing touches and detail here and there), I think it may be a move in the right direction since the current logo is a bit tacky/90's-style and may seem completely oddball when slapped on an otherwise polished website with a design that tries to stay modern. Redesigning it is then a matter of making it more compatible with other content -- that should be the primary reason and not just doing it for the sake of it.

> logo consistency is very important for branding

I agree; but while this is true in general, the subtle design details may have to be changed in order to keep in touch with the latest trends. E.g., Microsoft, Google and Apple:
- http://gizmodo.com/the-evolution-of-googles-iconic-logo-1582297667
- http://blogs.windows.com/bloggingwindows/2012/02/17/redesigning-the-windows-logo/
- http://blogs.tribune.com.pk/story/19792/the-evolution-of-the-apple-logo/

p0nce:
> the current logo is a bit tacky/90's-style and may seem completely oddball when slapped on an otherwise polished website with a design that tries to stay modern. Redesigning it is then a matter of making it more compatible with other content -- that should be the primary reason and not just doing it for the sake of it.

Exactly my point, that's why new logo proposal keep the "D + planet" shape.

PR #855, opened 2015-01-26
Issue 14047 - "this" missing from AsmPrimaryExp grammar

Hackerpilot:
https://issues.dlang.org/show_bug.cgi?id=14047

### DIFF
--- a/iasm.dd
+++ b/iasm.dd
@@ -307,6 +307,7 @@ $(GNAME AsmPrimaryExp):
     $(GLINK Register64)
     $(GLINK Register64) $(D :) $(I AsmExp)
     $(I DotIdentifier)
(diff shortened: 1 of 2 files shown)


iasm.dd
```diff
@@ -307,6 +307,7 @@ $(GNAME AsmPrimaryExp):
     $(GLINK Register64)
     $(GLINK Register64) $(D :) $(I AsmExp)
```
CyberShadow: Why is this only in `iasm.dd`?

Hackerpilot: Because this pull request for issue 13308 wasn't correct:
https://github.com/D-Programming-Language/dlang.org/pull/783/

PR #860, opened 2015-01-26
Issue 13301 - Inline ASM documentation does not allow string literals

Hackerpilot:
https://issues.dlang.org/show_bug.cgi?id=13301

### DIFF
--- a/grammar.dd
+++ b/grammar.dd
@@ -832,6 +832,12 @@ $(GNAME AsmInstruction):
     $(D df) $(I Operands)
     $(D dd) $(I Operands)
     $(D de) $(I Operands)
+    $(D db) $(GLINK2 lex, StringLiteral)
+    $(D ds) $(GLINK2 lex, StringLiteral)
+    $(D di) $(GLINK2 lex, StringLiteral)
+    $(D dl) $(GLINK2 lex, StringLiteral)
+    $(D dw) $(GLINK2 lex, StringLiteral)
+    $(D dq) $(GLINK2 lex, StringLiteral)
     $(I Opcode)
     $(I Opcode Operands)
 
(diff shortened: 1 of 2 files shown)


grammar.dd
```diff
@@ -832,6 +832,13 @@ $(GNAME AsmInstruction):
     $(D df) $(I Operands)
     $(D dd) $(I Operands)
     $(D de) $(I Operands)
+    $(D db) $(GLINK_LEX StringLiteral)
+    $(D ds) $(GLINK_LEX StringLiteral)
+    $(D di) $(GLINK_LEX StringLiteral)
+    $(D dl) $(GLINK_LEX StringLiteral)
+    $(D df) $(GLINK_LEX StringLiteral)
```
CyberShadow: How does this make sense? A floating-point string?

Hackerpilot: Each can have multiple operands. If an operand is a string literal, it is as if there were length operands, where length is the number of characters in the string

Though you're right that this doesn't make sense, and now that I'm reading the parser source it seems that I need to rewrite this part of the spec and also change the grammar.

MartinNowak:
How about updating `Operands` instead? Apparently Operand isn't defined anyhow.
And there is also https://github.com/D-Programming-Language/dlang.org/blob/0a92173a062697436456bb84a936ce0d0a3097f8/iasm.dd#L35.

JinShil:
@Hackerpilot, a follow-up to @MartinNowak 's comment?

DmitryOlshansky:
Ping @Hackerpilot

Hackerpilot:
I think that the reason was that string literals can't go in most of the places that an `Operand` can. I'd have to read through everything again to be sure.

CyberShadow:
@Hackerpilot @MartinNowak Is this good to go?

Hackerpilot:
I think so, I'm not sure if Martin agrees.

andralex:
I'll preemptively pull, @MartinNowak please undo (or build upon) if necessary.

PR #862, opened 2015-01-27
Cleanup and refactoring

CyberShadow:
- DRY/refactor common parts of std.ddoc and doc.ddoc (continuation of #852)
- Clean up unused macros

The cleanup and refactoring is an important step towards having a consistent style on *.dlang.org.


andralex:
Ah, so nice. I wanted to do the DDOC= consolidation forever, and I was planning on using exactly the same ROOT trick. Thanks!

PR #864, opened 2015-01-27
navbar for std.experimental.logger

burner:
phobos prerelease navbar entry for std.experimental.logger package

* add experimental top level entry below core
* add new macro MODULE4

### DIFF
--- a/std_navbar-prerelease.ddoc
+++ b/std_navbar-prerelease.ddoc
@@ -121,6 +121,17 @@ $(DIVID cssmenu, $(UL
 		  $(MODULE3 core, sync, semaphore)
 		)
 	  )
+
+	$(MENU_W_SUBMENU $(TT experimental))
+	  $(ITEMIZE
+		$(PACKAGE $(PACKAGE_NAME logger),
+		  $(MODULE4 std, experimental, logger, package),
+		  $(MODULE4 std, experimental, logger, core),
+		  $(MODULE4 std, experimental, logger, filelogger),
+		  $(MODULE4 std, experimental, logger, multilogger),
+		  $(MODULE4 std, experimental, logger, nulllogger)
+		)
+	  )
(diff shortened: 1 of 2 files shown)


std_navbar-prerelease.ddoc
```diff
@@ -121,6 +121,17 @@ $(DIVID cssmenu, $(UL
...
+	$(MENU_W_SUBMENU $(TT experimental))
+	  $(ITEMIZE
+		$(PACKAGE $(PACKAGE_NAME logger),
+		  $(MODULE4 std, experimental, logger, package),
+		  $(MODULE4 std, experimental, logger, core),
+		  $(MODULE4 std, experimental, logger, filelogger),
+		  $(MODULE4 std, experimental, logger, multilogger),
+		  $(MODULE4 std, experimental, logger, nulllogger)
+		)
+	  )
```
CyberShadow: I don't think you've added this in the right place.

CyberShadow:
#865

std_navbar-prerelease.ddoc
CyberShadow: Oh, I see now that was intentional. Did you do this because the CSS did not support it, or to avoid nesting things too deeply?

burner: I did it so it is clear that there is something called experimental and that is has a logger module.

having it below exception is sort of wired as exception is stable part of phobos and logger is not.

CyberShadow: OK, though that's a little inconsistent (package name is `std.experimental`, not `experimental`).

CyberShadow: https://github.com/D-Programming-Language/dlang.org/pull/866

burner: #866 works for me

PR #866, opened 2015-01-27
navbar for std.experimental.logger

CyberShadow:
Alternative with separate "std.experimental" menu

### DIFF
--- a/std_navbar-prerelease.ddoc
+++ b/std_navbar-prerelease.ddoc
@@ -121,6 +121,16 @@ $(DIVID cssmenu, $(UL
 		  $(MODULE3 core, sync, semaphore)
 		)
 	  )
+	$(MENU_W_SUBMENU $(TT std.experimental))
+	  $(ITEMIZE
+		$(MODULE3 std, experimental, logger, package),
+		$(PACKAGE
+		  $(MODULE4 std, experimental, logger, core),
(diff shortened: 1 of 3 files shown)


std_navbar-prerelease.ddoc
```diff
@@ -121,6 +121,16 @@ $(DIVID cssmenu, $(UL
 		  $(MODULE3 core, sync, semaphore)
 		)
 	  )
+	$(MENU_W_SUBMENU $(TT std.experimental))
+	  $(ITEMIZE
+		$(MODULE3 std, experimental, logger, package),
```
burner: shouldn't this be MODULE4

CyberShadow: No, the group caption should say "logger", not "package". Look e.g. at how std.container is presented in the navigation.

burner: I see

burner:
thanks

CyberShadow:
Um, the  idea was to merge either this on #865...

CyberShadow:
I think putting it in the main "std" menu is a little better than as a separate menu.

The rationale being that a user who knows about std.experimental would start looking for the module's documentation under the "std" menu - they might not notice the separate menu way below, and wonder why `experimental` is not in its alphabetically-sorted place.

PR #867, opened 2015-01-27
get rid of HTML duplication for mobile header

aG0aep6G:
Also replaces the jquery animation for the mobile search box with CSS.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1074,7 +1074,9 @@ input.runButton:active, input.resetButton:active, input.argsButton:active, input
 input.resetButton{display: none}
 /* Runnable-examples css -end */
 
-div#mobile-top, div#mobile-search {
+div#mobile-top, div#mobile-search, div#mobile-hamburger,
+div#mobile-search-button
+{
 	display: none;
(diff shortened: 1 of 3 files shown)


andralex:
"72 additions and 69 deletions" - not much code removal eh :)

css/style.css
```diff
@@ -1163,7 +1171,25 @@ div#mobile-top, div#mobile-search {
...
+		transition: height 0.5s;
+	}
+
+	div#search-box.open {
+		height: 2.225em;
+		/*
+			= 0.625em [<form>'s margin-top]
+			+ 1.4 em [#search-{query,dropdown,submit}'s height]
+			+ 0.2em [ca. #search-{query,dropdown,submit}'s border]
+		*/
```
andralex: is this comment any relevant?

aG0aep6G: It explains how the 2.225em above come together.

CyberShadow:
Hmm, such a big patch, but I don't see any difference, visually...

Good!

:+1:

PR #870, opened 2015-01-28
posix.mak: Use pdflatex instead of latex+dvipdf

CyberShadow:


CyberShadow:
CC @andralex

posix.mak
```diff
@@ -269,12 +269,10 @@ dlangspec.tex : $(DMD) $(DDOC) latex.ddoc dlangspec-consolidated.d
...
-dlangspec.dvi : dlangspec.tex
-	latex $^
-	latex $^
-
-$(DOC_OUTPUT_DIR)/dlangspec.pdf : dlangspec.dvi
-	dvipdf $^ $@
+$(DOC_OUTPUT_DIR)/dlangspec.pdf : dlangspec.tex
+	pdflatex -draftmode $^
+	pdflatex $^
+	-mv dlangspec.pdf $(DOC_OUTPUT_DIR)/dlangspec.pdf
```
andralex: The result of mv shouldn't be ignored. Better yet, use the `-output-directory` option of pdflatex to create the result straight at destination. Also in the rules it's better to use `$@` for the target instead of the explicit name. Thanks!

CyberShadow: > The result of mv shouldn't be ignored.

It  is to silence the error in the case that `$(DOC_OUTPUT_DIR)` is `.`.



> Better yet, use the -output-directory option of pdflatex to create the result straight at destination.

This will put all the intermediate junk there too.

andralex: Then let's remove the intermediate junk. Clearly we can't afford to silently ignore a failure. Possibility: generate in `/tmp/` then move.

CyberShadow: The `mv` also serves a second purpose, to protect against a fallacy that is common in makefiles. If the build process is interrupted, the partially-created PDF file will have an mtime newer than its dependencies, thus it will not be rebuilt.

CyberShadow: Using `/tmp` is not ideal because it is a global resource - junk created by one user can get in the way of another's. I think it should be only used with `mkstemp` or the like, which is not available from Makefiles, so I'll use a local `.tmp` instead.

CyberShadow:
Updated.

CyberShadow:
Ping?

PR #872, opened 2015-01-28
Update doc.ddoc

justanotheranonymoususer:
Any reason for that `$(NBSP)`?


CyberShadow:
Probably a leftover from the old design.

PR #873, opened 2015-01-28
chmgen overhaul

CyberShadow:
This is close to a rewrite of chmgen:

- Compatibility with the new layout
- Removing ancient hacks from the digitalmars.com/d/1.0/ days
- General code decrummification
- Removing KirkMcDonald's ancient index (it's no longer needed, anyway)
- Operating on entire files instead of per-line
- Slight clean-up of directory structure
- Adding `d.tag` output file for `dman`
- Navigation is parsed by rendering DDoc to almost-valid (trailing commas) JSON
- Explicit linting, which finds broken links etc.

I really wanted to do the HTML modification part with DDoc, but doing so would mean adding a few pages of copy-pasta to win32.mak (incl. a listing of Phobos modules, i.e. another unmaintainable mess). So regexes it is.

Does anyone still use the CHM documentation, anyway?


chmgen.d
```diff
@@ -566,418 +395,79 @@ main="D Programming Language","d.hhc","d.hhk","chm\index.html","chm\index.html",
...
+// ********************************************************************
 
+string forwardSlashes(string s)
+{
+	return s.replace(`\`, `/`);
+}
 
-----
+//                   '  *  '
+string backSlashes(string s)
```
andralex: why these one-liners?

dlang.org.ddoc
```diff
@@ -27,10 +27,10 @@ _=
 
 COMMON_SCRIPTS =
     $(SCRIPTLOAD https://ajax.googleapis.com/ajax/libs/jquery/1.7.2/jquery.min.js)
-    $(SCRIPT window.jQuery || document.write('<script src="$(ROOT)/js/jquery-1.7.2.min.js">\x3C/script>'))
-    $(SCRIPTLOAD $(ROOT)/js/codemirror-compressed.js)
-    $(SCRIPTLOAD $(ROOT)/js/run.js)
-    $(SCRIPTLOAD $(ROOT)/js/cssmenu.js)
+    $(SCRIPT window.jQuery || document.write('<script src="$(ROOT_DIR)js/jquery-1.7.2.min.js">\x3C/script>'))
```
andralex: This business with `$(ROOT_DIR)` including the slash is distinctly not awesome. A slash is always easy to add but more difficult to remove. Can we please go the other way, thanks.

dlang.org.ddoc
```diff
@@ -57,12 +57,12 @@ DDOC=
...
 <meta name="description" content="$(META_DESCRIPTION)" />
 $(T title, $(TITLE) - D Programming Language)
-<link rel="stylesheet" href="$(ROOT)/css/codemirror.css" />
-<link rel="stylesheet" href="$(ROOT)/css/style.css" />
-<link rel="stylesheet" href="$(ROOT)/css/print.css" media="print" />
-<link rel="stylesheet" href="$(ROOT)/css/cssmenu.css">
+<link rel="stylesheet" href="$(ROOT_DIR)css/codemirror.css" />
+<link rel="stylesheet" href="$(ROOT_DIR)css/style.css" />
+<link rel="stylesheet" href="$(ROOT_DIR)css/print.css" media="print" />
+<link rel="stylesheet" href="$(ROOT_DIR)css/cssmenu.css">
```
andralex: nononononoonono

dlang.org.ddoc
```diff
@@ -57,12 +57,12 @@ DDOC=
...
 <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.2.0/css/font-awesome.min.css">
-<link rel="shortcut icon" href="$(ROOT)/favicon.ico" />
+<link rel="shortcut icon" href="$(ROOT_DIR)favicon.ico" />
```
andralex: no

dlang.org.ddoc
```diff
@@ -245,7 +245,7 @@ TOP =
     $(SCRIPT document.body.className += ' have-javascript';)
     $(DIVID top,
         $(DIVID header,
-            <a class="logo" href="$(ROOT)"><img id="logo" width="125" height="95" alt="D Logo" src="$(ROOT)/images/dlogo.png"></a>
+            <a class="logo" href="$(ROOT)"><img id="logo" width="125" height="95" alt="D Logo" src="$(ROOT_DIR)images/dlogo.png"></a>
```
andralex: and no

doc.ddoc
```diff
@@ -1,3 +1,4 @@
+ROOT_DIR =
```
andralex: no, `./` should be fine

.gitignore
```diff
@@ -31,3 +31,5 @@ docs-prerelease.json
 docs.json
 .dub/
 dpl-docs/dpl-docs
+
+/d.tag
```
andralex: I'd rather use $TMP for junk instead of having it there and ignoring it. (That would be nice for the LaTeX-related stuff, too.)

chmgen.d
CyberShadow: I went back and forth between slashes and backslashes a few times. I'm not sure which way is more compatible with the CHM readers out there.

dlang.org.ddoc
CyberShadow: The ROOT macro is still there if you need it without the slash.

I need this for CHM generation. We can either do it this way, or with more regular expressions. You decide.

doc.ddoc
CyberShadow: The ROOT macro is literally still below this line.

.gitignore
CyberShadow: This file is needed to build the tools repository.

CyberShadow: Also, as mentioned in #870, $TMP is usually a global resource and should be used carefully. Early versions of rdmd used to have this problem - after one user used rdmd once, rdmd was not usable by any other user on the system.

dlang.org.ddoc
andralex: >I need this for CHM generation.

Where exactly?

CyberShadow: Microsoft's CHM generator seems to do funny things with `./`, it imports files in the current folder if it doesn't find a file in the folder of the file it's reading, thus a missed link causes it to pull in non-processed HTML files (meant for dlang.org) as well as preprocessed HTML files meant for the CHM.

MartinNowak:
Just tried to build a release from master and currently chmgen dies with an access violation.
So can we please go on with this?

CyberShadow:
Rebased.

@andralex Anything else?

chmgen.d
```diff
@@ -1,473 +1,305 @@
...
+		.replaceAll(re!`,\s*\]`, `]`)
+	;
+	scope(failure) std.file.write("error.json", text);
+	auto json = text.parseJSON();
+
+	Nav parseNav(JSONValue json)
+	{
+		if (json.type == JSON_TYPE.ARRAY)
+		{
+			auto nodes = json.array;
```
andralex: remove hard tabs pliz

CyberShadow:
> remove hard tabs pliz

Why? Why are you so concerned about the coding style of an utility that, in all likeliness, nobody except me is going to touch? Why bother if the only people who use this, use this indirectly because it's part of the build process? Why expend the effort, if we haven't heard anyone say anything about the CHM in years, likely because no one uses it any more? Why chmgen, when you have tabs all over the place elsewhere, in much more important and "trodden" parts, such as the .css files and spec .dd files and Dub and Ddox? And why now, when tabs in chmgen are nothing new, and when you know that this blocks fixing the build?

doc.ddoc
MartinNowak: Let's add `ROOT_SLASH=$(ROOT)/` and go in with this.

CyberShadow: What would be the point? There would be no reason to use `ROOT_SLASH` instead of `ROOT_DIR`, and then we would have three macros for one thing.

andralex: The point is `ROOT_SLASH` is more descriptive. Good makefile style is to NOT append slashes to directories in macros because on the use side it can be easily added and it's difficult to remove.

andralex:
>Why are you so concerned about the coding style of an utility that, in all likeliness, nobody except me is going to touch? 

No need to make a thing out of this. Our general guidelines avoid hard tabs, and this code is part of our repository. Please let's just do it? Thanks!

doc.ddoc
CyberShadow: OK I'm really confused, is the consensus that we want three macros for the same thing? I can add `ROOT_SLASH` but I see no practical point in it. Should it replace one of the two macros? Which one?

CyberShadow:
> No need to make a thing out of this.

Not the intention, it just got a little ... out of hand ... anyway, I did add a detab commit.

doc.ddoc
CyberShadow: For the case that there is a misunderstanding, in `doc.ddoc` there now is:

> ```
> ROOT_DIR = 
> ROOT = .

And in `std.ddoc`:

> ```
> ROOT_DIR = ../
> ROOT = ..

The `ROOT_DIR` macro is needed to avoid prefixing local URLs with `./`, which is not only ugly, but is also problematic for CHM generation.

The `ROOT` macro is now only used for linking to `index.html` (without actually putting `index.html` in the address bar, which is more ugly than it needs to be).

CyberShadow:
Ping? Does anyone want to explain or talk about the problem holding up this pull, and what needs to be done to fix it?

doc.ddoc
MartinNowak: I was suggesting ROOT_SLASH instead of ROOT_DIR, because it's meaning is clearer, and I suggested to define it as `$(ROOT)/`.

CyberShadow: > and I suggested to define it as $(ROOT)/.

No, that would make it pointless. It is only useful because instead of `./`, it will be an empty string.

MartinNowak: But you override it anyhow, so keeping the default in sync with ROOT seems to make sense.
I don't care too much about this though.

CyberShadow: ROOT_DIR is defined in exactly the same places (2) that ROOT is defined, right below the ROOT definitions. There is no overriding going on.

MartinNowak: Ah OK, I somehow thought there was a chm specific override.

MartinNowak:
I understand that some people on Windows expect a chm help in their start menu after installation.
But __this__ looks like our website in a crappy browser. Is this really worth all the effort we're putting into this?
chm

CyberShadow:
It shouldn't look like that. I'll look into it.

CyberShadow:
I'm now getting an access violation... did you not get one?

MartinNowak:
> It shouldn't look like that. I'll look into it.

That's not the point though, because I know you'll fix it. Since nobody looks at chm we'll always pile up regressions that we'll have to deal with before releasing, not funny.
And all of that for our website in a weird program that probably runs what, an IE7 engine?

BTW
Druntime isn't styled at all, but that might be an issue on my side.

CyberShadow:
> That's not the point though, because I know you'll fix it. Since nobody looks at chm we'll always pile up regressions that we'll have to deal with before releasing, not funny.

Yeah, I know. I wanted to do it in DDoc, but... win32.mak.

> And all of that for our website in a weird program that probably runs what, an IE7 engine?

Yeah, I know. I'm mostly maintaining it because Manu kept saying how important and widespread CHMs are in gamedev.

CyberShadow:
BTW, by "it shouldn't look like that" I didn't mean the brokedness by itself, I mean that the navigation and sidebars should be gone, as well as the header, i.e. the content fills up the entire frame.

MartinNowak:
> BTW, by "it shouldn't look like that" I didn't mean the brokedness by itself, I mean that the navigation and sidebars should be gone, as well as the header, i.e. the content fills up the entire frame.

I built that one manually, I'll hopefully will get one from the release builder in a few minutes.
The druntime pages filled the whole screen, but weren't styled.

CyberShadow:
Something is really really wrong with the navigation now.

MartinNowak:
> Yeah, I know. I'm mostly maintaining it because Manu kept saying how important and widespread CHMs are in gamedev.

Any comment on that @Turkeyman?

MartinNowak:
On 02/17/2015 08:21 PM, Vladimir Panteleev wrote:
> I'm now getting an access violation... did you not get one?
>
Wait an hour or so, I have a pile of fixes that I'm testing right now.

CyberShadow:
Sheesh, way to out a behind-the-back-talker ;) When I maintained my old CHM page, it was relatively popular ( http://thecybershadow.net/d/docs/ )

I should add a notice that it's in D now or something I guess.

CyberShadow:
The A/V was because I had an old Phobos checkout, I think. But I fixed the code too.

CyberShadow:
> Something is really really wrong with the navigation now.

Never mind, that's because I broke it. D'oh.

CyberShadow:
> The druntime pages filled the whole screen, but weren't styled.

Repro'd.

CyberShadow:
BTW, I think IRC might be a little more efficient. And Andrei was looking for you there the other day. ;)

CyberShadow:
https://github.com/D-Programming-Language/dlang.org/pull/910

We could create a poll on the NG, if no one uses the CHM then we can remove it. But this program would still be useful, for d.tag generation and as a lint tool.

MartinNowak:
You can find my build output here.
I'd be glad if you could test the windows zip a little.
https://dlang.dawg.eu/downloads/master/

MartinNowak:
> BTW, I think IRC might be a little more efficient.

I'll see what I can do.

CyberShadow:
> I'd be glad if you could test the windows zip a little.

Builds Digger, rabcdasm and DFeed just fine.

What happened with the libcurl download link in download.dd? I was just thinking we should add an sqlite3.lib/.dll download.

CyberShadow:
Oh, of course, curl.lib is included now. That leaves sqlite.

MartinNowak:
> Oh, of course, curl.lib is included now.

Yep :).

> I was just thinking we should add an sqlite3.lib/.dll download.

As long as we don't stuff everything onto the download page again.
https://web.archive.org/web/20141109181718/http://dlang.org/download.html

And I don't even know how to build it. There is http://wiki.dlang.org/Curl_on_Windows, but http://wiki.dlang.org/?search=sqlite is fairly empty. Something like a build manual in the wiki and a link to ftp.digitalmars.com would work. Or better yet one could host the bins on github.

CyberShadow:
The DLL download (http://www.sqlite.org/2015/sqlite-dll-win32-x86-3080802.zip)  contains a .dll and a .def.

CyberShadow:
SQLite is really easy to compile (as a static library, anyway) because in one of its forms it comes as a single .c file, and is compatible with DMC.

CyberShadow:
> The druntime pages filled the whole screen, but weren't styled.

This is still not fixed BTW, but the problem is not with chmgen. The Windows docs are built like that!

CyberShadow:
> The druntime pages filled the whole screen, but weren't styled.

Figured it out. You just need to force a rebuild of Druntime docs. As they're all built in Phobos' win32.mak, it can't track dependencies, so all you have to do is a clean build.

TurkeyMan:
@MartinNowak I still stand by that .chm claim. All gamedev docs are .chm files. It's familiar and comfortable to that industry.
I do think the importance is slowly regressing with time though, as people are more and more accustomed to just go straight for google than to worry about reading a document on their local machine.
I'd like to see it remain for the time being though. New comers from the games industry might find it useful and get a small confidence boost from it's presence :)

MartinNowak:
> @MartinNowak I still stand by that .chm claim. All gamedev docs are .chm files. It's familiar and comfortable to that industry.

Fair enough, I'm just worried that the quality we're delivering rather turns people off and that even though we're spending quite some time on it. Maybe it would be better spend improving the pdf?

TurkeyMan:
Speaking for myself, I would sooner go to the internet than ever open a pdf. But if it's absorbing significant time, then I'm not sure the value holds.
I'll say, I think the .chm is valuable, but I'll let someone who knows the maintenance burden decide if it should live.

MartinNowak:
I say we'll keep it for now since @CyberShadow just rewrote the chmgen tool (with the nice side-effect of making dman way more useful).
But in the longer-term we should reconsider our options.

CyberShadow:
Eh, I enjoy working on it.

I'll probably stop when the CHM stops carrying its weight (in zip filesize).

PR #874, opened 2015-01-28
Add note to install 'libevent.a' on OSX 

zachthemystic:
I needed to do it to build apidocs-... maybe others will need it too.


CyberShadow:
This is only needed for dpl-docs (the dub stuff), and then only the "serve" program, right?

zachthemystic:
I suspect you're right. Two of three compiles succeeded, so yeah.

PR #877, opened 2015-01-29
SVG logo

CyberShadow:
PNG logo is not removed, as various other sites, including *.dlang.org, hotlink (embed) it.

I don't have stats for dlang.org, but for forum.dlang.org, 99.7% of visitors use SVG-capable user agents.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-![D Logo](http://dlang.org/images/dlogo.png) D Language Website
+![D Logo](http://dlang.org/images/dlogo.svg) D Language Website
 ===============================================================
 
 This repository contains the source files for the [D Language
--- a/dlang.org.ddoc
+++ b/dlang.org.ddoc
@@ -245,7 +245,7 @@ TOP =
     $(SCRIPT document.body.className += ' have-javascript';)
     $(DIVID top,
         $(DIVID header,
-            <a class="logo" href="$(ROOT)"><img id="logo" width="125" height="95" alt="D Logo" src="$(ROOT)/images/dlogo.png"></a>
+            <a class="logo" href="$(ROOT)"><img id="logo" width="125" height="95" alt="D Logo" src="$(ROOT)/images/dlogo.svg"></a>
             <span id="d-language-mobilehelper"><a href="$(ROOT)" id="d-language">D Programming Language</a></span>
         )
         $(DIVID search-box,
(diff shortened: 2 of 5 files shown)


CyberShadow:
It looks slightly different.

MartinNowak:
Why to keep the PNG logo then?

CyberShadow:
> PNG logo is not removed, as various other sites, including *.dlang.org, hotlink (embed) it.

CyberShadow:
> Why to keep the PNG logo then?

Did you mean to ask "why not keep the PNG logo"?

Because it looks blurry on high-DPI screens.

CyberShadow:
It was requested here and a few other times, I think:

http://forum.dlang.org/post/qeiyyxfiajmlbkuxxbpu@forum.dlang.org

MartinNowak:
> Did you mean to ask "why not keep the PNG logo"?

No, I meant why not deleting the png file?

CyberShadow:
Yeah, because it's hotlinked in a few places. For example our README files do that, but I bet there's more than that.

andralex:
What are the advantages of svg over png? I figure it's scalable, but we're not taking advantage of that, right?

CyberShadow:
> Because it looks blurry on high-DPI screens.

Specifically: phones released in the past decade, high-DPI displays such as all the latest Macbooks, etc.

andralex:
@CyberShadow got relevant screenshots?

CyberShadow:
Sure I can make some.

andralex:
Also I recall there was that 6% of folks who won't see svgs?

CyberShadow:
No, that figure was off - it was based on global stats. For forum.dlang.org visitors (I have no dlang.org data) it's 0.3%, not 6%.

andralex:
mmmkay.. @WalterBright ?

CyberShadow:
If 0.3% is not acceptable, we can try to fall back to the old image (or progressively enhance to SVG), but all fallback methods I've seen were ugly hacks...

WalterBright:
If the 0.3% can't see it, what do they see instead? Is the rest of the page still good?

CyberShadow:
> If the 0.3% can't see it, what do they see instead?

Blank space, or a "broken image" placeholder.

> Is the rest of the page still good?

Yes.

WalterBright:
Then I'm fine with it.

CyberShadow:
### Note 4

PNG logo:

 

SVG logo:



### Windows 7 at 150% DPI

PNG logo:



SVG logo:

andralex:
@CyberShadow loud and clear thanks

PR #880, opened 2015-01-31
add "make all" hint to CONTRIBUTING.md

zachthemystic:
Redone. New pull.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,6 +11,7 @@ Make sure you have these prerequisites working:
 * The C++ compiler for your platform (invokable as `gcc`)
 * Optional: `latex` for building the pdf documentation
 * Optional: `kindlegen` for building the Kindle documentation
+* If you're running OSX, make sure you have some version of `libevent` installed
 
 ## Getting the code
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -100,8 +100,8 @@ make -f posix.mak apidocs-prerelease
 The output is in `$R/dlang.org/web/phobos-prerelease` and
 `$R/dlang.org/web/library-prerelease`.
 
-To rebuild your changes, use:
+To rebuild your dlang.org changes, use:
 
 ```
-make -f posix.mak all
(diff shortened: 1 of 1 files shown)


CyberShadow:
... Why are there two commits in your commit ilst?

But seriously, that is just a nitpick. It's not *the* reason for why the previous pull did not get merged, if such a reason even exists other than enough people approving the change.

As for the change, as minor as it is, it is a little suspicious. Building the `all` target will not only rebuild things that have already been built, but potentially also build things which have not been built yet. The same command that worked for the first build *should* work again for rebuilding the same targets.

Maybe we should fix the makefile instead? Can you describe the problem first?

zachthemystic:
My problem: I followed the instructions, then changed my files, and did "make -f posix.mak apidocs-prerelease" again, hoping it would rebuild the docs, but for some reason, it didn't because it thought they were up to date. I honestly don't know why it didn't work. "make all" and it did work, so I assumed that was what one was supposed to do. That's all I know.

CyberShadow:
> then changed my files

Which files did you change? 

> apidocs-prerelease

This is the DDox build target. Were you working on DDox, or on Phobos but testing your changes using DDox?

zachthemystic:
I wanted to build dlang.org. I had edited dlang.org/doc.ddoc, for example. What's the target for that?

CyberShadow:
`html`, which is included in `all`.

zachthemystic:
Okay, changed. Thanks for the guidance on this.

CyberShadow:
OK, but isn't your addition now redundant? The README now says:

> To build the main site, run this:
> ```
> cd $R/dlang.org
> make -f posix.mak html
> ```
>
> ......................
> 
>To rebuild your dlang.org changes, use:
>```
>make -f posix.mak html
>```

zachthemystic:
You're right. I just missed it. Sorry. Closing pull. Thanks for your patience.

CyberShadow:
> You're right. I just missed it. Sorry.

No problem.

> Thanks for your patience.

You too.

PR #884, opened 2015-02-02
add release date for 2.067

MartinNowak:


JakobOvrum:
Although tangential: currently the changes for 2.067 [are deployed to dlang.org and prominently displayed on the changelog page](http://dlang.org/changelog), complete with a download link to a non-existing archive. Surely this is just an honest mistake. Ping @andralex

andralex:
@JakobOvrum that's okay as long as the date is not there or is in the future, right?

JakobOvrum:
> @JakobOvrum that's okay as long as the date is not there or is in the future, right?

Hardly a crisis, but if I was a D newbie I'd find it very confusing and might even find it unprofessional.

From a general usability perspective I think it's intuitive that the latest release and its changes are the most relevant for the changelog page, and thus should be on the top.

MartinNowak:
> From a general usability perspective I think it's intuitive that the latest release and its changes are the most relevant for the changelog page, and thus should be on the top.

A small note that these are upcoming changes and a jump-link to the current release would be nice.

PR #886, opened 2015-02-03
increase max-widths

aG0aep6G:
'The big grey bar' made the menu and the inter-column spacing wider, and
I'd forgotten to update these.

Both get 4.5em more for margin-left: 18em now, 13.5em before.

Hiding #news gets an additional 1em for margin-right: 21em now, 20em
before.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1137,7 +1137,7 @@ div#mobile-top, div#mobile-search, div#mobile-hamburger
 	display: none;
 }
 
-@media only screen and (max-width: 64em) { /* 1024px for default font-size 16 */
+@media only screen and (max-width: 69.5em) { /* 1112px for default font-size 16 */
 	
 	div#news{
 		display: none;
@@ -1149,7 +1149,7 @@ div#mobile-top, div#mobile-search, div#mobile-hamburger
 	
 }
 
-@media only screen and (max-width: 50em) { /* 800px for default font-size 16 */
+@media only screen and (max-width: 54.5em) { /* 872px for default font-size 16 */
 
 	/* Disable desktop-size UI elements */
 


andralex:
So what's the effect of this? Screenshots?

CyberShadow:
Do we need this? `1112px` is wider than my usual browser viewport, and I think it looks fine right now.

aG0aep6G:
> So what's the effect of this? Screenshots?

This is all about how wide the page is, or more specifically how wide the content area is, just before the layout changes kick in. Comparative screenshots below.

> Do we need this? 1112px is wider than my usual browser viewport, and I think it looks fine right now.

[Nick Sabalausky complained about it in the forum](http://forum.dlang.org/post/maoc51$odj$1@digitalmars.com), reminding me that I didn't update the max-widths. Here I just did the math so that #content has the same widths as before 'the big grey bar' when the switches occur.

I think the one that hides the menu can use the bump, especially on phobos pages where indentations, tables, and code snippets need horizontal space.

The one that hides #news seems to be less of an issue to me, because it only affects the home page, and there's nothing that really needs more space. On the other hand, the difference isn't that big. Unless a viewport width between 1025px and 1112px is common, we could get away with dropping #news earlier than before.

###### Just wide enough that #news is not hidden
Before:
screen shot 2015-02-04 at 15 42 35
After:
screen shot 2015-02-04 at 15 47 00

###### Just wide enough that the menu is not hidden (std.array)
By the way, we should disable hyphenation in code snippets.
Before:
screen shot 2015-02-04 at 16 11 24
After:
screen shot 2015-02-04 at 16 11 28

andralex:
@CyberShadow what do you think?

CyberShadow:
I can't judge completely objectively because I have a small conflict of interest. Merge this if you think the text in the first screenshot is too narrow (although it will be more narrow still on small-screen devices).

MartinNowak:
News is important on the main page, and this won't change the docs, so let's not do it, even though the content is indeed very narrow on my screen.
before:
before
after:
after

PR #887, opened 2015-02-03
don't set 'background' without setting 'color' too

aG0aep6G:


Orvid:
This would explicitly set the text color to black, is that what it currently is?

Personally, I don't think this is needed, as the only reason I can think of that it would be needed is if a user has a custom CSS file that sets the default text color to something other than black. (As far as I know, the major browser's all default to black text)

aG0aep6G:
Forgot to mention: [Nick Sabalausky complained about this](<http://forum.dlang.org/post/maoc51$odj$1@digitalmars.com>).

> This would explicitly set the text color to black, is that what it currently is?

If no user stylesheet interferes, it inherits from \<body> and that's black, yes.

> Personally, I don't think this is needed, as the only reason I can think of that it would be needed is if a user has a custom CSS file that sets the default text color to something other than black.

You just gave the reason why it's needed. A user stylesheet can set a background/color pair for \<input>. If we set only one of them, it may end up being white/white for example.

Orvid:
Ah, alright, I'd failed to think of the probability of them styling all `<input>` by default, which isn't an unreasonable thing for someone to do. With that in mind, I do agree that this is a needed change.

Hackerpilot:
Users of dark themes thank you.

andralex:
screenshots pliz

CyberShadow:
This change will be invisible to almost all users, and will fix things for the remaining ones. Thanks.

PR #889, opened 2015-02-05
Get documentation of the documentation generator in sync with the compiler

Geod24:
This commit:
- Turns $(RAW_DOLLAR) into $(DOLLAR), affects only ddoc.dd;
- Remove RAW_DOLLAR definition, as it's the same as DOLLAR;
- Update DDOC macro, which was missing the footer.
- Documents DDOC_ENUM_BASETYPE (emitted on enum's base type);
- Documents DDOC_PSUPER_SYMBOL (emitted on class' base type(s));


Geod24:
Notes:
DDOC_ENUM_BASETYPE was introduced in https://github.com/D-Programming-Language/dmd/pull/2185
DDOC_PSUPER_SYMBOL was introduced in https://github.com/D-Programming-Language/dmd/pull/1247
Revised link is based against most current compiler doc.c (however, most of the changes are quite old, blame it): https://github.com/D-Programming-Language/dmd/blob/a7a7095bbbc30021852295eb7ffd1077cc5587c4/src/doc.c#L143

For what it's worth, I believe DDOC_PSUPER_SYMBOL and DDOC_ENUM_BASETYPE should be defined to the same by default.

9rnsr:
LGTM.

PR #890, opened 2015-02-06
add @trusted usage guidelines

WalterBright:
also consolidated whitespace guidelines into one section

### DIFF
--- a/dstyle.dd
+++ b/dstyle.dd
@@ -25,6 +25,29 @@ $(UL
     $(LI Use spaces instead of hardware tabs.)
 
     $(LI Each indentation level will be four columns.)
+
+    $(LI Braces should be on their own line. There are a few exceptions to this
+         (such as when declaring lambda functions), but with any normal function
+         block or type definition, the braces should be on their own line.)
+
+---
+void func(int param)
+{
+    if (param < 0)
+    {
+        ...
+    }
+    else
+    {
+        ...
+    }
+}
+---
+
+    $(LI Lines have a soft limit of 80 characters and a hard limit of 120
+         characters. This means that most lines of code should be no longer than
+         80 characters long but that they $(I can) exceed 80 characters when
+         appropriate. However, they can $(I never) exceed 120 characters.)
 )
 
 $(H3 Naming Conventions)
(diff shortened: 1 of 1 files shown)


dstyle.dd
```diff
@@ -241,47 +264,33 @@ $(P
...
-         appropriate. However, they can $(I never) exceed 120 characters.)
- )
-
-$(P
-    We are not necessarily recommending that all code follow these two rules.
-    They're both likely to be controversial in any discussion on coding
-    standards. However, they are required in submissions to Phobos and other
-    official D source code.
- )
+    assert(src.length == dest.length);
```
Panke: That's only a safe interface, when it's impossible to corrupt memory from @safe code. That's not the case if the assert gets removed. I'd say the check must stay in every build.

dstyle.dd
```diff
@@ -241,47 +264,33 @@ $(P
...
-$(UL
-    $(LI Braces should be on their own line. There are a few exceptions to this
-         (such as when declaring lambda functions), but with any normal function
-         block or type definition, the braces should be on their own line.)
+    $(P Trusted functions must expose a safe interface. For example:)
 
--------------------------------
-void func(int param)
+---
+static void trustedMemcopy(T[] dest, T[] src) @trusted
```
schveiguy: T isn't defined.

schveiguy:
It should be stated that `@trusted` nested functions that are not static should not be allowed inside `@safe` functions, and that such functions should be able to stand on their own (i.e. they should pass without any consideration of the context of the `@safe` function they can only be called from).

dstyle.dd
```diff
@@ -25,6 +25,29 @@ $(UL
     $(LI Use spaces instead of hardware tabs.)
 
     $(LI Each indentation level will be four columns.)
+
+    $(LI Braces should be on their own line. There are a few exceptions to this
```
klickverbot: This would make the "additional requirements for Phobos" a general guideline. I disagree, and in any case this should be in a separate PR.

dstyle.dd
klickverbot: That example isn't even correct if T has a postblit.

dstyle.dd
```diff
@@ -241,47 +264,33 @@ $(P
...
+} 
+---
+
+    $(P A trusted function with an unsafe interface is one that requires review
+    of the caller's context to determine if it is safe or not. Such functions are
+    not allowed. For example:)
+
+---
+static U trustedCast(U, V)(V v) @trusted { return cast(U) v; }
+return trustedCast!RetType(result);
```
klickverbot: Second line is irrelevant to the point you are trying to make.

dstyle.dd
WalterBright: I removed the "additional requirements for Phobos" heading. There are no additional requirements for Phobos. The whole thing is a D style guide, and Phobos is expected to abide by it. Other projects can create their own style guide by picking and choosing what they wish - this one is the official one for Phobos.

klickverbot: Please let's not start that discussion in this PR. It's potentially controversial, and IIRC we had to do quite some convincing to get everybody to accept even the current Phobos-only guidelines. Taking those changes out would make it easier to merge the `@trusted` additions.

dstyle.dd
WalterBright: We've had this exact debate in other contexts several times now. I do not wish to reignite it here, and will only say using asserts in this manner is acceptable practice.

dstyle.dd
WalterBright: right, will fix.

dstyle.dd
WalterBright: If you look a the original, all I did was remove the redundant wording, I did not actually change the semantic content.

dstyle.dd
WalterBright: I prefer to leave it as is because it is not a contrived example. It is cut and paste from Phobos.

klickverbot: Fine by me. Maybe I've done too much math lately so that I favor minimal examples.

WalterBright: I understand, and thanks for your indulgence.

dstyle.dd
schveiguy: Walter, I agree with @Panke here:

`src[] = dest[]` will fail in every compilation mode except for when no bounds check is turned on. Assert has different rules for when it is compiled. I think the memcpy should be equivalent to the slice assign in terms of safety.

dstyle.dd
schveiguy: I think it might be better then to include more context in the example:

```D
RetType foo(RetType)() @safe
{
    int result;
    ...
    static U trustedCast(U, V)(V v) @trusted { return cast(U) v; }
    return trustedCast!RetType(result);
}
```



> It is cut and paste from Phobos.

Just so you know, we know this code is going to be fixed. Without any note about this, someone in the future will come along and try to change this.

dstyle.dd
```diff
@@ -241,47 +264,33 @@ $(P
...
+    of the caller's context to determine if it is safe or not. Such functions are
+    not allowed. For example:)
+
+---
+static U trustedCast(U, V)(V v) @trusted { return cast(U) v; }
+return trustedCast!RetType(result); 
+---
+
+    $(P This requires review of what $(CODE RetType) and $(CODE result) are to determine
+    whether it is actually safe or not.)
```
schveiguy: It may be useful to note that RetType and result are only defined within `@safe` code.

andralex:
In wake of @klickverbot's recent findings, please also add the guideline that a generic function should use deduction instead of specifying `@trusted`.

andralex:
@WalterBright ping

WalterBright:
yeah, I need to rewrite this in the light of the Trusted Manifesto results

MartinNowak:
Ping, this would be really helpful to point people to.

MartinNowak:
Will close this for now, until it gets updated.

PR #900, opened 2015-02-12
d.tag

CyberShadow:
https://github.com/D-Programming-Language/tools/pull/156#issuecomment-74075394

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -434,3 +434,13 @@ ${STABLE_DMD}:
 
 ${DUB}: ${DUB_DIR}/.cloned ${STABLE_DMD}
 	cd ${DUB_DIR} && DC=$(abspath ${STABLE_DMD}) ./build.sh
+
+################################################################################
+# Dman tags
+################################################################################
+
+chmgen : chmgen.d $(DMD)
+	$(DMD) -I${PHOBOS_DIR} -g chmgen.d
+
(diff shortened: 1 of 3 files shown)


MartinNowak:
Thanks, needs a rebase though.

CyberShadow:
Rebased.

posix.mak
```diff
@@ -434,3 +434,13 @@ ${STABLE_DMD}:
...
 	cd ${DUB_DIR} && DC=$(abspath ${STABLE_DMD}) ./build.sh
+
+################################################################################
+# Dman tags
+################################################################################
+
+chmgen : chmgen.d $(DMD)
+	$(DMD) -I${PHOBOS_DIR} -g chmgen.d
+
+d.tag : chmgen $(ALL_FILES)
```
MartinNowak: I'm having a race condition because this target depends on the druntime/phobos documentation without stating so.
Would be fine to just add druntime-release and phobos-release as dependencies, or does it also depend on the prerelease stuff?

MartinNowak: https://github.com/D-Programming-Language/dlang.org/pull/1031

CyberShadow: Well, strictly speaking it doesn't care, as it will follow links recursively. But the CHM probably shouldn't contain prerelease documentation, hmm.

PR #901, opened 2015-02-15
Fix DDOX page layout and improve building dpl-docs on win32

s-ludwig:
Updates layout.dt to the latest dlang.org layout and cleans the dpl-docs executable (and intermediate build files) on `make -f win32 clean`. It also currently forces a rebuild of dpl-docs to work around D-Programming-Language/dub#331. The Posix makefile still needs to receive some fixes/improvements, which is subject to a separate pull request.

### DIFF
--- a/css/ddox.css
+++ b/css/ddox.css
@@ -149,7 +149,7 @@ a.inherited:after { content: url(../images/ddox/inherited.png); padding-left: 3p
 
 #symbolSearchResults {
 	position: absolute;
-	left: 136pt;
+	left: 158pt;
 	width: 400pt;
 	margin: 0;
 	background-color: white;
--- a/dpl-docs/views/layout.dt
+++ b/dpl-docs/views/layout.dt
@@ -16,6 +16,7 @@ html(lang='en-US')
     link(rel='stylesheet', href='#{root_dir}css/style.css')
     link(rel='stylesheet', href='#{root_dir}css/print.css', media='print')
     link(rel='stylesheet', href='#{root_dir}css/cssmenu.css')
+    link(rel='stylesheet', href='#{root_dir}css/ddox.css')
     link(rel='stylesheet', href='https://maxcdn.bootstrapcdn.com/font-awesome/4.2.0/css/font-awesome.min.css')
     link(rel='shortcut icon', href='#{root_dir}favicon.ico')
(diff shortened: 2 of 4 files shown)


CyberShadow:
Thanks a lot! I'm going to try to build & update the site right now.

CyberShadow:
OK, I got it to work after nuking `~/.dub`. I don't know if that's a one-time thing after the fix, or if there's still something odd with dependecies.

Looks good: http://dlang.org/library/

s-ludwig:
Still those linker errors or failure to rebuild the updated layout? I noticed the linker issues now, too, but didn't have the time to dig deeper. It happened for me when the `${STABLE_DMD}` in posix.mak was used instead of the locally installed DMD.

CyberShadow:
Failure to rebuild the updated layout. I think I cleaned/deleted everything except `~/.dub` before pulling/building.

CyberShadow:
I think the linker errors I saw were related to [issue 12572](https://issues.dlang.org/show_bug.cgi?id=12572) and [my workaround](https://issues.dlang.org/show_bug.cgi?id=12572#c5), which involves modifying `/etc/dmd.conf`. Could it sometimes be invoking dmd with a custom configuration, ignoring the system-wide one?

s-ludwig:
For the specific case of libcurl, I've added a different workaround to DUB's build.sh (appending -lphobos2 a second time, so that the curl symbols are dragged out eventually). But it still failed for the "stable" one and there have also been unresolved linker errors for libevent, although that should not be affected (it's passed in the right order to the linker command).

But in any case, there is nothing in the make file or in DUB (that I know of) that would explicitly switch to a different configuration. But configuration selection seems to be an issue anyway: https://issues.dlang.org/show_bug.cgi?id=14180

MartinNowak:
I wonder how hard it would be to integrate the symbol search into the old documentation search?

s-ludwig:
You mean using the symbol table for the DDOX search for searching/navigating in the Ddoc documentation? That should be relatively straight forward. But we should generally probably aim for an AJAX based solution, because the symbol table currently takes up almost 500 kB when transferred uncompressed (about 50 kB gzipped).

CyberShadow:
Oh. Um... What about `d.tag`?

PR #902, opened 2015-02-16
Fix navigation styling for DDOX based documentation

s-ludwig:
This fix uses the new cssmenu code for handling the styling and collapsing of the root navigation items (std/core/etc) to match the rest of the site.


CyberShadow:
Hmm, doesn't look quite right:

http://dlang.k3.1azy.net/library/std/net/curl.html

s-ludwig:
I think it's again that build dependency issue, sorry for that! You can run `dub clean --all-packages` to remove the cached build files. I'll look into a fix for this next.

CyberShadow:
```
make: *** No rule to make target `/home/dlang/d/dlang.org/web/images/ddox/tree-item-closed.png', needed by `html'. Stop.
```

The makefile needs to be updated when static files are added or deleted...

s-ludwig:
Oh I love this redundant makefile mess... I've tested on Windows and didn't notice that. Shall I open a quick fixup pull request?

CyberShadow:
I got it, https://github.com/D-Programming-Language/dlang.org/pull/904

s-ludwig:
Okay, thanks!

CyberShadow:
Well, I've done a clean build (deleting anything that might take part of the build), and it still looks off:

http://dlang.k3.1azy.net/library/std/net/curl.html

Does it look different on your side?

s-ludwig:
Looks fine for me in Firefox and Opera:

dlang

The display of nested modules is still not ideal, but I didn't yet have the chance to improve that (not sure if that was an intended use case of cssmenu).

CyberShadow:
Ah, OK. That's the problem I meant, yeah. The DDoc menu only highlights the current module, without parent packages.

PR #904, opened 2015-02-16
posix.mak: Fix build (remove deleted DDox images)

CyberShadow:


MartinNowak:
Thx, how about wildcard instead?

CyberShadow:
I don't know, I guess they have their benefits and drawbacks. And what about win32.mak? No wildcards there.

PR #906, opened 2015-02-16
Add .htaccess for std.container redirect

CyberShadow:
See comments for details.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -1,0 +1,23 @@
+# dlang.org .htaccess file (local Apache configuration)
+
+# Enable mod_rewrite
+RewriteEngine On
+
+# Indicate that this directory maps to the root
+# directory of the domain.
+RewriteBase /
+
+# Phobos rewrites:
+
+# Package files are now generated to std_foo.html
+# instead of std_foo_package.html. This change was
+# done to preserve the links to std.algorithm,
+# std.range etc. after they were split into packages.
+# However, we need redirects for packages in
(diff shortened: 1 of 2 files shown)


posix.mak
```diff
@@ -124,7 +124,7 @@ STYLES=$(addsuffix .css, $(addprefix css/, \
 PRETTIFY=prettify/prettify.css prettify/prettify.js
 
 PREMADE=appendices.html articles.html fetch-issue-cnt.php	\
-howtos.html language-reference.html robots.txt process.php
+howtos.html language-reference.html robots.txt .htaccess process.php
```
MartinNowak: win32.mak?

.htaccess
```diff
@@ -0,0 +1,23 @@
...
+RewriteEngine On
+
+# Indicate that this directory maps to the root
+# directory of the domain.
+RewriteBase /
+
+# Phobos rewrites:
+
+# Package files are now generated to std_foo.html
+# instead of std_foo_package.html. This change was
```
MartinNowak: Now means in the future? What do I need to pull?

MartinNowak: Oh, 1 hour in the past.

posix.mak
CyberShadow: win32.mak can't build many parts of the website anyway, not nearly enough as needed to the point where you can start it in Apache and get a dlang.org instance.

PR #908, opened 2015-02-17
SeeAlso heading should appear as two words.

quickfur:
This is a pretty dirty hack, but I don't see any other way around it, since ddoc doesn't support any string manipulation functions.


CyberShadow:
I think this solution is actually idiomatic (for DDoc).

PR #909, opened 2015-02-17
fix tag generation for non-empty docRoot

MartinNowak:


CyberShadow:
I must be losing my mind. Did I not test this at all?

MartinNowak:
Just had the same thought with https://github.com/MartinNowak/dlang.org/commit/678c57c8ed521fe64a01acab8f456401fb57e26a. I tested it yesterday on my Windows VirtualBox, but apparently I tested something else.

PR #912, opened 2015-02-17
missing phobos space

burner:
see title and file changelog

### DIFF
--- a/dstyle.dd
+++ b/dstyle.dd
@@ -259,7 +259,7 @@ $(UL
 -------------------------------
 void func(int param)
 {
-    if(param < 0)
+    if (param < 0)
     {
         ...
     }


CyberShadow:
```
C:\Projects\Extern\D\phobos\std> git grep \bif( | wc -l
2386

C:\Projects\Extern\D\phobos\std> git grep "\bif (" | wc -l
6115
```

There's clearly a majority, but no consensus.

burner:
can you get the average date on both. from what I see if( is old style.
I'm really not willing to write more than 2 comments about this, because this is totally trivial.

Anyway, I got PRs blocked for not having the space at the write position.

CyberShadow:
Actually, the space-less syntax seems to be a growing trend:

CyberShadow:
> Anyway, I got PRs blocked for not having the space at the write position.

Took me way too long to figure out what a write position is.

CyberShadow:
Going to merge this on the basis of majority + to avoid more time wasted on trivia.

burner:
thanks, I didn't wanted to sound rude, this is just one of the newsgroup endless discussion things. 

Anyway, can you tell me what shell magic you used to create this:
git blame | grep " .... "

CyberShadow:
There is no shell. There is only D.

```d
import std.algorithm;
import std.conv;
import std.datetime;
import std.file;
import std.path;
import std.process;
import std.range;
import std.regex;
import std.stdio;
import std.string;

import ae.sys.file;
import ae.utils.array;
import ae.utils.digest;
import ae.utils.regex;
import ae.utils.time;

enum PHOBOS = `C:\Projects\Extern\D\phobos\std`;

void blame(string source, string target)
{
    spawnProcess(
        ["git", "blame", "-p", source],
        stdin, File(target, "wb"),stderr,
        null, Config.init, source.dirName
    ).wait();
}

void main()
{
    int[string] a, b;
    string t;

    foreach (de; dirEntries(PHOBOS, "*.d", SpanMode.depth))
    {
        writeln(de);
        string blamed = "blame/" ~ getDigestString!MD5(de.name);
        cached!blame(de.name, blamed);

        auto f = File(blamed, "rb");
        while (!f.eof)
        {
            f.readln(); // commit line
            while (!f.eof)
            {
                auto line = f.readln().chomp();
                if (line.startsWith("\t"))
                {
                    if (line.match(re!`\bif\(`))
                        a[t]++;
                    else
                    if (line.match(re!`\bif \(`))
                        b[t]++;
                    break;
                }
                else
                if (line.eat("author-time "))
                    t = SysTime(unixTimeToStdTime(line.to!int())).formatTime!"Y-m";
            }
        }
    }

    foreach (month; chain(a.byKey, b.byKey).array.sort().release.uniq)
        writefln("%s,%s,%s", month, a.get(month, 0), b.get(month, 0));
}
```

burner:
nice, you should post this somewhere. this will kill so many useless newsgroups post. well, if trolls were to take facts. 

anyway, thanks

PR #907, opened 2015-02-17
generate module menu

MartinNowak:
- automatically keeps tree in sync
- avoids std_navbar-VERSION.ddoc files for every alpha/beta/rc release

PR #913, opened 2015-02-17
fix druntime dependencies

MartinNowak:


CyberShadow:
Hmm, deja vu, is this split off from #907?

MartinNowak:
Yes, Andrei asked me to split #907.

PR #915, opened 2015-02-17
update style rules for deeply nested packages

MartinNowak:
- missing styles for nested packages without package.d
- added MODULE5 styles for core.sys.posix.sys.*

### DIFF
--- a/css/cssmenu.css.dd
+++ b/css/cssmenu.css.dd
@@ -144,18 +144,31 @@ body.have-javascript #cssmenu ul ul ul {
   color: $(submenu_highlight);
 }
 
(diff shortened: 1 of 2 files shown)


CyberShadow:
> missing styles for nested packages without package.d

Um, that's intentional?

CyberShadow:
Oh, never mind, I was thinking of something else. So it was broken but we didn't notice because all packages without package.d were top-level?

MartinNowak:
> So it was broken but we didn't notice because all packages without package.d were top-level?

Yes, something like that and `core.stdc.*.*` broke it.

PR #918, opened 2015-02-21
Fix PHOBOS macro

e10s:
This PR fixes link text generation, e.g. the 2nd paragraph of http://dlang.org/arrays.html#strings-unicode.


MartinNowak:
thx

PR #919, opened 2015-03-04
fix dependencies for modlist targets

MartinNowak:
- they require checked out sources


CyberShadow:
BTW, see IRC (you might need to set up your IRC client to notify you when someone says your name).

MartinNowak:
> BTW, see IRC (you might need to set up your IRC client to notify you when someone says your name).

There is a dim light on the bottom of my emacs, but sometimes I use that editor to code ;).

PR #920, opened 2015-03-04
bugzilla changelog for 2.067.0

MartinNowak:


MartinNowak:
I generated the list of fixed bugzilla entries using [changed](https://github.com/D-Programming-Language/tools/blob/61a1f4fd3114b434e3825bedd65b9a4fa7425031/changed.d) with `--start-date=2014-08-18`, then found all the duplicate bugzilla entries using this.
```sh
grep -h BUGZILLA ../dlang.org/changelog.dd changelog.txt | sed 's|^[ \t]*||' | sort | uniq -cd | sed -n 's|.*\(BUGZILLA [0-9]*\).*|\1|p' > duplicates
```
I removed the duplicates from changelog.txt using this.
```sh
for line in (cat duplicates)
    sed -i "/$line/d" changelog.txt 
end
```
Finally inserted the changelog.txt manually into dlang.org's changelog.dd, rebuild dlang.org and looked at the changelog.html page.

MartinNowak:
We should hide those lists behind some sort of unfold button.

CyberShadow:
> We should hide those lists behind some sort of unfold button.

Then they wouldn't be searchable.

MartinNowak:
> Then they wouldn't be searchable.

Ah right. I still find this page fairly noisy, there should be clear bullet points or so.

MartinNowak:
This looks great https://golang.org/doc/go1.3.

braddr:
I don't have a good recommendation on how to do it, but I think the changelog should omit bugs that were introduced and fixed during the release cycle.  Those bugs are of almost zero value to list for anyone other than a branch follower, and those people are already aware.

ntrel:
> the changelog should omit bugs that were introduced and fixed during the release cycle

+1. Is there some way in future that we can mark this kind of bug report as a 'temporary' bug when it gets closed? Maybe with a new keyword, then the search query can be restricted to those that don't have that keyword.

I also wonder if we could use another keyword for marking an issue as needing an explanation in the changelog.

MartinNowak:
More manual annotations don't solve the problem because nobody is keen to add them.

I think we could start to manually curate a list of prominent fixed bugs, with a link to the full list in Bugzilla. The current list is too long anyhow and people personally affected by a particular bug should add themselves to it's CC list to get notified.

MartinNowak:
Updated, can we please merge this?

CyberShadow:
I don't see why not, but maybe you also want to update the release date?

MartinNowak:
see #933

MartinNowak:
thx

PR #921, opened 2015-03-04
add changelog entry for GC improvements

MartinNowak:


MartinNowak:
Preview: https://dlang.dawg.eu/changelog.html#faster-gc-1
Could use some help with the wording and the image overlap.
If the screen is too narrow the image will simply go over the border, looks horrible on my mobile.
screen

CyberShadow:
`style="max-width: 100%"` will fix the image width/overlap issue.

CyberShadow:
BTW you can preview mobile stuff in Firefox with Ctrl+Shift+M.

MartinNowak:
Nice trick, works like a charm.
screen

rainers:
I think you should give some information (e.g. in the title) on what platform the benchmark is done. I just happen to have generated this for Win32 showing more dramatic changes: [GC Benchmark Win32](https://gist.github.com/rainers/6cdf73b48837defb9f88/raw/20840cda4c3022d8593867ce3d3435a39ffefeae/gcbench32.png)
The slist and tree2 are way higher in dmd2.066 than shown, but that's probably due to false pointers. 

BTW: I had to change the vdparser test to make it work for 2.066, should we commit that?

MartinNowak:
> I had to change the vdparser test to make it work for 2.066, should we commit that?

Yes please, worked for me on 2.066.1, but the conmsg benchmark didn't.

MartinNowak:
> The slist and tree2 are way higher in dmd2.066 than shown, but that's probably due to false pointers.

Yes, the slist and dlist benchmarks are quite unstable. Sometime the GC runs with a bigger heap (22MB instead of 16MB or something like that) making the tests run much faster. I wouldn't want to include those results.

MartinNowak:
updated

rainers:
> Yes please, worked for me on 2.066.1, but the conmsg benchmark didn't.

https://github.com/D-Programming-Language/druntime/pull/1185

rainers:
> updated

It still doesn't show the architecture. Linux/64 in the title should be enough.

MartinNowak:
> Linux/64 in the title should be enough.

Updated.
Wondering though, who still uses 32-bit on linux. Might consume less memory, but only comes with half the registers.

changelog.dd
```diff
@@ -45,6 +46,28 @@ $(LI $(LNAME2 asm-attributes, $(D asm) statements can now be used in $(D pure),
 
 $(BUGSTITLE Library Changes,
 
+$(LI $(LNAME2 faster-gc-1, $(BIG The garbage collector is much faster.))
```
rainers: The performance gain still depends on the application, not all benchmarks show a notable improvement. So I'd be a little more humble and remove the "BIG" (being the first item makes this change sufficiently prominent) and rather make the second line the title, e.g. "GC optimizations".

rainers:
> Updated.

Thanks, looks good now.

> Wondering though, who still uses 32-bit on linux. Might consume less memory, but only comes with half the registers.

The win32/win64 benchmarks show very different results for me, with the Win32 versions sometimes much faster, sometimes slower:
```
                   Win32    Win64
 bulk             1.612 s  3.794 s
 resize           1.547 s  1.742 s
 string           0.052 s  0.047 s
 conalloc         0.153 s  0.205 s
 conappend        0.027 s  0.023 s
 concpu           0.234 s  0.255 s
 conmsg           1.166 s  1.106 s
 dlist            3.444 s  3.030 s
 huge_single      0.023 s  0.018 s
 rand_large       0.931 s  0.507 s
 rand_small       0.854 s  0.749 s
 slist            2.911 s  2.942 s
 testgc3          1.880 s  2.478 s
 tree1            0.968 s  0.929 s
 tree2            1.861 s  1.793 s
 vdparser         3.629 s  3.022 s
 words            1.806 s  1.849 s
```
BTW: the "words" benchmark is consistently slower with 2.067, this seems to be caused by std.array.split being made unicode correct. This might also have affected other benchmarks, e.g. slist and dlist.

MartinNowak:
Yeah it's hard to compare different compiler versions, because a lot of small things changed.
The slist and dlist benchmark should be fine though, because they hardly use phobos. They are unstable because the heap size sometimes differs (because of false pointers or something).

MartinNowak:
I weakened the statement a little, but it's still quite something we achieved here (especially for real-world apps) and we shouldn't be shy to say that.
http://forum.dlang.org/post/uuowzztfsyfafwfnsxol@forum.dlang.org

Also tagged all pull requests with GC and linked to the full list of changes.

rainers:
Looks better.now.

PR #923, opened 2015-03-09
Issue 13116: should reject returning `this` from a ref class member function

9rnsr:
https://issues.dlang.org/show_bug.cgi?id=13116


changelog.dd
```diff
@@ -58,6 +59,66 @@ $(LI $(LNAME2 package-protection, $(D package) protection attribute can be optio
     )
 )
 
+$(LI $(LNAME2 rvalue-class-this-super, $(D this) and $(D super) expressions in class method are changed to rvalue)
```
MartinNowak: Should add something like, `$(RED Breaking Change: )` in front of this.

9rnsr: Done.

MartinNowak:
Thanks anyway, returning this remains an error btw, because it escapes a local reference.

9rnsr:
I still think that passing `this` to a ref parameter should be an error immediately...

PR #925, opened 2015-03-10
Issue 14258 - SIGSEGV in dpldocs

s-ludwig:
This pull upgrades to the latest version of DDOX, which fixes the crash. The root cause was a missing termination condition for recursive macro instantiations.


CyberShadow:
Thanks!

PR #926, opened 2015-03-13
fix menu highlighting for dpl-docs and reference

MartinNowak:


CyberShadow:
LGTM, uploading.

MartinNowak:
We should contact @braddr about http://downloads.dlang.org/.

MartinNowak:
Actually it's here https://github.com/braddr/downloads.dlang.org/blob/master/src/gen_index.d.

PR #927, opened 2015-03-14
integrate news column in menu

MartinNowak:


MartinNowak:
Just an idea, this works nicer for narrow screens.
You can see a preview here https://dlang.dawg.eu/.
newsinmenu

MartinNowak:
BTW, would be nice if the forum iframes were available over https.

CyberShadow:
> BTW, would be nice if the forum iframes were available over https.

Both dlang.org and forum.dlang.org use self-signed certificates. The iframes will probably work via https if you first go to forum.dlang.org and accept the self-signed certificate.

CyberShadow:
Let's please not make the navigation *even wider* for all pages.

CyberShadow:
Is this necessary, considering the average viewport size? Over 93% of people use a resolution over 1280px wide according to [this](http://www.w3schools.com/browsers/browsers_display.asp).

MartinNowak:
Lots of programmers use their monitor in vertical mode. We could do this with a media query, depending on the available width.

CyberShadow:
OK. The other side of the coin is that this puts the news "below the fold", and I'm not sure if additional media queries are worth it from the complexity perspective. But we can get a third opinion.

MartinNowak:
Let's ask @aG0aep6G who did #886.

MartinNowak:
It looks a bit weird, clutters the menu and makes it fatter and generally looks a bit unbalanced.
So I'll close this, still would like to see a good idea how to not squeeze the main content.
In general it seems like we should clean up the index, the example is horrible and the important main points are drown in details.
A brief paragraph and fanouts to http://dlang.org/convenience, http://dlang.org/power, and http://dlang.org/efficiency might work much better.

PR #629, opened 2014-08-16
Dynamic/static arrays: Fix some doc errors

AndrewGodfrey:
Split out the less controversial fixes from https://github.com/D-Programming-Language/dlang.org/pull/623:

Fix some documentation errors, around dynamic array semantics and static array syntax.
- Add a note about the need for assumeSafeAppend in order to regrow a dynamic array.
- Add examples for static arrays as lvalues (for copying and setting).
- Soften an over-strong claim in ctod.dd.
- Also tweak some heading levels and such.

PR #928, opened 2015-03-15
Rebase PR #629 ("Dynamic/static arrays: Fix some doc errors")

AndrewGodfrey:
https://github.com/D-Programming-Language/dlang.org/pull/629

### DIFF
--- a/arrays.dd
+++ b/arrays.dd
@@ -12,7 +12,7 @@ $(SPEC_S Arrays,
         $(TROW $(ARGS $(I type)[$(I type)]), $(ARGS $(DDLINK hash-map, Associative Arrays, Associative arrays)))
     )
 
-$(H4 $(LNAME2 pointers, Pointers))
+$(H3 $(LNAME2 pointers, Pointers))
 
 ---------
 int* p;
@@ -29,7 +29,7 @@ int* p;
         and reference types.
         )
 
-$(H4 $(LNAME2 static-arrays, Static Arrays))
+$(H3 $(LNAME2 static-arrays, Static Arrays))
 
 ---------
 int[3] s;
@@ -56,7 +56,7 @@ int[3] s;
         )
 
 
-$(H4 $(LNAME2 dynamic-arrays, Dynamic Arrays))
+$(H3 $(LNAME2 dynamic-arrays, Dynamic Arrays))
 
 ---------
 int[] a;
@@ -152,8 +152,8 @@ p = q;     // p points to the same thing q does.
 p = s.ptr; // p points to the first element of the array s.
 p = a.ptr; // p points to the first element of the array a.
 
-s = ...;   // error, since s is a compiled in static
-           // reference to an array.
+s = t;     // Copies the elements of t to the elements of s.
+s = a;     // If a.length == 3, copies the elements of a to s. Otherwise, runtime error.
 
 a = p;     // error, since the length of the array pointed
            // to by p is unknown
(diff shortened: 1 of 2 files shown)


arrays.dd
```diff
@@ -218,9 +218,12 @@ $(H2 $(LNAME2 array-copying, Array Copying))
...
 int[3] s;
 int[3] t;
+int[] d = new int[3];
 
-s[] = t;           // the 3 elements of t[3] are copied into s[3]
-s[] = t[];         // the 3 elements of t[3] are copied into s[3]
+d[] = t;           // the 3 elements of t are copied into d (unlike "d = t")
+s[] = t;           // the 3 elements of t are copied into s
+s[] = t[];         // the 3 elements of t are copied into s
+s = t;             // also works (because s is a static array)
```
MartinNowak: because s __and__ t are static arrays of the same length

arrays.dd
```diff
@@ -263,9 +266,12 @@ $(H2 $(LNAME2 array-setting, Array Setting))
 
 ---------
 int[3] s;
+int[] d = new int[3];
 int* p;
 
+d[] = 3;     // same as d[0] = 3, d[1] = 3, d[2] = 3
 s[] = 3;     // same as s[0] = 3, s[1] = 3, s[2] = 3
+s = 3;       // also works (because s is a static array)
```
MartinNowak: Better not mention that, it might get deprecated soon.

arrays.dd
```diff
@@ -218,9 +218,12 @@ $(H2 $(LNAME2 array-copying, Array Copying))
 ---------
 int[3] s;
 int[3] t;
+int[] d = new int[3];
```
MartinNowak: Isn't the preferred syntax `new int[](3)` nowadays, because it also support multi-dimensional arrays, like `new int[][](3, 5)`. That syntax also looks a lot as if you'd allocate a static array on the heap, but it allocates a dynamic array of length 3.

JakobOvrum:
For the future: when rebasing a PR you don't need to file a brand new one; instead do a force push to the branch for the old PR: `git push -f myfork arrayDocFixes`

arrays.dd
AndrewGodfrey: That's true, but not unique to "s = t;" - it's also a requirement of "s[] = t;".
What I'm trying to point out is that "d = t;" would not mean the same thing.

arrays.dd
AndrewGodfrey: What I was trying to do here, is reduce the EXTREME confusion caused by the current documentation combined with the current semantics. I have noticed that the experienced members of this community don't generally see this problem anymore. I respectfully suggest that this problem - of the experienced members minimizing glaring inconsistencies in 'basic functionality' - is D's biggest threat. Here I'm focusing on the confusing differences in semantics between static and dynamic arrays.

The other areas I see this same threat are:
1) "dynamic arrays" which change their identity when you grow them.
2) the semantics of utf8 strings.

arrays.dd
AndrewGodfrey: I had no idea. I prefer that too.

AndrewGodfrey:
> For the future: when rebasing a PR you don't need to file a brand new one;
Even when the PR is closed? I don't know GitHub and couldn't find this in the help.

CyberShadow:
This is breaking the PDF build:
```

[206]
! You can't use `macro parameter character #' in horizontal mode.
l.11279 ...ful use of <a href="phobos/object.html#
                                                  assumeSafeAppend">\lstinli...
```

CyberShadow:
@AndrewGodfrey Do you plan to continue this PR?

To fix the PDF build, I think simply moving the macro declaration to html.ddoc should do.

AndrewGodfrey:
I may have time soon to fix that. But my impression is that new objections were raised blocking acceptance. Is that not the case?
________________________________

CyberShadow:
I haven't looked at the content of the PR. You can close it if you think that is the appropriate decision.

PR #929, opened 2015-03-16
mostly fix Issue 14280 - Links to command line tools have disappeared fr...

aG0aep6G:
...om navigation

Not considered here: overview.html and spec.html.

Adding a new menu section: Compilers & Tools. Adding dmd-windows.html,
rdmd.html, windbg.html, and htod.html. Also adding links to gdc and ldc.
dmd-linux.html, dmd-freebsd.html, and dmd-osx.html can be reached through
dmd-windows.html.

Adding const-faq.html, rationale.html, code_coverage.html, and everything
from SUBNAV_ARTICLES to Books & Articles.

Adding http://wiki.dlang.org/Editors, http://wiki.dlang.org/IDEs, and
glossary.html to Resources.

http://digitalmars.com/ctg/optlink.html is already linked from
dmd-windows.html, good enough.

http://digitalmars.com/ctg/trace.html is already linked from dmd-*.html,
good enough.

Leaving debugger.html dead.

Leaving features2.html and ascii-table.html orphaned, because they don't
add value.

Leaving appendices.html and memory.html orphaned, because they're just
redirects.

----

https://issues.dlang.org/show_bug.cgi?id=14280

A shot of the new menu section:
iconscreenshot1

### DIFF
--- a/doc.ddoc
+++ b/doc.ddoc
@@ -66,27 +66,54 @@ $(DIVID cssmenu, $(UL
         http://twitter.com/search?q=%23dlang, Twitter,
         http://digitalmars.com/d/dlinks.html, More Links
       )
+    $(MENU_W_SUBMENU Compilers $(AMP) Tools)
+      $(SUBMENU
+        dmd-windows.html, dmd $(NDASH) reference compiler,
+        http://gdcproject.org, gdc $(NDASH) gcc-based compiler,
+        http://wiki.dlang.org/LDC, ldc $(NDASH) LLVM-based compiler,
+        rdmd.html, rdmd $(NDASH) pseudo interpreter,
+        windbg.html, windbg $(NDASH) debugger,
+        htod.html, htod $(NDASH) .h to .d
+      )
     $(MENU_W_SUBMENU Books $(AMP) Articles)
       $(SUBMENU
         http://ddili.org/ders/d.en/index.html, Online Book (free),
         http://wiki.dlang.org/Books, More Books,
         howtos.html, How-tos,
         faq.html, FAQ,
+        const-faq.html, const(FAQ),
         comparison.html, Feature Overview,
         d-floating-point.html, Floating Point,
         wc.html, Example: wc,
         warnings.html, Warnings,
+        rationale.html, Rationale,
         builtin.html, Builtin Rationale,
         ctod.html, C to D,
         cpptod.html, C++ to D,
-        pretod.html, C Preprocessor vs D
+        pretod.html, C Preprocessor vs D,
+        code_coverage.html, Code coverage analysis,
+        exception-safe.html, Exception Safety,
+        hijack.html, Hijacking,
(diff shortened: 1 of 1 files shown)


aG0aep6G:
This has a little issue: on `dmd-{linux,freebsd,osx}.html` the Compiler & Tools section isn't expanded. That's because the URL doesn't match, of course. I'm not sure how to tackle this.

doc.ddoc
```diff
@@ -66,27 +66,54 @@ $(DIVID cssmenu, $(UL
         http://twitter.com/search?q=%23dlang, Twitter,
         http://digitalmars.com/d/dlinks.html, More Links
       )
+    $(MENU_W_SUBMENU Compilers $(AMP) Tools)
+      $(SUBMENU
+        dmd-windows.html, dmd $(NDASH) reference compiler,
+        http://gdcproject.org, gdc $(NDASH) gcc-based compiler,
+        http://wiki.dlang.org/LDC, ldc $(NDASH) LLVM-based compiler,
+        rdmd.html, rdmd $(NDASH) pseudo interpreter,
+        windbg.html, windbg $(NDASH) debugger,
```
CyberShadow: Using windbg is now outdated and should be expunged. On Windows, the canonical way to debug programs is now to use Visual Studio (with cv2pdb or Mago for 32-bit).

doc.ddoc
```diff
@@ -66,27 +66,54 @@ $(DIVID cssmenu, $(UL
         http://twitter.com/search?q=%23dlang, Twitter,
         http://digitalmars.com/d/dlinks.html, More Links
       )
+    $(MENU_W_SUBMENU Compilers $(AMP) Tools)
+      $(SUBMENU
+        dmd-windows.html, dmd $(NDASH) reference compiler,
+        http://gdcproject.org, gdc $(NDASH) gcc-based compiler,
+        http://wiki.dlang.org/LDC, ldc $(NDASH) LLVM-based compiler,
+        rdmd.html, rdmd $(NDASH) pseudo interpreter,
```
CyberShadow: rdmd is more important as a build tool than a pseudo interpreter (for which `dmd -run` is a closer fit).

CyberShadow:
In general we need to look closely at all the "missing" pages. Many of them were removed purposefully, because they are now outdated. Adding them back may not be a good move.

aG0aep6G:
> In general we need to look closely at all the "missing" pages. Many of them were removed purposefully, because they are now outdated. Adding them back may not be a good move.

After a fuck-up like this, I think the emergency action should be to bring stuff back. Weeding out can happen later.

andralex:
OK I'm fine with the additions. Thanks.

MartinNowak:
Only listing dmd-windows doesn't make sense.
I think those pages should become clear installation instruction and should be linked to from the download page, right after each OS.
And the rest should be submenus of Download.

PR #930, opened 2015-03-17
This comment is stale...

smandy:


CyberShadow:
It should probably be replaced with other DConf news?

@andralex

MartinNowak:
We should keep a link to dconf, maybe something like #512.

smandy:
either way - having an 'open till XXX' when 'XXX' is already in the past is bad. Needs to be cleaned up one way or the other

smandy:
addressed this (hopefully) in PR #931

MartinNowak:
Thx

PR #934, opened 2015-03-24
remove download links from changelog

MartinNowak:
- we have a link on the download page to the release archive
  and the one in the changelog always points to the bundled zip
  which is the least suitable download

- use permalink anchors instead

- also cleanup the anchor names (#2.066.1 instead of #new2_066.1)

### DIFF
--- a/changelog.dd
+++ b/changelog.dd
@@ -10711,7 +10711,7 @@ Macros:
 	<div class="version">
 	$(B $(LARGE <a name="new2_$1">
 	  Version
-	  <a HREF="http://ftp.digitalmars.com/dmd.2.$1.zip" title="D 2.$1">D 2.$1</a>
+	  <a HREF="http://downloads.dlang.org/releases/2.x/2.$1" title="D 2.$1">D 2.$1</a>
 	))
 	$(SMALL $(I $2, $3))
 	$5


braddr:
Don't remove the url to the release, change it to the downloads.dlang.org/releases/2.x/<version>/ directory.  It's a lot more useful to have a direct link there than having to hunt it down.

MartinNowak:
updated @braddr

MartinNowak:
I also made a pull to updated the downloads.dlang.org pages.
https://github.com/braddr/downloads.dlang.org/pull/1

CyberShadow:
Are the files mirrored all the way to the last version in the changelog?

MartinNowak:
Yes, they are.

CyberShadow:
Looks like it.

MartinNowak:
Nice, I still can't update dlang.org, would you take the honor?

CyberShadow:
OK, was this the last pull?

MartinNowak:
Yes, might want to look at https://dlang.dawg.eu, if you can find something else.

MartinNowak:
Wait, we should put a note on the main page.

CyberShadow:
I see .asc files in the download dir, but where's the public key?

MartinNowak:
updated index #936

MartinNowak:
I'll make a page
https://dlang.dawg.eu/downloads/d-keyring.gpg

CyberShadow:
> https://dlang.dawg.eu/downloads/d-keyring.gpg

Who is David Soria Parra <dsp@php.net> ?

MartinNowak:
Our other release manager, a coworker of Andrei who brings some experience from PHP and mercurial. Also see DIP75.

PR #935, opened 2015-03-24
fix links on download page

MartinNowak:


MartinNowak:
And another fix @CyberShadow, @braddr.

CyberShadow:
Good to see the year-in-URL scheme go away.

MartinNowak:
Yeah, can you also look at #934.

PR #937, opened 2015-03-24
Fix issue 14326 - Fixes JS error and missing syntax highlighting.

s-ludwig:
Was caused by missing imports of jQuery and the Prettify CSS. Also updates the dependencies to the latest vibe.d RC to avoid compilation issues with DMD 2.067.0 final.

https://issues.dlang.org/show_bug.cgi?id=14326


CyberShadow:
Let's give it a go

CyberShadow:
Looks good, but needs to be updated with 2.067 as the current version.

PR #938, opened 2015-03-24
add gpg signatures to downloads

MartinNowak:
- add gpg_keys page
- add download of d keyring
- link to gpg_keys from download page
- show sig download button only on hover


CyberShadow:
win32.mak too please?

MartinNowak:
See https://dlang.dawg.eu/download.html and https://dlang.dawg.eu/gpg_keys.html.

MartinNowak:
I thought we agree on dropping that, but OK.

CyberShadow:
The on-hover button shifts other buttons around. Not good.

On-hover effects should never move other things around.

MartinNowak:
Update for win32.mak

CyberShadow:
Maybe put the sig button under the main button?

MartinNowak:
Yes, makes the buttons pretty busy.

> Maybe put the sig button under the main button?

Can try that, will be difficult to align, and adds a lot of noise.

CyberShadow:
> Can try that, will be difficult to align, and adds a lot of noise.

I think I know how to do it without these problems.

MartinNowak:
That's really ugly.
sig

MartinNowak:
Well not that terrible, but I hope you have better idea.

CyberShadow:
https://github.com/D-Programming-Language/dlang.org/pull/939

wamp is acting up so can't put up a live preview ATM.

PR #939, opened 2015-03-24
Nicer .sig download buttons

CyberShadow:


MartinNowak:
Pretty nice.

css/style.css
```diff
@@ -28,17 +28,23 @@ body::before
 }
 
 /* show signature download on hover */
-span.sig_btn > a:last-child
+span.sig_btn
 {
+    position: relative;
```
MartinNowak: That's needed so position absolute in the child works, right?
I never understood that crap.

css/style.css
```diff
@@ -28,17 +28,23 @@ body::before
...
+    height: 4.4em;
```
MartinNowak: How about a `<br />` between the two buttons and changing opacity instead?

css/style.css
CyberShadow: Yeah.

css/style.css
CyberShadow: Dunno, I think opacity has a bigger performance impact on rendering. I'll check if I can avoid the hard-coded height.

CyberShadow: I think you meant visibility, not opacity

MartinNowak: The download page isn't that performance sensitive ;), but yeah avoiding the hard-coded height would be nice.

CyberShadow:
Ditched the hard-coded height but for some reason now I need a line-height. \*shrug\*

css/style.css
MartinNowak: > I think you meant visibility, not opacity

The thing where you don't change the layout. Apparently [`visibility`](https://css-tricks.com/almanac/properties/v/visibility/) can do that as well.

MartinNowak:
Yeah, had the same problem, because the buttons aren't blocks, I arrived at this dead-end.
Line-height is fine, because the buttons are fixed size as well (should use em though).

```diff
diff --git a/css/style.css b/css/style.css
index 864f736..2ebc682 100644
--- a/css/style.css
+++ b/css/style.css
@@ -1,4 +1,4 @@
-html
+ghtml
 {
 	height: 100%; /* of the viewport */
 }
@@ -30,21 +30,23 @@ body::before
 /* show signature download on hover */
 span.sig_btn
 {
-    position: relative;
-    height: 4.4em;
     display: inline-block;
 }
 
-span.sig_btn > a:first-child
+span.sig_btn > a
 {
-    position: absolute;
-    display: none;
-    bottom: 0;
+    display: block;
+    text-align: center;
 }
 
-span.sig_btn:hover > a:first-child
+span.sig_btn > a:last-child
 {
-    display: inline-block;
+    visibility: hidden;
+}
+
+span.sig_btn:hover > a:last-child
+{
+    visibility: visible;
 }
 
 a.btn
diff --git a/download.dd b/download.dd
index ff68e10..e97436f 100644
--- a/download.dd
+++ b/download.dd
@@ -70,7 +70,7 @@ Macros:
 	UBUNTU=$(LOGO ubuntu, Ubuntu)
 	WINDOWS=$(LOGO windows, Windows)
 
-    SBTN=$(SPANC sig_btn,$(BTN $1.asc,sig)$(BTN $1,$+))
+    SBTN=$(SPANC sig_btn,$(BTN $1,$+) $(BTN $1.asc,sig))
     BTN=<a href="$1" class="btn">$+</a>
     H3I=<h3 class="download">$0</h3>
     DLSITE=http://downloads.dlang.org/releases/2.x/$(DMDV2)/$0
```
sig

PR #945, opened 2015-03-26
Add std.range.enumerate and std.exception.handle to the changelog

JakobOvrum:
Note: after a couple of hours of trying I didn't manage to get the docs building so these are untested.

First I moved dlang.org alongside the dmd repository (even though it seems to download DMD anyway??), then I fixed DOS newlines in dub's build.sh, then I changed STABLE_DMD_VER in posix.mak to 2.067.0 because I got DMD errors about unrecognized `return`, then curl complained because it tried to look for it in the 2014 subdirectory... am I doing something basic horribly wrong?


JakobOvrum:
Hardcoding 2014 to 2015 in posix.mak on top of the other changes seems to have fixed it. The patch looks good in the HTML output.

So, do I have to run a local webserver now to view locally generated documentation with the proper styling?

CyberShadow:
I'm not sure why you're getting these problems. What OS is this? Try cloning dlang.org to a new empty directory, then running `make -f posix.mak`.

> then I fixed DOS newlines in dub's build.sh

That's strange, I don't see DOS newlines at any point in that file's history. Maybe git is converting newlines for some reason?

PR #947, opened 2015-03-29
changelog updates

MartinNowak:
- mention sealed references
- mention fixed with statement
- fix permalinks for enumerate and handle
- add missing paren


MartinNowak:
Thanks for updating the website right away, I finally got the ssh access.

PR #948, opened 2015-03-31
Nicer front page notices

CyberShadow:
The notices on the front page look a bit ad-hoc at the moment.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -370,6 +370,17 @@ a.anchor:hover:after
 	font-style: normal; /* ... and should not be italic */
 }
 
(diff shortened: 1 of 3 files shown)


CyberShadow:
It doesn't really fit into the rest of the page's style, but this is the best I could think of for the moment. If anyone has better design ideas they are welcome to share them.

MartinNowak:
fancy

MartinNowak:
Square corners, little less drop shadow and nicer colors would make it even better.

MartinNowak:
see #949

PR #949, opened 2015-03-31
improve notices

MartinNowak:
- square blocks with less shadow
- turn whole block into clickable link
- use dark background + accent color

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -370,15 +370,22 @@ a.anchor:hover:after
 	font-style: normal; /* ... and should not be italic */
 }
 
-div.notice
+a.notice
 {
-	border: 1px solid #CCC;
-	border-radius: 0.5em;
+	display: block;
 	padding: 0.5em;
 	text-align: center;
 	font-weight: bold;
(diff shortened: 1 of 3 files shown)


MartinNowak:
notice

CyberShadow:
That looks very jarring on my screen at least.

MartinNowak:
https://dlang.dawg.eu/

MartinNowak:
> That looks very jarring on my screen at least.

It just looks about right on mine and it's also supposed to be noticed.

MartinNowak:
Also the first time, that the website has a clear center of focus.

MartinNowak:
Two alternatives.
`white`
notice_white
[`Red A700 (#D50000)`](http://www.google.com/design/spec/style/color.html#color-color-palette)
notice_red_a700

CyberShadow:
White is much nicer.

Also, I just remembered that red-on-black is almost unreadable for people with protanopia, so we should avoid that.

CyberShadow:
#950

PR #950, opened 2015-04-01
Notice tweaks

CyberShadow:


MartinNowak:
Fine with me, the website won't win a design award anyhow.

PR #952, opened 2015-04-02
move dpl_rewrite rule to dpl specific .htaccess

MartinNowak:


dpl_htaccess
```diff
@@ -0,0 +1,12 @@
+# dlang.org .htaccess file (local Apache configuration)
+
+# Enable mod_rewrite
+RewriteEngine On
+
+# Indicate that this directory maps to the root
+# directory of the domain.
+RewriteBase /
```
CyberShadow: This is wrong, isn't it?

MartinNowak: I guess so, I hoped you'd know what goes here.

MartinNowak: I can test it locally though, just installed an apache.

CyberShadow: `/library/` ?

CyberShadow: I think `/library/` ought to work, plus it should also work for library-prerelease despite that the map file only has /library/.

MartinNowak: No, the mapfile also has library-prerelease, I used `sort`, so it's somewhere in the middle.
http://dlang.org/.dpl_rewrite_map.txt

MartinNowak: I think we're only doing absolute redirects and might not need `RewriteBase` at all, will test.

CyberShadow: Ideally the map file shouldn't have the library and library-prerelease prefixes...

CyberShadow: > I think we're only doing absolute redirects and might not need RewriteBase at all, will test.

We *definitely* need RewriteBase because the dlang.org server does not have a sensible RewriteBase for some reason.

MartinNowak: Works without RewriteBase.

CyberShadow: On digitalmars.com?

MartinNowak: On localhost, can't test the rewrite on digitalmars.

MartinNowak: Both the key and the value in the map are absolute, so we should be fine here.

CyberShadow: As far as I understand you need RewriteBase there, because the default is broken.

And, assuming that the /library/ and /library-prerelease/ rules are meaningfully different, we can't hard-code one here, because  this file will be copied to both directories.

CyberShadow: > Both the key and the value in the map are absolute, so we should be fine here.

OK... so it should work in theory... we should test before deploying this.

CyberShadow: BTW another thing, shouldn't we redirect only 404s? What if a new identifier appears that by the new convention has the same URL as a redirect?

MartinNowak: I'll make 2 files.

MartinNowak:
Done, I might mention that we probably just spend more energy creating this patch than it will ever save, not to mention our own time ;).

CyberShadow:
Yours mostly (I just grumble at everything) ;)

On that note, what of 

> BTW another thing, shouldn't we redirect only 404s? What if a new identifier appears that by the new convention has the same URL as a redirect?

dpl_htaccess
MartinNowak: > BTW another thing, shouldn't we redirect only 404s? What if a new identifier appears that by the new convention has the same URL as a redirect?

I created those redirects using the current modules. One would need to delete something and readd it under the same name (possibly varying the caps).
I guess then it would still be the correct page, but let me quickly think about this.

MartinNowak: If someone removed `OutputStream` and added `outputStream` or `output_stream` (which won't happen during the time we keep those redirects) it wouldn't be too big an issue.

CyberShadow: Hmm. OK.

MartinNowak:
I only have write access to dlang.org/data/, but I'll get some sleep now and figure out how to deploy this with Jan tomorrow.
BTW, the dbm file generated on the server is bigger than on my machine, so they might be incompatible. If that's the case we could check in the file anyhow.

MartinNowak:
Ah, the file won't change anyhow, so I might just ask Jan to put the dbm in a decent place.

PR #956, opened 2015-04-04
Update errors.dd

smcmurray:
I don't understand how the entire section of error handling never mentions try-catch or how to throw exceptions.

### DIFF
--- a/errors.dd
+++ b/errors.dd
@@ -173,6 +173,10 @@ $(DL
 Using exceptions to handle errors leads to another issue - how to write
 exception safe programs. $(DPLLINK exception-safe.html, Here's how).
 
+D uses $(DPLLINK statement.html#TryStatement, try, catch, finally statements)
+to handle exceptions. You can raise Exceptions using throw, as in
+throw new Exception("oops!");
+
 $(COMMENT
 	$(OL
 


smcmurray:
This is not much better. It would be nice if somebody could fill in more about try/catch/finally, and maybe some examples using assert().

errors.dd
```diff
@@ -173,6 +173,10 @@ $(DL
 Using exceptions to handle errors leads to another issue - how to write
 exception safe programs. $(DPLLINK exception-safe.html, Here's how).
 
+D uses $(DPLLINK statement.html#TryStatement, try, catch, finally statements)
+to handle exceptions. You can raise Exceptions using throw, as in
+throw new Exception("oops!");
```
ntrel: Could wrap this statement in backticks for code formatting.

smcmurray: I was hoping somebody who knows D would replace it with something much more
helpful.

MartinNowak:
Well it already cross-references http://dlang.org/statement.html#TryStatement, so you just need to find a nice transition to redirect people there.

errors.dd
MartinNowak: A basic try/catch/throw paragraph should be at the beginning not at the bottom of this text.

MartinNowak:
At best you expand the existing paragraph.
```
The solution is to use exception handling to report errors. All errors are objects derived from abstract class Error. Error has a pure virtual function called toString() which produces a char[] with a human readable description of the error.
```

MartinNowak:
> I don't understand how the entire section of error handling never mentions try-catch or how to throw exceptions.

The page is pretty old already.

andralex:
@smcmurray ping

smcmurray:
@andralex ?

andralex:
I'll close this because it doesn't fix the larger issues; created https://issues.dlang.org/show_bug.cgi?id=15521.

PR #959, opened 2015-04-05
rsync

MartinNowak:
- use rsync delete
- omit-directory-times


posix.mak
```diff
@@ -250,11 +250,13 @@ clean:
 	rm -f docs.json docs-prerelease.json dpl-docs/dpl-docs
 	@echo You should issue manually: rm -rf ${DMD_DIR}-${LATEST} ${DRUNTIME_DIR}-${LATEST} ${PHOBOS_DIR}-${LATEST} ${STABLE_DMD_ROOT} ${DUB_DIR}
 
+RSYNC_FILTER=-f 'P /Usage' -f 'P /.dpl_rewrite*'
```
CyberShadow: Did you verify that this is all we need? Ran it with --dry-run?

MartinNowak: Yes I did.
There is a test folder owned by you, but it doesn't seem to be needed.

andralex:
thx!

PR #960, opened 2015-04-06
move -boundscheck into sorted position

WalterBright:
switches are supposed to be in sorted order


CyberShadow:
Heh, just as I made https://github.com/D-Programming-Language/dmd/pull/4561

WalterBright:
That was the fastest pull ever!

PR #961, opened 2015-04-06
add lnames and links for all switches

WalterBright:
1. add href names for all switches instead of just a few
2. add cross linking for switch references
3. add helpful links for some other references
4. adjust styling for consistency


CyberShadow:
There is an unpleasant effect when you mouse over the switch names that take parameters (the rest of the line is shifted to make room for the paragraph symbol).

CyberShadow:
https://github.com/D-Programming-Language/dlang.org/pull/962

PR #964, opened 2015-04-07
Add changelog entry for issue4733

yebblies:
@MartinNowak

### DIFF
--- a/changelog.dd
+++ b/changelog.dd
@@ -2,6 +2,45 @@ Ddoc
 
 $(D_S D Change Log,
 
+$(VERSION 068.0, XXX ??, 2015, =================================================,
+
+$(BUGSTITLE Language Changes,
+$(LI $(RELATIVE_LINK2 array-bool-4733, The compiler now generates a warning when using an array as a boolean value))
+)
+
+$(BUGSTITLE Language Changes,
+
+$(LI $(LNAME2 array-bool-4733, The compiler now generates a warning when using an array as a boolean value)
+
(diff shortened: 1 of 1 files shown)


changelog.dd
```diff
@@ -2,6 +2,36 @@ Ddoc
...
+
+$(BUGSTITLE Language Changes,
+$(LI $(RELATIVE_LINK2 array-bool-4733, The compiler now generates a warning when using an array as a boolean value))
+)
+
+$(BUGSTITLE Language Changes,
+
+$(LI $(LNAME2 array-bool-4733, The compiler now generates a warning when using an array as a boolean value)
+
+    $(P The compiler currently allows using an expression with an array type in some places where a boolean is expected.  This is evaluated as `arr.ptr != null`, which may cause unintended behavior when dealing with empty slices, which have a non-null pointer but zero length.  The code should be changed to explicitly use `arr.ptr != null` if it really meant to check for a null array, or `arr.length != 0` or `arr.length != null` if another behavior was intended.  Details and discussion can be found $(LINK2 https://issues.dlang.org/show_bug.cgi?id=4733, here),)
```
MartinNowak: It lists `arr.length != null`, should be `arr !is null` I think.

MartinNowak:
Should we add [this](https://github.com/D-Programming-Language/dmd/pull/2885#issuecomment-71964785) table?
```
        $(TABLE_10 $(ARGS Boolean values for different arrays and expressions),
        $(VERTROW Expression/Array, non-empty array, empty array with valid ptr, null array)
        $(TROW $(D arr.length != 0), $(D true), $(D false), $(D false))
        $(TROW $(D !arr.empty), $(D true), $(D false), $(D false))
        $(TROW $(D arr != null), $(D true), $(D false), $(D false))
        $(TROW $(D arr.ptr != null), $(D true), $(D true), $(D false))
        )
```

yebblies:
Like this?

MartinNowak:
Looks OKish.

PR #965, opened 2015-04-10
remove self-reference to dlang.org

MartinNowak:
- it's pretty pointless to link from the
  getting started page back to the root


davideagen:
Looks good to me.

PR #966, opened 2015-04-11
Fix for 14428. Add epub and chm download buttons for the language spec.

davideagen:
https://issues.dlang.org/show_bug.cgi?id=14428

### DIFF
--- a/download.dd
+++ b/download.dd
@@ -38,7 +38,11 @@ $(H2 Other Downloads)
 
 $(H3 $(LINK2 http://github.com/D-Programming-Language/visuald/releases, VisualD - D Plugin for Visual Studio))
 
-$(H3I D Programming Language Specification) $(BTN http://dlang.org/dlangspec.pdf, pdf) $(BTN http://dlang.org/dlangspec.mobi, mobi)
+$(H3I D Programming Language Specification)
+$(BTN http://dlang.org/dlangspec.pdf, pdf)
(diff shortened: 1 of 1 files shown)


CyberShadow:
How are these updated? Wouldn't it be better to integrate them with the build process?

A CHM is already included with D, which also includes a Phobos reference, keyword index, and other things only possible because it is generated by a dedicated program (chmgen). It is also integrated with D's build process. I don't think we need two...

IMHO, it would be better to integrate something like http://manual.calibre-ebook.com/cli/ebook-convert.html into our build process for the .mobi and upload and link the CHM we already have.

Historically, links to 3rd-party resources have bitrotten rather quickly. For example, see how many links on http://www.digitalmars.com/d/dlinks.html still work.

davideagen:
Yes, it makes sense to host it on dlang.org and have it part of the build process. I see in posix.mak the d.tag target builds the chm file. I'll look into getting the epub built with ebook-convert and adding it and the chm to the rsync target.

CyberShadow:
Currently posix.mak can't build the CHM file, because the CHM program is Windows-only. It only creates d.tag, which is used by dman.

andralex:
Thanks!

MartinNowak:
Has there been any followup? The download links are already dead and I agree with Vladimir we shouldn't have 2 chms with similar content.

PR #967, opened 2015-04-18
Fix Issue 14418

tosttost:
Fix of issue 14418 ( https://issues.dlang.org/show_bug.cgi?id=14418 ) and an undefined identifier bug in the "C-style variadic function in D" example I found while testing the fix of 14418. I did not fill a bug report for the later.

This is my first pull request. I read http://wiki.dlang.org/Pull_Requests, but I perhaps something important. So any comments are welcomed.

I updated the extern(C) example only for consistency, so 'foo' is void in all examples - the extern(C) example does compile with 
extern (C) int foo(int x, int y, ...);
too, so that change is not important.

An alternative fix would be:
- foo returns int
- add something like "return 42;"
I think that would be rather confusing than helpful.

### DIFF
--- a/function.dd
+++ b/function.dd
@@ -1141,7 +1141,7 @@ void test()
 
 void foo(int x, int y, ...)
 {
-    va_list ap;
+    va_list args;
 
     version (X86)
         va_start(args, y);  // y is the last named parameter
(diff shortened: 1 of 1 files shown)


MartinNowak:
thx

PR #968, opened 2015-04-19
Bump DMD to 2.067.0.

jacob-carlborg:
We really need an URL always pointing to the latest version.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -25,10 +25,10 @@ REMOTE_DIR=d-programming@digitalmars.com:data
 GENERATED=.generated
 
 # stable dub and dmd versions used to build dpl-docs
-DUB_VER=0.9.22
-STABLE_DMD_VER=2.066.1
+DUB_VER=0.9.23
+STABLE_DMD_VER=2.067.0
 STABLE_DMD_ROOT=/tmp/.stable_dmd-$(STABLE_DMD_VER)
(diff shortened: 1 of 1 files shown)


CyberShadow:
The Dub and DDox versions referred to by this repo build fine with 2.067.0 then?

jacob-carlborg:
Oh, I didn't think of that. Why doesn't it use a pre-compiled version of Dub?

jacob-carlborg:
I had to update Dub to 0.9.23.

CyberShadow:
Does this fix anything for you? I can still build dlang.org fine without this pull...

jacob-carlborg:
Hmm, looks like I was a bit too quick to create this pull request. It actually do build for me without this PR as well. Someone had a problem in the newsgroup which I thought was related to this. I also misinterpreted how the makefile works. Apparently `STABLE_DMD_VER` is only used to build Dub. Initially I did have some problems building but I think that was rather a problem with me not having cleaned the dmd, druntime and phobos directories.

MartinNowak:
> We really need an URL always pointing to the latest version.

http://ftp.digitalmars.com/LATEST

jacob-carlborg:
That's awesome, thanks.

MartinNowak:
We added those to always know the latest version in Travis-CI.

PR #969, opened 2015-04-20
Fix Issue 14418 - D-style Variadic Function example does not compile

nomad-software:
https://issues.dlang.org/show_bug.cgi?id=14418


schveiguy:
LGTM

MartinNowak:
merged the more complete #967 instead

PR #973, opened 2015-04-21
Update garbage.dd

rsw0x:
fix broken link

was outputting 
    < a href="http://en.wikipedia.org/wiki/Garbage_collection_29"&gt;Wikipedia</a>

before. Maybe undefined macros should trigger a build failure(or warning?)

### DIFF
--- a/garbage.dd
+++ b/garbage.dd
@@ -383,7 +383,7 @@ $(H2 D Operations That Involve the Garbage Collector)
 $(H2 References)
 
 	$(UL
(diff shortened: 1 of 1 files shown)


CyberShadow:
Wait, I think this might be the wrong way to fix it. Percent signs apparently need to be escaped in LaTeX, so the correct fix is to define PERCENT in `html.ddoc`.

CyberShadow:
#974

PR #975, opened 2015-04-23
Fix a note that no longer reflects how D works

ibuclaw:
Reasoning:
```
while (1)
{
    try {
        break;
    }
    finally {
        goto label;   // Compiler error
        return 1;     // Compiler error
    }
}
```

### DIFF
--- a/statement.dd
+++ b/statement.dd
@@ -1170,9 +1170,8 @@ $(GNAME ContinueStatement):
         $(P Any intervening finally clauses are executed, and any intervening
         synchronization objects are released.)
 
-        $(P $(D Note:) If a finally clause executes a return, throw, or goto
-        out of the finally clause,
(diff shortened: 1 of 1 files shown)


statement.dd
```diff
@@ -1170,9 +1170,8 @@ $(GNAME ContinueStatement):
         $(P Any intervening finally clauses are executed, and any intervening
         synchronization objects are released.)
 
-        $(P $(D Note:) If a finally clause executes a return, throw, or goto
-        out of the finally clause,
-        the continue target is never reached.)
+        $(P $(D Note:) If a finally clause executes a throw own of the finally
```
CyberShadow: own? out?

ibuclaw: Fat fingers. :)

PR #976, opened 2015-04-23
Update deprecated features table to reflect errors that are infact also gone

ibuclaw:
Reasoning:
- Cast T[] to integral type
 - The error given is the same for all invalid casts.
- Using length in index expressions
 - The error given is "undefined identifier length" and not "use '$' instead of length".
- Variable shadowing inside functions
 - This will always be invalid code, and no changes to parsing/semantic will make this anything but.
- Upper case 'I' suffix for imaginary literals
 - This will never be gone until complex types leave the language.  Which is as good as never at this point.

### DIFF
--- a/deprecate.dd
+++ b/deprecate.dd
@@ -10,11 +10,11 @@ $(SPEC_S Deprecated Features,
 
     $(TABLE2 Deprecated Features,
         $(THEAD Feature,                                                          Spec,  Warn,   Dep,    Error,  Gone)
-        $(TROW $(DEPLINK Cast T[] to integral type),                              ?,     N/A,    2.060,  2.061,  &nbsp;)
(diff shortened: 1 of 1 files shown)


ibuclaw:
Thanks!  By the way @MartinNowak - is there anything that needs adding to the deprecated features list?  I know this table is not kept up-to-date and most things stuffed into the changelog.

MartinNowak:
Definitely the `if (arr)` thing.

ibuclaw:
OK - I'll write it up.  By the way, how long (by rule of thumb) should we be keeping 'gone' features in the deprecated page?

CyberShadow:
Should they be removed at all? They could come useful to people updating old codebases.

ibuclaw:
@CyberShadow - the question was 'How old is too old?'  5 years? 10 years? 20 years?

MartinNowak:
Sorting them by age should solve most of the problem.

PR #981, opened 2015-04-27
Reorder deprecated features.

ibuclaw:
Hitting two bird with one stone, as per discussion here: https://github.com/D-Programming-Language/dlang.org/pull/976#issuecomment-96258429


andralex:
so prolly this needs surgery as well after we undo `if(arr)`

ibuclaw:
Removed the ``if (arr)`` commit.

ibuclaw:
By the way @andralex @WalterBright - would there be any merit in adding a ``-transition`` for use of 'delete' and 'scope' classes?  Similar as per here https://github.com/D-Programming-Language/dmd/pull/4611

CyberShadow:
:+1: cc @MartinNowak

PR #983, opened 2015-04-29
Issue 14328 - The terms "lvalue" and "rvalue" should be added to the glossary

nomad-software:
Bugzilla: https://issues.dlang.org/show_bug.cgi?id=14328

### DIFF
--- a/glossary.dd
+++ b/glossary.dd
@@ -81,6 +81,16 @@ $(DL
 	maximize the portability of code.
 	)
 
+	$(DT $(LNAME2 lvalue, lvalue))
+	$(DD An lvalue is an abstract term referring to a value with an accessible
+	address (through e.g. the unary $(D &) operator). Typical examples of
+	lvalues include variables, constants introduced with `const` or `immutable`
+	(but not `enum`), and elements of arrays and associative arrays. Calls to
+	functions that return `ref` and pointer dereference expressions are also
+	lvalues. Lvalues can be passed to (or be returned from) functions by
+	reference. An lvalue is the counterpart to an rvalue.
+	)
+
 	$(DT $(LNAME2 nrvo, $(ACRONYM NRVO, Named Return Value Optimization)))
 	$(DD
 	$(P NRVO is a technique invented by Walter Bright around
@@ -195,6 +205,15 @@ void test()
 	RAII is commonly used for resources that are in short supply
 	or that must have a predictable point when they are released.
 	RAII objects in D are created using the $(D scope) storage class.
+	)
+
+	$(DT $(LNAME2 rvalue, rvalue))
+	$(DD An rvalue is an abstract term referring to a value resulting from an
+	expression that has no accessible memory address. This lack of address is
+	only conceptual; the implementation has the liberty to store an rvalue in
+	addressable memory. Rvalues cannot be changed and cannot be passed to (or
+	be returned from) functions by reference. An rvalue is the counterpart to
+	an lvalue.
 	)
 
 	$(DT $(LEGACY_LNAME2 sequential_consistency, sequential-consistency, Sequential Consistency))


glossary.dd
```diff
@@ -81,6 +81,14 @@ $(DL
 	maximize the portability of code.
 	)
 
+	$(DT $(LNAME2 lvalue, lvalue))
+	$(DD A lvalue is the counterpart to a rvalue. A lvalue is an abstract term
```
CyberShadow: Isn't it "An lvalue"? (Here we go again...)

The "X is the counterpart of Y" part should probably be at the end, to avoid inviting people to look up the other term only to get a reference to the first one.

nomad-software: An lvalue does read better but is it correct to write that? I've no idea. Has this discussion taken place before? Agreed the counterpart sentence is better appearing at the end.

nomad-software:
Please make comments and i'll make any necessary changes later.

glossary.dd
```diff
@@ -81,6 +81,14 @@ $(DL
...
 	)
 
+	$(DT $(LNAME2 lvalue, lvalue))
+	$(DD An lvalue is an abstract term referring to a value contained within a
+	variable (including elements of arrays and associative arrays).
+	Characteristics of lvalues are they have a memory address, they can be
+	assigned new values and can be passed to (or be returned from) functions by
+	reference. An lvalue is the counterpart to an rvalue.
+	)
+
```
aG0aep6G: > An lvalue is an abstract term

"Lvalue" is a (natural language) term, but *an* lvalue is an expression (in the programming language sense), no?

> referring to a value contained within a variable

Lvalues don't have to be variables. For example, `*(new int)` is an lvalue.

> Characteristics of lvalues are they have a memory address,

Plural "lvalues" conflicts with singular "memory address". So, "**an** lvalue" or "memory address**es**".
Also, "**that** they have"? I'm not sure.

> they can be assigned new values

They may be immutable, i.e. non-assignable.

> An lvalue is the counterpart to an rvalue.

This is about "lvalue" the term again. And maybe make "rvalue" the subject:
"The counterpart to the term lvalue is the term rvalue."
or
"The counterpart to lvalues are rvalues."
or something better I can't think of.
Also make "rvalues" a link.

Much of this applies to the rvalue entry as well.

andralex: Yah, "contained within a variable" is imprecise because it doesn't cover expressions yielding lvalues and things like immutable data. Proposed text:

```
$(DT $(LNAME2 lvalue, lvalue))
$(DD An lvalue is an abstract term referring to a value with an accessible address (through e.g. the unary $(D &) operator). Typical examples of lvalues include variables, constants introduced with `const` or `immutable` (but not `enum`), and elements of arrays and associative arrays. Calls to functions that return `ref` and pointer dereference expressions are also lvalues. Lvalues can be passed to (or be returned from) functions by
reference. An lvalue is the counterpart to an rvalue.
)
```

glossary.dd
```diff
@@ -197,6 +205,13 @@ void test()
 	RAII objects in D are created using the $(D scope) storage class.
 	)
 
+	$(DT $(LNAME2 rvalue, rvalue))
+	$(DD An rvalue is an abstract term referring to a value resulting from an
+	expression which has no memory address. This lack of address means rvalues
```
andralex: s/which/that/

glossary.dd
```diff
@@ -197,6 +205,13 @@ void test()
...
+	cannot be changed and cannot be passed to (or be returned from) functions
+	by reference. An rvalue is the counterpart to an lvalue.
+	)
+
```
andralex: Proposed text, a bit cleaned up:

```
$(DT $(LNAME2 rvalue, rvalue))
$(DD An rvalue is an abstract term referring to a value resulting from an
expression that has no accessible memory address. This lack of address is only conceptual; 
the implementation has the liberty to store in addressable memory. Rvalues cannot be 
changed and cannot be passed to (or be returned from) functions by reference. An rvalue is 
the counterpart to an lvalue.
)
```

nomad-software:
I'll make changes tomorrow after work and rebase. Thanks for the review.

andralex:
thx!

nomad-software:
@andralex done.

DmitryOlshansky:
No point in waiting on @andralex to just give a nod I guess.

PR #984, opened 2015-04-29
Issue 12803 - __traits(getFunctionAttributes) is not documented

nomad-software:
getFunctionAttributes is only currently missing from the grammar page. I notice another missing there too so i also added it.

Bugzilla: https://issues.dlang.org/show_bug.cgi?id=12803


Hackerpilot:
:white_check_mark:

CyberShadow:
Is `getPointerBitmap` part of the language? Sounds implementation-specific...

CyberShadow:
Well, it's listed on http://dlang.org/traits.html so I guess it is.

CyberShadow:
The bug report said that it is not listed on the traits page, but someone had already added it since then.

PR #988, opened 2015-05-04
Add your-code-here rotation for the front page plus new example

CyberShadow:
http://forum.dlang.org/post/mi0cel$10am$3@digitalmars.com


andralex:
thanks!

PR #989, opened 2015-05-04
update dpl-docs dependencies manually

MartinNowak:
- to avoid breakage by broken upstream package descriptions
- workaround for https://github.com/D-Programming-Language/dub/issues/564


CyberShadow:
Related? https://issues.dlang.org/show_bug.cgi?id=14661

PR #994, opened 2015-05-13
new XREF variant for sub-packages: XREF_PACK

aG0aep6G:
A PR for phobos follows that uses to this to fix all the broken package links.

A generic XREF variant for arbitrary package depth would be nice. But the best I could think of has a weird parameter order:

```d
XREF_DEEP=$(XREF_DEEP_NAMED std$(XREF_DEEP_PREPEND_DOTS $+,$1),$1,$+)
XREF_DEEP_NAMED=$(SPANC libref,$(AHTTP dlang.org/phobos/$(XREF_DEEP_FILE $+).html#$2,$1))
XREF_DEEP_FILE=std$(XREF_DEEP_PREPEND_UNDERSCORES $+)
XREF_DEEP_PREPEND_UNDERSCORES=$(UNDERSCORE)$1$(XREF_DEEP_PREPEND_UNDERSCORES $+)
XREF_DEEP_PREPEND_DOTS=.$1$(XREF_DEEP_PREPEND_DOTS $+)
```

So I went with another hard-coded version instead.

### DIFF
--- a/latex.ddoc
+++ b/latex.ddoc
@@ -85,7 +85,6 @@ _=
 
 FOOTER=
 FOOTNOTE=\footnote{$0}
-FULL_XREF=\href{phobos/std_$1.html#$2}{$(D std.$1.$2)}
 _=
 
 GLINK=$(LINK2 $0, $(I $0))
@@ -354,6 +353,10 @@ VTDX=& $(VERTICAL $1) $(VTDX $+)
 _=
 
 WHITE={\color{white}$0}
+_=
+
+XREF=\href{phobos/std_$1.html#$2}{$(D std.$1.$2)}
+XREF_PACK_NAMED=\href{phobos/std_$1_$2.html#$3}{$4}
 _=
 
 YELLOW={\color{yellow}$0}
--- a/arrays.dd
+++ b/arrays.dd
@@ -241,7 +241,8 @@ s[1..3] = s[0..2]; // error, overlapping copy
         semantics of C.
         )
         
-        $(P If overlapping is required, use $(FULL_XREF algorithm, copy):
+        $(P If overlapping is required, use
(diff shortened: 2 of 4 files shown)


latex.ddoc
```diff
@@ -85,7 +85,6 @@ _=
 
 FOOTER=
 FOOTNOTE=\footnote{$0}
-FULL_XREF=\href{phobos/std_$1.html#$2}{$(D std.$1.$2)}
```
CyberShadow: ?

aG0aep6G: `dlang.org.ddoc` defines `FULL_XREF=$(XREF $1, $2)` (i.e. `FULL_XREF` is an alias for `XREF`). The language reference only ever uses `FULL_XREF`, and hell would break loose in LaTeX if it used `XREF`. `FULL_XREF` seems to be a historical artifact.

By redefining `XREF` instead of `FULL_XREF` in `latex.ddoc`, they become synonymous there, too. Hell won't break loose if someone uses `XREF` instead of `FULL_XREF` in the language reference. And the `XREF` related definitions are next to each other in `latex.ddoc`.

CyberShadow:
Having some trouble testing this. Have you verified that the PDF is built correctly?

aG0aep6G:
> Having some trouble testing this. Have you verified that the PDF is built correctly?

I've checked that it's `std.algorithm.mutation.copy` now. And I did a `grep FULL_XREF *.dd` and checked that they didn't change in the pdf. Otherwise it looks as bad as before at a glance.

aG0aep6G:
Ping? Can I do anything to push this forward?

aG0aep6G:
Now on to https://github.com/D-Programming-Language/phobos/pull/3279 :)

PR #995, opened 2015-05-15
Don't load two additional fonts used solely for site navigation

CyberShadow:
cssmenu.css.dd was pulling in two extra fonts (carried over from [the original version](http://cssmenumaker.com/menu/modern-jquery-accordion-menu)). I'm not convinced we need them - they're used only for navigation, and thus are a mostly pointless gimmicks that increase page load time.

Before  | After
------------- | -------------
  |

### DIFF
--- a/css/cssmenu.css.dd
+++ b/css/cssmenu.css.dd
@@ -2,9 +2,7 @@ Ddoc
 
 /*Adapted from http://cssmenumaker.com/menu/modern-jquery-accordion-menu
 */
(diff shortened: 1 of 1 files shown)


Hackerpilot:
:white_check_mark:

CyberShadow:
Updated with a better font set. Can't provide screenshots because it'll look slightly different depending on what fonts are available and how they're rendered. The font set is nearly the same one Twitter uses for their widget.

PR #996, opened 2015-05-17
Issue 14579 - [SPEC] No specification on modifiers in TypeDelegate symbols

9rnsr:
https://issues.dlang.org/show_bug.cgi?id=14579

### DIFF
--- a/abi.dd
+++ b/abi.dd
@@ -539,7 +539,7 @@ $(GNAME TypeTypedef):
     $(B T) $(GLINK QualifiedName)
 
 $(GNAME TypeDelegate):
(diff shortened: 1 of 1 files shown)


JinShil:
I can't vouch for the technical accuracy, but everything else looks good.

JinShil:
@Hackerpilot, you're our grammar expert, right?  Would you mind doing a review?

Hackerpilot:
This is the change to the documentation for the symbol name mangling specification, not the language itself. I don't know if this is correct or not.

ibuclaw:
This is how my grammar works in the GNU Libiberty demangler (used in Binutils and GDB).

PR #997, opened 2015-05-19
Removed grammatical mistakes from  paragraph 'Here is where regular e…

Doado:
…xpressions'. 

Made some minor stylistic changes and also changed bold font to italic font (admittedly a matter of taste).

### DIFF
--- a/regular-expression.dd
+++ b/regular-expression.dd
@@ -19,7 +19,7 @@ $(D_S Regular Expressions,
      In fact, they are considered so useful that a number of languages provide built-in support
      for them. However, built-in support does $(I not) necessarily imply $(I faster) processing 
      or more features, it is rather a way of providing a
-     $(I convenient and user-friendly syntax) for typical operations and usage patterns.
+     convenient and user-friendly syntax for typical operations and usage patterns.
     )
     $(P The D programming language provides a standard library module $(STD regex).
      Being a highly expressive systems language, D allows regexes to be $(I efficiently) implemented 
--- a/regular-expression.dd
+++ b/regular-expression.dd
@@ -13,13 +13,13 @@ $(D_S Regular Expressions,
     Still no amount of fixed functionality could cover all needs, as naturally flexible text data
     needs flexible solutions.
     )
-    $(P Here is where $(LUCKY regular expressions) come in handy, often succinctly called as regexes.
-     Regexes are simple yet powerful language for defining patterns of strings, 
-     put together with a substitution mechanism, they form a Swiss Army knife of text processing.
-     They are considered so useful that a number of languages provides built-in support
-     for regular expressions. This does $(B not) nessary mean that
-     built-in implies $(B faster) processing or more features. It's more a matter of providing
-     $(B convenient and friendly syntax) for typical operations and usage patterns.
+    $(P Here is where $(LUCKY regular expressions) – often just called $(I regexes) – come in handy.
+     Regular expressions are a simple yet comprehensive language for defining string patterns. 
+     Combined with a mechanism for string substitution, regular expressions are a powerful tool for text processing.
+     In fact, they are considered so useful that a number of languages provide built-in support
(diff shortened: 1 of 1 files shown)


regular-expression.dd
```diff
@@ -13,13 +13,13 @@ $(D_S Regular Expressions,
...
-     put together with a substitution mechanism, they form a Swiss Army knife of text processing.
-     They are considered so useful that a number of languages provides built-in support
-     for regular expressions. This does $(B not) nessary mean that
-     built-in implies $(B faster) processing or more features. It's more a matter of providing
-     $(B convenient and friendly syntax) for typical operations and usage patterns.
+    $(P Here is where $(LUCKY regular expressions) – often just called $(I regexes) – come in handy.
+     Regular expressions are a simple yet comprehensive language for defining string patterns. 
+     Combined with a mechanism for string substitution, regular expressions are a powerful tool for text processing.
+     In fact, they are considered so useful that a number of languages provide built-in support
+     for them. However, built-in support does $(I not) necessarily imply $(I faster) processing
```
McSherry: I don't think the italics on this line + lines 22 add anything to the doc. IMO, the emphasis the italics provide makes the sentence read a bit oddly.

McSherry:
Along with my comment on the diff, this should probably be a pull on the Phobos repo ([this file](https://github.com/D-Programming-Language/phobos/blob/master/std/regex/package.d)) so documentation generated from in-code docs are correct.

I imagine that the docs in this repo are autogenerated from the source code, so changing the documentation comments in the Phobos repo will update the docs in here on the next release.

regular-expression.dd
Doado: Sure, I agree. They don't really add anything.

Doado:
*Along with my comment on the diff, this should probably be a pull on the
Phobos repo (this file
<https://github.com/D-Programming-Language/phobos/blob/master/std/regex/package.d>)
so documentation generated from in-code docs are correct.*

I don't get this bit (yet). I'm new to PRs.

McSherry:
The Phobos repository contains the source code for D's standard library. D supports, in its source code, documentation comments. For example:

```D
/**
 * Multiplies a number by four.
 * 
 * Params:
 *     number = The number to quadruple.
 * 
 * Returns:
 *     The provided number multiplied by four.
*/
auto quadruple(int number)
{
    return number * 4;
}
```

From these in-code comments, you can then generate documentation (like that on the D website, and in this repository). Because the documentation is generated from the source code, the change needs to be made to the source code, otherwise regenerating the documentation would overwrite the change with the old text.

To make the change to the source code, you'll need to pull the [Phobos](https://github.com/D-Programming-Language/phobos/) repository and change the documentation in the `std\regex\package.d` file. Then, when the documentation is next generated from the source code, your change will be there.

Doado:
I understand that the documentation is generated from the Phobos source
files with ddoc. But I don't understand why a change to a page that is not
in Phobos (or any other library doc), in this case
http://dlang.org/regular-expression.html, needs to be pulled in Phobos.

CyberShadow:
> I imagine that the docs in this repo are autogenerated from the source code, so changing the documentation comments in the Phobos repo will update the docs in here on the next release.

No.

This pull request changes:

http://dlang.org/regular-expression.html

This is a dlang.org article that stands separately from the Phobos module documentation:

http://dlang.org/phobos/std_regex.html
http://dlang.org/library/std/regex.html

Doado:
Thanks for clarifying this, Vladimir. I was already wondering.

McSherry:
> No.
>
> This pull request changes:
>
> http://dlang.org/regular-expression.html

Whoops, didn't see that. My mistake, I thought this was a change to the docs rather than an article.

regular-expression.dd
CyberShadow: If there is consensus, let's remove them and merge?

PR #999, opened 2015-05-24
Fix documented order of evaluation

tgehr:
In accordance with http://forum.dlang.org/thread/pgrfvbffrviwjudeawvz@forum.dlang.org#post-mjtihl:2424a3:241:40digitalmars.com.

### DIFF
--- a/expression.dd
+++ b/expression.dd
@@ -13,66 +13,24 @@ $(SPEC_S Expressions,
 
 $(H3 $(LNAME2 order-of-evaluation, Order Of Evaluation))
 
-    $(P The following binary expressions are evaluated in strictly
-        left-to-right order:
-    )
-
-    $(P
-        $(GLINK OrExpression),
-        $(GLINK XorExpression),
-        $(GLINK AndExpression),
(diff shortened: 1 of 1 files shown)


andralex:
Thanks! We should make this pull contingent upon the appropriate compiler changes.

tgehr:
It still includes the warning that DMD does not comply yet, so why not pull it now?

JinShil:
I would like to see this merged with or without changes to the compiler.  It is also my opinion that more pull requests like this should be encouraged, as proper definition should precede implementation.  With the specification nailed down, users don't have to wonder "Is this a bug, or by design?".  Having a clear specification will enable users to file bug reports with confidence and actually aid the implemention by providing test cases that compiler devs can use to verify their work (i.e. teamwork in disguise).  I have passed, a few times, on filing bug reports simply because I could't reconcile what the language designers intended.

Hackerpilot:
I agree. The note that DMD does not currently follow the standard has not been removed, so I don't see any harm in pulling this.

ibuclaw:
:+1:

PR #1002, opened 2015-06-01
remove stream from the index

burner:
the next first step in removing stream from phobos

https://github.com/D-Programming-Language/phobos/pull/3352


burner:
thanks

PR #1006, opened 2015-06-02
split up the changelog

aG0aep6G:
The per-version changelog files (`changelog/2.foo.dd`) have been extracted from the full log by this little ugly script: <https://gist.github.com/aG0aep6G/014339e25fd39ae638f1>.

Preview: http://d-ag0aep6g.rhcloud.com/changelog/2.067.1.html


CyberShadow:
I think we discussed splitting the changelog once before... the main disadvantage is that it breaks Ctrl+F. But now that it's in its own directory, we could do Google "site:dlang.org/changelog" searches.

I'm not against the idea but it's a non-trivial change and should be approved by more people.

aG0aep6G:
Made a thread on the forum: http://forum.dlang.org/post/thvnxyvvtqjrutzdugtf@forum.dlang.org

aG0aep6G:
Moved the list to the site menu, as suggested by Steven Schveighoffer in the forum discussion. Unfortunately, [issue 14717](https://issues.dlang.org/show_bug.cgi?id=14717) makes it fall short. That will have to be fixed first.

Legacy URLs like `changelog.html#2.067.0` now redirect to `changelog/legacy.html` which is not going to be updated with new versions.

Preview: http://d-ag0aep6g.rhcloud.com/accordion/changelog/2.067.1.html

aG0aep6G:
> Unfortunately, issue 14717 makes it fall short.

PR to fix issue 14717: https://github.com/D-Programming-Language/dmd/pull/4799

andralex:
please rebase

.htaccess
```diff
@@ -23,3 +23,5 @@ Redirect 301 /dll.html http://wiki.dlang.org/Win32_DLLs_in_D
 Redirect 301 /htomodule.html http://wiki.dlang.org/Converting_C_.h_Files_to_D_Modules
 Redirect 301 /memory.html http://wiki.dlang.org/Memory_Management
 Redirect 301 /windows.html http://wiki.dlang.org/D_for_Win32
+
+RewriteRule ^changelog.html$ changelog/legacy.html [R=301,L]
```
aG0aep6G: Should this be  `Redirect`? I'm not that well-versed in Apache configuration.

aG0aep6G:
Rebased. Preview is at http://d-ag0aep6g.rhcloud.com again.

Issue 14717 has been fixed, so this is good to go from my side.

.htaccess
CyberShadow: We can use either. But we'll need to test if the `^` will work as expected, IIRC there were some issues with `RewriteBase`.

MartinNowak:
I like it. The changelog page is too big, takes ages to compile, and ages to load.
We could also add some js to legacy.html to directly redirect #2.067.0 to the 2.067.0 page?

MartinNowak:
What about other anchors on the page? I just recently cleaned up the version anchors to be 2.067.0, there is still some legacy anchor new2067 or so.

aG0aep6G:
> I like it. The changelog page is too big, takes ages to compile, and ages to load.
> We could also add some js to legacy.html to directly redirect #2.067.0 to the 2.067.0 page?

I guess that's possible. Will look into it.

> What about other anchors on the page? I just recently cleaned up the version anchors to be 2.067.0, there is still some legacy anchor new2067 or so.

legacy.html has both kinds of anchors. I.e. these two take you to the same link:
http://d-ag0aep6g.rhcloud.com/changelog.html#2.067.1
http://d-ag0aep6g.rhcloud.com/changelog.html#new2_067.1

aG0aep6G:
Did the JS redirects. My JS-foo isn't that great, so someone should look over it.

http://d-ag0aep6g.rhcloud.com/changelog.html#2.067.1 and
http://d-ag0aep6g.rhcloud.com/changelog.html#new2_067.1 should take you to
http://d-ag0aep6g.rhcloud.com/changelog/2.067.1.html now.

aG0aep6G:
Updated for 2.068. I'd appreciate if we could get this over with.

aG0aep6G:
Ping? This has been sitting idle for more than a week now. It's still good to go from my side.

CyberShadow:
This link currently works (unfortunately):

http://dlang.org/changelog

aG0aep6G:
> This link currently works (unfortunately):
> 
> http://dlang.org/changelog

Redirecting that one to `changelog/legacy.html` too now. I.e., these work:
http://d-ag0aep6g.rhcloud.com/changelog (-> /changelog/legacy.html)
http://d-ag0aep6g.rhcloud.com/changelog#2.068.0 (-> /changelog/legacy.html#2.068.0 -> /changelog/2.068.0.html)

CyberShadow:
http://dtest.thecybershadow.net/artifact/website-bb011af464e0c6aac7fcff1d65dc61ed499f2b82-06e16d9f514ddd1641dd267af85ea145/web/changelog/2.068.0.html

This is completely broken because it still thinks that css and js files are under css/ and js/, they are now at ../css/ and ../js/

CyberShadow:
I'm not sure we need the entire changelog index in the sidebar. Instead, I suggest...

- An index page
- At the top and bottom of each version's changelog, links to the next and previous versions' changelogs, and to the index

CyberShadow:
Apart from being unstyled, the per-version changelog pages appear to be all devoid of actual change logs.

aG0aep6G:
> This is completely broken because it still thinks that css and js files are under css/ and js/, they are now at ../css/ and ../js/

> Apart from being unstyled, the per-version changelog pages appear to be all devoid of actual change logs.

I don't know what's going on here. When I build it, those problems don't come up.

> I'm not sure we need the entire changelog index in the sidebar. Instead, I suggest...
> 
> An index page
> At the top and bottom of each version's changelog, links to the next and previous versions' changelogs, and to the index

This started with something similar to that. The current layout was a [suggestion from Steven Schveighoffer](http://forum.dlang.org/post/mm1q4r$13s4$1@digitalmars.com) which people seemed to like (at least no one spoke up against it).

Also, I don't see what a dedicated index page would buy us over having all versions in the sidebar. Having them in the sidebar, highlighting of the current page works trivially, and it acts as a "you are here" indicator in the changelog history.

aG0aep6G:
> I don't know what's going on here. When I build it, those problems don't come up.

There's no mention of `changelog.ddoc` in the [build log](http://dtest.thecybershadow.net/results/5be0b23086c1cd667e6d60547deea4a3311f1e37/7682c947adbdb1ad98340d7f45dea744c978660d/build.log). A missing `changelog.ddoc ` would explain the issues, I think. No idea why it's not there, though.

CyberShadow:
`make` is probably picking up this rule:

```
$(DOC_OUTPUT_DIR)/%.html : %.dd $(DDOC) $(DMD)
	$(DMD) -conf= -c -o- -Df$@ $(DDOC) $<
```
before this one:
```
$(DOC_OUTPUT_DIR)/changelog/%.html : changelog/%.dd $(CHANGELOG_DDOC) $(DMD)
	$(DMD) -conf= -c -o- -Df$@ $(CHANGELOG_DDOC) $<
```

aG0aep6G:
Yup, swapped them around, and it seems to work now. Is it ok to rely on the order like this?

CyberShadow:
> This started with something similar to that. The current layout was a suggestion from Steven Schveighoffer which people seemed to like (at least no one spoke up against it).
> 
> Also, I don't see what a dedicated index page would buy us over having all versions in the sidebar. Having them in the sidebar, highlighting of the current page works trivially, and it acts as a "you are here" indicator in the changelog history.

I don't know, I don't think I agree with that. Having miles-long navigation is just poor UI. You shouldn't have to scroll the page to navigate around a website. The list is only going to get longer, as well. Do you know any other websites that do this?

I think the sidebar should only list the latest few versions. Everything else's significance is mostly historical, and shouldn't be there any more than 1.x and 0.x changelogs should.

aG0aep6G:
> You shouldn't have to scroll the page to navigate around a website. The list is only going to get longer, as well. Do you know any other websites that do this?

The section can be collapsed, and "D reference" and "Books & Articles" force scrolling, too. But it's true that the changelog section is already longer than those and will only get even longer.

> I think the sidebar should only list the latest few versions. Everything else's significance is mostly historical, and shouldn't be there any more than 1.x and 0.x changelogs should.

I wouldn't mind changing it back to a single item (or a few items) in the sidebar and a dedicated listing page. People didn't seem to like it, though. However, I think the main point was that the landing page of the changelog should be the latest entry rather than the listing, which should be doable.

That would mean: clicking "Change Log" in the sidebar takes you to `changelog/2.068.0.html` where you have "previous/next version" and "all versions" which takes you to a long list of links to all the changelog entries. Would everyone be happy with that? @andralex @schveiguy

schveiguy:
> Would everyone be happy with that?

If there is a way to get to a prior version (any prior version) aside from actually scrolling the page down to that version via visual scan (current model), then I'm fine with it. A list of versions with clickable links to each version is a must somewhere, I don't care where, and I don't care how it works (either anchors or individual pages). Putting it on the sidebar, or making it its own page, whatever works for everyone else is fine with me with those requirements satisfied.

aG0aep6G:
> That would mean: clicking "Change Log" in the sidebar takes you to changelog/2.068.0.html where you have "previous/next version" and "all versions" which takes you to a long list of links to all the changelog entries. Would everyone be happy with that? @andralex @schveiguy

Did that.

CyberShadow:
Looks nice. @andralex ?

schveiguy:
LGTM

MartinNowak:
> clicking "Change Log" in the sidebar takes you to changelog/2.068.0.html where you have "previous/next version" and "all versions" which takes you to a long list of links to all the changelog entries.

Much nicer than the long menu.

MartinNowak:
Thanks for the very good work.

CyberShadow:
Well done and thanks for persevering, @aG0aep6G.

PR #1008, opened 2015-06-04
fix wrong usage of "lvalue" and "rvalue"

aG0aep6G:
These meant left-hand/right-hand side or operand.

### DIFF
--- a/arrays.dd
+++ b/arrays.dd
@@ -208,11 +208,11 @@ int[] b = p[0..8];
 
 $(H2 $(LNAME2 array-copying, Array Copying))
 
-        $(P When the slice operator appears as the lvalue of an assignment
-        expression, it means that the contents of the array are the
+        $(P When the slice operator appears as the left-hand side of an
+        assignment expression, it means that the contents of the array are the
         target of the assignment rather than a reference to the array.
-        Array copying happens when the lvalue is a slice, and the rvalue
-        is an array of or pointer to the same type.
+        Array copying happens when the left-hand side is a slice, and the
+        right-hand side is an array of or pointer to the same type.
         )
 
 ---------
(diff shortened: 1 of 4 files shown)


arrays.dd
```diff
@@ -346,16 +346,17 @@ a[] = b[] + 4;
...
+        The right-hand side can be an expression consisting either of an array
+        slice of the same length and type as the left-hand side or an expression
+        of the element type of the left-hand side, in any combination.
         The operators supported for vector operations are the binary
         operators +, -, *, /, %, ^, $(AMP) and |, and the unary operators
         - and ~.
         )
 
-        $(P The lvalue slice and any rvalue slices must not overlap.
+        $(P The slice on the left and any slices on the right must not overlap.
```
Hackerpilot: This should probably say "all" instead of "any"

Hackerpilot:
Other than the "any/all" comment this looks good.

arrays.dd
aG0aep6G: "any" makes sense to me. I take it to mean that if there are *any* slices on the right, then they must not overlap with the one on the left.

"all" sounds weirder to me. Would that mean that the the slices on the right must not overlap with each other? Is that so? dmd doesn't mind it.

I don't know if right-hand slices may overlap with each other, and I'm not a native speaker, so I'd like to leave this for someone else, if it needs changing.

Hackerpilot: Any here reads like "at least one", which makes it "The slice on the left and at least one of the slices on the right must not overlap". I'm pretty sure that's not what is intended because it implies that some of the slices on the right **can** overlap.

aG0aep6G: > Any here reads like "at least one", which makes it "The slice on the left and at least one of the slices on the right must not overlap".

I'm not seeing that. How about this:

    The slice on the left and the slices on the right, if any, must not overlap.

The point to express is that there may not be any slices on the right at all. If we don't care about that we could just drop the "if any":

    The slice on the left and the slices on the right must not overlap.

Hackerpilot: Your second suggestion looks good.

aG0aep6G: Would you be ok with the first one, too? I think it's worth acknowledging that there may not be any slices on the right.

aG0aep6G: Ping? Yes/no?

andralex:
I think the text is clear enough as is.

PR #1011, opened 2015-06-07
posix.mak: Fix clean builds

CyberShadow:
The modlist-*.ddoc files require DMD, so the first make run will fail.


CyberShadow:
Actually, this is not enough - we also need to build the Phobos static library to be able to use source DMD.

PR #1012, opened 2015-06-08
Give suggestions on both error and success after fetching LATEST

andralex:


posix.mak
```diff
@@ -64,9 +64,9 @@ LATEST:=$(shell cd ${DMD_DIR} && \
   git tag | grep '^v[0-9][0-9.]*$$' | sed 's/^v//' | sort -nr | head -n 1)
 endif
 ifeq (,${LATEST})
-  $(error Could not fetch latest version)
+  $(error Could not fetch latest version, place LATEST=2.xyz in the command line)
```
CyberShadow: `2.xyz.t` ?

CyberShadow:
>  per title 

The pull request title is not visible when browsing commits.

posix.mak
andralex: It's either one or the other. Which is more common?

andralex:
@CyberShadow fixed

posix.mak
CyberShadow: Um? All recent releases are in the form `2.NNN.N`. Not sure what you mean.

PR #1013, opened 2015-06-09
Replace "Page wiki" button with new "Report a bug" button

CyberShadow:
See also: https://issues.dlang.org/show_bug.cgi?id=14668


dlang.org.ddoc
```diff
@@ -211,6 +211,12 @@ _=
 PAGE_TOOLS=
 $(DIVID tools,
 	$(SPANC tip,
+		<a href="https://issues.dlang.org/enter_bug.cgi?bug_file_loc=http%3A%2F%2Fdlang.org/$(SELF_PATH)&bug_severity=enhancement&component=$(BUGZILLA_COMPONENT)&op_sys=All&priority=P3&product=D&rep_platform=All&short_desc=%5B$(TITLE)%5D&version=D2" class="button">Report a bug</a>
```
andralex: `$(HTMLTAG3 a, href="..." class="button", Report a bug)`

CyberShadow: Eh, done, but I don't see the point frankly. It's longer and carries no semantic meaning (there is no meaningful alternative value for HTMLTAG3 one may want to use)..

CyberShadow: Plus it's now inconsistent with the other button's HTML code. Such changes are probably better done separately.

doc.ddoc
```diff
@@ -1,5 +1,7 @@
 ROOT_DIR =
 ROOT = .
+SELF_PATH =
+BUGZILLA_COMPONENT = websites
```
andralex: Oops, these just changed, please replace: http://beta.forum.dlang.org/post/ml5tr4$1na4$1@digitalmars.com

std.ddoc
```diff
@@ -1,5 +1,7 @@
 ROOT_DIR = ../
 ROOT = ..
+SELF_PATH = phobos/
+BUGZILLA_COMPONENT = Phobos
```
andralex: ditto

doc.ddoc
CyberShadow: Done - turns out we don't need `BUGZILLA_COMPONENT` now, because we already have `PROJECT` :)

PR #1015, opened 2015-06-09
refactor, add links

WalterBright:
And since gawd hath invented alpha sort, so shall it be done, and it was good.

Added navigation links.

Converted tabs to spaces.

No text changes.

### DIFF
--- a/pragma.dd
+++ b/pragma.dd
@@ -57,11 +57,19 @@ $(H3 $(LEGACY_LNAME2 Predefined-Pragmas, predefined-pragmas, Predefined Pragmas)
 
         $(P All implementations must support these, even if by just ignoring
         them:
+
+        $(UL
+            $(LI $(LINK2 #inline, pragma inline))
+            $(LI $(LINK2 #lib, pragma lib))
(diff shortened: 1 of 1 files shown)


pragma.dd
```diff
@@ -57,11 +57,19 @@ $(H3 $(LEGACY_LNAME2 Predefined-Pragmas, predefined-pragmas, Predefined Pragmas)
 
         $(P All implementations must support these, even if by just ignoring
         them:
+
+        $(UL
+            $(LI $(LINK2 #inline, pragma inline))
```
andralex: Need `$(HASH)` instead of `#`. Run `make pdf` to make sure the pdf makes it through.

CyberShadow:
Looks fine on the doc tester:

http://dtest.thecybershadow.net/artifact/website-9bbe7b16aaa20f28b879722609308a024a12ae52-88b9585ccbd203d7a6fba092a71f9acd/web/dlangspec.pdf (page 128)

andralex:
@CyberShadow how come latex didn't choke on the pound sign?

andralex:
apparently it works! pulling

PR #1017, opened 2015-06-09
Issue 14661 - Add workaround for dpl-docs build failure

s-ludwig:


CyberShadow:
Thanks. My understanding is that `dub --upgrade` will connect online and fetch the latest information about packages? If so, does this mean that a change in an online package will change Dub's behavior? (If possible, it would be ideal if we do not have dependencies on mutable external resources, as that makes things such as bisecting regressions very troublesome.)

s-ludwig:
The `--missing-only` switch will cause it to download only what is already present in dub.selections.json, without looking for actual upgrade possibilities. Together with `dub build --nodeps` it basically just resembles the default behavior of `dub build` without `--nodeps`.

CyberShadow:
Perfect, thanks. Testing now.

CyberShadow:
Dub seems to be fixed!

Now I'm getting a PDF build error, but that's unrelated :)

CyberShadow:
Testing the staging area now

CyberShadow:
All fixed, yay!

s-ludwig:
:+1:

PR #1018, opened 2015-06-11
Use code font for module titles in phobos

andralex:


css/style.css
```diff
@@ -206,6 +206,11 @@ h1
 	margin-top: 0;
 }
 
+h1.d_inlinecode
+{
+	font-family: "Consolas", "Lucida Console", "Lucida Sans Typewriter";
+}
```
CyberShadow: Why? We already have *two* `.d_inlinecode` rules

CyberShadow: And you need to terminate the font list with `monospace`

PR #1019, opened 2015-06-11
Use stable rdmd instead of assuming libphobos.a is built

CyberShadow:
Previously, posix.mak assumed libphobos.a inside ../phobos was built and usable with dmd -L-L. As there is no rule in the makefile to enforce this dependency, a clean build of dlang.org would always fail.

This change causes the stable rdmd (which is downloaded and unpacked to /tmp) to be used instead.


andralex:
alrighty then

CyberShadow:
Wait

andralex:
well?

CyberShadow:
https://github.com/D-Programming-Language/dlang.org/pull/1020

PR #1023, opened 2015-06-12
Add forum skin

CyberShadow:
This pull request adds the Makefile rules, DDoc hooks, and other changes to build forum-template.html, which is used on the new forum.dlang.org.


andralex:
thx!

PR #1024, opened 2015-06-16
dlangspec.pdf: table of contents

aG0aep6G:
before/after:
iconscreenshot1

### DIFF
--- a/latex.ddoc
+++ b/latex.ddoc
@@ -3,6 +3,7 @@ _=Macros for LaTeX generation. The following section is ordered alphabeticaly.
 _=*****************************************************************************
 
 A=\ref{$1}{$+}
+AHTTP=\href{http://$1}{$+}
 AMP=\&
 _=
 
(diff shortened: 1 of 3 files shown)


CyberShadow:
Nice. I thought the TOC was broken because of some Latex invocation weirdness (you have to run it twice or something)

latex.ddoc
```diff
@@ -3,6 +3,7 @@ _=Macros for LaTeX generation. The following section is ordered alphabeticaly.
 _=*****************************************************************************
 
 A=\ref{$1}{$+}
+AHTTP=\href{http://$1}{$+}
```
CyberShadow: Question, why do we need both `AHTTP` and `WEB`?

aG0aep6G: I have no idea. There's also `HTTP`. The differences (if any) between them are not clear to me at all.

CyberShadow:
Thanks!

PR #1025, opened 2015-06-17
Improve hamburger icon load time

CyberShadow:
As the icon is one of the first things in view for the mobile view, it should be displayed as quickly as possible.


andralex:
rox

PR #1029, opened 2015-06-19
Update index.dd: copy editing, wording suggestions

gmfawcett:
Nothing major here; just a few punctuation fixes, and some suggestions for reworking a few awkward or ambiguous phrases.

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -96,9 +96,9 @@ efficiency, control, and modeling power, with safety and programmer productivity
 
 $(SECTION3 Convenience,
 
-$(P D allows writing large code fragments without redundantly specifying types,
-like dynamic languages do. On the other hand, static inference deduces types and other
-code properties, giving the best of both the static and the
+$(P D allows you to write code without redundantly specifying types: it feels like you're writing
+in a dynamic language. But static type inference deduces types and other
+code properties, giving you the best of both the static and the
 dynamic worlds. $(EXAMPLE 1,
 ----
 void main()
@@ -126,7 +126,7 @@ $(P Automatic memory management makes for safe, simple, and robust code.
 D also supports scoped resource management (aka the
 $(HTTP en.wikipedia.org/wiki/Resource_Acquisition_Is_Initialization, RAII) idiom)
 and $(LINK2 statement.html#ScopeGuardStatement, $(D scope) statements) for
-deterministic transactional code that is easy to write and read. $(EXAMPLE 2,
+deterministic, transactional code that is easy to write and read. $(EXAMPLE 2,
 
 ----
 import std.stdio;
@@ -150,8 +150,8 @@ void main()
 )
 )
 
-$(P Built-in linear and associative arrays, slices, and ranges make daily
-programming simple and pleasant for tasks, both small and large. $(EXAMPLE 3,
+$(P Built-in linear and associative arrays, slices, and ranges make everyday
+programming simple and pleasant, for tasks both small and large. $(EXAMPLE 3,
 ----
 #!/usr/bin/env rdmd
 import std.range, std.stdio;
@@ -175,7 +175,8 @@ void main()
 
 $(SECTION3 Power,
 
-$(P The best paradigm is to not impose something at the expense of others.
+$(P The ideal programming paradigm lets you choose the features you need today, 
+without sacrificing the ones you might need tomorrow.
 D offers classic polymorphism, value semantics, functional
 style, generics, generative programming, contract programming,
 and more&mdash;all harmoniously integrated. $(EXAMPLE 4,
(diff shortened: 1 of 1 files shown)


index.dd
```diff
@@ -96,9 +96,9 @@ efficiency, control, and modeling power, with safety and programmer productivity
 
 $(SECTION3 Convenience,
 
-$(P D allows writing large code fragments without redundantly specifying types,
-like dynamic languages do. On the other hand, static inference deduces types and other
-code properties, giving the best of both the static and the
+$(P D allows you to write code without redundantly specifying types: it feels like you're writing
+in a dynamic language. But static type inference deduces types and other
+code properties, giving you the best of both the static and the
```
CyberShadow: I don't like this change. Isn't there a grammar/style rule not to use "you" or something?

index.dd
```diff
@@ -126,7 +126,7 @@ $(P Automatic memory management makes for safe, simple, and robust code.
 D also supports scoped resource management (aka the
 $(HTTP en.wikipedia.org/wiki/Resource_Acquisition_Is_Initialization, RAII) idiom)
 and $(LINK2 statement.html#ScopeGuardStatement, $(D scope) statements) for
-deterministic transactional code that is easy to write and read. $(EXAMPLE 2,
+deterministic, transactional code that is easy to write and read. $(EXAMPLE 2,
```
CyberShadow: OK

index.dd
```diff
@@ -150,8 +150,8 @@ void main()
 )
 )
 
-$(P Built-in linear and associative arrays, slices, and ranges make daily
-programming simple and pleasant for tasks, both small and large. $(EXAMPLE 3,
+$(P Built-in linear and associative arrays, slices, and ranges make everyday
+programming simple and pleasant, for tasks both small and large. $(EXAMPLE 3,
```
CyberShadow: OK

index.dd
```diff
@@ -175,7 +175,8 @@ void main()
 
 $(SECTION3 Power,
 
-$(P The best paradigm is to not impose something at the expense of others.
+$(P The ideal programming paradigm lets you choose the features you need today, 
+without sacrificing the ones you might need tomorrow.
```
CyberShadow: "you" again

index.dd
```diff
@@ -246,11 +247,11 @@ $(SECTION3 Efficiency,
 
 $(P D compiles naturally to efficient native code.)
 
-$(P D is designed such that most "obvious" code is fast $(I and)
-safe. On occasion a function might need to escape the confines of type
-safety for ultimate speed and control. For such rare cases D offers
+$(P D is designed so that the most "obvious" way to write your program produces fast $(I and)
+safe code. In the rare cases that you need to sacrifice type
+safety for ultimate speed and control, D offers
```
CyberShadow: "you" again

index.dd
```diff
@@ -302,7 +303,7 @@ void main()
 
 $(P The $(D @safe), $(D @trusted), and $(D @system) function
 attributes allow the programmer to best decide the safety-efficiency
-tradeoffs of an application, and have the compiler check for
+tradeoffs of an application, and have the compiler guarantee
```
CyberShadow: I don't think we're yet at the point where we can guarantee these things. There's a few holes to close yet.

index.dd
```diff
@@ -347,7 +348,7 @@ $(P Many D enthusiasts are not only using, but also contributing to
 the language, its implementation, and its standard library. The main
 hub of D development is $(HTTP github.com/D-Programming-Language,
 D-Programming-Language on github). The project has been awash in
-contributions ever since creating the github repository on January 23,
+contributions ever since the creation of the Github repository on January 23,
```
CyberShadow: OK

index.dd
gmfawcett: "giving the best" is fine. The main concern in this paragraph was the ambiguity of the first sentence ("redundantly specifying types, like dynamic languages do").

index.dd
gmfawcett: Heh, it's not that I like "you" so much, honest. :) But the first sentence is awkward -- it reads to me like, "the best viewpoint is the one where other people don't pay for your decisions." Even if we wrote, "The best [programming] paradigm imposes nothing at the expense of other things," it's still not true: all paradigms impose limits. I think the new wording is truer to what we are trying to express here.

index.dd
gmfawcett: I concede this one. :)

index.dd
gmfawcett: fair point, I thought this was accidentally written in a weak way.

index.dd
aG0aep6G: I think adding the comma would be wrong. I don't think the sentence is supposed to say that D allows for transactional code, but rather that D allows such code to be deterministic.

index.dd
```diff
@@ -336,9 +337,9 @@ digitalmars.com/pnews/indexing.php?server=news.digitalmars.com&group=digitalmars
 [browse]) $(LINK2 news://news.digitalmars.com/digitalmars.D, [usenet])
 $(HTTP lists.puremagic.com/cgi-bin/mailman/listinfo/digitalmars-d,
 [mailing list])) is the best place to discuss anything and everything
-related to D: language design ideas, suggestions, status and future,
-contributions to the language and its standard library. Reach and
-engage all major D contributors, including the very creator of D,
+related to D: language design ideas, suggestions, current and future status,
```
aG0aep6G: "future status" sounds weird, doesn't it? How about "current status and future direction"?

gmfawcett: I would be fine with "future direction"

index.dd
gmfawcett: Ah, good point! I hadn't thought of it that way

gmfawcett:
I'm going to withdraw this pull request. There are a few points of valid discussion/disagreement, and the remaining changes are fairly minor.

PR #1032, opened 2015-06-23
Update simd.dd

jmh530:
Fleshing out the importance of runtime checks with SIMD by means of example.

### DIFF
--- a/simd.dd
+++ b/simd.dd
@@ -60,6 +60,9 @@ import core.simd;
 	the compiler is targetting. If a particular CPU family has varying
 	support for vector types, an additional runtime check may be necessary.
 	The compiler does not emit runtime checks; those must be done by the programmer.
+    For instance, the Windows version of DMD only implements support for SIMD when
+    compiling with the 64bit flag. If a program does not contain runtime checks,
+    then there might be an error if the 64bit flag was omitted during compilation.
 	)
 
 	$(P The types defined will all follow the naming convention:)


simd.dd
```diff
@@ -60,6 +60,9 @@ import core.simd;
 	the compiler is targetting. If a particular CPU family has varying
 	support for vector types, an additional runtime check may be necessary.
 	The compiler does not emit runtime checks; those must be done by the programmer.
+    For instance, the Windows version of DMD only implements support for SIMD when
```
CyberShadow: Does the OS really matter?

simd.dd
```diff
@@ -60,6 +60,9 @@ import core.simd;
...
+    compiling with the 64bit flag. If a program does not contain runtime checks,
```
CyberShadow: "the 64bit flag" is ambiguous

CyberShadow: `core.simd` is wrapped in `version(D_SIMD)`, so it shouldn't even compile without `-m64`

simd.dd
jmh530: I don't know how true it is on other machines. I was just providing an example for what I was working on.

simd.dd
jmh530: You're right that the "the 64bit flag" could be replaced with `-m64`. I'm new to github (I literally joined to make this change) and it's not entirely obvious to me how to fix it. 

Nevertheless, the general point remains. That page doesn't mention the importance of compiling with `-m64` anywhere and the error messages aren't particularly clear either. I realized I was making a mistake and thought the documentation would be improved if other people know that SIMD might require adding a compiler flag.

CyberShadow: > You're right that the "the 64bit flag" could be replaced with -m64. I'm new to github (I literally joined to make this change) and it's not entirely obvious to me how to fix it.

I'm not sure you can fix it from the website alone, you'll probably need a Git client. [GitHub for Windows](https://windows.github.com/) should be simple enough for new Git users.

> I realized I was making a mistake and thought the documentation would be improved if other people know that SIMD might require adding a compiler flag.

OK, but we shouldn't conflate runtime errors with compiler errors.

CyberShadow: Maybe such a change belongs in the documentation for the `core.simd` module instead, or to the documentation of the `D_SIMD` built-in version.

jmh530: I had decided to add it there because that documentation is the most
detailed. The other ones are more terse. Nevertheless, I would recommend
adding some kind of information about this to all three of those locations.
I hadn't thought of it until you mentioned it.

I have a Github client on my home computer. I doubt it will work on my work
computer (firewall).

jmh530: I'm not sure how to modify this one, but I made some changes and will submit in a separate pull request. I'm closing this one.

PR #1033, opened 2015-06-25
Minor updates to SIMD.dd documentation

jmh530:
A few changes to SIMD.dd, mainly continuing my earlier attempt to provide some additional details on getting it to work in practice with compiler flags. In addition, I fixed an issue with one of the examples not compiling (at least for me). I also fixed some minor spelling issues.

### DIFF
--- a/simd.dd
+++ b/simd.dd
@@ -6,7 +6,7 @@ $(SPEC_S Vector Extensions,
 	operations, sometimes called "media instructions".
 	Vector types are a fixed array of floating or integer types, and
 	vector operations operate simultaneously on them, thus achieving
-	great speedups.)
+	great speed-ups.)
 
 	$(P When the compiler takes advantage of these instructions with standard D code
 	to speed up loops over arithmetic data, this is called auto-vectorization.
@@ -57,9 +57,14 @@ import core.simd;
 ---
 
 	$(P These types and operations will be the ones defined for the architecture
-	the compiler is targetting. If a particular CPU family has varying
+	the compiler is targeting. If a particular CPU family has varying
 	support for vector types, an additional runtime check may be necessary.
-	The compiler does not emit runtime checks; those must be done by the programmer.
+	The compiler does not emit runtime checks; those must be done by the 
+	programmer.
+	)
+	
+	$(P Depending on the architecture, compiler flags may be required to  
+	activate support for SIMD types.
 	)
 
 	$(P The types defined will all follow the naming convention:)
@@ -93,7 +98,7 @@ $(H3 Conversions)
 
 ---
 int4 v = 7;
-v = 3 * v;   // multiply each element in v by 3
+v = 3 + v;   // add 3 to each element in v
 ---
 
 $(H3 Accessing Individual Vector Elements)


simd.dd
```diff
@@ -57,9 +57,25 @@ import core.simd;
...
 	support for vector types, an additional runtime check may be necessary.
-	The compiler does not emit runtime checks; those must be done by the programmer.
+	The compiler does not emit runtime checks; those must be done by the 
+	programmer.
+	)
+	
+	$(P It is possible to test if the architecture the compiler is targeting
+	supports SIMD types by using:)
+---
+version(D_SIMD)
```
ibuclaw: This is not true, as I've mentioned time and again.

When SIMD support was introduced, I made sure that ``version(D_SIMD)`` only applied to DMD's vendor-specific ``__simd`` intrinsic.

ibuclaw: See http://dlang.org/version.html#predefined-versions

The bit that says: ``Vector extensions (via __simd) are supported``

ibuclaw: The *correct* way to determine whether or not the architecture you are targeting supports a given vector natively is by using the method here: https://github.com/D-Programming-Language/druntime/blob/master/src/core/simd.d#L41

Of course, if you just want to test if the ``core.simd`` alias exists, then ``__traits(compiles, float4 f)``will work just as well.

jmh530: I removed the version(D_SIMD) text. Sorry if you've mentioned something "time and again". I only first started using SIMD types like 3 days ago and was not aware of this. Some thing you may find obvious in the documentation are not as obvious to people with less experience. I've been trying to add more details to areas where I had some difficulty (I probably spent a few hours trying to figure out why it wouldn't compile when I didn't use -m64).

ibuclaw: No problem.  I don't mind me if I get a bit emotional when I see dmd-specific information being put into the spec.  There should be a dmd page for this kind of content.

> I've been trying to add more details to areas where I had some difficulty (I probably spent a few hours trying to figure out why it wouldn't compile when I didn't use -m64).

That is a limitation of using dmd you've found there, other compilers have a wider range of SIMD support.

simd.dd
```diff
@@ -57,9 +57,20 @@ import core.simd;
...
 	support for vector types, an additional runtime check may be necessary.
-	The compiler does not emit runtime checks; those must be done by the programmer.
+	The compiler does not emit runtime checks; those must be done by the 
+	programmer.
+	)
+	
+	$(P Depending on the architecture, compiler flags may be required to  
+	activate support for SIMD types. For instance, DMD does not support 
+	SIMD types for X86, but it does for X86_64. Thus, compiling with the)
+---
```
andralex: Use `$(D -m64)` here, not a full code paragraph.

simd.dd
```diff
@@ -92,8 +100,8 @@ $(H3 Conversions)
 	to their vector equivalents:)
 
 ---
-int4 v = 7;
```
DmitryOlshansky: So `int4` doesn't work ? Seems like a bug.

jmh530: It's not that int4 v = 7; doesn't work, it's the next line that doesn't work. v = 3 * v; does not work because int4 does not have multiplication defined. See the table at the bottom. That's why I made the change. v = 3 + v; does work. But I changed it to a float so that I kept it as multiplication. I watched Manu's talk from DConf 2013 last night and he actually recommended not even doing this kind of thing anyway, so I'm tempted to add that in as well.

DmitryOlshansky: What's the recommended way to multiply a float vector by a constant using SIMD then? I mean a compile ought to be able to replicate a vector of 3's and do the normal vector x vector SIMD multiply.

jmh530: Float multiplication works just fine. It's int multiplication that doesn't
tend to work for ints and longs. I'm tempted to just change it to adding
int4's instead of multiplying. It avoids the confusion and is the smallest
change from the existing code.

DmitryOlshansky: Sounds good.

DmitryOlshansky:
> Having some issues with building this with DMD reference. Just removing text and keeping it more general.

What issues? It would be nice to be able to check things on DMD I believe.

DmitryOlshansky:
@jmh530 also a note of wisdom - ping reviewers once you're done changing pull request and want feedback.

jmh530:
@DmitryOlshansky I don't really know what the issue was. I hadn't actually tried to build the docs on my own machine, I just depended on what github was saying. While I agree that it would be nice to have the DMD reference in there, someone else above had said that they would prefer NOT to have references to specific compilers. So once I narrowed it down to that text not working, I wasn't really that bothered with changing it. Also, wrt to your second point, I'm a github newbie, so thanks for the advice.

DmitryOlshansky:
> While I agree that it would be nice to have the DMD reference in there, someone else had said that they would prefer NOT to have references to specific compilers.

Works for me.

DmitryOlshansky:
Please sqash commits then it should be good to go

jmh530:
I had been using the GitHub GUI. I saw no way to squash in there. I ended
up using this:
http://blog.steveklabnik.com/posts/2012-11-08-how-to-squash-commits-in-a-github-pull-request
I think I did it right. Let me know if there's a problem.

DmitryOlshansky:
Thanks!

PR #1036, opened 2015-06-28
Add 301 redirect for pages moved to the wiki.

JinShil:
I'll delete the temporary page if this gets accepted.  I have a couple more I'd like to do, but this is an initial test that I'm doing things right.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -17,3 +17,4 @@ RewriteBase /
 # previous stable releases (namely, std.container).
 
 RewriteRule ^(phobos(-prerelease)?/std_container)_package\.html $1.html [R=301,L]
+RewriteRule 301 /memory.html http://wiki.dlang.org/Memory_Management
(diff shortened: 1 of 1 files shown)


CyberShadow:
That doesn't look like the correct RewriteRule syntax. The response code doesn't go in front.

> I'll delete the temporary page if this gets accepted.

The website should work offline without a web server, so the redirect page should probably stay.

CyberShadow:
BTW, I didn't say that a HTTP redirect wouldn't be an improvement.

Also, we can probably remove the .html page if there are no links to it in our current documentation.

JinShil:
> That doesn't look like the correct RewriteRule syntax. The response code doesn't go in front.

I didn't intend to use mod_rewrite, but I forgot to change `RewriteRule` to `Redirect`.  No rewrite is needed for these, so a simple redirect seems more straightforward.  

> The website should work offline without a web server, so the redirect page should probably stay.

If this PR is merged I'll submit a separate pull request to delete those pages, and the decision can be made there.

andralex:
Alrighty let's see...

PR #1039, opened 2015-07-19
Add documentation for interfacing with Objective-C

jacob-carlborg:
I have not manged to get `core.attribute` to show up in the left menu.

@michelf


objc_interface.dd
```diff
@@ -0,0 +1,370 @@
...
+        NSObject init() @selector("init");
+    }
+    ---
+
+    $(P
+        Currently all Objective-C classes needs to be declared as interfaces in
+        D. All Objective-C classes that should be accessible from within D
+        needs to be declared with the $(LINK2 #objc-linkage, Objective-C)
+        linkage.
+    )
```
michelf: s/needs/need/ (2 instances)

objc_interface.dd
```diff
@@ -0,0 +1,370 @@
...
+
+    $(SECTION2 $(LNAME2 instance-method, Calling an Instance Method))
+
+    $(P
+        Calling an Objective-C instance method uses the same syntax as calling
+        regular D methods:
+    )
+
+    ---
+    NSObject result = object.init();
```
michelf: This could need clarification because usually in D `init` is a property for the default initialization value. Maybe just use a more "normal" method name as an illustration to avoid unnecessary explanations at this point.

jacob-carlborg: Good point, I'll change to some other method.

objc_interface.dd
jacob-carlborg: Fixed

objc_interface.dd
jacob-carlborg: Replaced with a different method.

michelf:
I find it sad that most of the code in this is going to become the wrong way to do things as soon as the rest of the Objective-C support is merged:

- Using `interface` and defining `Class` is going to be superseded by native support for classes
- Using `initWith` and `alloc` is going to be superseded by constructors;
- Calling `objc_lookupClass`  is going to be superseded by the dynamic linker providing that pointer for us;
- Calling `str.initWithUTF8String("Hello World!")` is going to be superseded by native string literal support.

So in the end it's going to be an example of something not to do. But it's still good as intermediary documentation.

CyberShadow:
PDF build is failing but I'm not sure why. Seems to be caused by `$(LINK2 #objc-version-identifier, D_ObjectiveC)`.

jacob-carlborg:
@michelf Yes, unfortunately. Should I add a note about that? But the full support for Objective-C won't break this, right?

jacob-carlborg:
@CyberShadow there's nothing wrong with that link?

CyberShadow:
I think it doesn't like the undescore.

http://stackoverflow.com/questions/2476831/getting-the-error-missing-inserted-in-latex

michelf:
@jacob-carlborg 
> But the full support for Objective-C won't break this, right?

No. It won't break this. The only thing I see that could break this is if ARC gets implemented at some point for `extern (Objective-C)` types (which would be extremely nice, but clearly backward-incompatible).

jacob-carlborg:
@CyberShadow could the `Escapes` macro be used to fix this.

CyberShadow:
I don't know. I don't really know that much about LaTeX. You'll probably need to fiddle with it to get it to  work. Maybe try putting the link text in `$(D ...)`.

jacob-carlborg:
It seems like adding `$(D ...)` worked, but it's an ugly workaround, since the HTML version won't look as good.

@andralex do you know anything about this. I have underscores in the text, mostly links, that causing problems in LaTeX when generating the PDF. It might be that underscores have a special meaning in LaTeX and needs to be escaped.

CyberShadow:
Please rebase?

jacob-carlborg:
@CyberShadow done. I still don't know what to do about the error when generating the PDF.

CyberShadow:
Why not the `$(D ...)` macro?

IIRC, when I looked at the output with that macro, it looked fine to me, maybe there's a font/rendering issue on your side? Can you wrap it in `$(D ...)` again?

jacob-carlborg:
It doesn't look that bad. I don't exactly remember what the problem was. I'm guessing the text `The attribute is only defined when the version identifier D_ObjectiveC is enabled` where `D_ObjectiveC` is a link. If I wrap that in `$(D)` it gets a bold font.

I guess we have to live with that, but just the fact that it's needed and that it doesn't look exactly like the other links annoy me :smiley:.

jacob-carlborg:
Anyway, updated the code.

CyberShadow:
Well, you could change the other links to match instead :) After all, `core.attribute`, `object`, `version` are all bits of D code.

CyberShadow:
Hmm, looks like you'll also need to do this with `objc_lookUpClass` and other identifiers inside links containing underscores.

jacob-carlborg:
:(

jacob-carlborg:
Everything is passing now.

CyberShadow:
Great. But I agree with your earlier observation about inconsistency (some identifiers are in monospace font, some aren't). However, not only is the current situation inconsistent with this document, it is also to some degree inconsistent with the rest of the documentation, as we tend to use code font for identifiers everywhere. I think all identifiers should be in code font (`$(D ...)` or `$(CPPCODE ...)` etc.). This improves readability and clarity (in situations where it's not clear whether a word refers to an English word or an identifier that exists with that exact name somewhere in the code).

So, maybe also make the rest of the identifiers in code font?

- `core.attribute`
- `object`
- `version`
- `release`
- `Class`
- `NSString` (this one is in different fonts across the document, actually)
- `initWithUTF8String:`
- `NSLog`

jacob-carlborg:
I think I've been pretty consistent, at least the way I'm thinking.

* `core.attribute` Never present in code
* `object` Never present in code
* `version` Never present in code
* `release` The references are not referred to the `release` in the code, it's referred to the Apple documentation
* `Class` same as as with `release`
* `NSString` has the code font/style in all cases but once, where it's a link to the Apple documentation
* `initWithUTF8String:` same as with `NSString`
* `NSLog` same as with `NSString`

In general I used the standard style for links for external documentation, even if they refer to (what might be) code. I've also only linked all symbols once, otherwise the code style is used.

CyberShadow:
> Never present in code

I don't understand what you mean here, not present in what code?

jacob-carlborg:
> I don't understand what you mean here, not present in what code?

"In situations where it's not clear whether a word refers to an English word or an identifier that exists with that exact name **somewhere in the code**"

All the code examples here [1]. Isn't that what you're referring to?

[1] http://dtest.thecybershadow.net/artifact/website-9fdead997e4c7af195b1bdd39dccd72cdf15955e-a864c6dad008f3bcd7643506e5cc145e/web/objc_interface.html

CyberShadow:
No, I mean C/C++/Objective-C/D code that the text refers to in general, not necessarily included in this text.

Whether or not an identifier occurs in code on the same page seems like a pretty weird metric to decide what font to use for said identifier.

CyberShadow:
> I've also only linked all symbols once, otherwise the code style is used.

I don't see why code style and links should be mutually exclusive. They change different attributes, thus aren't.

jacob-carlborg:
@CyberShadow update now. I think I found all of them, and a couple more.

CyberShadow:
Looks great!

Can this be merged now (this documents things already supported by the compiler), or is it blocked by an open DMD pull?

If this needs to end up in the HTML docs included with 2.069, then it needs to target the stable branch.

jacob-carlborg:
> Can this be merged now (this documents things already supported by the compiler), or is it blocked by an open DMD pull?

It's already been merged and it's in the stable branch.

> If this needs to end up in the HTML docs included with 2.069, then it needs to target the stable branch.

I'll resubmit the PR.

PR #1047, opened 2015-07-26
state that casting away const and then mutating is undefined behaviour

aG0aep6G:
Related discussion: http://forum.dlang.org/post/riiehqozpkyluhhifwha@forum.dlang.org

I think there's a consensus now that casting away const and then mutating is always undefined behaviour.

Drive-by style fixes: title case, "immutableness" -> "immutability"

### DIFF
--- a/const3.dd
+++ b/const3.dd
@@ -284,10 +284,10 @@ auto p = s.idup;
 p[0] = ...;	  // error, p[] is immutable
 ---
 
-$(H2 Removing Immutable With A Cast)
+$(H2 Removing Immutable or Const with a Cast)
 
 	$(P
-	The immutable type can be removed with a cast:
+	An immutable or const type qualifier can be removed with a cast:
 	)
 
 ---
@@ -312,6 +312,23 @@ int* q = cast(int*)p;
 	the responsibility to ensure the immutableness of the data, as
 	the compiler will no longer be able to statically do so.
 	)
+
+	$(P
+	Note that casting away a const qualifier and then mutating is undefined
+	behavior, too, even when the referenced data is mutable. This is so that
+	compilers and programmers can make assumptions based on const alone. For
+	example, here it may be assumed that $(D f) does not alter $(D x):
+	)
+
+---
+void f(const int* a);
+void main()
+{
+    int x = 1;
+    f(&x);
+    assert(x == 1); // guaranteed to hold
+}
+---
 )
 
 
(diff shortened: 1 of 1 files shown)


Hackerpilot:
:white_check_mark:

MartinNowak:
> Note that casting away a const qualifier and then mutating is undefined behaviour, too, even when the referenced data is mutable.

This sounds too harsh to me, I'd like to have a word on this from @WalterBright.

jmdavis:
> This sounds too harsh to me, I'd like to have a word on this from @WalterBright.

@MartinNowak I totally agree that it would be a good idea to get Walter's approval on this, but I don't see how it's too harsh. It's been brought up time and time again that D's const is physical const and that any attempt to mutate it is undefined behavior. Allowing anything else makes it so that you cannot rely on a const object not being mutated via a const reference. Walter has made it abundantly clear in the past that D's const is actually supposed to provide the guarantee that an object cannot be mutated via a const reference and that the compiler is free to assume that fact. Allowing the casting away of const and then mutating to _ever_ be defined behavior violates this and makes it so that we're back to C++'s const except that it's transitive, and you have to be careful about the possibility of the object actually being immutable. But it's still basically C++'s const and not physical const at all if casting away and mutating is defined behavior. No D program should _ever_ be casting away const to mutate. If it's doing that, it should not be using const.

aG0aep6G:
Ping @WalterBright? (Does he get notices about these @ mentions?)

Btw, I just came across this on <http://dlang.org/cpp_interface.html>:
```
// No:
// ptr to const int
const(int)* p;
int* q = cast(int*)p;
*q = 3;   // undefined behavior
```

const3.dd
```diff
@@ -309,9 +309,26 @@ int* q = cast(int*)p;
...
 	Casting is, as always, a blunt and effective instrument, and
 	when using it to cast away immutable-correctness, one must assume
-	the responsibility to ensure the immutableness of the data, as
+	the responsibility to ensure the immutability of the data, as
 	the compiler will no longer be able to statically do so.
 	)
+
+	$(P
+	Note that casting away a const qualifier and then mutating is undefined
+	behaviour, too, even when the referenced data is mutable. This is so that
```
WalterBright: Please use American spelling, i.e. `behavior`

WalterBright:
LGTM after spell correction.

aG0aep6G:
> Please use American spelling, i.e. behavior

fixed

aG0aep6G:
Ping @MartinNowak @jmdavis. Walter gave his ok. Anything else to do here?

PR #1049, opened 2015-08-03
Remove dependency of dmd on .clone

andralex:
cc @MartinNowak 

Sadly the dependency of `dmd` on `.clone` does not work out.

Consider a user who wants to build `dlang.org` and has a preexisting `dmd` created by other means (i.e. there's no `../dmd/.cloned` directory) and built properly. Then, trying to make dlang.org will follow an unpleasant sequence of events:

1. There's no `../dmd/.cloned` so it will execute the rule for it. The rule uses `touch` to create a `../dmd/.cloned` that is NEWER than the prebuilt `../dmd/src/dmd`. 

2. Following that, the rule attempts to build `../dmd/src/dmd`. The rule will issue `make` in `dmd`'s directory, which in turn requires that either dmd had been installed so it's usable for bootstrapping, or AUTO_BOOTSTRAP=1 is used in the command line (which isn't).

It's nice to have the build of dlang.org attempt to build the entire world of dependencies needed for it to work, but having it require boostrapping paraphernalia etc. seems excessive, particularly when `../dmd/src/dmd` is present and functional. So this PR removes the dependency on `.cloned`. The build will be attempted only if `../dmd/src/dmd` is actually absent.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -30,9 +30,8 @@ STABLE_DMD_VER=2.067.1
 STABLE_DMD_ROOT=/tmp/.stable_dmd-$(STABLE_DMD_VER)
 STABLE_DMD_URL=http://downloads.dlang.org/releases/2.x/$(STABLE_DMD_VER)/dmd.$(STABLE_DMD_VER).$(OS).zip
 STABLE_DMD=$(STABLE_DMD_ROOT)/dmd2/$(OS)/$(if $(filter $(OS),osx),bin,bin$(MODEL))/dmd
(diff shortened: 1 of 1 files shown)


andralex:
Whaaa? No idea why those spurious changes.

andralex:
bleh I need to redo all that

MartinNowak:
> It's nice to have the build of dlang.org attempt to build the entire world of dependencies needed for it to work, but having it require boostrapping paraphernalia etc. seems excessive, particularly when ../dmd/src/dmd is present and functional.

I absolutely agree, have been struggling with the fact that dlang.org comes with it's own build system for quite a while in the release scripts.

PR #1050, opened 2015-08-03
Remove dependency of dmd on .cloned

andralex:
cc @MartinNowak 

Sadly the dependency of `dmd` on `.clone` does not work out.

Consider a user who wants to build `dlang.org` and has a preexisting `dmd` created by other means (i.e. there's no `../dmd/.cloned` directory) and built properly. Then, trying to make dlang.org will follow an unpleasant sequence of events:

1. There's no `../dmd/.cloned` so it will execute the rule for it. The rule uses `touch` to create a `../dmd/.cloned` that is NEWER than the prebuilt `../dmd/src/dmd`. 

2. Following that, the rule attempts to build `../dmd/src/dmd`. The rule will issue `make` in `dmd`'s directory, which in turn requires that either dmd had been installed so it's usable for bootstrapping, or AUTO_BOOTSTRAP=1 is used in the command line (which isn't).

It's nice to have the build of dlang.org attempt to build the entire world of dependencies needed for it to work, but having it require boostrapping paraphernalia etc. seems excessive, particularly when `../dmd/src/dmd` is present and functional. So this PR removes the dependency on `.cloned`. The build will be attempted only if `../dmd/src/dmd` is actually absent.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -331,10 +331,10 @@ dlangspec.verbatim.txt : $(DMD) verbatim.ddoc dlangspec-consolidated.d
 # dmd compiler, latest released build and current build
 ################################################################################
 
-$(DMD) : ${DMD_DIR}/.cloned
+$(DMD):
 	${MAKE} --directory=${DMD_DIR}/src -f posix.mak -j 4
 
-$(DMD_REL) : ${DMD_DIR}-${LATEST}/.cloned
+$(DMD_REL):
 	${MAKE} --directory=${DMD_DIR}-${LATEST}/src -f posix.mak -j 4
 
 ################################################################################


CyberShadow:
Why not simply use touch's `-d` parameter to make the .cloned file as old
as possible?

andralex:
@CyberShadow that doesn't seem like a solution to me. Then why bother with `.cloned` in the first place.

CyberShadow:
I don't understand the problem?

AFAICS making the .cloned file very old solves the problem in your post
while also preserving the Makefile dependency (clone repo if it doesn't
exist before building).

andralex:
@CyberShadow The problem is it adds complexity on an approach that is already of debatable utility. I don't think we need to develop monsters by patching over one little matter at a time.

Building the documentation should have no business building the compiler, which is a somewhat involved process. It's somewhat reasonable to build druntime and phobos seeing as documentation is built from those.

Please don't debate this ad nauseam. I know several solutions are possible that kinda keep `.cloned` around just for the sake of... pretending it's any good. That doesn't make them desirable. When I posted this two-liner I thought "here's the beginning of Yet Another Great Debate. Please let's get work done instead of debating the slightest change.

CyberShadow:
It was just a question/suggestion, no need to get confrontational.

CyberShadow:
https://github.com/D-Programming-Language/phobos/pull/3542#issuecomment-129572745

I think this needs to be reverted (at least the second change).

aG0aep6G:
This introduced a regression: https://issues.dlang.org/show_bug.cgi?id=14915

aG0aep6G:
Made a PR to revert this: #1074

MartinNowak:
> There's no ../dmd/.cloned so it will execute the rule for it. The rule uses touch to create a ../dmd/.cloned that is NEWER than the prebuilt ../dmd/src/dmd.
> 
> Following that, the rule attempts to build ../dmd/src/dmd. The rule will issue make in dmd's directory, which in turn requires that either dmd had been installed so it's usable for bootstrapping, or AUTO_BOOTSTRAP=1 is used in the command line (which isn't).

Seems the conclusion should have been to use `../dmd/src/dmd` instead of `../dmd/.cloned`.

CyberShadow:
`../dmd/src/dmd` is `$(DMD)` in this patch

PR #1038, opened 2015-07-08
Compile-time Argument Lists - Wordmithing, Grammar, and Makefiles

JinShil:
This is a followup to https://github.com/D-Programming-Language/dlang.org/pull/986 specifically to @andralex's [request](https://github.com/D-Programming-Language/dlang.org/pull/986#issuecomment-119320721)

The purpose of this pull request is simply to fix some minor grammar issues and wordsmithing.  There is currently a [discussion taking place on the forum](http://forum.dlang.org/post/mnhfhs$2b9o$1@digitalmars.com) and a [Phobos pull request](https://github.com/D-Programming-Language/phobos/pull/3463) which may result in different terminology, but I don't think they should prevent this PR from being pulled.  Rather a separate PR should be created when that discussion reaches a conclusion. 

While creating this pull request I also found that the ctarguments.dd/html file was not added to the makefiles, so I modified win32.mak and posix.mak.  I don't understand the structure of those makefiles, so if it is not correct, please advise me and I will make the necessary changes.  These changes should also make it easier to review as the ctarguments.html file should show up in the documentation auto-tester.

I also had to do some tabbing to make it easier to understand the DDoc markup.   [View these changes without whitespace](https://github.com/D-Programming-Language/dlang.org/pull/1038/files?w=1)

PR #1052, opened 2015-08-06
fix tuple.dd and ctarguments.dd (issue 14879)

aG0aep6G:
Fix link to std.typecons.tuple.
Add ctarguments.dd to make files so that ctarguments.html gets built.
Add missing parenthesis in ctarguments.dd.
Add TITLEs to tuple.dd and ctarguments.dd.
Add ctarguments.html to the menu.

Fixes issue 14879 - tuple documentation broken link.

https://issues.dlang.org/show_bug.cgi?id=14879

### DIFF
--- a/ctarguments.dd
+++ b/ctarguments.dd
@@ -219,4 +219,4 @@ $(P TODO)
 )
 
 Macros:
-TITLE=Compile-Time Lists
+    TITLE=Compile-Time Lists
--- a/ctarguments.dd
+++ b/ctarguments.dd
@@ -215,3 +215,8 @@ static assert ([ Arguments!(1, 2, 3) ] == [ 1, 2, 3 ]);
 $(H3 Further reading)
 
 $(P TODO)
+
+)
+
+Macros:
+TITLE=Compile-Time Lists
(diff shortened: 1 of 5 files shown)


ctarguments.dd
```diff
@@ -215,3 +215,8 @@ static assert ([ Arguments!(1, 2, 3) ] == [ 1, 2, 3 ]);
 $(H3 Further reading)
 
 $(P TODO)
+
+)
+
+Macros:
+TITLE=Compile-Time Lists
```
MartinNowak: We usually indent macros.

MartinNowak:
ctarguments needs to be updated to match the smaller std.meta change we have right now.
Also AliasSeq instead of Arguments.

ctarguments.dd
aG0aep6G: done

aG0aep6G:
> ctarguments needs to be updated to match the smaller std.meta change we have right now.
> Also AliasSeq instead of Arguments.

This made me remember that there's #1038 which aims to improve the ctarguments article, including any naming updates. This PR and #1038 have some overlap, but this one is focused on getting things working, not on the content. So I'd leave AliasSeq vs Arguments and similar things to #1038.

MartinNowak:
replaced by #1038

PR #1055, opened 2015-08-07
changelog entries for 2.068.0

MartinNowak:


MartinNowak:
> Let's use same order in the 'headers' list and 'descriptions'.

Yes, that was unintentional.

9rnsr:
Other than nitpicks, looks good.

rainers:
> Is the object.di -> object.d change important? I'd just like to know why you list that change.

A user visible change is the renaming of field next to base in TypeInfo_Const. Some more small fixes are listed in https://github.com/D-Programming-Language/druntime/pull/1222

9rnsr:
@rainers Thanks, I understand.
@MartinNowak I think the changelog should notice such the actual effects to end users has introduced.

MartinNowak:
Updated

MartinNowak:
> Is the object.di -> object.d change important? I'd just like to know why you list that change.

I think it is, b/c there are often issues with correctly setting up import paths and or finding object.di.
Also a stale object.di would override the new object.d.

MartinNowak:
wait

MartinNowak:
OK
I added macros to reference PRs
The details of the object.d change shouldn't impact many people, so the PR ref should suffice.

MartinNowak:
more phobos changelog entries

MartinNowak:
Preview https://dlang.dawg.eu/changelog.html.

WalterBright:
Under Compiler Changes:
```
Added compiler switch -profile=gc which instruments calls to memory allocation and writes
a report to the file profilegc.log upon program termination.
```

9rnsr:
The fine-grained commits are much better to review. LGTM.

PR #1058, opened 2015-08-10
also update DConf 2015 notice

MartinNowak:


MartinNowak:
I'll go ahead with this so I can rebuild the website.

PR #1062, opened 2015-08-10
portability.dd: assert should be static assert

luismarques:
This can be asserted at compile time, and that is generally want you want.

(should I open a bugzilla issue before submitting these kinds of changes?)


ZombineDev:
LGTM

PR #1063, opened 2015-08-10
Issue 14900 - fix changelog example

MartinNowak:
[Issue 14900 – 2.068.0 change log example does not compile](https://issues.dlang.org/show_bug.cgi?id=14900)

### DIFF
--- a/changelog.dd
+++ b/changelog.dd
@@ -228,10 +228,15 @@ $(LI $(LNAME2 index-type-followed-ident, Indexed type tuples can now be followed
             alias U = TypeTuple!(long, string);
         }
 
-        alias Types = Tuple!(S, S);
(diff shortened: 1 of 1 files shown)


MartinNowak:
The feature doesn't completely work, so change the example to only show what works.

9rnsr:
it would be the best we can do just now.

MartinNowak:
> it would be the best we can do just now.

Website is updated, it's still in the released docs and chm file, but I doubt that many people are reading that anyhow.

9rnsr:
Thank you again for your lightning work. > Website is updated,

> it's still in the released docs and chm file

If someone post bugzilla issue, I'll humbly receive the complaint.

PR #1068, opened 2015-08-18
Put bugtracker link under Community instead of Resources.

quickfur:
Per: http://forum.dlang.org/thread/kxvzuikgxmoxpduxdjel@forum.dlang.org

### DIFF
--- a/doc.ddoc
+++ b/doc.ddoc
@@ -58,6 +58,7 @@ $(DIVID cssmenu, $(UL
     $(MENU http://code.dlang.org, DUB $(NDASH) The D package registry)
     $(MENU_W_SUBMENU Community)
       $(SUBMENU
+        bugstats.php, Bug Tracker,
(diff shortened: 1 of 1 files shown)


quickfur:
Thanks!

CyberShadow:
```
[2015-09-17 02:44:28] [Destructionator] why the hell is bugzilla linked to under community?
[2015-09-17 02:44:34] [Destructionator] it was the third place i looked to find it
```

adamdruppe:
for the record, the first thing i checked was actually Compilers & Tools, since bug tracking is in my mind a tool relating to the compiler. Then I checked resources. Community was next only because the other options were even worse (it obviously isn't D reference or a Book)

ZombineDev:
Wouldn't it be simpler for the users if the Bugzilla was a top-level menu item, right after DUB, and before community?

PR #1069, opened 2015-08-19
fix issue 14933

BBasile:


pragma.dd
```diff
@@ -124,8 +124,9 @@ pragma(lib, "foo.lib");
 
 
     $(DT $(LNAME2 mangle, $(D mangle)))
-    $(DD Overrides the default mangling for a symbol. This allows
-        linking to a symbol which is a D keyword, which would normally
+    $(DD Overrides the default mangling for a symbol. In DMD, it's only effective
```
9rnsr: `In DMD` is not necessary.

BBasile:
Unfortunately the PR was pushed before your explanation on the bug tracker, after what it was clear that neither GDC or LDC can implement it differently. Anyway it's ok now.

9rnsr:
@BBasile Thanks for your contribution!

9rnsr:
Link to the bugzilla ticket: https://issues.dlang.org/show_bug.cgi?id=14933

PR #1070, opened 2015-08-19
Fix #14935

majiang:


majiang:
The corresponding issue is closed invalid.

PR #1071, opened 2015-08-20
LEADINGROW should extend across entire table, not just 2 columns.

quickfur:
Thankfully, the HTML spec specifies that a `rowspan` value of `0` will do exactly this.

This affects ddoc tables wider than 2 columns, for example the container primitives table in `phobos/std/container/package.d` (see: http://dlang.org/phobos-prerelease/std_container.html).

### DIFF
--- a/std.ddoc
+++ b/std.ddoc
@@ -16,6 +16,7 @@ LREF = <a href="#$1">$(D $1)</a>
 _=
 
 LEADINGROW = <tr class=leadingrow><td colspan=2><b><em>$(NBSP)$(NBSP)$(NBSP)$(NBSP)$0</em></b></td></tr>
+LEADINGROWN = <tr class=leadingrow><td colspan=$1><b><em>$(NBSP)$(NBSP)$(NBSP)$(NBSP)$+</em></b></td></tr>
 TABLE = <table cellspacing=0 cellpadding=5><caption>$1</caption>$2</table>
 TD = <td>$0</td>
 TDNW = <td class="donthyphenate" nowrap>$0</td>


aG0aep6G:
> Thankfully, the HTML spec specifies that a rowspan value of 0 will do exactly this.

HTML5 defines a `0` value for `rowspan` [1]. But you're using `colspan` here. For `colspan` the HTML5 spec says that the value must be greater than zero [2]. HTML4 allowed `0`, though [3].

The best rationale for that I could find is this: "rowspan is supported in the spec. colspan is not because only Gecko supports it and it can't be done incrementally." [4]

[1] http://www.w3.org/TR/html5/tabular-data.html#attr-tdth-rowspan
[2] http://www.w3.org/TR/html5/tabular-data.html#attr-tdth-colspan
[3] http://www.w3.org/TR/html4/struct/tables.html#adef-colspan
[4] https://www.w3.org/Bugs/Public/show_bug.cgi?id=13770

quickfur:
Hmm. That's unfortunate. I guess I'll just have to resort to defining multiple `LEADINGROW` macros, depending on how many columns need to be spanned...

aG0aep6G:
A large value like `1000` might work as long as we don't use `table-layout: fixed` anywhere (I don't know if we do). See <http://stackoverflow.com/questions/398734/colspan-all-columns>.

quickfur:
I think it's safer to just have the caller specify how many columns should be spanned.

CyberShadow:
@quickfur Is this good to go?

quickfur:
Should be, unless people have any objections to having two `LEADINGROW` macros.

quickfur:
Thanks! So now https://github.com/D-Programming-Language/phobos/pull/3569 should be ready to go too.

PR #1072, opened 2015-08-22
use binary gpg signatures instead of ASCII armored ones (.asc -> .sig)

MartinNowak:
- makes the signatures downloadable by default
  instead of showing them in the browser

### DIFF
--- a/download.dd
+++ b/download.dd
@@ -76,7 +76,7 @@ Macros:
 	UBUNTU=$(LOGO ubuntu, Ubuntu)
 	WINDOWS=$(LOGO windows, Windows)
 
-    SBTN=$(SPANC sig_btn,$(BTN $1,$+)<br>$(BTN $1.asc,sig))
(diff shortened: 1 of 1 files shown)


MartinNowak:
I already `gpg --dearmor`'ed all the .asc files and uploaded the .sig files.

MartinNowak:
Anyone?

CyberShadow:
Is one format more widely used than the other?

> makes the signatures downloadable by default instead of showing them in the browser

If this was the only motivation, we could've done the same with .htaccess (though not sure if those work on wherever the downloads are hosted).

MartinNowak:
The use case for ASCII armoring is embedding in text which we don't do and the download page already names them `sig`. This should go into stable though.

MartinNowak:
See #1078

MartinNowak:
> If this was the only motivation, we could've done the same with .htaccess (though not sure if those work on wherever the downloads are hosted).

I'd have to delete and reupload all the signatures to S3 in order to change the content-type.

PR #1073, opened 2015-08-22
Specify that pragma(inline, true) can omit stand-alone functions

jpf91:
As discussed here: http://forum.dlang.org/post/mailman.391.1440162972.13986.digitalmars-d@puremagic.com

If we want to replace some of the druntime RTTI interfaces with templated functions we need to make sure that these template instances don't bloat executables. One solution is to use a force-inline  mechanism and do not generate function code in object files for these functions.

This change is also useful in other cases (heavily templated code with very small functions). We could also decide that `pragma(inline, true)` should always emit a function body. Either way we should specify whether taking the address of a `pragma(inline, true)` function is legal.

ping @WalterBright @andralex

### DIFF
--- a/pragma.dd
+++ b/pragma.dd
@@ -94,10 +94,15 @@ pragma(inline, false)
 ---
 pragma(inline, true)
 ---
-        If a function cannot be inlined with the $(DDSUBLINK dmd-windows, switch-inline, $(TT -inline))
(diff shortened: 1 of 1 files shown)


jpf91:
As Iain pointed out we can emit the function into COMDAT if (and only if) the address is taken.

PR #1074, opened 2015-08-22
Revert "Remove dependency of dmd on .cloned"

aG0aep6G:
Reverts D-Programming-Language/dlang.org#1050

Fixes [issue 14915 - [REG2.068.0] can't build phobos-release](https://issues.dlang.org/show_bug.cgi?id=14915)

I don't know how to do properly what #1050 aimed for. In the short term, I think it's more important to get a working build again.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -337,10 +337,10 @@ dlangspec.verbatim.txt : $(DMD) verbatim.ddoc dlangspec-consolidated.d
 # dmd compiler, latest released build and current build
 ################################################################################
 
-$(DMD):
+$(DMD) : ${DMD_DIR}/.cloned
 	${MAKE} --directory=${DMD_DIR}/src -f posix.mak -j 4
 
-$(DMD_REL):
+$(DMD_REL) : ${DMD_DIR}-${LATEST}/.cloned
 	${MAKE} --directory=${DMD_DIR}-${LATEST}/src -f posix.mak -j 4
 
 ################################################################################


posix.mak
```diff
@@ -337,10 +337,10 @@ dlangspec.verbatim.txt : $(DMD) verbatim.ddoc dlangspec-consolidated.d
 # dmd compiler, latest released build and current build
 ################################################################################
 
-$(DMD):
+$(DMD) : ${DMD_DIR}/.cloned
 	${MAKE} --directory=${DMD_DIR}/src -f posix.mak -j 4
 
-$(DMD_REL):
+$(DMD_REL) : ${DMD_DIR}-${LATEST}/.cloned
```
MartinNowak: Please make those depend on `${DMD_DIR}/src/dmd` to address the main point of #1500.

MartinNowak: Never mind, it's already like it should be. We depend on dmd/src/dmd which itself depends on .cloned.

CyberShadow:
I think only reverting the second part would fix the build problem at hand while also not touching whatever @andralex was actually trying to fix.

CyberShadow:
I still don't see any issue with adding a `touch` invocation, I think it would fix everything.

CyberShadow:
An alternative would be to instead of depending on a `.cloned` file we made up, to depend on a source file we *know* is going to be in the tree and that is part of the `dmd` build process.

MartinNowak:
> An alternative would be to instead of depending on a .cloned file we made up, to depend on a source file we know is going to be in the tree and that is part of the dmd build process.

I have a better solution already.

MartinNowak:
See #1077

PR #1075, opened 2015-08-23
fix Issue 8846 - Specs for Inline Assembler don't include cmpxchg16b

WalterBright:
https://issues.dlang.org/show_bug.cgi?id=8846


aG0aep6G:
Those empty cells in the middle of the table don't look nice.

WalterBright:
Rewriting the whole table makes the diffs look bad.

andralex:
Not sure what the matter here is. @WalterBright please just add the entries and call it a day?

WalterBright:
I did just add the entries. The issue is the table is very large, and inserting a couple entries in the middle means redoing the entire rest of the table, resulting in large rather gratuitous diffs. So I inserted a couple blanks to avoid this.

andralex:
I'll merge this as an improvement and build more on top of it if needed.

PR #1077, opened 2015-08-23
use folders instead of /.cloned dependency

MartinNowak:
- that's what actually gets created when cloning
- will always be older than the content


CyberShadow:
Let's not merge this until the dlang.org autotester is fixed.

https://github.com/D-Programming-Language/dmd/pull/4923

CyberShadow:
OK, let's try it.

CyberShadow:
Tried a clean build, seems to work!

(aside from demanding that `LATEST` is specified on the command line, we should get rid of that some day)

PR #1078, opened 2015-08-24
use binary gpg signatures instead of ASCII armored ones (.asc -> .sig)

MartinNowak:
- makes the signatures downloadable by default
  instead of showing them in the browser

### DIFF
--- a/download.dd
+++ b/download.dd
@@ -76,7 +76,7 @@ Macros:
 	UBUNTU=$(LOGO ubuntu, Ubuntu)
 	WINDOWS=$(LOGO windows, Windows)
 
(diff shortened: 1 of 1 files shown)


CyberShadow:
What's the point of targeting `stable` here?

MartinNowak:
So it'll be in the docs for 2.068.1.

CyberShadow:
You mean in the download.html included with the zip files? Wouldn't think anyone would have any reason to open it.

MartinNowak:
You're right, I think it's even excluded from the release. Still this is going to be the only signatures for the next release, so it kinda makes sense to me to do it on stable.

PR #1081, opened 2015-08-25
remove documentation for -property

aG0aep6G:
Related dmd PRs:
https://github.com/D-Programming-Language/dmd/pull/4955
https://github.com/D-Programming-Language/dmd/pull/4862 (follow-up to the above)

### DIFF
--- a/dcompiler.dd
+++ b/dcompiler.dd
@@ -676,9 +676,6 @@ dmd -cov -unittest myprog.d
 			to the file $(TT profilegc.log) upon program termination.)
 		)
 	  )
-	  $(SWITCH $(SWNAME -property),
-		enforce use of $(CODE @property) on property functions
-	  )
 
 	  $(SWITCH $(SWNAME -quiet),
 		suppress non-essential compiler messages


aG0aep6G:
This should maybe not be pulled before the 2.069 release.

aG0aep6G:
Uhh, no waiting until 2.069 then? 2.068 still has a working `-property`.

DmitryOlshansky:
> Uhh, no waiting until 2.069 then? 2.068 still has a working -property.

Well 2.068 is out, all we merge afterwards is bound for 2.069

aG0aep6G:
The release cycle isn't followed with the website, or is it now? I'm sure I've seen website updates without a release.

MartinNowak:
> The release cycle isn't followed with the website, or is it now? I'm sure I've seen website updates without a release.

We do update the website in between releases, that won't here though.

aG0aep6G:
With the 2.068.1 release this has now been pushed online. That means http://dlang.org/dmd-windows.html and friends don't document `-property` anymore, but the current compiler still has the switch.

PR #1082, opened 2015-08-26
Issue 14963: XREF and XREF_PACKED_NAMED should use relative, not absolute, URLs

quickfur:
Fixes: https://issues.dlang.org/show_bug.cgi?id=14963

Using absolute URLs break the doc tester, local dlang.org installations, as well as cross-module links in `phobos-prerelease`, as it forces everything to point to `http://dlang.org/phobos/`.

This reverts the generated links back to relative URLs.

### DIFF
--- a/dlang.org.ddoc
+++ b/dlang.org.ddoc
@@ -301,9 +301,9 @@ WIKI=$(TITLE)
 _=
 
 XREF=$(XREF2 $1, $2, $(D std.$1.$2))
-XREF2=$(SPANC libref, $(AHTTP dlang.org/phobos/std_$1.html#$2, $(TAIL $+)))
+XREF2=$(SPANC libref, $(A /phobos/std_$1.html#$2, $(TAIL $+)))
 XREF_PACK=$(XREF_PACK_NAMED $1, $2, $3, $(D std.$1.$2.$3))
-XREF_PACK_NAMED=$(SPANC libref, $(AHTTP dlang.org/phobos/std_$1_$2.html#$3, $4))
+XREF_PACK_NAMED=$(SPANC libref, $(A /phobos/std_$1_$2.html#$3, $4))
 _=
 
 YELLOW=$(SPANC yellow, $0)
--- a/dlang.org.ddoc
+++ b/dlang.org.ddoc
@@ -301,9 +301,9 @@ WIKI=$(TITLE)
 _=
 
 XREF=$(XREF2 $1, $2, $(D std.$1.$2))
-XREF2=$(SPANC libref, $(A /phobos/std_$1.html#$2, $(TAIL $+)))
+XREF2=$(SPANC libref, $(A $(ROOT_DIR)phobos/std_$1.html#$2, $(TAIL $+)))
 XREF_PACK=$(XREF_PACK_NAMED $1, $2, $3, $(D std.$1.$2.$3))
-XREF_PACK_NAMED=$(SPANC libref, $(A /phobos/std_$1_$2.html#$3, $4))
+XREF_PACK_NAMED=$(SPANC libref, $(A $(ROOT_DIR)phobos/std_$1_$2.html#$3, $4))
 _=
 
 YELLOW=$(SPANC yellow, $0)
--- a/std.ddoc
+++ b/std.ddoc
@@ -37,3 +37,7 @@ JOIN_DOT_TAIL=$(JOIN_DOT $+)
 JOIN_DOT=.$1$(JOIN_DOT $+)
 TAIL=$+
 _=
+
+XREF2=$(SPANC libref, $(A std_$1.html#$2, $(TAIL $+)))
+XREF_PACK_NAMED=$(SPANC libref, $(A std_$1_$2.html#$3, $4))
+_=


aG0aep6G:
This breaks links from outside the Phobos docs. E.g., the link to `std.algorithm.mutation.copy` on [`/arrays.html`](http://dtest.thecybershadow.net/artifact/website-cb8d983e5c777a8ff9598c1cb9b41d624845e804-72d5a9f6c9b23e93b05cac499b45f667/web/arrays.html#overlapping-copying)

Defining `XREF` differently in `std.doc`/`doc.ddoc` may be a way out. 

`/library/` has a deep directory structure, but doesn't seem affected by this (I didn't check thoroughly, though). I guess they have their own `XREF` already.

quickfur:
Yeah there are several different definitions of `XREF`. I guess the correct solution would be to override(?) it in `std.doc` or `doc.ddoc`.

aG0aep6G:
As far as I see, `/phobos-prerelease/` pages are the only ones that shouldn't link to `/phobos/`.

So I think a good solution would be to link to `/phobos/` by default, and use links to the current directory on Phobos pages (release and prerelease).

That is, in `dlang.org.ddoc` define `XREF` with `$(A $(ROOT_DIR)phobos/std_...)`, and then override it in `std.ddoc` with `$(A std_...)`.

quickfur:
Fixed and rebased. Please review. Thanks!

dlang.org.ddoc
```diff
@@ -301,9 +301,9 @@ WIKI=$(TITLE)
 _=
 
 XREF=$(XREF2 $1, $2, $(D std.$1.$2))
-XREF2=$(SPANC libref, $(AHTTP dlang.org/phobos/std_$1.html#$2, $(TAIL $+)))
+XREF2=$(SPANC libref, $(A /phobos/std_$1.html#$2, $(TAIL $+)))
 XREF_PACK=$(XREF_PACK_NAMED $1, $2, $3, $(D std.$1.$2.$3))
-XREF_PACK_NAMED=$(SPANC libref, $(AHTTP dlang.org/phobos/std_$1_$2.html#$3, $4))
+XREF_PACK_NAMED=$(SPANC libref, $(A /phobos/std_$1_$2.html#$3, $4))
```
aG0aep6G: Absolute paths (starting with `/`) are bad. They don't work when you open `.html` files directly, without a web server in between, or when the site is built in a sub-directory. I'd expect the links to be broken in the documentation tester output, but it spits 500s right now, so can't check.

Instead of a leading `/`, use the `ROOT_DIR` macro to get a relative path to the site root: `$(ROOT_DIR)phobos/...`

quickfur: You're quite right, it messes up the path when viewed locally. It also messes up on a local webserver where it isn't installed to the web root.

Fixed. Thanks for the tip about `$(ROOT_DIR)`, I wasn't aware of it before.

aG0aep6G:
LGTM

DmitryOlshansky:
Same here ddoc builder though spews error 500 so it's hard to asses the results. @CyberShadow ?

CyberShadow:
*sigh* fixed

DmitryOlshansky:
@CyberShadow  are you sure? I got another one at:
http://dtest.thecybershadow.net/artifact/website-4918c9d6ef9ceb055decd05eec6fc38c8fd26229-8d87c04a1173413135bdd41c727037e9/web/library/std/algorithm/iteration/each.html

CyberShadow:
Thanks, gave it a harder kick. Maybe it'll stick this time :)

quickfur:
Seems to work now.

Checked a few links in `phobos/`, `phobos-prerelease/`, `array.html`, seems to point to the right places now.  Reviewers?

aG0aep6G:
> Reviewers?

Still looks good to me.

quickfur:
ping @MartinNowak @andralex @jmdavis @CyberShadow 

This has been stagnating, and in the meantime links on dlang.org are leading to the wrong places. Can we move forward with this please?

CyberShadow:
Looking at the diff, some links in DDox are getting broken, e.g. in http://dtest.thecybershadow.net/artifact/website-6f1c238c51e767f03f0887a0aab3d2f252fba173-8dfaecbab23ee25ffaf19d15442aee16/web/library-prerelease/std/algorithm/comparison/among.html the link to `find` now leads to a 404. Not sure what to do about this though, does DDox override some macros to fix links to use its URL structure or something?

quickfur:
Hmm no idea.  The link macros are a mess... they are defined somewhere and then variously overridden in other places, and it's not obvious which file(s) use which override.

aG0aep6G:
> some links in DDox are getting broken

As far as I can tell, the problem is that `XREF_PACK_NAMED` isn't being overridden in `std-ddox.ddoc`. It's also missing `XREF2`. They currently generate links to `/phobos/` which too is wrong, of course (but better than 404 links).

ddox It seems to recognize the dotted names and generate links for them. I don't know how to make named links that way.

aG0aep6G:
Other links from `/library/` to the rest of dlang.org are broken, too. E.g. the link "Introduction to std.datetime" on http://dlang.org/library/std/datetime.html.

As far as I can tell, ddox doesn't provide a straight-forward way to link from the generated pages to other parts of the site. And I can't think of a non-straight-forward one.

As a band-aid, I suggest to add overrides like the following to `std-ddox.ddoc`:

    XREF2 = $3 (std.$1.$2)
    XREF_PACK_NAMED = $4 (std.$1.$2.$3)

This way, the short names are not links, but the fully qualified names in the parentheses are recognized by ddox and linked properly. It's ugly, but at least we get proper, relative links, and all the information is there.

We should find a way to make nicer XREF links (and fix the other broken ones). But I think that shouldn't block this PR. We're better of with ugly links that point to the right pages than the other way around.

CyberShadow:
> As a band-aid, I suggest to add overrides like the following to std-ddox.ddoc:

That sound like the way to go. @quickfur ?

CyberShadow:
@quickfur Ping

andralex:
@quickfur ?

andralex:
It's been a good while, could someone take this over? Thanks.

quickfur:
Sorry, been too busy to get back to this. Yes, please let someone else take it over.

PR #1084, opened 2015-08-31
add a search option for the changelog

aG0aep6G:
It's only available on changelog pages. And there it's the default.


CyberShadow:
Thanks!

PR #1089, opened 2015-09-05
Field inheritance explanation

MeinMein:


class.dd
```diff
@@ -113,13 +113,14 @@ $(H3 Fields)
 	the base class followed by a dot:)
 
 ---
-class A { int a; }
+class A { int a; int a2}
```
aG0aep6G: Should be `int a2;` (missing semicolon).

9rnsr:
There is corresponding forum discussions or others?

CyberShadow:
I think the example would be improved in a better way by renaming the `int a` field to something else, to avoid confusion with the `class A` type.

JackStouffer:
ping @MeinMein any chance of updating this?

MeinMein:
@9rnsr None that I know of. This is it  :sweat_smile:
@CyberShadow I think the field naming was chosen to show the origin of the field. For the purpose of my addition ```a2``` works in showing that you can access fields defined only in the parent class.

andralex:
OK as is, everyone feel free to build further on this.

PR #1096, opened 2015-09-09
remove override of struct toHash and opEquals in example.

growlercab:
The struct MyString example didn't compile because the methods were marked as override.


9rnsr:
Oh... The redundant `override` keyword has been removed in #1106.

PR #1097, opened 2015-09-09
add link to beta downloads (commented out)

MartinNowak:


CyberShadow:
We have enough horizontal space to simply duplicate the download buttons for all platforms in a second column.

andralex:
So what would be the process of uncommenting?

CyberShadow:
Unwrap the contents of the `$(COMMENT ...)` block.

MartinNowak:
> We have enough horizontal space to simply duplicate the download buttons for all platforms in a second column.

Right, will work on that.

PR #1098, opened 2015-09-12
Various front-page example fixes

CyberShadow:
https://issues.dlang.org/show_bug.cgi?id=12210


CyberShadow:
Instead of removing the two examples requiring network access, they could be added to the exclusion list, which would hide the edit/run buttons. Of course that would mean that the first example visitors see might not be runnable online. Thoughts?

andralex:
I'd say we go for simplicity. Just put code that works in there, and don't put code that doesn't.

MartinNowak:
> Instead of removing the two examples requiring network access, they could be added to the exclusion list, which would hide the edit/run buttons. Of course that would mean that the first example visitors see might not be runnable online. Thoughts?

Regarding http://forum.dlang.org/post/wuchoevfdcnwvpuldiqd@forum.dlang.org I though network access was now allowed. Is that correct @nazriel?

MartinNowak:
I put you on the trello card (https://trello.com/c/JAYt8VG7/91-front-page-code-samples-frequently-crash-when-run-in-browser).

CyberShadow:
> Regarding http://forum.dlang.org/post/wuchoevfdcnwvpuldiqd@forum.dlang.org I though network access was now allowed. Is that correct @nazriel?

If network programs don't work then I guess it's not live yet.

However, my opinion is that we should NOT allow networked programs to run. It creates a very appealing target for amplifying DoS attacks, using the site as a proxy, etc. I think similar sandboxes also block network access for similar reasons.

> I think it's a major downer to not provide any network examples, and instead show some pointless line counting examples. Maybe when the backend is renewed, you can use LXC/docker to limit the network traffic to something reasonable and at least allow http(s).

I don't know. I'm not sure the websites we do allow network access to will be happy about it. If we get an influx of users that use up their website's resources without them getting anything back (like ad views).

PR #1099, opened 2015-09-19
Document the ELFv1 and ELFv2 version identifiers.

redstar:
These identifiers are defined for some time but I did not update
the documentation.
ELFv1 is defined on most systems using ELF.
ELFv2 is defined on Linux/PPC64le.


CyberShadow:
Nit, please capitalize "The" - all other lines in that column start with an upper-case letter. LGTM otherwise

redstar:
It's changed now.

PR #1100, opened 2015-09-19
Introduce SystemZ version identifier and deprecate S390X.

redstar:
The new 64-bit architecture is called System z. IMHO we should
use this name instead the cryptic S390X.


redstar:
See https://github.com/D-Programming-Language/dmd/pull/5101 and https://github.com/D-Programming-Language/dlang.org/pull/1100.

yebblies:
Whoops I forgot I don't have merge rights.  Anyway LGTM.

redstar:
I already changed the druntime source: https://github.com/D-Programming-Language/druntime/pull/1392

PR #1102, opened 2015-09-21
Update the version of Windows that we support.

jmdavis:
It was decided even before XP went EOL that we're not going to
officially support it anymore. It was decided that we officially support
Windows 7 and later, but the download page was not updated, and it still
says XP.

http://forum.dlang.org/post/ktfgps$2ghh$1@digitalmars.com

So, this fixes it so that it says Windows 7.

This just came up in D.Learn, and folks are giving the incorrect answer thanks to the current state of the download page: http://forum.dlang.org/post/uxqojxjkuarkaowpyrzz@forum.dlang.org

### DIFF
--- a/dcompiler.dd
+++ b/dcompiler.dd
@@ -38,7 +38,7 @@ $(H2 $(LNAME2 requirements, Requirements and Downloads))
 	$(LI $(LINK2 http://dlang.org/download.html, Download D Compiler))
 
     $(WINDOWS
-	$(LI Windows operating system, Windows XP or later, 32 or 64 bit)
+	$(LI Windows operating system, Windows 7 or later, 32 or 64 bit)
 
 	$(LI Download
 	 <a href="http://ftp.digitalmars.com/dmc.zip" title="download dmc.zip">


DmitryOlshansky:
>  It was decided that we officially support Windows 7 and later

Is there any code that requires Win7 and not say Vista? I don't mind supporting only Win7+.

jmdavis:
> Is there any code that requires Win7 and not say Vista? I don't mind supporting only Win7+.

AFAIK, we don't currently have any code that won't work with Vista, but something may have been added that I don't know about. I don't pay as much attention to pull as I used to. I don't even remember if we have code that won't work with XP (though at some point, I should probably rewrite some of `WindowsTimeZone` to use functionality that Vista added). However, I'm pretty sure that we have code that won't work with any Windows older than XP.

I think that Walter picked Windows 7 with the idea that we were testing that, whereas we're not testing Vista and it's not a great idea to say that we're supporting something that we haven't tested (though it's not like we're testing every Linux distro or Mac OS X release or FreeBSD release either).

I don't really care if we say Vista or 7, but I don't think that it makes any sense to say that we support an OS that the maintainers of that OS no longer support, and Vista _did_ add a lot of functions to the Win32 API. So, at some point, some calls that use it will probably end up in something that Phobos does, and that code won't work in XP. It could be that there's already some portion of Phobos that's using a function from Vista, and I don't know about it.

But Walter said Windows 7, so I put Windows 7 in the PR.

I suspect that there aren't many Windows Vista uses though. The bigger issue is folks that refuse to upgrade from XP. But they're free to download and use dmd. We just don't support it on those systems, and if they find a problem that is specific to older versions of Windows, then we won't fix it.

DmitryOlshansky:
Fair enough

PR #1110, opened 2015-09-25
unlink references to the long gone std.date

aG0aep6G:
Eventually, intro-to-datetime should be updated/rewritten/removed, as it
references the long obsolete std.date throughout. For now, just fixing dead
links.


9rnsr:
Nice fix for 404.

PR #1113, opened 2015-09-30
provide input for examples

MartinNowak:
- also add pow (^^) to the rpn calculator

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -89,7 +89,8 @@ void main()
 
     void process(in char[] token)
     {
-        alias Ops = AliasSeq!("+", "-", "*", "/", "%");
+        alias Ops = AliasSeq!(
+            "+", "-", "*", "/", "%", "^^");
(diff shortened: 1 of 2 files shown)


CyberShadow:
https://github.com/D-Programming-Language/dlang.org/pull/1098 ?

MartinNowak:
Updated @CyberShadow

CyberShadow:
Too bad the "Run" button doesn't appear on the autotester, because it would involve a cross-origin request...

nazriel:
@CyberShadow I guess we can allow auto-tester URL for cross-origin?

CyberShadow:
I guess? Just keep in mind that anyone can create a PR...

nazriel:
Hmm, not sure if they can do any more harm than just using dpaste.dzfl.pl ?
Although you may be right

PR #1118, opened 2015-10-08
changelog 2.069.0

MartinNowak:
- fix links in pre-release changelog
- fix curl anchor
- add note about self-hosted ddmd


MartinNowak:
fixed, should work now

PR #1119, opened 2015-10-08
[2.069] mention @property in the changelog

aG0aep6G:
Also put a note on the spec page stating that @property is not implemented
and that its details may change.

I'm targetting the stable branch here. Is that right? Do we do that with the dlang.org repository?

### DIFF
--- a/function.dd
+++ b/function.dd
@@ -371,7 +371,8 @@ $(H4 $(LNAME2 property-functions, Property Functions))
 
         $(P Property functions are tagged with the $(D @property)
         attribute. They cannot be called with parentheses (hence
-        they act like fields except in some cases).
+        they act like fields except in some cases) ($(B note:) dmd does not
+        enforce this currently).
         )
 
         $(P If a property function has no parameters, it works as a getter.
--- a/changelog/2.069.0_pre.dd
+++ b/changelog/2.069.0_pre.dd
@@ -70,6 +70,11 @@ $(LI $(LNAME2 property-switch-deprecated, The $(TT -property) switch has
     $(P Since the behaviour of the $(TT -property) switch was not well-liked,
         it's been deprecated and made to have no effect when used.
     )
+
+    $(P The $(A $(ROOT_DIR)function.html#property-functions,$(D @property)
+        attribute) is not affected, and remains as a mostly cosmetic keyword.
+        Its semantics are going to be revisited in the future.
+    )
 )
 
 $(LI $(LNAME2 backend-improvements, DMD's codegen has improved.)


CyberShadow:
Why 2.069's changelog specifically? What changed in 2.069?

function.dd
```diff
@@ -369,6 +369,11 @@ $(H4 $(LNAME2 inout-functions, Inout Functions))
 
 $(H4 $(LNAME2 property-functions, Property Functions))
 
+        $(P $(B Note:) dmd does not implement the behaviour described in this
+        section. And the details of what is described here may change in the
```
CyberShadow: Please don't start sentences with "And"

aG0aep6G:
> Why 2.069's changelog specifically? What changed in 2.069?

We're deprecating `-property` in 2.069.

function.dd
aG0aep6G: > Please don't start sentences with "And"

Just start without it then?

CyberShadow:
Oh, I missed that this is an addition to an existing item, never mind.

function.dd
CyberShadow: Yes, or I think you can just use a comma instead of a new sentence here.

changelog/2.069.0.dd
```diff
@@ -53,6 +53,14 @@ $(LI $(LNAME2 property-switch-deprecated, The $(TT -property) switch has
     $(P Since the behaviour of the $(TT -property) switch was not well-liked,
         it's been deprecated and made to have no effect when used.
     )
+
+    $(P The $(A $(ROOT_DIR)function.html#property-functions,$(D @property)
+        attribute) has not been deprecated. But since it only had an effect with
```
CyberShadow: Again, "But", consider "However,"

changelog/2.069.0.dd
```diff
@@ -53,6 +53,14 @@ $(LI $(LNAME2 property-switch-deprecated, The $(TT -property) switch has
...
+        the $(TT -property) switch, it has no effect anymore. The future of
+        $(D @property) is uncertain. It's clear that there is a desire for
```
CyberShadow: Should we really announce to the entire world in the release changelog that we are uncertain about things?

jmdavis:
> Why 2.069's changelog specifically? What changed in 2.069?

Yeah. Nothing has changed with `@property`. It's the same that it's always been - and pretty close to what it's probably always going to be.

The only thing that's changed is that the -property switch has been deprecated, because it attempted to enforce behavior that we're never going to enforce.

It may be that the description of `@property` needs to be updated - but not to claim that it's not implemented. In all likelihood, `@property` will be adjusted to affect property functions that return callables so that only one set of parens is required rather than two, and everything else about it will stay the same. _Maybe_ we'll end up doing something to restrict setters, but it's pretty clear at this point that for the most part `@property` is what it's going to be, and nothing about that has changed with 2.069.

changelog/2.069.0.dd
jmdavis: We should mention that the `-property` flag has been deprecated - and maybe mention that we're deprecating it because it enforces behavior that we don't want to enforce - but I don't think that there's any need to discuss `@property` beyond that.

aG0aep6G:
> but it's pretty clear at this point that for the most part @property is what it's going to be, and nothing about that has changed with 2.069.

Maybe I'm missing something about `@property`. I'm under the impression that it doesn't have any effect at all. Does it actually do anything? dmd doesn't seem to treat the examples from the spec different with/without `@property`.

changelog/2.069.0.dd
aG0aep6G: I think an explanation why we're keeping a nop attribute around is in order. I don't know how to give a more positive spin.

aG0aep6G:
Followed @CyberShadow's wording suggestions. (The force push also made the "uncertain" discussion "outdated" to GitHub, sorry.)

CyberShadow:
> I don't know how to give a more positive spin.

How about:

$(P The $(A $(ROOT_DIR)function.html#property-functions,$(D @property) attribute) is unaffected, and remains as a purely cosmetic keyword. It is being preserved for backwards compatibility.

jmdavis:
> Maybe I'm missing something about @property. I'm under the impression that it doesn't have any effect at all. Does it actually do anything? dmd doesn't seem to treat the examples from the spec different with/without @property.

It affects the signature of the function, so it affects overloading. It also affects what you get from `typeof`. Until recently, `save` was required to be `@property` by `isForwardRange`. It's mostly corner case stuff that it does much to at this point. What it _needs_ to do - but doesn't yet - is affect functions that return callables so that something like `myCallableProp()` will work instead of having to do `myCallableProp()()`. It _might_ at some point prevent something like

`writeln = 7;`

but that's not a given. Ultimately, it will be mostly cosmetic, but it won't be completely cosmetic, because without it, property syntax won't work with functions that return callables. And maybe the spec should be updated, but I don't think that it makes sense to slap on a note about how the spec is wrong rather than fixing the spec. And I don't think that the changelog really needs to discuss `@property` beyond stating that the `-property` flag has been deprecated, because nothing else has changed with regards to `@property`.

> $(P The $(A $(ROOT_DIR)function.html#property-functions,$(D @property) attribute) is unaffected, and remains as a purely cosmetic keyword. It is being preserved for backwards compatibility.

Except that it isn't purely cosmetic. It doesn't affect much, but it does affect a few things. And it's pretty much a given that at some point it's going to affect functions which return delegates and other callables - though it hasn't yet. I don't think that we really need to say much about `@property` at this point. We're currently keeping the status quo with `@property`. It's just that `-property` has been deprecated and now does nothing.

jmdavis:
>> It affects the signature of the function, so it affects overloading.
> Huh?

I meant overriding, not overloading, though either I misremembered, or that changed at some point, because when I just tested it, it worked.

aG0aep6G:
> $(P The $(A $(ROOT_DIR)function.html#property-functions,$(D @property) attribute) is unaffected, and remains as a purely cosmetic keyword. It is being preserved for backwards compatibility.

I could live with that, but it's a bit light on the explanation part, isn't it? Actually we're keeping `@property` around so that we can fill it with meaning in the future. But we can't say what exactly we're going to do with it, because the discussions about it have not been conclusive.

> It also affects what you get from typeof.

Ha, indeed. Weird.

> I don't think that it makes sense to slap on a note about how the spec is wrong rather than fixing the spec.

dmd's current behaviour is not what we want in the spec (right?). And what's in the spec is not implemented by dmd at all. I think a warning is in order in such a situation.

If there's a plan for `@property`, it would definitely be better to describe that instead of just stating that things "may change in the future". I'm not aware of one, but I'm also not very deeply involved in D discussion culture, so there may be one.

aG0aep6G:
By the way, we can still paddle back: re-add `-property`, fix phobos to be compatible with it, pretend nothing ever happened.

jmdavis:
> dmd's current behaviour is not what we want in the spec (right?). And what's in the spec is not implemented by dmd at all. I think a warning is in order in such a situation.

Maybe, but looking over that section now, the only parts that look wrong are the part about @property functions not being callable with parens (which would be fantastic, but I don't expect that we'll ever actually get that) and the part about optional parens not working with the setter syntax (since unfortunately, they do - though that may change; we'll have to wait and see). Arguably, the only parts that are wrong are parts that folks shouldn't be doing in code anyway.

A note on the two bits that are wrong saying that they're not yet implemented might make sense, but marking the whole section the way that you are is just going to make people wonder what's implemented and what isn't and make the whole section unclear.

As it stands, worst case, someone accidentally uses parens on an @property function or uses a non-@property function with the setter syntax and is surprised that it works. So, I think that it would actually be less confusing to change none of it that to put a note on the top like you're proposing - particularly since it's unclear exactly what changes will be happening to @property in the future. It's pretty clear that those changes will be minimal, but we can't give concrete information at this point, and I don't think that injecting uncertainty into the docs is going to help. It's possible that what's in the spec _will_ end up being the truth, and it's possible that the only change that will happen will be with regards to property functions which return callables. We really don't know.

> By the way, we can still paddle back: re-add -property, fix phobos to be compatible with it, pretend nothing ever happened.

Yikes, no. `-property` should have been removed ages ago. It's been known for quite some time that it was enforcing behavior that we did not intend to enforce in the long run, and it should have been removed some time ago. For better or worse, UFCS killed any chance of forcing non-@property functions to be called with parens.

jmdavis:
By the way, it would really suck for anyone on github to have an account with the same name as a D attribute that started with @. Someone with the name property or safe would probably end up with tons of messages. :)

JakobOvrum:
@jmdavis, wrap those attribute references in `code tags` and it won't notify anyone.

jmdavis:
> @jmdavis, wrap those attribute references in code tags and it won't notify anyone.

Yeah. I generally try and do that, but it's still the case that anyone with a name like that risks getting a bunch of extra notifications due to when someone doesn't wrap the attribute reference in code tags (and when the comment was generated from a commit message, it's not going to have markdown in it). I just commented on it because of how many times I was having to type `@property`, and it occurred to me that someone with the username property probably got more stray notifications from the D community than they might like.

aG0aep6G:
Alright, how about this:

Changelog: "The `@property` attribute is not affected, and remains as a mostly cosmetic keyword. Its semantics are going to be revisited in the future."

"Mostly cosmetic" is a little vague, and the spec doesn't really help there, but that's a battle for another day, I suppose.

On the spec page, add a note only to the first paragraph: "They cannot be called with parentheses (hence [...]) (note: dmd does not enforce this currently)."

jmdavis:
> "Mostly cosmetic" is a little vague, and the spec doesn't really help there, but that's a battle for another day, I suppose.

I guess that such a comment is fine, though honestly, I wouldn't say anything about `@property` in the changelog other than to say that it hasn't changed in spite of the removal of `-property`, because nothing about `@property` has changed.

PR #1122, opened 2015-10-08
[issue 15078] GC documentation should reflect 2.067 changes

JackStouffer:
Copy-pasted from the 2.067 change log.


MartinNowak:
There is something wrong with the latex generation, I think you need to use LPAREN and RPAREN or so.

JackStouffer:
Well, I was about to rebase to get rid of the garbage commits. ¯\_(ツ)_/¯

PR #1126, opened 2015-10-10
Disambiguate type tuple names in 2.068.0 changelog

ntrel:


CyberShadow:
It's possible this should have targeted the stable branch. We build dlang.org from master, but as I understand, the changes in master will currently not end up in 2.069 zip files unless cherry-picked. @MartinNowak ?

PR #1128, opened 2015-10-13
Remove dpl-docs build workarounds on Windows.

s-ludwig:
- Removes excluding format.d which used to fail compilation, but works now
- Removes the workaround for D-Programming-Language/dub#331, which has been fixed for a few releases now


DmitryOlshansky:
LGTM

PR #1129, opened 2015-10-13
  Add documentation for interfacing with Objective-C

jacob-carlborg:
See https://github.com/D-Programming-Language/dlang.org/pull/1039 for reference and comments.


jacob-carlborg:
@CyberShadow

CyberShadow:
Thanks!

attribute.dd
```diff
@@ -91,6 +92,9 @@ $(GNAME LinkageType):
         $(D C) and $(D D) must be supplied, the others are what
         makes sense for the implementation.
         $(D C++) offers limited compatibility with C++.
+        $(D Objective-C) offers limited compatibility with Objective-C,
+        see the $(LINK2 objc_interface.html, Interfacing to Objective-C)
```
MartinNowak: Guess this should have used `$(ROOT_DIR)objc_interface.html`.

jacob-carlborg: Why is that? The link works. Or is for an other format?

PR #1133, opened 2015-10-15
Fix for issue 13624 (Part 1)

JackStouffer:
https://issues.dlang.org/show_bug.cgi?id=13624

* add link in getting started to overview page
* add links in sections to pages with more info
* make it clear that D supports 64 bit
* slight change in suggestions for a first language (does anyone still learn Basic anymore?)
* convert the example program to use more D style code

To mark this issue as fixed, I think that the another example program should be added showing range based code, and a section on ranges should be added. I broke up the changes to make it easier to review.

### DIFF
--- a/overview.dd
+++ b/overview.dd
@@ -314,8 +314,8 @@ $(UL
 	$(LI Support for 16 bit computers.
 	No consideration is given in D for mixed near/far pointers and all the
 	machinations necessary to generate good 16 bit code. The D language
-	design assumes at least a 32 bit flat memory space. D will fit smoothly
-	into 64 bit architectures.
+	design assumes at least a 32 bit flat memory space and supports 64 bit as
+    well.
 	)
 
 	$(LI Mutual dependence of compiler passes. In C++, successfully parsing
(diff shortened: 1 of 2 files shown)


overview.dd
```diff
@@ -1044,23 +1067,25 @@ $(SECTION2 Sample D Program (sieve.d),
...
-int main()
+void main()
 {
-    int i, count, prime, k, iter;
+    size_t i, count, prime, k;
+    bool[8191] flags;
 
-    writeln("10 iterations");
-    for (iter = 1; iter <= 10; iter++)
+    "10 iterations".writeln;
```
CyberShadow: Please no, we don't use UFCS for writeln.

`writefln` and `format` are a bit different in that the first argument is special, but I don't think there is much point in confusing newcomers here.

overview.dd
```diff
@@ -1044,23 +1067,25 @@ $(SECTION2 Sample D Program (sieve.d),
...
+    
+    // using _ as a throwaway variable
+    foreach(_; 1 .. 11)
```
CyberShadow: Please no, just use `iter`.

`_` is not what people might expect, if they come from languages where it is an actual language syntactical feature. You can't define more than one `_` in a function in D, for example, because it's just a variable.

overview.dd
```diff
@@ -1044,23 +1067,25 @@ $(SECTION2 Sample D Program (sieve.d),
 
 import std.stdio;
 
-bool[8191] flags;
-
-int main()
+void main()
 {
-    int i, count, prime, k, iter;
+    size_t i, count, prime, k;
```
CyberShadow: `i` is unused

CyberShadow: Move `k` and `prime` to where they are used, declaring all your variables at the top of the function is a C-ism

JackStouffer:
@CyberShadow addressed issues

CyberShadow:
> @CyberShadow addressed issues

Oops, thanks

PR #1137, opened 2015-10-23
Fix some typos

schveiguy:
Also, there was a statement that some "characters" were not valid symbol characters, but then listed only one character, the colon. The way this looked in the docs was very confusing (I wasn't sure if there was something missing after the colon).


JackStouffer:
LGTM

PR #1116, opened 2015-10-07
beta preparation

MartinNowak:
- changelog for 2.069.0
- add beta channel to download page

PR #1139, opened 2015-10-24
one-click download button

aG0aep6G:
snapshot1

(To avoid confusion, it looked [like this](https://cloud.githubusercontent.com/assets/9287500/10711239/4fb76968-7a75-11e5-8677-8a27eebdc952.png) before [this point in the discussion](https://github.com/D-Programming-Language/dlang.org/pull/1139#issuecomment-155489317).)

I've only tested a couple browser/OS combinations:
Windows 7: Internet Explorer, Firefox and Chrome all get it right (.exe download).
Ubuntu 15.10: Firefox gets it right (.deb download), Chrome and Konqueror only get Linux (.tar.xz download).

I'm not sure if going for the generic Linux download is the right thing to do when no specific distribution could be detected. It may be better to lead people to download.html where they may find a proper package.

Would be nice if someone could test OS X, Debian, Fedora, CentOS, openSUSE, and FreeBSD.

~~The first commit fixes a couple issues with the current notices. See the commit message for details.~~ (Maybe later.)

### DIFF
--- a/js/one-click-download.js
+++ b/js/one-click-download.js
@@ -1,0 +1,102 @@
+(function() {
+    var platform = navigator.platform.toLowerCase();
+
+    var Arch = {unknown: "unknown", x86: "x86", x86_64: "x86_64"};
+        // poor man's enum
+    var arch = Arch.unknown;
+    if (platform.match('x86_64|amd64'))
+    {
+        arch = Arch.x86_64;
+    }
+    else if (platform.match('i.86'))
+    {
+        arch = Arch.x86;
+    }
+
+    var name = '';
+    var file = '';
+
+    if (platform.match('win'))
+    {
+        name = 'Windows';
+        file = 'dmd-' + LATEST + '.exe';
+    }
+    else if (platform.match('mac'))
+    {
+        name = 'OS X';
+        file = 'dmd.' + LATEST + '.dmg';
+    }
+    else if (platform.match('linux'))
+    {
+        name = 'Linux';
+        file = 'dmd.' + LATEST + '.linux.tar.xz';
+
+        // See if the linux distribution is one for which we have a special package.
+        var ua = navigator.userAgent.toLowerCase();
+        if (ua.match('ubuntu|debian'))
+        {
+            if (arch === Arch.x86)
+            {
+                name = 'Ubuntu/Debian i386';
+                file = 'dmd_' + LATEST + '-0_i386.deb';
+            }
+            else if (arch == Arch.x86_64)
+            {
+                name = 'Ubuntu/Debian x86_64';
+                file = 'dmd_' + LATEST + '-0_amd64.deb';
+            }
+        }
+        else if (ua.match('fedora|centos'))
+        {
+            if (arch === Arch.x86)
+            {
+                name = 'Fedora/CentOS i386';
+                file = 'dmd-' + LATEST + '-0.fedora.i386.rpm';
+            }
+            else if (arch == Arch.x86_64)
+            {
+                name = 'Fedora/CentOS x86_64';
+                file = 'dmd-' + LATEST + '-0.fedora.x86_64.rpm';
+            }
+        }
+        else if (ua.match('suse'))
+        {
+            if (arch === Arch.x86)
+            {
+                name = 'openSUSE i386';
+                file = 'dmd-' + LATEST + '-0.openSUSE.i386.rpm';
+            }
+            else if (arch == Arch.x86_64)
+            {
+                name = 'openSUSE x86_64';
+                file = 'dmd-' + LATEST + '-0.openSUSE.x86_64.rpm';
+            }
+        }
+    }
+    else if (platform.match('freebsd'))
+    {
+        if (arch === Arch.x86)
+        {
+            name = 'FreeBSD i386';
+            file = 'dmd.' + LATEST + '.freebsd-32.tar.xz';
+        }
+        else if (arch == Arch.x86_64)
+        {
+            name = 'FreeBSD x86_64';
+            file = 'dmd.' + LATEST + '.freebsd-64.tar.xz';
+        }
+    }
+
+    if (file !== "")
+    {
+        var url = 'http://downloads.dlang.org/releases/2.x/' + LATEST  + '/' +
+            file;
+        $('.notice.download a.download').attr('href', url);
+        $('.notice.download a.download').append(' for ' + name);
+        $('.notice.download')
+            .append(' &middot; ')
+            .append($('<a>')
+                .attr('href', 'download.html')
+                .text('Other Downloads'));
+    }
+})();
--- a/dlang.org.ddoc
+++ b/dlang.org.ddoc
@@ -72,7 +72,10 @@ $(COMMON_HEADERS_DLANG)
 $(EXTRA_HEADERS)
 </head>
 <body id='$(TITLE)' class='$(BODYCLASS)'>
-$(SCRIPT document.body.className += ' have-javascript')
+$(SCRIPT
+    document.body.className += ' have-javascript';
+    var LATEST = "$(LATEST)";
+)
 $(DIVID top,
 	$(DIVID header,
 		<img src="$(STATIC images/hamburger.svg)" id="mobile-hamburger">
@@ -203,8 +206,6 @@ NG_digitalmars_D_announce = <a href="$(NEWS digitalmars.D.announce,$0)">D.announ
 NOTRANSLATE=$(SPANC notranslate, $0)
 NO=
 NOT_EBOOK=$0
-NOTICE_L=<a href="$1" class="notice">$2</a>
-NOTICE=<span class="notice">$1</span>
 _=
 
 OPT=$(SUBSCRIPT opt)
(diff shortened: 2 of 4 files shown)


adamdruppe:
BTW I wrote this script http://arsdnet.net/dcode/install-d.sh to help
a friend install the zip. Maybe we should tweak it a bit more and offer
this on the download too.

She had a ridicuously difficult time actually getting the compiler to
work. The Linux install instructions are notoriously awful, the packages
in the OS tend to be out of date... we really need to work on all this.

JackStouffer:
OS X with safari, chrome, and firefox works.

I like this new feature, but I'm not really a fan of the new design. The fact that the link was at the top of the page and inside a special call-out outbox already adds emphasis to its functionality. Making the text bigger and inside of a red button is just sensory overload at that point. Why not just keep the current download button but change its behavior?

aG0aep6G:
> Why not just keep the current download button but change its behavior?

I think the extra emphasis is warranted. It's supposed to be the very first thing a new visitor notices. Go to dlang.org, click download, follow the instructions, get hacking.

But if I'm the outlier here and everyone else thinks it's too much, I wouldn't mind toning it down to previous levels or somewhere in between.

CyberShadow:
Agreed with @JackStouffer, we use too many font styles already. The unbolding is also not an improvement.

js/one-click-download.js
```diff
@@ -0,0 +1,103 @@
+(function() {
+    var platform = navigator.platform.toLowerCase();
+
+    var Arch = {unknown: "unknown", x86: "x86", x86_64: "x86_64"};
+        // poor man's enum
+    var arch = Arch.unknown;
+    if (platform.match('x86_64|amd64'))
+	{
```
CyberShadow: Nit, mixing tabs and spaces

CyberShadow:
This kinda goes in the opposite direction of https://github.com/D-Programming-Language/dlang.org/pull/1067 :) The point of that is to present people with a choice *before* they see a download link for the reference compiler.

CyberShadow:
Unprefixed global support for flexbox seems to be about 75%, is it perhaps too early to start using flexbox on our front page?

http://caniuse.com/#feat=flexbox

aG0aep6G:
> Agreed with @JackStouffer, we use too many font styles already.

Back to the less flashy style then.

> The unbolding is also not an improvement.

To me, the TWID summary looks way better not bold. The download box also looks nicer to me with only the heading bold. I.e., I think the bottom one looks better than the top one here:

snapshot1

> This kinda goes in the opposite direction of #1067 (edit: also #1116) :) The point of those is to present people with a choice before they see a direct download link for the reference compiler.

I think there's room for both. A quick download button is good for newcomers that just want to try D out. It makes sense on the homepage.

A good comparison on the download page is nice, but forcing newbies to read through it in order to get a compiler is asking too much of them. We don't want D to be for experts only. If one doesn't know what GCC, LLVM, or GDB are, what their computer architecture is, and if "fast compilation" or "strong optimization" should be important to them, then they're going to be lost.

And even if one can make sense of the table, for a first contact it's more important to get any compiler installed quickly. I know that if I'm trying some language out, then I don't care for maximum performance or anything other than that it works, preferably now.

> Unprefixed global support for flexbox seems to be about 75%, is it perhaps too early to start using flexbox on our front page?

Can add prefixes, of course. The fallback isn't so bad either: the boxes are simply laid out vertically.

CyberShadow:
> A good comparison on the download page is nice, but forcing newbies to read through it in order to get a compiler is asking too much of them. We don't want D to be for experts only. If one doesn't know what GCC, LLVM, or GDB are, what their computer architecture is, and if "fast compilation" or "strong optimization" should be important to them, then they're going to be lost.

I made a point to have "Simple installation" in the list.

aG0aep6G:
I dropped all style changes, including the flexbox. Now, do we want a one-click download button or not?

MartinNowak:
Good idea, but didn't quite work for me. My `navigator.platform` using chrome is `Linux x86_64` so I don't get rpm packages on a Fedora Box. I don't think that finding our current download page is a problem though (fat on sidebar) and it explains in more detail how to choose between different compilers.
Also a one-click download that doesn't redirect you to a step-by-step installation guide is a dead end for someone in a hurry.

MartinNowak:
Why was the install script idea dismissed? No need to pipe curl to bash, but downloading and executing an installer script seems more sensible than downloading and showing install instructions.
I'm seeing a good opportunity to get both nightlies and an install done.
http://forum.dlang.org/post/n1v541$r99$1@digitalmars.com

An example of a nice download/instruction instruction btw.
https://www.flockport.com/start/
https://www.flockport.com/install/

aG0aep6G:
> Good idea, but didn't quite work for me. My navigator.platform using chrome is Linux x86_64 so I don't get rpm packages on a Fedora Box.

Yeah, apparently Chrome doesn't include the distro in `platform.userAgent`. Do you agree that linking to download.html would be better than linking to the generic linux .zip when the distro cannot be determined?

> Also a one-click download that doesn't redirect you to a step-by-step installation guide is a dead end for someone in a hurry.

The idea was that it'd be enough to follow the instructions on the screen, because the default action would be to install the package with some GUI. On Windows, that should work with the .exe. On Ubuntu, Chrome gets this part right, while Firefox just saves the file. The default action in the file manager (KDE) is then to install with a little GUI. I'm not sure if that's good enough.

Aside from installation instructions, what about usage instructions? "I installed that D thing. Now what?"

> Why was the install script idea dismissed?
[...]
> http://forum.dlang.org/post/n1v541$r99$1@digitalmars.com

This PR is older than that forum thread.

I gotta say, I'm a bit demotivated about this whole thing, because the browsers are fighting back, and it turns into more work than I had hoped. I wouldn't mind if we went in a different direction, e.g. with an install script.

MartinNowak:
> I wouldn't mind if we went in a different direction, e.g. with an install script.

I'm currently working on nightly builds and tend to prefer a small script that manages downloaded compilers. This what I've in mind at the moment.

```sh
dlang-installer --get dmd # downloads&unpacks LATEST dmd to `~/dlang/dmd-2.069.0` and write env vars into activate script
dlang-installer --get gdc # same for gdc
dlang-installer --list    # ls -lh ~/dlang
dlang-installer --remove dmd-2.069.0 # rm -rf ~/dlang/dmd-2.069.0
dlang-installer --use dmd # does `source ~/dlang/dmd-2.069.0/activate` to setup env vars [virtualenv](http://virtualenv.readthedocs.org/en/latest/userguide.html#usage)
dlang-installer --get dmd-nightly
```
Most of this already exists (and is tested) for [Travis-CI](https://github.com/travis-ci/travis-build/blob/04a5212c35af9f8ba78d2dc5d28fc22855c99f5f/lib/travis/build/script/d.rb) and [Heroku](https://github.com/MartinNowak/heroku-buildpack-d). It would just require some bash work for the command line and some polishing (e.g. to handle cases where people don't have xz installed).
Not sure how feasible something similar is for Windows (maybe with PowerShell?).

CyberShadow:
I am in favor of merging this PR now as-is (or with minor tweaks). An installer download is a workflow that Windows users (or Linux users coming from Windows) expect. Perhaps this could be expanded to show the friendliest installation method for the detected platform (installer download for Windows, RPM/Deb for Debian/Fedora and derivatives, and an install script for everything else).

> Why was the install script idea dismissed?

An install script is definitely not an universal installation method because Windows.

MartinNowak:
I wrote a nice installer script, see https://github.com/D-Programming-Language/installer/pull/162.
Something similar could be done using PowerShell and we already have "environment" shells for dmd and VS.

JackStouffer:
@MartinNowak @CyberShadow given Martin's script, what's the decision on this? I could honestly go either way.

andralex:
@MartinNowak leaving decision about this to you. Thanks all for the work,

MartinNowak:
I think `curl -fsSL https://dlang.org/install.sh | bash -s` is the better one-click installer b/c it actually does the installation and it also covers `dmd-beta`, `dmd-nightly`, `gdc`, and `ldc`.
Will look at integrating that with the download page and we'll get a Powershell version as well.

Thanks for your initiative @aG0aep6G.

CyberShadow:
@MartinNowak 

An install script will still not be a preferred installation method for Windows users, PowerShell or not. 

Also I'm still skeptical whether we really should push an install script as the preferred method. Maybe a newsgroup poll would be appropriate.

MartinNowak:
> Also I'm still skeptical whether we really should push an install script as the preferred method. Maybe a newsgroup poll would be appropriate.

It's an alternative simple install method. There are reasons to prefer a real/package installer, but to simply try out D or switch versions, a script is just about right.

> Maybe a newsgroup poll would be appropriate.

I asked several times for feedback on the installer scripts, this comes a bit late.

CyberShadow:
To clarify, I'm totally fine with adding it, I just don't think we should advertise it as the preferred way. E.g. I think what Rust is doing is fine.

MartinNowak:
> To clarify, I'm totally fine with adding it, I just don't think we should advertise it as the preferred way. E.g. I think what Rust is doing is fine.

Yip, nothing else intended, see here https://github.com/D-Programming-Language/dlang.org/pull/1219.

CyberShadow:
OK, sorry I must've misunderstood that your goal was to put the install script front-and-center instead of the one-click download button :)

PR #1141, opened 2015-10-25
clarify what @property actually does

WalterBright:
I gleaned these from reading the dmd source code. How they worked had changed around so much over the years, I did not know just how @property interacted with UFCS. This clarifies it, or at least enumerates the effects.


aG0aep6G:
`@property` is also documented at http://dlang.org/function.html#property-functions which is linked from <http://dlang.org/attribute.html#property>. We should probably merge things into one.

property.dd
```diff
@@ -247,6 +247,25 @@ $(SECTION3 $(LNAME2 classproperties, User Defined Properties),
...
 
+	$(P Simple getter and setter properties can be written using $(LINK2 function.html#pseudo-member, UFCS).
+	These can be enhanced with the additon of the $(CODE @property) attribute to the function, which
+	adds the following behaviors:
+	)
+
+	$(UL
+	$(LI $(CODE @property) functions cannot be overloaded with non $(CODE @property) functions with the same name.)
+	$(LI $(CODE @property) functions can only have zero, one or two parameters.)
+	$(LI $(CODE @property) functions cannot have variadic parameters.)
```
aG0aep6G: The three points above are also mentioned later on, between the code samples. And it says "Properties may only have zero or one parameter" (not two) there.

WalterBright:
You're right, I need to some more work here.

andralex:
thx for doing this! lmk when ready

WalterBright:
I merged the two descriptions in property.html and function.html into one in function.html. Had to revamp the 'Optional Parentheses' section as well. Removed bogus examples.

function.dd
```diff
@@ -123,7 +123,7 @@ $(H4 $(LNAME2 pure-functions, Pure Functions))
 
         $(P Pure functions are functions which cannot access global or static, mutable
         state save through their arguments. This can enable optimizations based on the fact
-        that a pure function is guaranteed to mutate nothing which isn't passed to it,
+        that a pure function is guaranteed to mutate nothing which is not passed to it,
```
andralex: The (previous) phrasing is quite awkward. Let's take this opportunity to change the text to:

```
This enables optimizations based on the fact that a pure function may at most mutate state reachable through its parameters.
```

function.dd
```diff
@@ -367,146 +367,164 @@ $(H4 $(LNAME2 inout-functions, Inout Functions))
...
-        attribute. They cannot be called with parentheses (hence
-        they act like fields except in some cases) ($(B note:) dmd does not
-        enforce this currently).
-        )
+$(H4 $(LNAME2 optional-parenthesis, Optional Parentheses))
 
-        $(P If a property function has no parameters, it works as a getter.
-        If has exactly one parameter, it works as a setter.
+        $(P If a function call does not take any arguments syntactically,
+        it is callable without parentheses, like a getter $(RELATIVE_LINK2 property-functions, property function).
```
andralex: I get what you're saying here, i.e. include cases like default arguments etc. Here's a more direct way of saying it: replace

"If a function call does not take any arguments syntactically, it is callable without parentheses, like a getter."

with

"If a function call passes no explicit argument, i.e. it would syntactically use $(D ()), then these parentheses may be omitted, similar to a getter invocation."

function.dd
```diff
@@ -367,146 +367,164 @@ $(H4 $(LNAME2 inout-functions, Inout Functions))
...
 
         ---
-        struct S
-        {
-            int m_x;
-            @property
-            {
-                int x() { return m_x; }
-                int x(int newx) { return m_x = newx; }
+        void foo() {}   // no arguments
```
andralex: Add an example with a default parameter, e.g.

```D
void fun(int x = 10) {}
```

function.dd
```diff
@@ -367,146 +367,164 @@ $(H4 $(LNAME2 inout-functions, Inout Functions))
...
+	being the value it is set to.
+	)
+
+	$(P Simple getter and setter properties can be written using $(LINK2 pseudo-member, UFCS).
+	These can be enhanced with the additon of the $(CODE @property) attribute to the function, which
+	adds the following behaviors:
+	)
+
+	$(UL
+	$(LI $(CODE @property) functions cannot be overloaded with non $(CODE @property) functions with the same name.)
```
andralex: Replace space after "non" with en dash.

function.dd
```diff
@@ -367,146 +367,164 @@ $(H4 $(LNAME2 inout-functions, Inout Functions))
...
+	$(LI $(CODE @property) functions can only have zero, one or two parameters.)
```
andralex: use D instead of CODE

andralex:
lgtm after these nits

PR #1144, opened 2015-10-28
function.html: add simple navigation

WalterBright:
Sorely needed.

### DIFF
--- a/function.dd
+++ b/function.dd
@@ -2,6 +2,41 @@ Ddoc
 
 $(SPEC_S Functions,
 
+$(P
+$(RELATIVE_LINK2 contracts, Contracts)
+$(RELATIVE_LINK2 function-return-values, Function Return Values)
+$(RELATIVE_LINK2 function-declarations, Functions Without Bodies)
+$(RELATIVE_LINK2 pure-functions, Pure Functions)
+$(RELATIVE_LINK2 nothrow-functions, Nothrow Functions)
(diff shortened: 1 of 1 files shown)


CyberShadow:
Doesn't look all that great... how about a list?

WalterBright:
I like having it be horizontal because there are a lot of items and a long vertical list is rather unpleasant.

In any case, nobody ever likes my styling. But these pages desperately need navigation, nobody ever adds it, so I suggest pulling it and someone who is ~~more popular~~ better at styling can then style it.

CyberShadow:
Heh, OK.

WalterBright:
thanks

CyberShadow:
Actually, is the navigation at the top even needed? With the `LNAME2` changes by themselves, you can get a link to the section by clicking the section title.

CyberShadow:
https://github.com/D-Programming-Language/dlang.org/pull/1145

WalterBright:
> Actually, is the navigation at the top even needed? With the  LNAME2  changes by themselves, you can get a link to the section by clicking the section title.

I don't know what that means. There was no existing navigation.

CyberShadow:
> I don't know what that means. There was no existing navigation.

This PR changes two things:

1) Each heading is now an anchor, and thanks to some JS is also clickable. You can get the link to a section simply by clicking the heading.

2) It adds the "navigation" blob at the top.

What I meant is that 2) isn't that necessary because of 1).

PR #1145, opened 2015-10-28
function.dd: Table of Contents

CyberShadow:
https://github.com/D-Programming-Language/dlang.org/pull/1144 followup

### DIFF
--- a/function.dd
+++ b/function.dd
@@ -2,40 +2,50 @@ Ddoc
 
 $(SPEC_S Functions,
 
-$(P
-$(RELATIVE_LINK2 contracts, Contracts)
-$(RELATIVE_LINK2 function-return-values, Function Return Values)
-$(RELATIVE_LINK2 function-declarations, Functions Without Bodies)
-$(RELATIVE_LINK2 pure-functions, Pure Functions)
-$(RELATIVE_LINK2 nothrow-functions, Nothrow Functions)
-$(RELATIVE_LINK2 ref-functions, Ref Functions)
-$(RELATIVE_LINK2 auto-functions, Auto Functions)
-$(RELATIVE_LINK2 auto-ref-functions, Auto Ref Functions)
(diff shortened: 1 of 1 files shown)


WalterBright:
I'd rather see it styled like the `Jump to:` menu used in Phobos: http://dlang.org/phobos/std_array.html

I believe that is using a list too, but styled as being horizontal.

CyberShadow:
> I'd rather see it styled like the Jump to: menu used in Phobos: http://dlang.org/phobos/std_array.html

That's really more of a hack than a solution. Ideally we would be using summary tables / cheat sheets like those written manually for std.algorithm/std.range, or those generated by DDox. These horizontal lists are too dense to be usable, are difficult to scan (even though they're sorted you have to search two dimensions), their main benefit right now is that they're the first Ctrl+F occurrence that will lead to the sought symbol.

JackStouffer:
IMO this takes up way too much vertical space. It should be like the "Contents" box on Wikipedia pages, see example here: https://en.wikipedia.org/wiki/D_(programming_language).

Bonus points for making it hide-able.

CyberShadow:
@JackStouffer Do you want to take over this PR? I think this PR is still an improvement over the current intractable mess.

JackStouffer:
> Do you want to take over this PR?

Sure, I'll whip something up tonight.

> I think this PR is still an improvement over the current intractable mess.

It is an improvement, but I think that the correct solution here is to make something that is reusable on other pages, as I noticed when working on the overview page that it needed a contents section as well. The problem with putting something like this on other pages is that users decide in less than one second after the page is loaded whether they will stay on a web page or if they will do something else. Having the actual content of the page below the fold significantly reduces the chances that people will keep reading.

CyberShadow:
> Sure, I'll whip something up tonight.

Any news?

JackStouffer:
@CyberShadow sorry it took so long: https://github.com/D-Programming-Language/dlang.org/pull/1151

PR #1146, opened 2015-10-28
Fix for issue 13624 (Part 3)

JackStouffer:
Rounding out the changes to the overview page, I added in a section that covers the basic ideas of ranges with some examples and a links to more thorough introductions. I think this is enough to mark this issue as fixed.

### DIFF
--- a/overview.dd
+++ b/overview.dd
@@ -757,6 +757,213 @@ long[] func(int x); // declares a function returning an array of longs
     )
 )
 
+
+$(SECTION3 Ranges,
+    $(P D uses the concept of a range in lieu of iterators or generators found in
+    other languages. A range is any type that provides a common interface to a
+    sequence of values. The purpose of a range is to allow for a simpler way to
+    write code that works on arbitrary data, thereby making it reusable.
+    )
+
+    $(P The most basic type of range is called an input range,
+    which provides three methods.
+    )
+
+-----------------------
+struct MyRange
+{
+    auto front()
+    {
+        // return the next value in the sequence
+    }
+
+    void popFront()
+    {
+        // move the front of the sequence to the next value
+    }
+
+    bool empty()
+    {
+        // true if the range has no more values to return
+    }
+}
+-----------------------
+
+    $(P To understand the power of this simple interface, let's run through an
+    example. Say we wanted to write a program that took all of the employees in
+    a company, took out the ones younger than 40, and grouped the remaining into
+    an array of arrays by their organization.
+    )
+
+-----------------------
+struct Employee
+{
+    uint id;
+    uint organization_id;
+    string name;
+    uint age;
+}
+
+struct Employees
+{
+    Employee[] data;
+
+    this(Employee[] employees)
+    {
+        employees = data;
+    }
+
+    Employee front()
+    {
+        return data[0];
+    }
+
+    void popFront()
+    {
+        data = data[1 .. $];
+    }
+
+    bool empty()
+    {
+        return data.length == 0;
+    }
+}
+-----------------------
+
+    $(P Here the data is coming from a constructor as a simple example, but it
+    could come from any source, like a CSV or a database.
+    )
+
+    $(P Please note that this code should not be used in actual code, because In D,
+    the basic dynamic array also acts as a range, so any algorithm that accepts
+    ranges also accepts arrays. However, static arrays are not considered ranges,
+    as the operation $(D popFront) is based on mutating the length of the range,
+    which is impossible with static arrays. To get a range out of a static array,
+    create a slice containing all of its elements, like so:
+    )
+
+-----------------------
+int[4] array = [1, 2, 3, 4]; // not a range
+array[]; // valid range
+-----------------------
+
+
+    $(P Now that the range is defined, we can populate it and write the filtering code.
+    )
+
+-----------------------
+void main()
+{
+    import std.algorithm.iteration : filter, chunkBy;
+
+    Employees employees = Employees([
+        Employee(1, 1, "George", 50),
+        Employee(2, 3, "John", 65),
+        Employee(3, 2, "David", 40),
+        Employee(4, 1, "Eli", 40),
+        Employee(5, 2, "Hal", 35)
+    ]);
+
+    auto older_employees = employees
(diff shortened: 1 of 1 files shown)


JackStouffer:
Fun fact: with this change, the overview page is the only place in the entire website (outside of Phobos Docs) that describes what ranges are or what they are for. I don't know how the site went this long without a page on ranges.

overview.dd
```diff
@@ -757,6 +757,199 @@ long[] func(int x); // declares a function returning an array of longs
     )
 )
 
+
+$(SECTION3 Ranges,
+    $(P D uses the concept of a range in lieu of iterators or generators found in
+    other languages. A range is any type that provides a common interface to a
+    sequence of values. The purpose of a range is to allow for a simpler way to
+    write code that works on an arbitrary data, thereby making it reusable.
```
CyberShadow: "an arbitrary data" - data is not a singular noun (remove "an")

overview.dd
```diff
@@ -757,6 +757,199 @@ long[] func(int x); // declares a function returning an array of longs
...
+    {
+        // return the next value in the sequence
+    }
+
+    void popFront()
+    {
+        // move the front of the sequence to the next value
+    }
+
+    bool empty()
```
CyberShadow: Should this and `front` be `@property`?

overview.dd
```diff
@@ -757,6 +757,199 @@ long[] func(int x); // declares a function returning an array of longs
...
+    {
+        // true if the range has no more values to return
+    }
+}
+-----------------------
+
+    $(P To understand the power of this simple interface, let's run through an
+    example. Say we wanted to write a program that took all of the employees in
+    a company, took out the ones younger than 40, and grouped the remaining into
+    and array of arrays by their organization.
```
CyberShadow: s/and/an/

overview.dd
JackStouffer: Probably not since function attributes aren't covered in this page.

overview.dd
```diff
@@ -757,6 +757,199 @@ long[] func(int x); // declares a function returning an array of longs
...
+-----------------------
+struct Employee
+{
+    uint id;
+    uint organization_id;
+    string name;
+    uint age;
+}
+
+struct Employees
```
CyberShadow: I'm not sure this is a great example. It is essentially wrapping an array, which is already a range by itself.

overview.dd
```diff
@@ -757,6 +757,199 @@ long[] func(int x); // declares a function returning an array of longs
...
+
+    auto older_employees = employees
+        .filter!(a => a.age > 40)
+        .chunkBy!((a,b) => a.organization_id == b.organization_id);
+}
+-----------------------
+
+    $(P All of the algorithms in std.algorithm work with ranges to avoid the
+    problem of rewriting common functionality for every project. std.algorithm
+    implements sorts, filters, maps, reduces, and more.
```
CyberShadow: "reduces"? "reductions"? Better to rephrase, I think

overview.dd
```diff
@@ -757,6 +757,199 @@ long[] func(int x); // declares a function returning an array of longs
...
+    writeln(employee);
+}
+-----------------------
+
+    $(P which is equivalent to)
+
+-----------------------
+for(; !employees.empty; employees.popFront())
+{
+    writeln(employees.front());
```
CyberShadow: `front()` won't work for ranges where `front` is an actual field.

overview.dd
JackStouffer: Other than introducing the reader to a Phobos module, I'm not sure what else could be used as I want to have code that someone can copy/paste and have it compile immediately so people can play with it.

CyberShadow: What about a generative range? E.g. the Fibonacci sequence?

JackStouffer: Then we would have to go into finite ranges vs infinite ranges. Also, it's nice to show something to the reader which they are likely to have to do at some point, in this case filtering and grouping. 

I see only three satisfactory solutions to this:

1. Leave it the way it is
2. Show the values coming from a file or a csv and the range parsing the values and returning employee structs from `front`
3. Show code that is just an example and does not actually compile

CyberShadow: What did Ali use in his book?

If this example is kept, at least the text should say near the beginning that this is an illustrative example. We don't want people to think that they need to write a range by hand just to map/filter an array.

JackStouffer: > What did Ali use in his book?

This same code, but with a student and school structs instead of employee and employees.

```
struct Student {
    string name;
    int number;

    string toString() const {
        return format("%s(%s)", name, number);
    }
}

struct School {
    Student[] students;

    @property bool empty() const {
        return students.length == 0;
    }

   @property ref Student front() {
        return students[0];
    }

    void popFront() {
        students = students[1 .. $];
    }
}
```

> If this example is kept, at least the text should say near the beginning that this is an illustrative example.

Ok, I have a sentence later saying that arrays are ranges, so I'll just move that.

CyberShadow:
BTW, I don't want to appear as a gatekeeper for these PRs. What I post here is my personal opinion and not really requirements for the PR to be merged. I can only guess that everyone else is either too busy or don't even know that this PR exists to add their opinions.

JackStouffer:
Added sentence before the example to clarify.

overview.dd
```diff
@@ -757,6 +757,206 @@ long[] func(int x); // declares a function returning an array of longs
...
+}
+-----------------------
+
+    $(P Here the data is coming from a constructor as a simple example, but it
+    could come from any source, like a CSV or a database.
+    )
+
+    $(P Please note that this code should not be used in actual code, because In D,
+    the basic dynamic array also acts as a range, so any algorithm that accepts
+    ranges also accepts arrays. However, static ranges are not considered ranges,
```
CyberShadow: s/static ranges/static arrays/ ?

JackStouffer:
@CyberShadow fixed, and added small code section on how to get a range from a static array.

CyberShadow:
Test comment please ignore

CyberShadow:
No feedback from anyone else? Let's merge.

JackStouffer:
Thanks!

PR #1149, opened 2015-11-10
Fix small bug in lazy-evalution.dd

JohanEngelen:
Fix a small bug in a code example, so that the code matches with the given console output.


JackStouffer:
LGTM

PR #1151, opened 2015-11-16
function.dd: Table of Contents 2

JackStouffer:
Alternative solution to https://github.com/D-Programming-Language/dlang.org/pull/1145.

~~I was unable to float the contents section to the right or left and have the contents of the page reflow around it without making the `pre` sections become only as wide as their content. This is *fine* as it is, but if someone has a solution to that, that would be awesome.~~

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1,20 +1,20 @@
 html
 {
-	height: 100%; /* of the viewport */
+    height: 100%; /* of the viewport */
 }
 
 body
 {
-	padding: 0;
-	border: 0;
-	color: black;
-	background-color: #1f252b;
-	background-repeat: no-repeat;
-	font-size: 100%;
-	font-family: Verdana, "Deja Vu", "Bitstream Vera Sans", sans-serif;
-	margin: 0px auto;
-	position: relative; /* So that absolute/fixed positions cling to this. */
-	min-height: 100%;
+    padding: 0;
+    border: 0;
+    color: black;
+    background-color: #1f252b;
+    background-repeat: no-repeat;
+    font-size: 100%;
+    font-family: Verdana, "Deja Vu", "Bitstream Vera Sans", sans-serif;
+    margin: 0px auto;
+    position: relative; /* So that absolute/fixed positions cling to this. */
+    min-height: 100%;
 }
 
 /* Pseudo-padding. Can't use padding-top, because that's added to height.
@@ -22,9 +22,9 @@ Can't use margin-top on children, because it collapses with body's, and
 consequently is added to height. */
 body::before
 {
-	content: "";
-	display: block;
-	height: 1em;
+    content: "";
+    display: block;
+    height: 1em;
 }
 
 /* show signature download on hover */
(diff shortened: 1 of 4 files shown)


JackStouffer:
Another thing: barring the ability to properly float this, it would be best if there was content above the contents box as to draw people into the page a-la Wikipedia pages, but I don't think that is too much of a problem on this page as it's mainly for reference for those already familiar to the language.

JackStouffer:
@CyberShadow I'm not quite sure how to parse that Latex error as it doesn't seem to give the problem location.

CyberShadow:
For one, you can't use raw HTML in the spec pages, because latex is not some variant of HTML.

JackStouffer:
Any way to exclude it from the Latex generation? I'm already excluding it from the printed version of the HTML.

CyberShadow:
I see a `NOT_EBOOK` macro that could do that.

CyberShadow:
I don't think you need the raw HTML, though. You can use one CSS class with `DIVC` and `.page-contents > div` / `.page-contents > ol` CSS rules.

JackStouffer:
Everything looks good now.

aG0aep6G:
> I was unable to float the contents section to the right or left and have the contents of the page reflow around it without making the pre sections become only as wide as their content. This is fine as it is, but if someone has a solution to that, that would be awesome.

I guess the problem is that when you give `float` to the contents box, the first grammar boxes get behind it, right?

```css
.page-contents {float: right; margin: 0 0 2em 2em;}
pre {overflow: auto; /* i.e. not 'visible', so that it gets its own block formatting context */}
```

JackStouffer:
> I guess the problem is that when you give float to the contents box, the first grammar boxes get behind it, right?

>.page-contents {float: right; margin: 0 0 2em 2em;}
pre {overflow: auto; /* i.e. not 'visible', so that it gets its own block formatting context */}

That fixed it. Thanks!

Code is now updated with this style.

aG0aep6G:
For consistency's sake I suggest to style the contents box similar to the jump links on phobos pages (e.g. http://dlang.org/phobos/std_algorithm.html). That is, give the box the `quickindex` class. And either drop the borders or imitate tables with `border: solid black; border-style: 2px 0;`.

aG0aep6G:
I think the font-size is too tiny. It's also different from all the others. If it has to be smaller than normal text, then maybe make it the same as in the dropdown menus (12px).

JackStouffer:
I changed the font size, but adding a 2px black border to the box made it very loud when you should be able to ignore it if you don't need it. Also, adding the `quickindex` class does nothing.

css/style.css
```diff
@@ -1335,6 +1336,47 @@ div#mobile-top, div#mobile-search, img#mobile-hamburger
...
+	display: inline;
+	color: #444;
+	font-weight: bold;
+	margin-top: 0;
+}
+
+.page-contents li
+{
+	line-height: 1.3;
+	padding-bottom: 0 !important;
```
aG0aep6G: Instead of taking out the `!important` sledgehammer, I suggest making the selector more specific so that it beats the other one. The other one is `div#content li`, so `div#content .page-contents li` should get us there.

css/style.css
```diff
@@ -1335,6 +1336,47 @@ div#mobile-top, div#mobile-search, img#mobile-hamburger
...
+}
+
+.page-contents li a
+{
+	text-decoration: none;
```
aG0aep6G: `.quickindex` should take care of this already.

css/style.css
```diff
@@ -1335,6 +1336,47 @@ div#mobile-top, div#mobile-search, img#mobile-hamburger
...
+}
+
+.page-contents li a:hover
+{
+	text-decoration: underline;
```
aG0aep6G: Ditto, `.quickindex` does this.

function.dd
```diff
@@ -2,40 +2,56 @@ Ddoc
...
-$(RELATIVE_LINK2 safe-functions, Safe Functions)
-$(RELATIVE_LINK2 trusted-functions, Trusted Functions)
-$(RELATIVE_LINK2 system-functions, System Functions)
-$(RELATIVE_LINK2 function-attribute-inference, Function Attribute Inference)
-$(RELATIVE_LINK2 pseudo-member, Uniform Function Call Syntax (UFCS))
+$(NOT_EBOOK 
+$(DIVC page-contents quickindex,
+    $(DIVC page-contents-header,
+        $(H3 Contents)
+        <span><a href="javascript:void(0);">[hide]</a></span>
```
aG0aep6G: I prefer to insert buttons like this with JS. Otherwise users that don't have JS end up with a button that does nothing. Could be I'm just being pedantic for no good reason, though.

JackStouffer: I generally think it's best practice to avoid putting HTML generation into JS, unless you have separate view files a la most JS MVC frameworks. Separation of concerns and all that.

JackStouffer:
Addressed concerns. Also converted tabs to four spaces as per the style guide in a separate commit.

function.dd
aG0aep6G: I don't agree. I think interface elements that only have a function with JS should be generated by JS, with or without a helping framework. I don't see that separation of concerns is a good reason to let it slide here.

But I'm a bit of a purist when it comes to supporting non-JS users, and I understand that others put less priority to it than I do.

CyberShadow: We have a `have-javascript` CSS class, which is only enabled if JS is actually working.

andralex: At any rate, please prefer macros to raw HTML tags.

JackStouffer:
Updated to insert hide button with JS.

CyberShadow:
(Reposting on the correct PR this time)

screen shot 2015-11-23 at 19 14 32

Doesn't look great at certain page widths.

JackStouffer:
@CyberShadow fixed.

CyberShadow:
Sorry, this broke with #896's merge, please rebase.

JackStouffer:
@CyberShadow Successfully rebased.

CyberShadow:
LGTM!

CyberShadow:
Thanks!

JackStouffer:
Thank you.

ntrel:
Hi, is the table of contents manually generated? I think we need some kind of automated process.

PR #1152, opened 2015-11-16
Left Align Text

JackStouffer:
I know that discussing styling decisions on the internet is a road full of personal opinion and flame wars, so I want to keep this discussion purely practical by stating this: justified text is harder to read on the web than left aligned text.

A good summary of the problems and the evidence behind them can be found here: https://kaiweber.wordpress.com/2010/05/31/ragged-right-or-justified-alignment/. TL;DR is that it makes little difference for people who can read well, but for people with dyslexia or other reading disabilities, justified text is much harder for them to comprehend.

Also, the reason that most books are justified is they are done by hand or reviewed and edited by someone after a automatic process is finished. Automatic justifying software is just not up to the task in most cases, creating lots of weird hyphenations, off line breaks, and rivers in text, e.g. https://en.wikipedia.org/wiki/River_(typography)#/media/File:Typographic_river.svg 

For some less scientific evidence, if you look on the web at sites that are purely designed to have a good reading experience for long form content, Medium, Feedly, Pocket, Instapaper, Readability, and Safari's built-in Reader View, they all use left aligned text.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -801,8 +801,18 @@ div#content
     background-color: #f6f6f6;
     font-size: 0.875em;
     line-height: 1.4em;
-    text-align: justify;
     border: 0.25em solid #ccd5d5;
+}
+
+/*  This is a CSS hack to only target Blink based browsers
+    Targets Chrome 28 and later and Opera 14 and later
+*/
+@media all and (-webkit-min-device-pixel-ratio:0) and (min-resolution: .001dpcm) {
+    div#content
+    {
+        /* because Blink based browsers don't support hyphens */
+        text-align: left;
+    }
 }
 
 body#Home div#content


eco:
@andralex is passionate about hyphenated, double justified text (there is quite a lot of discussion on the forums about this). Chrome is more or less the only major browser left that doesn't support `hyphens`. I've starred the issue on Chromium about implementing it and it sounds like it'll be early next year.

We used, for a long time, a javascript hyphens emulating script that was terrible and we finally ditched it. If you want to try to convince Andrei of doing this I suspect it'll have to be Chrome only and have to be reverted once Chromium implements `hyphens`.

andralex:
I'm okay with left aligned on Chrome only. Thanks.

CyberShadow:
IIRC Andrei said that an article written by a typography / web design expert could change his mind, but when we discussed the issue last time there was no such piece that would conclusively weigh it either way. Perhaps it might be worth asking on StackExchange / Quora. I personally also agree that hyphenation doesn't belong on the web, but as it is it's opinion versus Andrei's background in typography.

JackStouffer:
@eco @andralex @CyberShadow changed CSS to only target chrome.

This was tested on Chrome and Safari on OS X, so if you can tell me if it works on your system that would be great.

CyberShadow:
screen shot 2015-11-23 at 19 14 32

Doesn't look great at certain page widths.

JackStouffer:
@CyberShadow Did you accidentally comment on the wrong PR?

CyberShadow:
Oops, yes :)

andralex:
I see this removes `text-align: justify;` and adds a left aligned under certain circumstances, so how does this preserve justification for non-Chrome browsers?

JackStouffer:
@andralex If you look in the diff section that's not shown by github by default, you'll see that for some reason, the `text-align: justify;` was included twice, so I removed the extra declaration. The original is still at the top of the CSS rule.

css/style.css
```diff
@@ -801,10 +801,20 @@ div#content
...
 	font-size: 0.875em;
 	line-height: 1.4em;
-	text-align: justify;
 	border: 0.25em solid #ccd5d5;
 }
 
+/*  This is a CSS hack to only target Blink based browsers
+    Targets Chrome 28 and later and Opera 14 and later
+*/
+@media all and (-webkit-min-device-pixel-ratio:0) and (min-resolution: .001dpcm) {
```
CyberShadow: Ouch at this hack. There has to be a better way, even if it involves JS. Ideally a bit of JS will detect if hyphenation is actually supported by the browser or not, I think hyphenate.js has such a chechk.

JackStouffer: I wanted to avoid JS if at all possible because I don't like putting style inside of anything but the CSS, as it has a high probability of being forgotten and left in there for a while. And I figured that it wasn't so bad as this will probably be removed soon when Chrome supports hyphens.

If you still want me to change this to a JS solution even with the above problems, please let me know.

CyberShadow:
Please rebase! (I guess due to detabbing in your other PR?)

JackStouffer:
@CyberShadow fixed

JackStouffer:
Thanks.

PR #1153, opened 2015-11-18
Advise std.typecons.Typedef instead of typedef

JohanEngelen:
typedef is deprecated (http://dlang.org/deprecate.html#typedef) and so this text needed updating.


JackStouffer:
LGTM

andralex:
Thanks!

PR #1154, opened 2015-11-19
Update cpp_interface.dd for recent developments

JakobOvrum:
* Removes old assumptions
 * Specifies the new type and template stuff, with examples
 * Updated memory allocation section
 * Updated data type table

Depends on: [issue 15372](https://issues.dlang.org/show_bug.cgi?id=15372). The template example doesn't link.

The first commit contains only the tabs to spaces changes, so it's probably easier to review using the diff for the second commit.

### DIFF
--- a/latex.ddoc
+++ b/latex.ddoc
@@ -11,6 +11,7 @@ B=\textbf{$0}
 BIG=\large{$0}
 BLACK={\color{black}$0}
 BLUE={\color{blue}$0}
+BUGZILLA=\href{https://issues.dlang.org/show_bug.cgi?id=$0}{Bugzilla $0}
 _=
 
 BLOCKQUOTE_BY=\begin{quote}
--- a/spec/cpp_interface.dd
+++ b/spec/cpp_interface.dd
@@ -2,68 +2,68 @@ Ddoc
 
 $(SPEC_S Interfacing to C++,
 
-	$(P
-	While D is fully capable of
-	$(DDLINK spec/interfaceToC, Interfacing to C, interfacing to C),
-	its ability to interface to C++ is much more limited.
-	There are three ways to do it:
-	)
+    $(P
+    While D is fully capable of
+    $(DDLINK spec/interfaceToC, Interfacing to C, interfacing to C),
+    its ability to interface to C++ is much more limited.
+    There are three ways to do it:
+    )
 
-	$(OL
+    $(OL
 
-	$(LI Use C++'s ability to create a C interface, and then
-	use D's ability to
-	$(DDLINK spec/interfaceToC, Interfacing to C, interface with C)
-	to access that interface.
-	)
+    $(LI Use C++'s ability to create a C interface, and then
+    use D's ability to
+    $(DDLINK spec/interfaceToC, Interfacing to C, interface with C)
+    to access that interface.
+    )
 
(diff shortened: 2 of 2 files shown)


cpp_interface.dd
```diff
@@ -296,328 +340,406 @@ class E
...
+    $(ARGS $(B unsigned char))
+    )
+
+    $(TROW
+    $(ARGS $(B char)),
+    $(ARGS $(B char) (chars are unsigned in D))
+    )
+
+    $(TROW
+    $(ARGS $(B core.stdc.stddef.wchar_t)),
```
JackStouffer: This line seems to be causing some error in the LaTex generation.

andralex: It's because of the underscore I think. Use `$(UNDERSCORE)` instead.

CyberShadow: Or use `$(D ...)`, as is more appropriate here because you're referring to an identifier.

JakobOvrum: Yeah, that makes sense. I changed it to use `$(D ...)` and the test passes now.

JakobOvrum:
I filed the linking issue: https://issues.dlang.org/show_bug.cgi?id=15372

JackStouffer:
So this is blocked by 15372? If so it would be useful if someone added the "Blocked" tag to this.

JakobOvrum:
Well, I could just remove function templates from the article. I just can't claim it works when it doesn't.

JackStouffer:
I'd say add a note about bug 15372 and then merge this, as this is how it should work baring bugs and that's how we document everything else.

JackStouffer:
@JakobOvrum rebase please

JakobOvrum:
Is the `BUGZILLA` macro broken for latex output?

CyberShadow:
> Is the BUGZILLA macro broken for latex output?

Seems so.

deadalnix:
Is there a way to see the page somewhere ?

JakobOvrum:
This PR now depends on #1194 for `MREF`.

> Is there a way to see the page somewhere ?

The build is failing because of a missing LaTeX macro (added in #1194), so I guess not.

adamdruppe:
I just merged 1194, if the test re-runs here, ping me and I'll merge this too.

tsbockman:
@adamdruppe The test has passed now.

tsbockman:
Thanks for this. When will this actually go live on the website?

CyberShadow:
The next time I, Martin or Andrei syncs it.

Would you like me to sync it now?

tsbockman:
@CyberShadow 
> Would you like me to sync it now?

That would be nice. Thanks.

People ask about C++ support from time to time on the forums, and I'd like to finally be able to just say, "go read the spec".

CyberShadow:
Done: http://dlang.org/spec/cpp_interface.html

tsbockman:
Thanks. Any idea why "*COM interface*" at the top of the page is a dead link?

CyberShadow:
It seems to be using the wrong macro. But http://dlang.org/COM.html is just a redirect to a wiki page with one line in it. The link should probably just be removed.

tsbockman:
@CyberShadow 
> The link should probably just be removed.

I made a PR: https://github.com/D-Programming-Language/dlang.org/pull/1249

PR #1155, opened 2015-11-23
Fix version number in DDOX based navigation section.

s-ludwig:
The version "2.067.1" stayed hard-coded since the new website layout was converted to Diet format.


CyberShadow:
Thanks!

PR #1156, opened 2015-11-23
Update doc.ddoc

bachmeil:
- Promoted Ali's book and the wiki to top level of the sidebar menu [as discussed in the forum](http://forum.dlang.org/post/n2ttcv$uti$1@digitalmars.com). 
- Removed "Change Log" from top level, as that is already available under "Downloads" and  there is no obvious reason to leave it at top level. (Also discussed at the link above.)

Note: This is my first real pull request here so let me know of any problems.


JackStouffer:
Looks like Andrei already pseudo-signed off on this change, and the code looks good so LGTM.

bachmeil:
> Looks like Andrei already pseudo-signed off on this change, and the code looks good so LGTM. 

He did on the first. The second was suggested by someone else and I thought it made sense.

andralex:
thanks!

PR #1157, opened 2015-11-23
Remove wc example from Books & Articles submenu

bachmeil:
It doesn't fit there and I don't understand why that example is in the menus at all. wc.dd is still there, so a link can be added elsewhere if desired.


andralex:
alrighty

PR #1158, opened 2015-11-25
Cleaned up css files

JackStouffer:
* ~~removed overqualifed css rules~~
* removed commented out old code
* changed `0px` and `0em` to `0`
* added some extra prefixed css rules
* made long lines under 80 characters

### DIFF
--- a/css/print.css
+++ b/css/print.css
@@ -11,7 +11,16 @@ pre
 	font-size: 100%;
 }
 
-#search-box, #header, #navigation, #lastupdate, div#tools, #footernav, #copyright, #github-ribbon, div#top, .page-contents
+#search-box,
+#header,
+#navigation,
+#lastupdate,
+#tools,
+#footernav,
+#copyright,
+#github-ribbon,
+#top,
+.page-contents
 {
 	display: none;
 }
--- a/css/style.css
+++ b/css/style.css
@@ -12,7 +12,7 @@ body
     background-repeat: no-repeat;
     font-size: 100%;
     font-family: Verdana, "Deja Vu", "Bitstream Vera Sans", sans-serif;
-    margin: 0px auto;
+    margin: 0 auto;
     position: relative; /* So that absolute/fixed positions cling to this. */
     min-height: 100%;
 }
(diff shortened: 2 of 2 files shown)


JackStouffer:
This speeds up rendering time on Chrome by 76ms (I only did one test though).

CyberShadow:
I have not measured any performance improvements at all. As such, I think dequalifying CSS rules provides no benefit, and worsens readability. It might also change behavior in style sheets that build upon style.css, such as forum.dlang.org's stylesheet.

JackStouffer:
>I have not measured any performance improvements at all.

It may very per browser, but these rules are performing less searches than their original versions, as

```
element#selector
```

does two searches while

```
#selector
```

performs one, so it must be faster.

> As such, I think dequalifying CSS rules provides no benefit, and worsens readability. 

I understand that this goes down to personal preference, but for me overqualified rules are just noise, as 

```
div#content
```
doesn't give you anymore *useful* information than
```
#content
```

CyberShadow:
> It may very per browser,

I tested with Chrome.

> so it must be faster

You don't know that. The browser might be optimizing this as there can be at most one ID per document anyway.

Also, if it's faster by an insignificant amount, then it has no practical gain.

> but for me overqualified rules are just noise

Maybe in that example. In some other cases, it can be informative, e.g. when the element is something other than a div or span.

JackStouffer:
@CyberShadow Ok, so I removed all of my changes where the element being styled was not either a div or a span, where the class name did not include the type of element being styled, or where the rule was not an id. I think these changes help remove noise while keeping things unambiguous for most rules.

CyberShadow:
You haven't addressed the main concern:

> It might also change behavior in style sheets that build upon style.css, such as forum.dlang.org's stylesheet.

These changes change CSS specificity and as such are potentially breaking changes. Please just leave them out.

JackStouffer:
@CyberShadow fixed

css/print.css
```diff
@@ -11,12 +11,21 @@ pre
...
+#copyright,
+#github-ribbon,
+#top,
+.page-contents
 {
 	display: none;
 }
 
-div#content
+#content
```
CyberShadow: Still not fixed

css/style.css
```diff
@@ -184,7 +163,7 @@ table.download-compilers h3
     text-align: center;
 }
 
-div#download-choose
+#download-choose
```
CyberShadow: Ditto

css/style.css
```diff
@@ -274,7 +255,7 @@ h3.download
     font-weight: bold;
 }
 
-div.summary, div.description, div.keyval
+.summary, .description, .keyval
```
CyberShadow: Ditto. I see more in the diff.

JackStouffer:
Sorry about that. I tried using a GUI git tool and it seems most of my discards didn't take.

JackStouffer:
Seems to have worked this time

CyberShadow:
Thx!

PR #1159, opened 2015-11-26
set ROOT_DIR=/ for DDox pages

aG0aep6G:
DDox generates a nested directory structure, so a relative ROOT_DIR, as is
used for /phobos/ pages, doesn't work.

This fixes links on http://dlang.org, but not in local mirrors (releases,
git clones). There it actually even breaks some links that happened to
work before.


CyberShadow:
Thanks!

PR #1161, opened 2015-12-03
bugstats.php.dd: Add hidden iframe to trigger generating the image

CyberShadow:
I've regularly noticed that the graph image linked to from bugstats.php
would not display - however, it would start displaying for a while when
you visit the page it links to. Since the image URL seems to link to a
static file, I guess that the graphs are generated by the CGI request,
and the static files are cleaned up periodically. Thus, to have the image
display on another website (dlang.org), we also need to regularly trigger
its generation, one way being to make the CGI request to generate the
image at the same time as when requesting the image.


andralex:
Cool solution, I noticed the problem but didn't know how to fix it.

andralex:
thx!

JackStouffer:
This did not fix the problem, I just had to regenerate the image by following the link posted on bugzilla manually.

CyberShadow:
Have you tried reloading the page first? It might be that the PNG was requested before the webpage.

JackStouffer:
I believe I did, but it's to late to test now.

No luck finding the root cause of this? If push comes to shove we could just have something that pings the graph page on bugzilla every three days.

CyberShadow:
We don't even know what the expiration time is.

Probably the "proper" way to fix this would involve looking at the Bugzilla
source.

PR #1162, opened 2015-12-06
add search option for language reference

aG0aep6G:
And make it the default on language reference pages.

Google has picked up the new locations, time to do this.


CyberShadow:
Thanks!

PR #1166, opened 2015-12-13
use https: for wikipedia links

WalterBright:


CyberShadow:
:+1:

WalterBright:
We should try https: for all the external links, and switch to them for ones that work.

PR #1174, opened 2015-12-18
Typo error, bar instread of foo

Bolpat:
Typo error, bar instread of foo


Hackerpilot:
:white_check_mark:

andralex:
thx!

Bolpat:
The [Functions page](http://dlang.org/spec/function.html#inout-functions) will be updated, right? Or is there something else to do?

PR #1177, opened 2015-12-30
MREF macro for linking to modules

adamdruppe:
Usage: `$(MREF std,algorithm,searching)` or `$(MREF std,file)`. Returns a link to the given module. Unlike the existing cases of `$(LINK2)`, this also works in ddox as well as ddoc.

### DIFF
--- a/std-ddox.ddoc
+++ b/std-ddox.ddoc
@@ -3,9 +3,7 @@ XREF = std.$1.$2
 XREF_PACK = std.$1.$2.$3
 CXREF = core.$1.$2
 ECXREF = etc.$1.$2
-MREF_HELPER=/$1$(MREF_HELPER $+)
-MREF_HELPER2=.$1$(MREF_HELPER2 $+)
-MREF=<a href="/library/$1$(MREF_HELPER $+).html">$1$(MREF_HELPER2 $+)</a>
+MREF=<a href="/library/$1$(SLASH_PREFIXED $+).html">$1$(DOT_PREFIXED $+)</a>
 LREF = $1
 ROOT_DIR=/
 _=
--- a/std-ddox.ddoc
+++ b/std-ddox.ddoc
@@ -3,6 +3,9 @@ XREF = std.$1.$2
 XREF_PACK = std.$1.$2.$3
 CXREF = core.$1.$2
 ECXREF = etc.$1.$2
+MREF_HELPER=/$1$(MREF_HELPER $+)
+MREF_HELPER2=.$1$(MREF_HELPER2 $+)
+MREF=<a href="/library/$1$(MREF_HELPER $+).html">$1$(MREF_HELPER2 $+)</a>
 LREF = $1
 ROOT_DIR=/
 _=
--- a/std.ddoc
+++ b/std.ddoc
@@ -13,6 +13,9 @@ _=
 CXREF = <a href="core_$1.html#$2">$(D core.$1.$2)</a>
 ECXREF = <a href="etc_c_$1.html#$2">$(D etc.c.$1.$2)</a>
 LREF = <a href="#$1">$(D $1)</a>
+MREF_HELPER=_$1$(MREF_HELPER $+)
+MREF_HELPER2=.$1$(MREF_HELPER2 $+)
+MREF=<a href="$1$(MREF_HELPER $+).html">$1$(MREF_HELPER2 $+)</a>
 _=
 
 LEADINGROW = <tr class="leadingrow"><td colspan="2"><b><em>$(NBSP)$(NBSP)$(NBSP)$(NBSP)$0</em></b></td></tr>
--- a/std.ddoc
+++ b/std.ddoc
@@ -13,9 +13,10 @@ _=
 CXREF = <a href="core_$1.html#$2">$(D core.$1.$2)</a>
 ECXREF = <a href="etc_c_$1.html#$2">$(D etc.c.$1.$2)</a>
 LREF = <a href="#$1">$(D $1)</a>
-MREF_HELPER=_$1$(MREF_HELPER $+)
-MREF_HELPER2=.$1$(MREF_HELPER2 $+)
-MREF=<a href="$1$(MREF_HELPER $+).html">$1$(MREF_HELPER2 $+)</a>
+SLASH_PREFIXED=/$1$(SLASH_PREFIXED $+)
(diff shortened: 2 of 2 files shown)


JakobOvrum:
Tangential to this PR, can we use the same trick to make `XREF` variadic?

adamdruppe:
Sort of, but you hit the problem where at a certain point, a name stops being the module and starts being a reference inside it.

Consider `$(XREF std,stdio,File,writeln)` vs `$(XREF std,algorithm,searching,countUntil)`. They are both valid references and both have the same number of arguments, but the correct link for the former is `std_stdio.html#File.writeln` - with two args past the hash, and the correct link for the latter is `std_algorithm_searching.html#countUntil`, with just one arg past the hash. There's a similar worry on ddox, where the link is `http://dlang.org/library/std/stdio/file.writeln.html`. Where's the dot go rather than slashes?

(And oh the horror, ddox converts camel case to underscore names! How are we possibly going to express *that* in a ddoc macro?)


How do you tell the difference?  I think you'd be forced to use several macros to account for that (unless we changed the link structure... which I'm doing with my new docs, to avoid these problems. But I don't think that's realistic for Phobos right now, we still want some kind of transition macro at least to avoid breaking all the links in the mean time).

adamdruppe:
BTW note that my recursive trick does *not* work when referencing `object`, or any other single-name module. But since `object` is the only case here, all the others start with `core`, `etc`, or `std`, I'm ok with special casing that when writing docs.

Unless any of you know an easy trick that makes it work in the recursive macro, but I don't.

CyberShadow:
If we are going to introduce new macros, can we get rid of some old ones (STDMODREF, COREMODREF) at the same time? This will also put the additions to the test.

adamdruppe:
I have never even seen those before. Grep tells me they have only been used in the changelog... and seem to be only defined in the changelog.

Alas though, that's spread across three repos, so I'll leave that to someone else. It would probably be better to make that change after merging the new ones so there's no breakage in the mean time.

CyberShadow:
OK, my (minor) issues with this PR are that:

- it does not convert any existing DDoc to use the new macros
- as a consequence of the above, it includes no tests for the added macros (so it's not obvious the macros are correct e.g. via the doc autotester)
- it increases the total number of macros and [different ways to link to things](http://forum.dlang.org/post/xpsppxdmmnifknhlgpyb@forum.dlang.org), when that number needs to go down, not up.

@adamdruppe Would be great if you could address those by e.g. converting and removing STDMODREF / COREMODREF as well as any other links to modules that you may find, but I would be OK with merging this PR if you don't want to do that.

adamdruppe:
> it does not convert any existing DDoc to use the new macros

All the ones I want to change are in the Phobos repo and I'd rather not break it in the mean time.

Every single instance of `$(LINK2 std_*.html, std.*)` in Phobos needs to be converted. The current situation is broken links, like the first line here:


>

CyberShadow:
> All the ones I want to change are in the Phobos repo and I'd rather not break it in the mean time.

What about the ones you found in the changelog?

CyberShadow:
> Every single instance of $(LINK2 std_*.html, std.*) in Phobos needs to be converted.

Are there no such links anywhere on the website?

std-ddox.ddoc
```diff
@@ -3,6 +3,9 @@ XREF = std.$1.$2
 XREF_PACK = std.$1.$2.$3
 CXREF = core.$1.$2
 ECXREF = etc.$1.$2
+MREF_HELPER=/$1$(MREF_HELPER $+)
```
andralex: Could you please rename this as `SLASH_PREFIXED`? I reckon there are other good uses for it.

std-ddox.ddoc
```diff
@@ -3,6 +3,9 @@ XREF = std.$1.$2
...
+MREF_HELPER2=.$1$(MREF_HELPER2 $+)
```
andralex: This would be `DOT_PREFIXED`.

std.ddoc
```diff
@@ -13,6 +13,9 @@ _=
 CXREF = <a href="core_$1.html#$2">$(D core.$1.$2)</a>
 ECXREF = <a href="etc_c_$1.html#$2">$(D etc.c.$1.$2)</a>
 LREF = <a href="#$1">$(D $1)</a>
+MREF_HELPER=_$1$(MREF_HELPER $+)
```
andralex: And this would be `UNDERSCORE_PREFIXED`.

andralex:
Good stuff, @adamdruppe. Could you please rename as suggested, and also take a look at the existent `LINK2` that can be replaced?

andralex:
Aside: I just edited @CyberShadow's comment. Trippy!

andralex:
ping

adamdruppe:
I renamed them, though in my eyes the new names just promoted them from a private function to a public function since they are now generically named....

There are no hits on my grep for existing LINK2s that can be correctly changed by this change in dlang.org.

CyberShadow:
Broken record here, but what about COREMODREF/STDMODREF?

adamdruppe:
On Wed, Jan 06, 2016 at 11:43:00AM -0800, Vladimir Panteleev wrote:
> Broken record here, but what about COREMODREF/STDMODREF?

I'm not interested in modifying the changelog. It seems to work OK as-is.

andralex:
Hmmm.. not sure what the status is now. I see the older macros here and the new ones in the other PR. What gives?

adamdruppe:
On Wed, Jan 06, 2016 at 01:41:00PM -0800, Andrei Alexandrescu wrote:
> Hmmm.. not sure what the status is now. I see the older macros here and the new ones in the other PR. What gives?

The other one is a branch off this one.

adamdruppe:
BTW: after using `$(REF ...)`, the magic macro in my doc generator, XREF, LREF, and MREF (and all their variants) are dead to me.

I *think* the `DSymbols* a` (what a descriptive name) argument in dmd has the potential to form a similar magic macro if someone wanted to port that over.

It'd be a bit of a pain to do generic linking in ddoc. You might need six different macros anyway to account for the user-defined urls in the implementation, but I'm sure it is possible.

REF in dmd would prolly be implemented in terms of MREF by the way, so this is still surely a necessary step.

andralex:
thx!

PR #1178, opened 2015-12-31
New macros

adamdruppe:
This is a follow-up to https://github.com/D-Programming-Language/dlang.org/pull/1177 and assumes that one gets merged first.

I like to use these new macros when writing documentation to call out certain parts of the text (or in the case of SIDEBAR, to put it there with*out* drawing attention to it).

The look can be defined elsewhere.

I want to change cases like `$(RED Important note:) blah blah` in some Phobos functions to `$(DANGER blah blah)` and `$(BLUE This function is Windows only)` to `$(NOTE this function is Windows only)` to we can restyle them more easily.

I also intend to use these in writing my doc additions, which hopefully we can get into Phobos itself too with minimal disruption to the dlang.org site

### DIFF
--- a/std-ddox.ddoc
+++ b/std-ddox.ddoc
@@ -8,4 +8,9 @@ MREF_HELPER2=.$1$(MREF_HELPER2 $+)
 MREF=<a href="/library/$1$(MREF_HELPER $+).html">$1$(MREF_HELPER2 $+)</a>
 LREF = $1
 ROOT_DIR=/
+NOTE = <div class="note">$0</div>
+WARNING = <div class="warning">$0</div>
+PITFALL = <div class="pitfall">$0</div>
+TIP = <div class="tip">$0</div>
+SIDEBAR = <div class="sidebar">$0</div>
 _=
--- a/std-ddox.ddoc
+++ b/std-ddox.ddoc
@@ -3,8 +3,6 @@ XREF = std.$1.$2
 XREF_PACK = std.$1.$2.$3
 CXREF = core.$1.$2
 ECXREF = etc.$1.$2
-MREF_HELPER=/$1$(MREF_HELPER $+)
-MREF_HELPER2=.$1$(MREF_HELPER2 $+)
-MREF=<a href="/library/$1$(MREF_HELPER $+).html">$1$(MREF_HELPER2 $+)</a>
+MREF=<a href="/library/$1$(SLASH_PREFIXED $+).html">$1$(DOT_PREFIXED $+)</a>
 LREF = $1
 ROOT_DIR=/
--- a/std-ddox.ddoc
+++ b/std-ddox.ddoc
@@ -8,9 +8,3 @@ MREF_HELPER2=.$1$(MREF_HELPER2 $+)
 MREF=<a href="/library/$1$(MREF_HELPER $+).html">$1$(MREF_HELPER2 $+)</a>
 LREF = $1
 ROOT_DIR=/
-NOTE = <div class="note">$0</div>
-WARNING = <div class="warning">$0</div>
-PITFALL = <div class="pitfall">$0</div>
-TIP = <div class="tip">$0</div>
-SIDEBAR = <div class="sidebar">$0</div>
-_=
--- a/std-ddox.ddoc
+++ b/std-ddox.ddoc
@@ -3,6 +3,9 @@ XREF = std.$1.$2
 XREF_PACK = std.$1.$2.$3
 CXREF = core.$1.$2
 ECXREF = etc.$1.$2
+MREF_HELPER=/$1$(MREF_HELPER $+)
+MREF_HELPER2=.$1$(MREF_HELPER2 $+)
+MREF=<a href="/library/$1$(MREF_HELPER $+).html">$1$(MREF_HELPER2 $+)</a>
 LREF = $1
 ROOT_DIR=/
 _=
--- a/std.ddoc
+++ b/std.ddoc
@@ -13,6 +13,9 @@ _=
 CXREF = <a href="core_$1.html#$2">$(D core.$1.$2)</a>
 ECXREF = <a href="etc_c_$1.html#$2">$(D etc.c.$1.$2)</a>
 LREF = <a href="#$1">$(D $1)</a>
+MREF_HELPER=_$1$(MREF_HELPER $+)
+MREF_HELPER2=.$1$(MREF_HELPER2 $+)
+MREF=<a href="$1$(MREF_HELPER $+).html">$1$(MREF_HELPER2 $+)</a>
 _=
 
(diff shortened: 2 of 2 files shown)


std-ddox.ddoc
```diff
@@ -3,6 +3,14 @@ XREF = std.$1.$2
 XREF_PACK = std.$1.$2.$3
 CXREF = core.$1.$2
 ECXREF = etc.$1.$2
+MREF_HELPER=/$1$(MREF_HELPER $+)
+MREF_HELPER2=.$1$(MREF_HELPER2 $+)
+MREF=<a href="/library/$1$(MREF_HELPER $+).html">$1$(MREF_HELPER2 $+)</a>
 LREF = $1
 ROOT_DIR=/
+NOTE = <div class="note">$0</div>
```
CyberShadow: I don't think you need to redefine things in std-ddox.ddoc if the values are the same as in std.ddoc, AFAIU this file is essentially an override of std.ddoc for DDox.

CyberShadow:
Please, no more unused macro additions. We have too many unused/underused macros already. Please include use cases for these additions by providing justifiable examples of where they will be useful enough to justify their existence.

adamdruppe:
There's the first one that comes to mind:

http://dlang.org/phobos/std_string.html#toStringz

That "important note" would be a PITFALL with the new macro.

http://dlang.org/phobos/std_file.html#getTimesWin

That blue text can be a NOTE.

It changes presentation embedded in ddoc to be semantic for easier restyling, while enriching the vocabulary for authors.

std-ddox.ddoc
adamdruppe: On Thu, Dec 31, 2015 at 02:54:52PM -0800, Vladimir Panteleev wrote:
> I don't think you need to redefine things in std-ddox.ddoc if the values are the same as in std.ddoc, AFAIU this file is essentially an override of std.ddoc for DDox.

Oh, cool.

CyberShadow:
OK, thanks for the clarification. Do you want to create the Phobos PR as well?

Also, I don't understand why this is a follow-up to #1177, I don't understand how the two are related.

CyberShadow:
Do we really need 5 macros?

Do we need all NOTE / WARNING / PITFALL? Why would you style all three differently?

Do we need special styling for TIP?

SIDEBAR carries no semantic meaning, I'm not sure when it could be used. Maybe I need to see your Phobos PR first.

Maybe I'm just overly cynical but it seems to me that you're inventing macros and then trying to find use cases for them, which seems backwards to me as it should be in the other direction - finding areas where the documentation can be improved and coming up with solutions that balance benefit and bloat.

adamdruppe:
This related to the other one just because I forgot to git checkout master before git branching to this, so the commit is included. But they are related to my thought process anyway - I'm trying to get some of these macros sane for alternate generators.

CyberShadow:
I think it would also be prudent to ask that you include proposed styling for the macros, because as it is this only increases the macro count with no visible/effective gain.

adamdruppe:
Here's the changes I've made to Phobos locally so far. https://github.com/D-Programming-Language/phobos/pull/3895

It isn't complete, but this shouldn't be blocked on that (quite the contrary, pulling the Phobos changes without the new macro will result in text disappearing from the website).

adamdruppe:
> Do we really need 5 macros?

Not *need* per se, but I like to use them all when I'm writing about D and having a small, but comprehensive set means people won't be tempted to invent their own ways over and over again in each module. This set matches my expanded vocabulary when writing things of this nature.

> Maybe I'm just overly cynical but it seems to me that you're inventing macros and then trying to find use cases for them

You're right. I've never written anything before and have no idea how I'd do it. To be honest, I've never actually read anything before either.

But I had a dream last night where these words popped into my mind. I have no clue what they mean, but my gut is telling me that dream had meaning.

CyberShadow:
> You're right. I've never written anything before and have no idea how I'd do it. To be honest, I've never actually read anything before either.
> 
> But I had a dream last night where these words popped into my mind. I have no clue what they mean, but my gut is telling me that dream had meaning.

Sorry, I don't understand what you mean. Maybe I should clarify.

Here is what I think should be the right way to approach such problems:

"These bits of text in particular are all styled similarly for the same reason but in an ad-hoc way. Let's create a macro that replaces ad-hoc styling with a semantic macro so the DDoc is easier to write/understand/maintain, and it's easier to tweak the styling of all such text in the future."

What I understand your process to be:

"Our DDoc could use some macros to convey semantic meaning in some bits of text. Here are some ideas. Here is an implementation of these ideas. I'll see if I can find some use cases for these ideas later."

Am I wrong?

std.ddoc
```diff
@@ -10,9 +10,25 @@ $(SCRIPT jQuery(document).ready(listanchors);)
...
+_= SIDEBAR is something for users who are interested into diving much deeper than they have to to understand the function. It should just contain a short sentence and probably a link to learn more.
+SIDEBAR = <div class="sidebar">$0</div>
+
+_=
+
 CXREF = <a href="core_$1.html#$2">$(D core.$1.$2)</a>
 ECXREF = <a href="etc_c_$1.html#$2">$(D etc.c.$1.$2)</a>
 LREF = <a href="#$1">$(D $1)</a>
+MREF_HELPER=_$1$(MREF_HELPER $+)
+MREF_HELPER2=.$1$(MREF_HELPER2 $+)
```
andralex: Per my comment in the other PR, these would be `SLASH_PREFIXED` and `DOT_PREFIXED`.

std-ddox.ddoc
```diff
@@ -3,6 +3,8 @@ XREF = std.$1.$2
 XREF_PACK = std.$1.$2.$3
 CXREF = core.$1.$2
 ECXREF = etc.$1.$2
+MREF_HELPER=/$1$(MREF_HELPER $+)
+MREF_HELPER2=.$1$(MREF_HELPER2 $+)
```
andralex: Per my comment in the other PR, these would be `SLASH_PREFIXED` and `DOT_PREFIXED`.

std.ddoc
andralex: Another thing: `dlang.org.ddoc` is commonly used by all documentation, so you could hoist these declarations in there and avoid duplication.

adamdruppe: On Tue, Jan 05, 2016 at 07:12:42AM -0800, Andrei Alexandrescu wrote:
> Another thing: `dlang.org.ddoc` is commonly used by all documentation, so you could hoist these declarations in there and avoid duplication.

Don't forget though that these links won't work in different directories because they are relative to the current directory.

Absolute linking would work on the site but would cause problems elsewhere like users trying to view offline...

andralex: I'm only talking about the generic macros `XXX_PREFIXED`. They don't prescribe absolute vs. relative linking etc.

andralex:
Thanks. I think PREFIXED_XXX and MREF are great. How about you hold off on the others until you add documentation that actually uses them? Thanks!

andralex:
My understanding after chatting with @adamdruppe is that we close this now.

PR #1180, opened 2016-01-04
Issue 15471: Associative Arrays: add example of static initialization

markisaa:
https://issues.dlang.org/show_bug.cgi?id=15471

screen shot 2016-01-03 at 7 24 07 pm

FWIW: The extra whitespace was auto-removed by my setup, I checked and it does not change the rendered html page.


markisaa:
Errr... typo in there.. embarrassing. 1 sec.

markisaa:
Better :)

andralex:
Nice. Thanks! I see you use that "remove trailing spaces" in emacs/vim :o).

PR #1181, opened 2016-01-04
Fix typo in docs for number literal ranges

schveiguy:
All the other rows clearly show inclusive ranges, while uint decimal shows an exclusive range (i.e. `uint.max + 1` as the top value)


andralex:
Thanks!

PR #1183, opened 2016-01-05
Wrong variable naming in the example

tcak:
On top the variable is defined as "flag" but on the bottom, it is used as "flags" while both of them should have been same.


JackStouffer:
LGTM

Hackerpilot:
:white_check_mark:

andralex:
thx!

PR #1184, opened 2016-01-08
Add new macro REF, a variadic and more general XREF

JakobOvrum:
Usage is like:
```
$(REF File.writeln, std, stdio)
$(REF splitter, std, algorithm, iteration)
$(REF_ALTTEXT forward range, isForwardRange, std, range, primitives)
```
Added to dlang.org.ddoc because that's where XREF is defined. Can be used to link to symbols rooted in `std`, `core`, `etc` or any other root package.

Implementation is a bit messy. I don't know quite why, but using TAIL instead of those helper macros didn't work. Commas showed up in the output.

Will implement for DDox output ASAP.

### DIFF
--- a/dlang.org.ddoc
+++ b/dlang.org.ddoc
@@ -133,6 +133,8 @@ DDOCKEYVAL2=$(DIVC keyval $1, $(SPANC key key$1, $2:) $(DIVC val val$1, $(TAIL $
 DDSUBLINK=$(LINK2 $(ROOT_DIR)$1.html#$2, $3)
 _=
 
+DOT_PREFIXED=.$1$(DOT_PREFIXED $+)
+DOT_PREFIXED_SKIP=$(DOT_PREFIXED $+)
 DRUNTIMESRC=$(HTTPS github.com/D-Programming-Language/druntime/blob/master/src/$0, $0)
 DPLLINK=$(LINK2 $1,$+)
 _=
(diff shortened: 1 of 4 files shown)


JakobOvrum:
Pushed a DDox implementation but I have no clue how to test it.

JakobOvrum:
Added D-Programming-Language/phobos/pull/3911 as an example of how it replaces existing macros.

CyberShadow:
> I don't know quite why, but using TAIL instead of those helper macros didn't work. Commas showed up in the output.

IIRC the situation is something along the lines of - some of the parameter macros (don't remember which exactly, one of $0 / $+ / etc.) pass their content as a series of parameters (which get expanded to $1/$2/...), but some pass it as a single string. This is what allows $(COMMA) to work inside nested macros AFAIU.

dlang.org.ddoc
```diff
@@ -241,6 +241,13 @@ PHOBOSSRC=$(SPANC phobos_src, $(AHTTPS github.com/D-Programming-Language/phobos/
 _=
 
 RED=$(SPANC red, $0)
+PREFIX_DOT=.$1$(PREFIX_DOT $+)
+PREFIX_DOT_SKIP=$(PREFIX_DOT $+)
+PREFIX_UNDERSCORE=_$1$(PREFIX_UNDERSCORE $+)
+PREFIX_UNDERSCORE_SKIP=$(PREFIX_UNDERSCORE $+)
+PREFIX_UNDERSCORE_SKIP2=$(PREFIX_UNDERSCORE_SKIP $+)
```
andralex: See @adamdruppe 's recent PR for simple macros that prefix with dots, underscores etc.

JakobOvrum:
Moved the new `*_PREFIXED` macros from std.ddoc to macros.ddoc and used them in `REF` (which is in dlang.ddoc).

JakobOvrum:
OK, can confirm that the DDox override works.

JakobOvrum:
Removed duplicate `DOT_PREFIXED_SKIP` definition from std-ddox.ddoc.

JakobOvrum:
Alright, now `REF` and `MREF` and all helper macros are all in dlang.org.ddoc. Also changed `MREF` to use `$(D …)` for the link text.

JakobOvrum:
`REF` and `MREF` now use `ROOT_DIR` in their paths.

CyberShadow:
LGTM :)

JakobOvrum:
Thanks!

PR #1185, opened 2016-01-08
Remove leading dot from DDOC_ANCHOR

JakobOvrum:
The leading dot was introduced in #179 to avoid conflicts with legacy fragment names.

However, the legacy fragment name scheme is not compatible with HTML. Fragment names must be unique in the document, whether using `name` (removed in HTML5) or `id`, otherwise behaviour is undefined. For example, for http://dlang.org/phobos/std_stdio#writeln, Firefox links to the `name="writeln"` appearing earliest in the document, which is actually `std.stdio.File.writeln`, as opposed to the `std.stdio.writeln` one would expect.

With this PR, only the qualified name (rooted in the module) emits a fragment, so we can remove the leading dot.

Apropos `name`, it should be changed to `id` everywhere for HTML5 compatbility. This PR only does it in one place.

### DIFF
--- a/dlang.org.ddoc
+++ b/dlang.org.ddoc
@@ -123,8 +123,8 @@ DDOC_PARAM_ID  = <td class="param_id">$0</td>
 DDOC_PARAM_DESC  = <td class="param_desc">$0</td>
 DDOC_PARAMS    = $(DDOCKEYVAL Parameters, <table class="params">$0</table>)
 DDOC_BLANKLINE	= $(P)
-DDOC_PSYMBOL = $(ADEF $0)$(SPANC ddoc_psymbol, $0)
-DDOC_ANCHOR = $(ADEF .$1)$(DIVCID quickindex, quickindex.$1, )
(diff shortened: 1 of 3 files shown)


adamdruppe:
Looks good to me. If there's no objections by the end of the day, someone ping me and I'll hit the merge myself.

CyberShadow:
Does this fix anything in practice? Because obviously it does break legacy links, and breaking links is bad practice.

aG0aep6G:
How about generating both dot-prefixed and non-prefixed anchors via `DDOC_ANCHOR`? I.e.:

    DDOC_PSYMBOL = $(SPANC ddoc_psymbol, $0)
    DDOC_ANCHOR = $(ADEF .$1)$(ADEF $1)$(DIVCID quickindex, quickindex.$1, )

This should get us rid of the duplicates and keep all links working.

adamdruppe:
I suspect those legacy links were already really broken. I never used them because they wouldn't link to the right place anyway!

LightBender:
PR closed as stalled. If you wish to resurrect it you are welcome to do so.

PR #1186, opened 2016-01-10
Better DDOC_BLANKLINE companion

andralex:
This works in tandem with https://github.com/D-Programming-Language/dmd/pull/5344. Together, they produce the same html output for the site and standard library.

I had to do a few tweaks to the source files.


macros.ddoc
```diff
@@ -2,9 +2,11 @@ _=Fundamental macros that apply to all generated formats
 
 ARGS = $0
 COMMA = ,
-COMMENT = 
+COMMENT =
```
andralex: This is spurious (due to my editor setup), but spaces at the end of the line are suspicious anyway.

std.ddoc
```diff
@@ -37,3 +37,6 @@ JOIN_DOT_TAIL=$(JOIN_DOT $+)
 JOIN_DOT=.$1$(JOIN_DOT $+)
 TAIL=$+
 _=
+
+DDOC_BLANKLINE = $(BR)$(BR)$(LF)
```
andralex: The extra LF in there is for compatibility with the existing generated HTML. We may eliminate it later if we find it fit.

macros.ddoc
JakobOvrum: [It's part of our .editorconfig](https://github.com/D-Programming-Language/dlang.org/blob/master/.editorconfig), but I think we can override it for specific files if we need to.

andralex:
Let me retire this and redo it after the bug push.

andralex:
Back in swing!

andralex:
egh, lemme work off a different branch

PR #1188, opened 2016-01-10
correct hash for "round floating point numbers" example

aG0aep6G:
This fixes the default input.


aG0aep6G:
With the redesign reverted, this can be worked into that. Closing.

PR #1189, opened 2016-01-10
Revert "Ivan Smirnov's redesign"

andralex:
Reverts D-Programming-Language/dlang.org#1187. Guess we're not ready yet, at least on account of http://downloads.dlang.org not working.


aG0aep6G:
LGTM ;)

MartinNowak:
Oh, that's a good opportunity to redo that subdomain :).
I already tried to integrate it w/ dlang.org at some point and we really need to do this even though braddr didn't like the idea [¹](https://github.com/braddr/downloads.dlang.org/pull/1).
But now we want to integrate [nightlies](https://builds.dawg.eu/) as well [²](http://forum.dlang.org/post/97209668-DAED-4BCA-B66F-41AEAC09E021@me.com) so we should come up w/ something better. I'd be in favor of completely integrating downloads.dlang.org with dlang.org (the repo and maybe even the subdomain). The tricky part of course is to keep that in sync w/ an ever changing list of files.
Any ideas or preferences?

CyberShadow:
Thanks all. I agree that this is great work that we must act upon, but let's pace ourselves.

> Oh, that's a good opportunity to redo that subdomain :).

The way I did this for forum.dlang.org is to include a copy of dlang.org as a Git submodule. This way, it is still based on dlang.org CSS/DDoc, can be easily updated to the latest design, but still locked to a specific version to avoid dlang.org changes breaking the forum.

andralex:
@CyberShadow please don't forget to also sync the site

CyberShadow:
> @CyberShadow please don't forget to also sync the site

OK, done (was hoping someone else could do it as I'm not home yet)

PR #1187, opened 2016-01-10
Ivan Smirnov's redesign

aG0aep6G:
Forum thread: http://forum.dlang.org/post/n6pdes$1upn$1@digitalmars.com

The last commit is the actual redesign. The others are independent small fixes. I can extract them and make individual pull requests if needed.

PR #1190, opened 2016-01-11
Ivan smirnov's redesign (2nd try)

aG0aep6G:
**Do not merge.** This breaks downloads.dlang.org.

Forum thread: http://forum.dlang.org/post/n6pdes$1upn$1@digitalmars.com

This being a revert of a revert, the little commit information that was there is gone. It can be seen over at #1187.

Reitering that someone should probably recreate the Twitter widget from a different account. https://github.com/D-Programming-Language/dlang.org/pull/1187#issuecomment-170360243


andralex:
@CyberShadow are you on this? Let me know if you need help. Thx!

articles.dd
```diff
@@ -0,0 +1,163 @@
...
+            $(DIVC item,
+                $(H4 $(LINK2 $(ROOT_DIR)d-array-article.html, D Slices))
+                $(P Steven Schveighoffer writes about slices and dynamic arrays
+                    in D. Unfortunately, this article uses slightly different
+                    terminology than the language reference: What the article
+                    calls a $(I slice) is a $(I dynamic array) in the language
+                    reference, and what the article calls a $(I dynamic array)
+                    is not named specifically by the language reference. Still,
+                    the article is a great help in getting a good understanding
+                    of D's array semantics.)
```
JakobOvrum: Unfortunately the language spec uses different terminology from this article ;)

documentation.dd
```diff
@@ -0,0 +1,6 @@
+Ddoc
+
+$(D_S Documentation, $(NAVIGATION_DOCUMENTATION))
+
+Macros:
+    TITLE=Documentation
```
JakobOvrum: What is this file for?

dpl-docs/views/layout.dt
```diff
@@ -32,97 +33,182 @@ html(lang='en-US')
...
+          #cssmenu
+            ul
+              li
+                a(href="#{root_dir}getstarted.html")
+                  span Learn
+              li.expand-container
+                a.expand-toggle(href="#{root_dir}documentation.html")
+                  span Documentation
+                ul
+                  li
```
JakobOvrum: Hm, how did ddox manage without this duplication before?

documentation.dd
adamdruppe: That's the page if you open the documentation link: http://d-ag0aep6g.rhcloud.com/documentation.html
right-click "documentation" in the menu and open in a new tab to get there from the home page.

JakobOvrum: Thanks, I see

aG0aep6G: That is, it's a fallback page for when there's no JS. community.dd, resources.dd, search.dd, and menu.dd are fallback pages, too.

dpl-docs/views/layout.dt
aG0aep6G: It simply doesn't feature the main menu.

MartinNowak:
Thanks, any ideas for downloads.dlang.org? We could operate w/ a stripped down version until we redesign it as well.

aG0aep6G:
I've made a companion PR for forum.dlang.org: <https://github.com/CyberShadow/DFeed/pull/51>. That needs some changes to dlang.org's forum-template.dd. I've added them here.

aG0aep6G:
> Thanks, any ideas for downloads.dlang.org? We could operate w/ a stripped down version until we redesign it as well.

Where is the git repository of downloads.dlang.org? Updating forum.dlang.org went pretty smoothly, so maybe I can just do downloads.dlang.org over the next couple days, and then we can make the switch for everyone at the same time.

Alternatively, we can freeze downloads.dlang.org in the old design by using copies of the dlang.org files (via git submodule) instead of using the online files. This may be a good idea anyway so that we don't accidentally break things again.

CyberShadow:
As an option I guess we could leave style.css etc. alone, and use new filenames for files used in the new design. But of course we should update it instead if possible.

So I don't know these things either:

 - Where is the source code for the downloads.dlang.org page templates?
 - How can we build and preview it?
 - Who can update it and push it to the live site?

MartinNowak:
Unfortunately it's not possible to generate the pages w/o an S3 account.
https://github.com/MartinNowak/downloads.dlang.org
Please don't put too much effort into this b/c we have to come up w/ something else in order to integrate the nightlies. I'd say updating the [template](https://github.com/MartinNowak/downloads.dlang.org/blob/9fe27c68990bb3240c57bd2114c6c836e7ea6250/src/gen_index.d#L158) so it's operable should be good enough for now.

CyberShadow:
> Unfortunately it's not possible to generate the pages w/o an S3 account.

Doesn't look so bad, maybe a second main inside `version(test)` which supplies dummy data to `buildIndex`.

MartinNowak:
> Doesn't look so bad, maybe a second main inside version(test) which supplies dummy data to buildIndex.

I can just do that, so you guys can work on the template.

download.dd
```diff
@@ -1,9 +1,5 @@
 Ddoc
 
-<!-- Github ribbon -->
-<a href="http://github.com/D-Programming-Language"><img id="github-ribbon"
-src="images/github-ribbon.png" alt="Fork D on GitHub"></a>
-
```
CyberShadow: Since the GitHub ribbon is removed, I think there should be a link to the source code somewhere (e.g. near the release archive / GPG keys links).

dpl-docs/views/layout.dt
```diff
@@ -32,97 +33,182 @@ html(lang='en-US')
...
+        - else if (hidx > 0 && title.startsWith("Module ") && info.mod)
+          h1 Module <code>#{info.mod.qualifiedName}</code>
+        - else
+          h1= title
+        block body
 
-      |     
-      #quickindex.quickindex
+        |     
+        #quickindex.quickindex
```
CyberShadow: We really should do something about needing to keep this in sync by hand. It's the same content in two different languages.

CyberShadow:
This seems to break building the CHM. I'll look into it.

MartinNowak:
> I can just do that, so you guys can work on the template.

There you go https://github.com/MartinNowak/downloads.dlang.org/compare/eed9a5190abfdf83f20f3747dcd4c4c2c5cdf711...MartinNowak:9bd901c95681ce827da46fcd5aacc708266f953b, now generates html from a json file.
```sh
git clone git://github.com/MartinNowak/downloads.dlang.org.git
cd downloads.dlang.org
make -C src
cp test_index.json index.json
./src/build-gen-index generate
```
Not sure if it works on Windows though (it's using `/` to construct paths everywhere).

CyberShadow:
How about a changelog SUBNAV?

CyberShadow:
> This seems to break building the CHM. I'll look into it.

https://github.com/aG0aep6G/dlang.org/pull/2

CyberShadow:
What's with the link to the non-existing /menu.html ?

CyberShadow:
Main page screenshot with restyled forum widgets:

screen shot 2016-01-13 at 06 32 29

MartinNowak:
Nice

CyberShadow:
The D logo is broken in IE 11 (only top right corner is shown), maybe the SVG needs to be "resized".

download.dd
aG0aep6G: I don't think we need a link to GitHub on the downloads page, but here you go. Added a commit: https://github.com/aG0aep6G/dlang.org/commit/2d1d95dd7a475d5d258332f7bb01796e14966186

dpl-docs/views/layout.dt
aG0aep6G: I think I've found a way. Added a commit: https://github.com/aG0aep6G/dlang.org/commit/93e3240804441e666c3af78eba6201367a586bc3

I had tried generating a .dt file with Ddoc before. The significant whitespace doesn't make that easy.

aG0aep6G:
> What's with the link to the non-existing /menu.html ?

/menu.html should be there. It's a no-JS fallback page.

menu.dd is here: https://github.com/D-Programming-Language/dlang.org/pull/1190/files#diff-39b0dfe44d7b515b991a96b903b95f9b

And here it is in posix.mak: https://github.com/D-Programming-Language/dlang.org/pull/1190/files#diff-228b518f4545d060f900578830ade5daR172

aG0aep6G:
Incorporated @CyberShadow's CHM fixes. Can't say that I have reviewed them. I don't know a thing about the CHM stuff.

dpl-docs/views/layout.dt
aG0aep6G: Ah crap. `$(ROOT_DIR)` isn't set correctly then. The D code in .dt files is run at request time, right? Is it acceptable to call dmd for every request? I guess not. Back to square one.

I reverted the commit. When we find a solution (or settle for the duplication), I should clean the commit history up, but for now I'm leaving it up for reference.

aG0aep6G:
> How about a changelog SUBNAV?

Added: https://github.com/aG0aep6G/dlang.org/commit/43ca3df1357fdc876879a863bcf444ec03cd96d7

dpl-docs/views/layout.dt
CyberShadow: > The D code in .dt files is run at request time, right?

Only in case of dpldocs-serve. It is compiled once then executed by a vibe web server. We use static file generation on dlang.org though.

> Is it acceptable to call dmd for every request?

I don't see how your patch would do that.

Aside std.file.read, you could use string imports.

CyberShadow:
> /menu.html should be there. It's a no-JS fallback page.

Hm, OK. Not sure what I did to get a 404, but I can see it's all there.

> Added: aG0aep6G@43ca3df

Nice!

CyberShadow:
> Incorporated @CyberShadow's CHM fixes. Can't say that I have reviewed them. I don't know a thing about the CHM stuff.

Most of it has been shuffling the navigation macros to allow emitting a complete navigation tree in a JSON-ish format that chmgen can parse.

dpl-docs/views/layout.dt
aG0aep6G: > We use static file generation on dlang.org though.

Right. What I have in mind might work then: Generate a .ddoc file and run dmd on it from the .dt template. Otherwise I don't know how to get the correct root dir set up.

I'm going to look further into this tomorrow.

andralex: Thanks. Let's keep the ball rolling towards the goal folks. Ideally I plan to announce this on or before Friday.

CyberShadow:
@aG0aep6G Will you be looking into updating the downloads.dlang.org template? That currently has the highest priority.

aG0aep6G:
Pull request for downloads.dlang.org - quick and dirty, not making it more robust against future breakage: https://github.com/MartinNowak/downloads.dlang.org/pull/1

aG0aep6G:
And here's a PR for the Visual D pages: https://github.com/D-Programming-Language/visuald/pull/57

aG0aep6G:
> The D logo is broken in IE 11 (only top right corner is shown), maybe the SVG needs to be "resized".

Added a commit to fix that, and one to fix the search dropdown style.

andralex:
Thanks all for the hard work! @CyberShadow @MartinNowak are we on track to push the bits to dlang.org on Friday (essentially 12 hours from now)?

CyberShadow:
Blockers:

- [ ] downloads.dlang.org (waiting on Martin)

Non-blocking issues:

- [ ] forum.dlang.org tweaks (see discussion on https://github.com/CyberShadow/DFeed/pull/51 and [forum thread](http://forum.dlang.org/post/odbtivxmrggaywcbebqx@forum.dlang.org))

Possible issues (needs more discussion):

- [ ] Change the web font (or not) to a standard one

aG0aep6G:
Added two commits:
* move #copyright into #content - Looks a bit nicer with the sidebar. Also helps forum.dlang.org with height calculations.
* give the subnav a left border too - People complained that it looked weird before: http://forum.dlang.org/post/qcubliszoceqjchcvqrl@forum.dlang.org

MartinNowak:
> downloads.dlang.org (waiting on Martin)

Merged https://github.com/MartinNowak/downloads.dlang.org/pull/1, anything else?

CyberShadow:
> Merged MartinNowak/downloads.dlang.org#1, anything else?

I guess no, we just need to deploy this PR and downloads.dlang.org in tandem.

andralex:
@CyberShadow wanna do the honors?

CyberShadow:
Well, I think @MartinNowak should do that since I can't update downloads.dlang.org.

andralex:
@MartinNowak you got the red button

aG0aep6G:
Added a last minute fix as per https://github.com/CyberShadow/DFeed/issues/52#issuecomment-172328137

MartinNowak:
Will deploy this in an hour.

MartinNowak:
> Will deploy this in an hour.

Done, don't forget to press F5.

PR #1193, opened 2016-01-13
Fix MREF and REF for phobos-prerelease target

JakobOvrum:
Let's see if this does the trick.


JakobOvrum:
Yeah, this works (tested locally).

DDox `MREF` links are still hard-coded to `/library/`, but that was always the case. I think fixing that would require a new `.ddoc` file.

CyberShadow:
:+1:

PR #1194, opened 2016-01-13
Let DDox handle module references

JakobOvrum:
It looks like DDox detects and handles module references in the same way it detects symbol references. This fixes module references in `/library-prerelease/` and removes `<a>`-inside-`<a>` shenanigans.

DDox doesn't appear to detect module references if they appear the end of a punctuated sentence, like `see std.algorithm.`; I'll file that over at the DDox repository.

### DIFF
--- a/latex.ddoc
+++ b/latex.ddoc
@@ -213,6 +213,7 @@ MULTICOL_CELL=\multicolumn{$1}{l}{$+}
 MULTICOL_HEADER=\multicolumn{$1}{l}{$+}
 MULTICOLS=\begin{multicols}{$1}$+ \end{multicols}
 MULTIROW_HEADER=\multirow{$1}{*}{$+}
+MREF=$(D $1$(DOT_PREFIXED $+))
 _=
 
 NBSP=~
@@ -240,6 +241,7 @@ _=
 
 RCURL=$\}$
 RED={\color{red}$0}
+REF=$(D $2$(DOT_PREFIXED_SKIP $+).$1)
 REG=$(COPY)
 RELATIVE_LINK2=$(LINK2 $1,$+)
 _=
--- a/std-ddox.ddoc
+++ b/std-ddox.ddoc
@@ -3,10 +3,10 @@ XREF = std.$1.$2
 XREF_PACK = std.$1.$2.$3
 CXREF = core.$1.$2
 ECXREF = etc.$1.$2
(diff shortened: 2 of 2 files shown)


JakobOvrum:
Filed: rejectedsoftware/ddox/issues/110

CyberShadow:
> DDox doesn't appear to detect module references if they appear the end of a punctuated sentence, like see std.algorithm.; I'll file that over at the DDox repository.

Maybe wrap it into a `<span>` (with no adornments)?

JakobOvrum:
Hm, it still doesn't recognize it with `<span>`.

s-ludwig:
The planned future way is to only recognize symbols within backticks, or within `$(D)`. So in this case I'd define
```
MREF=`$+`
```

I'll have a look at the end-of-sentence issue.

JakobOvrum:
`$(MREF a, b, c)` now resolves to ``` `a.b.c` ``` for DDox.

s-ludwig:
Hm, is there an example MREF use somewhere? Everywhere I looked, I found only things like `$(LINK2)` or `$(B)` used.

JakobOvrum:
@s-ludwig, PR for Phobos links: D-Programming-Language/phobos/pull/3926

No PR for druntime or dlang.org yet.

s-ludwig:
Okay, tested locally and `MREF=$(D $1$(DOT_PREFIXED $+))` works. I need to check how Ddox behaves for backticks within macro definitions.

BTW, `std.algorithm` currently doesn't emit a link, because it's detected as a package. I'll add support for forwarding to "package.d" there.

JakobOvrum:
Thanks, changed to `MREF=$(D …)`.

CyberShadow:
@JakobOvrum How about `REF` too, as per https://github.com/D-Programming-Language/phobos/pull/3926#issuecomment-171837220 ?

JakobOvrum:
Done, seems to work, but I'm getting confused as to exactly where the `apidocs-*` targets look for Phobos. Should be `../phobos`, no?

CyberShadow:
> but I'm getting confused as to exactly where the apidocs-* targets look for Phobos. Should be ../phobos, no?

Looks like it reads its tasks from a JSON file (`docs.json` / `docs-prerelease.json`), which is generated using `${PHOBOS_DIR}`, which by default is defined as `../phobos`, so yeah.

adamdruppe:
So, this thing works? The code looks fine to me, I'll pull tomorrow if there's no objections.

s-ludwig:
Looks good to me, too. The diff is also fine.

PR #1197, opened 2016-01-17
correct cmp to opCmp and add example for opBinary

jamadagni:
Documentation refers to `cmp` i.o. `opCmp`. Corrected that. Also provided `opBinary` example which also well illustrates usefulness of `op*` system rather than `operator?` system.

### DIFF
--- a/rationale.dd
+++ b/rationale.dd
@@ -14,10 +14,32 @@ $(H3 Why not name them operator+(), operator*(), etc.?)
 	to refer to overloading $(SINGLEQUOTE +) with $(SINGLEQUOTE operator+). The trouble is
 	things don't quite fit. For example, there are the
 	comparison operators <, <=, >, and >=. In C++, all four must
-	be overloaded to get complete coverage. In D, only a cmp()
+	be overloaded to get complete coverage. In D, only an opCmp()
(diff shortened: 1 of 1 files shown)


JackStouffer:
LGTM

@CyberShadow could you explain the reason for added the spec label to this?

CyberShadow:
It's a content change

JackStouffer:
But it has nothing to do with the language specification (which is what I thought the label was for).

CyberShadow:
Yes.

JackStouffer:
> Yes.

To which part of my statement?

CyberShadow:
Well, I wouldn't know what your thoughts were, would I?

All changes to dlang.org can be categorized as follows:
 1. technical (style, design, macro, build process, etc.)
 2. content (usually entirely constrained to `.dd` files)
   1. spec (the spec/ directory and what goes into the language spec PDF)
   2. non-spec (articles, style guides, etc.)

I concern myself with only the first category, with the other ones usually needing review/approval from someone else. Since 2.ii changes are relatively rare, I've been lumping them together with 2.i ones.

CyberShadow:
I've renamed the label.

JackStouffer:
Thanks for the clarification.

PR #1199, opened 2016-01-17
use no-as-needed to workaround dmd's link order issues

MartinNowak:
- mainly affects Ubuntu which uses as-needed by default


andralex:
OK, awesome! I assume this works well on all non-OSX posixes including FreeBSD etc.

MartinNowak:
> I assume this works well on all non-OSX posixes including FreeBSD etc.

Yes, should work http://www.unix.com/man-page/freebsd/1/ld/, though we don't have anyone building dlang.org on FreeBSD.

PR #1201, opened 2016-01-17
don't overwrite install.sh/install.sh.sig when rsyncing

MartinNowak:
- meant for the https://dlang.org/install.sh script

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -279,7 +279,7 @@ clean:
 	rm -f docs.json docs-prerelease.json dpl-docs/dpl-docs
 	@echo You should issue manually: rm -rf ${DMD_DIR}-${LATEST} ${DRUNTIME_DIR}-${LATEST} ${PHOBOS_DIR}-${LATEST} ${STABLE_DMD_ROOT} ${DUB_DIR}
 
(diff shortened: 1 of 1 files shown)


MartinNowak:
As a preparation for https://github.com/D-Programming-Language/installer/pull/162.
IMO it's a must to have the install script on the main site (dlang.org), but maintaining it in the installer repo makes more sense. It would then be deployed separately and thus we need to teach dlang.org to not delete it.

MartinNowak:
If you want to try it out, use this to get the latest beta.
```sh
curl -fsSL https://dlang.org/install.sh | bash -s dmd-2.070.0-b2
```

CyberShadow:
Sorry @MartinNowak, my rsync just deleted it. I've merged so it doesn't happen again :)

MartinNowak:
> Sorry @MartinNowak, my rsync just deleted it. I've merged so it doesn't happen again :)

Already noticed and fixed.
```sh
rsync -av --chmod=u=rw,go=r install.sh install.sh.sig martin@digitalmars.com:/usr/local/www/dlang.org/data/
```
The script makes testing old dmd releases really handy, we could still use a PowerShell version ;).

PR #1205, opened 2016-01-18
Fix example on the front page

9il:
`.sort` is deprecated property. In the same time `.sort()` looks weird.

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -58,7 +58,7 @@ void main()
     stdin
         .byLineCopy
         .array
-        .sort
+        .sort!((a, b) => a > b) // descending order
         .each!writeln;
 }
 ----
--- a/js/run-main-website.js
+++ b/js/run-main-website.js
@@ -10,7 +10,7 @@ Authors:   Damian Ziemba
 var mainPage = new Array();
 
 // Top rotation - sort lines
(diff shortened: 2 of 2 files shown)


JackStouffer:
I wish the `sort` property would get removed some day soon.

LGTM

CyberShadow:
> In the same time .sort() looks weird.

IMHO it's less weird than scaring people into thinking that they must always specify a sort predicate.

9il:
> IMHO it's less weird than scaring people into thinking that they must always specify a sort predicate.

Reverse order? `.sort!((a, b) => a > b)`

quickfur:
LGTM.

adamdruppe:
@9il Add the comment about reverse order and I'll merge this too.

9il:
> @9il Add the comment about reverse order and I'll merge this too.

done

CyberShadow:
I think you need to also update the example hash in `js/run-main-website.js` so that the input is carried along. (See the commentary at the top of `js/run.js` for instructions.)

9il:
> I think you need to also update the example hash in js/run-main-website.js so that the input is carried along. (See the commentary at the top of js/run.js for instructions.)

done

CyberShadow:
Thanks

PR #1206, opened 2016-01-19
Update CodeMirror

CyberShadow:
http://forum.dlang.org/post/kterugwpstrrlybqlrdl@forum.dlang.org

https://issues.dlang.org/show_bug.cgi?id=15548


aG0aep6G:
Bastiaan Veelo, who reported the issue to bugzilla, has confirmed the fix, and the files are just those from CodeMirror. So, LGTM.

CyberShadow:
Of note is that to get `codemirror-compressed.js`, you have to follow a wizard, where you select which syntax, themes etc. you want to support. I don't know what settings were used originally, but the ones I used (D syntax and nothing else) seem to work fine in practice.

aG0aep6G:
Has this been pushed online? Can the issue be marked fixed?

CyberShadow:
It is now.

PR #1209, opened 2016-01-24
fixed issue 15599

BBasile:


JackStouffer:
This is a page I won't mind seeing removed honestly. I mean, why do we need this page? Who is still on the fence about D2 that doesn't know all of the improvements? You can't even download D1 on dlang.org anymore. 

IMO the proper "fix" for this issue is to delete the page and any references to it.

BBasile:
IDC. I've just seen the issue when checking what was up since latest hours.

schuetzm:
I merged #1211, closing this one.

PR #1216, opened 2016-01-29
Link to alpha mobile compiler releases

joakim-noah:
~~Only missing [the Android logo](https://commons.m.wikimedia.org/wiki/File:Android_robot.png), would appreciate if someone would process that image as needed and submit a PR with the Android logo.  I've used the openSUSE logo as a placeholder for now.~~ No logos needed anymore, much simpler.

### DIFF
--- a/download.dd
+++ b/download.dd
@@ -37,7 +37,11 @@ $(TABLEC download-compilers,
       $(UL
         $(LI $(LINK2 http://llvm.org/, LLVM)-based D compiler)
         $(LI Strong optimization)
-        $(LI Architectures: i386, amd64, $(LINK2 https://wiki.dlang.org/Compilers#Comparison,others))
+        $(LI Alpha support for mobile:
+          $(LINK2 https://github.com/smolt/ldc-iphone-dev/releases, iOS),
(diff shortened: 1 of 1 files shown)


MartinNowak:
I wouldn't want to have those in our dmd release channels.
The download page is already too confusing, presenting visitors which a choice among dmd, ldc, and gdc.
Maybe under other downloads?

CyberShadow:
Yes, other downloads please, and I don't think this warrants a table for 2 links.

joakim-noah:
I like the way the current download page is done, presenting all three compilers and their differing strengths.  I don't see why a D compiler for mobile'd go under `Other downloads`, which aren't even D compilers.  However, I don't think it looked good the way I had it before, so I just moved it to the top, under the ldc banner.

schuetzm:
I agree with @joakim-noah that it belongs in the LDC section. Is there still opposition against the updated proposal? IMO it looks nice.

CyberShadow:
No oppositions here, just recommendations for improval. And yes, that looks nice to me, though I'd move it to the bottom line (after the Architectures line).

joakim-noah:
Thanks, hopefully these alpha releases get more usage now and beaten into shape before merging upstream.

PR #1217, opened 2016-01-29
Add documentation for new clear method (for 2.071)

schveiguy:
Also added a section on construction and reference semantics.

### DIFF
--- a/spec/hash-map.dd
+++ b/spec/hash-map.dd
@@ -327,9 +327,9 @@ Properties for associative arrays are:
         the program is done loading up a symbol table and now needs
         fast lookups in it. Returns a reference to the reorganized array.))
         $(TROW $(D .clear), $(ARGS Removes all remaining keys and values from an associative array.
-        The array is not rehashed after removal, to allow for the existing space to be reused. This
-        is not equivalent `destroy(aa)` which sets the reference to `null`, and will affect other
-        references to the instance.))
+        The array is not rehashed after removal, to allow for the existing storage to be reused.
+        This will affect all references to the same instance and is not equivalent to `destroy(aa)`
+        which only sets the current reference to `null`))
         $(TROW $(D .byKey()), $(ARGS Returns a forward range suitable for use
         as a $(I ForeachAggregate) to a $(GLINK2 statement, ForeachStatement)
         which will iterate over the keys of the associative array.))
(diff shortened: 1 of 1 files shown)


schveiguy:
See druntime PR: https://github.com/D-Programming-Language/druntime/pull/1459

CyberShadow:
Since dlang.org does not have master/stable, shall we merge this closer to release? Added a milestone.

schveiguy:
> Since dlang.org does not have master/stable

Yeah, why don't we have that? I thought we did?

CyberShadow:
> Yeah, why don't we have that? I thought we did?

Well, it's a bit complicated. If we did have that, essentially it would mean that dlang.org would always be built off the "stable" branch, and we'd have next.dlang.org OSLT for master. This implies that all changes that are not tied to specific D releases, and should go live on the next site reupload, would need to target the "stable" branch - which actually applies to the overwhelming majority of PRs. Although we could make things easier by making "stable" the default branch for the dlang.org repository, that would be inconsistent with the other components' repositories.

It's something worth thinking about, though, considering that Andrei really wants us to have docs for all past versions of D available on the website somewhere.

spec/hash-map.dd
```diff
@@ -307,6 +326,10 @@ Properties for associative arrays are:
         are more efficient. $(D rehash) is effective when, for example,
         the program is done loading up a symbol table and now needs
         fast lookups in it. Returns a reference to the reorganized array.))
+        $(TROW $(D .clear), $(ARGS Removes all remaining keys and values from an associative array.
+        The array is not rehashed after removal, to allow for the existing space to be reused. This
+        is not equivalent `destroy(aa)` which sets the reference to `null`, and will affect other
+        references to the instance.))
         $(TROW $(D .byKey()), $(ARGS Returns a forward range suitable for use
```
MartinNowak: Better "This will affect all references to the same instance and is not equivalent to `destroy(aa)` which only sets the current reference to `null`".

schveiguy:
What about a branch for each future release that gets merged for the release? I hate having a lingering PR that's "approved but not merged until we get there." What happens if there are several PRs that have conflicts? You'll have to sort that out at the release, which won't be fun.

I agree that the master branch should be the "stable" version.

schveiguy:
Updated table per @MartinNowak

schveiguy:
Note, this already appears on the web site's [documentation](http://dlang.org/spec/hash-map.html#properties), even though 2.071 is not out. Resulting issue: https://issues.dlang.org/show_bug.cgi?id=15753

PR #1221, opened 2016-01-31
HTTP -> HTTPS conversion for several links

sarneaud:
The following domains support HTTPS, so I've converted links:

code.dlang.org
dlang.org
forum.dlang.org
issues.dlang.org
wiki.dlang.org

bitbucket.org
creativecommons.org
docs.oracle.com
docs.sun.com
en.wikipedia.org
github.com
goo.gl
google.com
groups.google.com
i.creativecommons.org
msdn.microsoft.com
stackoverflow.com
twitter.com
w3.org
www.amazon.com
www.gnu.org
www.w3.org


schuetzm:
Thank you!

PR #1224, opened 2016-02-03
Add caveat for opApply and exceptions.

Shachar:
As discussed on the forum, add explanation about exception handling inside opApply, and their interaction with exceptions thrown from within the opApply delelgate.

### DIFF
--- a/spec/statement.dd
+++ b/spec/statement.dd
@@ -682,6 +682,12 @@ $(CONSOLE
         }
         --------------
 
+    $(P It is important to make sure that, if $(D opApply) catches any exceptions, that those
+        exceptions did not origin at the delegate passed to $(I opApply). Catching an exception
+        that escaped from the delegate would cause a condition that the user expected would terminate
+        the loop to not do so.
+     )
+
 $(H4 $(LEGACY_LNAME2 foreach_with_ranges, foreach-with-ranges, Foreach over Structs and Classes with Ranges))
 
     $(P If the aggregate expression is a struct or class object, but the


schveiguy:
Link to discussion: https://forum.dlang.org/post/n8ptpu$1ncu$1@digitalmars.com

spec/statement.dd
```diff
@@ -682,6 +682,12 @@ $(CONSOLE
         }
         --------------
 
+    $(P It is important to make sure that, if $(D opApply) catches any exceptions, that those
+        exceptions did not origin at the delegate passed to $(I opApply). Catching an exception
+        that escaped from the delegate would cause a condition that the user expected would terminate
+        the loop to not do so.
```
schveiguy: 1. "origin at" -> originate from
2. The last sentence is cryptic. Possibly better: "The user would expect exceptions thrown from a `foreach` body to both terminate the loop, and propagate outside the `foreach` body."

schuetzm: Also, _that_ appears twice: "make sure **that**, ..., **that** those"

adamdruppe:
@Shachar  if you update the wording a bit from the comments, i'll pull this.

AndrejMitrovic:
Hi @Shachar, could you amend the suggested fixes? If you do not have the time we can take over. Thanks!

andralex:
I'll pull now and make the changes as a fixup.

andralex:
https://github.com/dlang/dlang.org/commit/ca83d45b1d8e7576639faac50f67d5bce0737277

PR #1226, opened 2016-02-04
document macros REF and REF_ALTTEXT

aG0aep6G:


CyberShadow:
Not that it really matters, but this kinda seems like a misuse of the `COMMENT` macro, since it's not doing anything here and someone looking at the code might think it does. Maybe we could use `_=` as a sort of comment prefix (like `//` in C/D).

aG0aep6G:
> Not that it really matters, but this kinda seems like a misuse of the COMMENT macro, since it's not doing anything here and someone looking at the code might think it does. Maybe we could use _= as a sort of comment prefix (like // in C/D).

I think `COMMENT` is the equivalent of `//`. It's [defined as nothing in macros.ddoc](https://github.com/D-Programming-Language/dlang.org/blob/a34cd8e26a92c5f3b932b8f2afeb8c8dd9d667ff/macros.ddoc#L5), and it doesn't seem to be overridden anywhere. There's also `DDOC_COMMENT` which is [defined as an HTML comment in html.ddoc](https://github.com/D-Programming-Language/dlang.org/blob/a34cd8e26a92c5f3b932b8f2afeb8c8dd9d667ff/html.ddoc#L112). I would have named them the other way around, though.

PR #1227, opened 2016-02-08
Increased the size of the header and logo.

nomad-software:
I've increased the size of the D logo in the header as I feel it was a little on the small side. I've also adjusted the navigation and spacing in the header to account for the new size. It should also work fine in different sizes for mobile and desktop, etc. This is purely a CSS stylesheet change.

Preview: http://nomad.so/misc/d/dlang.org/increase-header-size/

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -110,14 +110,14 @@ body > .container
 {
     display: block;
     float: left;
-    margin-left: -5.4em;
-    margin-right: -4em;
-    padding-top: 1px;
+    margin-left: -10.75em;
+    margin-right: -7.25em;
+    padding-top: 6px;
 }
 
 #top img#logo
 {
-    height: 2.533em;
+    height: 5em;
     vertical-align: middle;
     width: auto;
 }
@@ -168,6 +168,11 @@ body > .container
     display: block;
 }
 
+#top #cssmenu .expand-container > .expand-content
+{
+    min-width: 180px;
+}
+
 #top #cssmenu ul
 {
     list-style: none;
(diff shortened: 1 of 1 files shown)


aG0aep6G:
I think I like the smaller one better.

CyberShadow:
Oof, no, I think this is not an improvement. It's much too big, especially on small screens. As-is, it would mark a massive decrease of usability on the forum on phones.

There is also an implementation problem - on some screen widths, the navigation bar wraps and takes two "rows".

nomad-software:
> Oof, no, I think this is not an improvement. It's much too big, especially on small screens. As-is, it would mark a massive decrease of usability on the forum on phones.

Have you got an example of lack of usability on small screen devices? Also the forums are as of yet unaffected.

> There is also an implementation problem - on some screen widths, the navigation bar wraps and takes two "rows".

Fixed.

JackStouffer:
Sorry, I'm really not a fan of this. IMO this is a waste of precious screen space, and navigation should be less obtrusive not more.

CyberShadow:
> Have you got an example of lack of usability on small screen devices?

Just go to forum.dlang.org on your phone, rotate the phone to portrait orientation, and select the horizontal-split view mode.

> Also the forums are as of yet unaffected.

The styles of the two websites should stay in sync, so any change in dlang.org's style implies it will need to trickle down to the forum too.

CyberShadow:
> I think I like the smaller one better.

> Sorry, I'm really not a fan of this.

> I prefer the current version as well.

Well, I think we have consensus here. Sorry.

eco:
One alternative might be to have this on the front page and the smaller one on the more content rich pages.

nomad-software:
Ok, the community has spoken but I will add one point.

We shouldn't be designing the site from a mobile point of view, especially because it's a site that will be used by developers while developing on desktop PC's. The site should scale to whatever device you are on. I personally think the logo should be bigger on desktop and this in no way affects the mobile experience as that can be a different size, etc.

Also, the comments about the forums are moot as they haven't been touched in this PR. That was going to be addressed in a later PR if this one was ok'ed.

p0nce:
I liked the proposal better than the current. It would be cool to have something intermediate.

ZombineDev:
+1 for something intermediate in terms of size.

d-random-contributor:
[screenshot](http://abload.de/img/tmprxqkw.png)

PR #1230, opened 2016-02-10
Add clarification for abstract function bodies.

JohanEngelen:
I find it clearer to explicitly state that `abstract int foo() { return 0; }` renders the encompassing class un-instantiatable, even though all function bodies are available.


adamdruppe:
I like it.

JackStouffer:
LGTM

JohanEngelen:
Thanks for merging.
(I notice that I adhered to the paragraph's style to use "still" in every sentence)

PR #1231, opened 2016-02-10
Update ctod.dd

Kozzi11:
Fix static array declaration

### DIFF
--- a/ctod.dd
+++ b/ctod.dd
@@ -266,7 +266,7 @@ for (i = 0; i &lt; ARRAY_LENGTH; i++)
 $(H4 The D Way)
 
 ----------------------------
-int array[17];
+int[17] array;
 array[] = value;
 ----------------------------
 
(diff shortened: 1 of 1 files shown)


ctod.dd
```diff
@@ -266,8 +266,7 @@ for (i = 0; i &lt; ARRAY_LENGTH; i++)
 $(H4 The D Way)
 
 ----------------------------
-int array[17];
-array[] = value;
+int[17] array = value;
```
aG0aep6G: I think the assignment should stay separate. The initialization way might be a bit misleading, as it only works with fixed-size arrays.

Kozzi11: Ok, what is better:
int[17] array;
array[] = 7; // or array = 7;

aG0aep6G: > array[] = 7; // or array = 7;

I'm for `array[] = value;`. The other variant only works with fixed-size arrays again.

ZombineDev:
LGTM

9rnsr:
Looks good, too.

PR #1233, opened 2016-02-14
Rework expandable code example boxes

sarneaud:
Accessibility improvements:
* controls now accessible by pressing tab
* WAI-ARIA attributes added
* still has graceful degradation without JavaScript

Other improvements:
* boxes now open with quick sliding animation
* (subjective) less work in the macro means easier maintainability by
  web developers new to D

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1523,13 +1523,16 @@ div.changelog-nav
     display:none;
 }
 
-.answer-nojs {
-    display:block;
-}
-
 .nobr {
     white-space:nowrap;
 }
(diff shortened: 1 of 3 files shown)


sarneaud:
I've tested this with Firefox (38), Chromium (48) and Midori (0.5.10).  Judging by http://caniuse.com/ it should be usable even in very old browsers.  You might want to try it out in other browsers if you have them, though.

CyberShadow:
Nice, LGTM

PR #1235, opened 2016-02-16
added example for the trait getVirtualIndex

BBasile:


Kapps:
While it is just an example, it seems questionable to encourage such code to work around @disable by providing an example. I don't know if it's necessarily guaranteed that code marked @disable even gets compiled in to the generated program, let alone its own vtable index, or that this is allowed at all since @disable is supposed to prevent referencing the declaration.

Personally I used the trait to implement reflection to allow something like:
````
class Foo {
	int vmethod() { return 0; }
}

// Defined elsewhere:
class FooBar : Foo {
	override int vmethod() { return 1; }
}

void main() {
	// Get the index of the method using the type we know at compile-time.
	ptrdiff_t index = __traits(getVirtualIndex, Foo.vmethod);

	Foo foo = new FooBar;
	// Set up a delegate using the vtbl
	int delegate() dg;
	dg.funcptr = cast(int function()) foo.__vptr[index];
	dg.ptr = cast(void*) foo;
	// Returns the value of FooBar rather than Foo.
	assert(dg() == 1);
}
````

BBasile:
I see what you mean but I don't agree, I think that my example shows well that "traits" are something special.

I'm also the author of https://github.com/D-Programming-Language/dmd/pull/5289
Just to say that I think that something is not quite right with traits.

Anyway if it's not ok, no problem you can close.

Kapps:
Shrug, it's just my own opinion on it. Best to wait for someone else to weigh in.

PR #1237, opened 2016-02-18
Fix issue 15697 - The TWID script is always served over HTTP

Kapps:
This changes the script to use // for the protocol to fix issues like the script being ignored by Chrome and warnings being displayed in Edge / IE.

Would like to get approval from @adamdruppe that it's okay to serve the script from his website over HTTPs, in case the extra load is an issue.

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -90,7 +90,7 @@ $(DIVC boxes,
                 before it's sold out!
             )
             $(P Stay updated with
-                $(LINK2 http://arsdnet.net/this-week-in-d,
+                $(LINK2 //arsdnet.net/this-week-in-d,
                     <cite>This Week in D</cite>),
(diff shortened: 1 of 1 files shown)


JackStouffer:
> in case the extra load is an issue.

HTTPS hasn't been a resource drag since 2005. This should be fine.

LGTM

Kapps:
Err, I made a mistake. Did this over the web interface and was waiting for the bot to generate the documentation to test it, but I updated the wrong link. Will fix in a second.

JackStouffer:
No need to close, just force push. It's too late now though

Kapps:
Didn't realize I wouldn't be able to reopen after a force push. Recreated with pull #1238.

As for the HTTPs, I suspect it's fine as well, but figured I'd ask in case there is any issues.

PR #1238, opened 2016-02-18
Fix issue 15697 - The TWID script is always served over HTTP

Kapps:
This changes the script to use // for the protocol to fix issues like the script being ignored by Chrome and warnings being displayed in Edge / IE.

Recreated pull after changing wrong link originally. I didn't test locally, so will need to wait for the documentation bot to generate it, I'll update the pull when checked.


Kapps:
Checked the generated documentation, looks correct now.

adamdruppe:
Yeah, we didn't do it before because I didn't have a https certificate installed, but I fixed that a couple weeks ago (and thought you guys already changed this!).

So all good, I'll merge it myself.

PR #1245, opened 2016-03-03
use https for google search

greenify:
If I currently visit https://dlang.org, I get the following error:

```
Mixed Content: The page at 'https://dlang.org/' was loaded over a secure connection, but contains a form which targets an insecure endpoint 'http://google.com/search'. This endpoint should be made available over a secure connection.
```

Btw since 2014 Google is [preferring HTTPS sites](https://googlewebmastercentral.blogspot.fi/2014/08/https-as-ranking-signal.html), so why not make SSL the default?

### DIFF
--- a/dpl-docs/views/layout.dt
+++ b/dpl-docs/views/layout.dt
@@ -115,7 +115,7 @@ html(lang='en-US')
             a.expand-toggle(href="#{root_dir}search.html", title="Search")
               span Search
             #search-box
-              form(method='get', action='http://google.com/search')
+              form(method='get', action='https://google.com/search')
                 input#domains(type='hidden', name='domains', value='dlang.org')
                 |             
                 input#sourceid(type='hidden', name='sourceid', value='google-search')


Hackerpilot:
:white_check_mark:

CyberShadow:
> If I currently visit https://dlang.org, I get the following error:

The patch you submitted is only for the DDox template, and only affects pages under http://dlang.org/library.

greenify:
@CyberShadow thanks - #1247 should then do the trick?

CyberShadow:
Yep - you can use the doc autotester to preview/test your changes BTW.

Generally when replacing HTTP links they should probably be replaced with protocol-relative URLs, but in this case HTTPS is fine as Google always redirects to HTTPS (I think).

greenify:
> Yep - you can use the doc autotester to preview/test your changes BTW.

I am quite okay with compiling `dlang.org` locally - it works now for all pages ;-)

> Generally when replacing HTTP links they should probably be replaced with protocol-relative URLs, 

A couple of years ago it was a common pattern to have `<a href=//foobar.de>`. This is changing now, because many sites have realized that there is no good reason not to enforce SSL.

>but in this case HTTPS is fine as Google always redirects to HTTPS (I think).

Yup Google will _always_ redirect HTTP to HTTPS.
http://googlesystem.blogspot.fi/2013/09/google-encrypted-search-for-everyone.html

PR #1246, opened 2016-03-04
add missing anchors from "functions" to "Embedded Documentation"

BBasile:
add more missing anchors, from "functions to "Embedded Documentation"

following the order in the site menu is remaining to do:
 -  operator overloading: DAutotestfailure because ?
 -  Const and Immutable: const3.dd, need rewritting to allow LNAME, paragraphs are surrounded with $(SECTION ...), the markup style is not super friendly
 -  _Interfacing to C_ to _Vector Extensions_.
 -  maybe the pages that are not located in the spec folder.

If acceptable, this PR can be merged as it or you can wait for the remaining pages to be processed.

### DIFF
--- a/spec/contracts.dd
+++ b/spec/contracts.dd
@@ -29,7 +29,7 @@ $(COMMENT <img src="../images/d4.gif" alt="Contracts make D bug resistant" borde
 	out of date, wrong, or non-existent. Moving the contracts into the code makes them
 	verifiable against the program.)
 
-$(H2 Assert Contract)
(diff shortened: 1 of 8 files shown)


BBasile:
This is not a pleasant work to do but, the problem is so: sometimes you want to help someone in **.learn** and you want to put a link. But you can't because the there's no anchor at a particular place. To prevent this problem we have to put some anchors in the dd files.

And it's not a joke. It really happens.
Last time:

https://issues.dlang.org/show_bug.cgi?id=15683#c3
> (seriously, still not linked? ugh)

PR #1248, opened 2016-03-11
We should be including core.stdc in the menu

schveiguy:


schveiguy:
Note, the vast number of additions is just because the menu appears in every single html file, and we are adding many links to the menu. The files themselves were already generated and on the server.

JackStouffer:
I don't see why not.

LGTM

CyberShadow:
LGTM though can't vouch for the Windows patch.

schveiguy:
Windows uses c too, no?

CyberShadow:
Yeah. What I meant is that I can't test the win32.mak patch right now.

schuetzm:
The win32.mak change is trivial, let's just risk it.

schveiguy:
thanks!

PR #1250, opened 2016-03-12
add missing anchors - complete the pages in the language reference

BBasile:


spec/abi.dd
```diff
@@ -9,16 +9,15 @@ $(SPEC_S Application Binary Interface,
 	D binaries built by other implementations.
 	)
 
-$(SECTION3 C ABI,
+$(H3 $(LNAME2 c_abi, C ABI))
```
CyberShadow: Why not e.g. `$(SECTION3 $(LNAME2 c_abi, C ABI)),`?

BBasile: personal taste, SECTION<X> tend to lead to right paren error since the content is surrounded, which is not the case when using H<X>.

But I've checked the output on your site and it looks OK.

CyberShadow:
LGTM, IDK about the SECTION3 changes though. I thought some output formats might use the nesting info but it doesn't seem so. Would be nice if the whitespace changes were separated too.

BBasile:
I'll accept to take the blame if one day someone comes and complains about that. I thought that SECTIOn is the same as H.

CyberShadow:
I am a master nitpicker so feel free to ignore me :)

PR #1251, opened 2016-03-15
use https for internal documentation links

greenify:
why don't we use `https` for the links within phobos's documentation?


JakobOvrum:
All the macros edited here should be removed in the near future, but other than that, LGTM.

greenify:
With what will it be replaced if I may ask?

JakobOvrum:
`REF`.

PR #1252, opened 2016-03-21
Additional template doc example

earthfront:
Added content to template doc page.

### DIFF
--- a/spec/template.dd
+++ b/spec/template.dd
@@ -1406,7 +1406,7 @@ $(GNAME Constraint):
         import std.traits;
         
         struct Bar(T)
-            if( isIntegral!T )
+            if (isIntegral!T)
         {
             ...
         }
--- a/spec/template.dd
+++ b/spec/template.dd
@@ -1381,7 +1381,7 @@ $(GNAME Constraint):
         The $(I Expression) is computed at compile time
         and returns a result that is converted to a boolean value.
         If that value is true, then the template is matched,
-        otherwise the template is not matched and compilation fails.
+        otherwise the template is not matched.
     )
 
     $(P For example, the following function template only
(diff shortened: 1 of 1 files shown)


spec/template.dd
```diff
@@ -1381,7 +1381,7 @@ $(GNAME Constraint):
         The $(I Expression) is computed at compile time
         and returns a result that is converted to a boolean value.
         If that value is true, then the template is matched,
-        otherwise the template is not matched.
+        otherwise the template is not matched and compilation fails.
```
CyberShadow: The compilation only fails if no template matches. There might be other overloads with different constraints.

spec/template.dd
```diff
@@ -1394,11 +1394,27 @@ $(GNAME Constraint):
...
+    )
+    
+    $(P Template constraints can be used with aggregate types (structs, classes, unions). 
+        Constraints are effectively used with library module "std.traits":
+    
+        ---
+        import std.traits;
+        
+        struct Bar(T)
+            if( isIntegral!T )
```
CyberShadow: Nit, please use Phobos style in examples

earthfront: Can you be more specific? 
This seems to abide, according to "https://dlang.org/dstyle.html". Further, I just copied the style from the example immediately above.

spec/template.dd
earthfront: Good call. Changed.

spec/template.dd
CyberShadow: Sorry, this isn't in the style guide yet:

https://github.com/D-Programming-Language/dlang.org/pull/1191

Essentially we place the spaces outside the parens, not inside.

The same style is used in other examples on this page, so it would be nice to avoid inconsistent code style.

earthfront: Got it. Seems to apply to conditional statements only?

PR #1257, opened 2016-03-27
Implement REF_ALTTEXT macro for DDox

JakobOvrum:
For DDoc,
```
$(REF_ALTTEXT writeln, File.writeln, std, stdio)
```
expands to
```
<a href="std_stdio.html#.File.writeln">writeln</a>
```
which relies on the fact that `File.writeln` is part of the URL. To provide identical functionality for DDox, it would have to expand to:
```
a href="/library/std/stdio/File/writeln>writeln</a>
```
`REF` handles this by expanding to `std.stdio.File.writeln` and letting DDox recognize it as a symbol path, but I'm not aware of an equivalent feature in DDox that works like `REF_ALTTEXT`.

This patch makes it expand to:
```
writeln (std.stdio.File.writeln)
```
It's not ideal but it's an improvement over the broken output it currently generates.

### DIFF
--- a/std-ddox.ddoc
+++ b/std-ddox.ddoc
@@ -8,5 +8,6 @@ ROOT_DIR=/
 _=
 
 REF=$(D $2$(DOT_PREFIXED_SKIP $+).$1)
+REF_ALTTEXT=$1 ($(REF $+))
 MREF=$(D $1$(DOT_PREFIXED $+))
 _=


CyberShadow:
LGTM though it doesn't look like it's used anywhere?

aG0aep6G:
DDOX recently got a macro `DDOX_ROOT_DIR` ([commit](https://github.com/rejectedsoftware/ddox/commit/e4fcead4a6e25cd8da7602e4709f250e55869240)). With that it should be possible to define the DDOX `REF_ALTTEXT` similarly to the Ddoc one.

That would require updating DDOX to 0.13.3 in the dependencies of dpl-docs ([here](https://github.com/D-Programming-Language/dlang.org/blob/51746598464affbbea860ba27d1bc99ccde3eb1a/dpl-docs/dub.json#L4)).

JakobOvrum:
@CyberShadow, it's used in a couple of PRs, including D-Programming-Language/Phobos#3911. It's intended as a companion to `REF` replacing all the `_NAMED` macros... but it's not a straight upgrade considering the less-than-ideal DDox implementation. Right now those macros simply link from DDox to DDoc.

@aG0aep6G, the root directory isn't the problem - we have `ROOT_DIR` for dlang.org. There's no way to "replace" dots with slashes using DDoc macros.

aG0aep6G:
> @aG0aep6G, the root directory isn't the problem - we have ROOT_DIR for dlang.org. There's no way to "replace" dots with slashes using DDoc macros.

Ah, right. Sorry for the noise. This sucks.

JakobOvrum:
Yeah, it sucks. DDoc macros can only have one variadic component and we have no way of "splitting" one argument into many.

Markdown has `[text here](link here)`. There's been some murmuring about Markdown before and maybe it would provide a better way out. For now the implementation in this PR is all I got.

PR #1260, opened 2016-04-04
fix Issue 15873 - compile time info about CPU specifics is needed

WalterBright:
https://issues.dlang.org/show_bug.cgi?id=15873

https://github.com/D-Programming-Language/phobos/pull/2862

Fix documentation, as apparently the description for `D_SIMD` is inadequate.


DmitryOlshansky:
Doesn't say anything about SSE2 while I belive it should.

PR #1262, opened 2016-04-09
Update spec to follow .outer property behavior change introduced in 2.071

9rnsr:
See: https://issues.dlang.org/show_bug.cgi?id=15839
Issue 15839 - this.outer is of wrong type

### DIFF
--- a/spec/class.dd
+++ b/spec/class.dd
@@ -1241,33 +1241,69 @@ int bar()
         )
 
         $(P The property $(D .outer) used in a nested class gives the
-        $(D this) pointer to its enclosing class. If the enclosing
-        context is not a class, the $(D .outer) will give the pointer
-        to it as a $(D void*) type.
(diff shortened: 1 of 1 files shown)


spec/class.dd
```diff
@@ -1241,33 +1241,69 @@ int bar()
         )
 
         $(P The property $(D .outer) used in a nested class gives the
-        $(D this) pointer to its enclosing class. If the enclosing
-        context is not a class, the $(D .outer) will give the pointer
+        $(D this) pointer to its enclosing class. If there is no  enclosing
```
klickverbot: "If there is no enclosing class context ... pointer to it?"

9rnsr: Fixed to: "If there is no enclosing class context, .outer would return a pointer to enclosing function frame with void*."

DmitryOlshansky:
LGTM

PR #1265, opened 2016-04-12
posix.mak: Update MOD_EXCLUDES lists

CyberShadow:
http://forum.dlang.org/post/ptnoowrvpsliwtxcmfoi@forum.dlang.org

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -36,11 +36,11 @@ STABLE_RDMD=$(STABLE_DMD_ROOT)/dmd2/$(OS)/$(if $(filter $(OS),osx),bin,bin$(MODE
 
 # exclude lists
 MOD_EXCLUDES_PRERELEASE=$(addprefix --ex=, gc. rt. core.internal. core.stdc.config core.sys.	\
-	std.c. std.algorithm.internal std.internal. std.regex.internal. std.typelist		\
+	std.c. std.algorithm.internal std.internal. std.regex.internal. 			\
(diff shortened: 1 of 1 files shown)


CyberShadow:
`_PRERELEASE` / `_RELEASE` macros seem like an additional point of failure for the release process. Perhaps they should be named according to the versions they imply to (e.g. `MOD_EXCLUDES_2071`) so that the build will fail if they are not updated.

posix.mak
```diff
@@ -36,11 +36,11 @@ STABLE_RDMD=$(STABLE_DMD_ROOT)/dmd2/$(OS)/$(if $(filter $(OS),osx),bin,bin$(MODE
 
 # exclude lists
 MOD_EXCLUDES_PRERELEASE=$(addprefix --ex=, gc. rt. core.internal. core.stdc.config core.sys.	\
-	std.c. std.algorithm.internal std.internal. std.regex.internal. std.typelist		\
+	std.c. std.algorithm.internal std.internal. std.regex.internal. 			\
 	std.windows.iunknown std.windows.registry etc.linux.memoryerror std.stream std.cstream	\
```
ZombineDev: Somewhat off topic, but why is `std.windows.registry` excluded? It doesn't have a deprecation note. (The other excluded modules are correct).

CyberShadow: Not at all off-topic. It's excluded because it's a Windows-only module, and unlike the other Windows-only modules not excluded here, it doesn't have a `version(StdDdoc)` block which only has the documentation.

PR #1267, opened 2016-04-13
all language reference should link to spec.html NOT intro.html

Passw:
for spec.html we can see contents of spec and download link to *.pdf and *.mobi


schuetzm:
Makes sense IMO.

schuetzm:
Or alternatively, could we integrate the download links into the introduction page and remove spec.html? The table of contents is redundant.

CyberShadow:
LGTM.

PR #1270, opened 2016-04-18
Move package.d modules navigation entries into their package menu.

s-ludwig:
See https://forum.dlang.org/thread/nerjg7$2tqv$1@digitalmars.com


CyberShadow:
Well, it's an improvement!

s-ludwig:
Yeah, not ideal, but better than nothing.

CyberShadow:
Placing the package documentation on the package node is not possible because the click is used to expand/collapse the node. Perhaps that action could be restricted to just the arrow? Alternatively, the DDoc documentation seems to fare well enough without the whole accordion thing, but its pages are longer since it puts all module symbols on one page.

PR #1274, opened 2016-04-19
Upgrade dpl docs dependencies

s-ludwig:
Fixes:
- Types now have some CSS styling
- Cross references within backtick code work again
- Fixed invalid nesting of HTML elements in some cases
- Function attributes are highlighted
- Automatic cross-referencing outside of D code is now disabled
- `@...` built-in attributes are now highlighted as keywords

Still missing: sometimes keywords are not recognized as such, seems to be [a bug](https://github.com/Hackerpilot/libdparse/issues/94) in libdparse.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1119,6 +1119,12 @@ These classes come from prettify.js:
     color: #4040DC;
 }
 
+.typ
+{
+    color: #4040DC;
+}
+
+
 .d_param
 {
     font-style: italic;
--- a/dpl-docs/dub.json
+++ b/dpl-docs/dub.json
@@ -1,7 +1,7 @@
 {
 	"name": "dpl-docs",
 	"dependencies": {
-		"ddox": "~>0.14.0"
+		"ddox": "~>0.15.0"
 	},
 	"versions": ["VibeCustomMain"]
 }
--- a/dpl-docs/dub.json
+++ b/dpl-docs/dub.json
@@ -1,7 +1,7 @@
 {
 	"name": "dpl-docs",
 	"dependencies": {
-		"ddox": "~>0.13.0"
+		"ddox": "~>0.14.0"
 	},
 	"versions": ["VibeCustomMain"]
 }
--- a/dpl-docs/dub.selections.json
+++ b/dpl-docs/dub.selections.json
@@ -2,7 +2,7 @@
 	"fileVersion": 1,
 	"versions": {
 		"backtrace-d": "~master",
-		"ddox": "0.13.5",
+		"ddox": "0.14.2",
 		"experimental_allocator": "2.70.0-b1",
 		"hyphenate": "1.1.1",
 		"libasync": "0.7.9",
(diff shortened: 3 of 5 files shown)


CyberShadow:
> Automatic cross-referencing outside of D code is now disabled

Awesome. Now we just need to change the spec to match :)

s-ludwig:
Just saw that `@..` is currently rendered incorrectly - should have waited for the CI to pass.

CyberShadow:
```
Selected package ddox 0.14.2 doesn't exist. Using latest matching version instead.
Root package dpl-docs contains reference to invalid package ddox >=0.14.0 <0.15.0
```
Hmm, do I need to do something on the server again?

s-ludwig:
> Hmm, do I need to do something on the server again?

We really need to get out a new version of DUB. This is still the metadata cache issue (it doesn't recognize new package versions before 24h have passed, or if `dub clean-caches` is run).

> Now we just need to change the spec to match

That would be really nice, with support for backtick code, the current "Emphasis"  handling appears like a useless crutch. BTW, looks like the spec got actually changed to allow links to be used - Walter used to oppose that idea when I first told him how DDOX works.

CyberShadow:
Hmm, with no links, the cheat sheets in module documentation are now useless:
http://dtest.thecybershadow.net/artifact/website-506f641ae1b0c3bdf4d23fa5150af8077b24db0f-d574e16d12b7bce09f483558c3467b83/web/library/std/traits.html

CyberShadow:
Maybe `LREF` in `std-ddox.ddoc` should be changed to `$(D $1)`?

CyberShadow:
Same with XREF and other macros there.

CyberShadow:
Looks good!

Style nit, there are some links with blue text and a red underline, which clash a bit.

CyberShadow:
Ouch, there's some HTML visible:
http://dtest.thecybershadow.net/artifact/website-7f3804b809d871c676c056e8def98a0cc7bf208c-99de2fdd99ed0a6a8cdb703744440eb6/web/library-prerelease/std/range.html

CyberShadow:
Looks good. BTW, in the future add a comment to a PR after updating it, as there's otherwise no notification, and I may forget to come back to it.

CyberShadow:
dlang.org updated.

s-ludwig:
I forgot about it after waiting for the test build to finish (got company).                                                                                                                                                                                                                                                                                                                                        Gesendet von meinem BlackBerry 10-Smartphone.                                                                                                                                                                                                                Von: Vladimir PanteleevGesendet: Mittwoch, 20. April 2016 04:15An: dlang/dlang.orgAntwort an: dlang/dlang.orgCc: Sönke Ludwig; AuthorBetreff: Re: [dlang/dlang.org] Upgrade dpl docs dependencies (#1274)Looks good. BTW, in the future add a comment to a PR after updating it, as there's otherwise no notification, and I may forget to come back to it.

—You are receiving this because you authored the thread.Reply to this email directly or view it on GitHub

PR #1277, opened 2016-04-21
update old links to digitalmars.com

wilzbach:
Hey,

while browsing dlang.org, I just stumbled across a couple of old links.
Imho there's still much to find, so if someone has a bit of time, search e.g. with this:

```
grep "digitalmars.com" -n --exclude-dir=web --exclude='*.json' -r .
```


howto-promote.dd
```diff
@@ -83,7 +85,6 @@ $(H3 Promoting the Project)
 	)
 
 	$(UL
-	$(LI $(LINK2 http://www.ddj.com, Dr. Dobb's Journal))
```
wilzbach: afaict it's archive-only:
http://www.drdobbs.com/architecture-and-design/farewell-dr-dobbs/240169421

CyberShadow:
Looks good, though those pages need some heavy lifting in general.

howto-promote.dd
```diff
@@ -55,13 +55,15 @@ $(H3 Promoting the Project)
 	)
 
 	$(UL
-	$(LI $(LINK2 http://www.digitalmars.com/pnews/indexing.php?server=news.digitalmars.com$(AMP)group=digitalmars.D.announce, digitalmars.D.announce) newsgroup)
+	$(LI $(LINK2 http://forum.dlang.org/group/announce) newsgroup)
+	$(LI $(LINK2 https://wiki.dlang.org/) D wiki)
```
CyberShadow: The `LINK2` syntax is wrong here.

wilzbach: ah thanks - we should really just use markdown ... :/

PR #1278, opened 2016-04-21
add third-party pages to downloads

wilzbach:
I think it would be nice to mention all the [hard work](https://www.archlinux.org/groups/x86_64/dlang/) that @Dicebot is doing to distribute official dlang packages to Arch Linux users.

While I was editing the page, I also saw that you never mention the "d-apt" repository. I imagine it to be quite convenient for Ubuntu/Debian users as they automatically get the updates.

It looks like this on my machine:

image

PS: I would like to see a more prominent place for both distribution forms, but I couldn't find a better place.


CyberShadow:
It should be made clear that these are third party download links and are not maintained by the same people maintaining the official downloads. Distribution package maintainers have come and went in the past.

What's the plan for inclusion criteria? Arch and Debian aren't the only distros with D packages.

wilzbach:
> It should be made clear that these are third party download links and are not maintained by the same people maintaining the official downloads. 

Done.

> What's the plan for inclusion criteria?

I would argue for a very relaxed criteria - e.g. everything that has been updated within the last six months.
Most people not using these distro probably won't care and for those who do it's an additional convenience.

>Arch and Debian aren't the only distros with D packages.

What else is there?

CyberShadow:
> What else is there?

I know of Gentoo:

https://wiki.gentoo.org/wiki/Dlang

CyberShadow:
CC @MartinNowak @ximion @mleise

wilzbach:
> I know of Gentoo:

added ;-)

ximion:
You can get gdc, ldc and dub in Debian and Ubuntu from the official sources now.
The Debian D team is a bit, well, dead though ;-) Doesn't mean that this can't be fixed :)

mleise:
:+1: Looks good.

wilzbach:
> You can get gdc, ldc and dub in Debian and Ubuntu from the official sources now.
> The Debian D team is a bit, well, dead though ;-) Doesn't mean that this can't be fixed :)

I just opened a [PR](https://github.com/ldc-developers/ldc/pull/1451) at the ldc repo to show an overview of it's packages in distributions.
I am not sure whether it's possible to also display it here too (would be nice though).

Moreover "dub" isn't mentioned at all on the entire page, but maybe this is on purpose?

JackStouffer:
> What else is there?

Homebrew

wilzbach:
> Homebrew

Added. Thanks!
What about [Fedora](https://build.opensuse.org/package/show/devel:languages:D/dmd)?

CyberShadow:
While we're enumerating package managers, there's also Chocolatey:

https://chocolatey.org/packages/dmd

CyberShadow:
> Moreover "dub" isn't mentioned at all on the entire page, but maybe this is on purpose?

There's a "Packages" link at the very top (in the navigation) :)

Eventually dub will be packaged with dmd (once it hits 1.0 AIUI, which should be really soon AIUI).

wilzbach:
> there's also Chocolatey

Done - I picked their new logo. Is there any specific order I should in which I sort them?

wilzbach:
> Is there any specific order I should in which I sort them?

1) I just sorted them alphabetically.
2) I added the CLI instructions to add the Ubuntu/Debian repo
3) My PR at LDC to improve the installation guide was merged, so we now have a better ldc installation page to link to for the download link.

wilzbach:
Update: I just realized that you host a [copy of the d-apt GPG](http://dlang.org/d-keyring.gpg) on dlang.org -> updated debian instructions

wilzbach:
Ping - so can we move forward with this?

[Preview](http://dtest.thecybershadow.net/artifact/website-3144137b28c420e388e635502d98a21dd496a3d2-2b9174c95bdfe92255624e234eb998e0/web/download.html)

CyberShadow:
I thought the Chocolatey logo looked more like this:
https://chocolatey.org/Content/Images/nugetlogo.png

wilzbach:
> Done - I picked their new logo

http://codebetter.com/robreynolds/2016/04/19/chocolatey-has-a-new-logo/

CyberShadow:
Ah, OK.

CyberShadow:
I have to say that this all looks nice except for the Debian instructions, which are like ~13 times longer than the other ones.

Weren't there some helpers that make adding apt repositories easier and in one command?

CyberShadow:
CC @jordisayol

ximion:
`dub`, `ldc` and `gdc` are part of the Debian distribution, additional repositories can also be added using the `add-apt-repository` command, but only if the package `software-properties-common` is installed, which might not be the case.

wilzbach:
> Weren't there some helpers that make adding apt repositories easier and in one command?

Afaik not, but @jordisayol might know better.

> dub, ldc and gdc are part of the Debian distribution, additional repositories can also be added using 

Ideally we should provide the install instructions for ldc & gdc directly on this download page, but I guess this is part of the longer step to get them under the D foundation brand.

Dub is currently not displayed at all at this site, but for the next release it will be bundled with dmd and thus I expect we will change this page anyways?

So @CyberShadow - how do we proceed?

CyberShadow:
> Ideally we should provide the install instructions for ldc & gdc directly on this download page, but I guess this is part of the longer step to get them under the D foundation brand.

Well, there is the matter of them wanting to keep this page up to date, which implies that this page gets tied in with *their* release process. I think they should take the initiative.

PR #1282, opened 2016-04-25
Better DDOC_BLANKLINE companion

andralex:
Reboot of https://github.com/dlang/dlang.org/pull/1186


CyberShadow:
I'm hazy on the details by now but it seems there are some open questions remaining on https://github.com/dlang/dmd/pull/5344.

PR #1284, opened 2016-04-28
Include current file name in "report a bug" template

CyberShadow:


wilzbach:
Nice idea!

CyberShadow:
The links will have `.dd` / `.d` extensions, unfortunately, so when interpreted literally will be 404s. There is a `DOCFILENAME` macro, but it's the absolute path on the file system the documentation is built on.

The title of the page is already included in the template.

It's a tradeoff between generating working links vs. links that are closer to the intended ones. Not sure if this is an improvement.

PR #1285, opened 2016-04-28
update template mixin and opAssign spec to reflect what compiler actually does

WalterBright:
Based on suggestions from https://issues.dlang.org/show_bug.cgi?id=15957

### DIFF
--- a/spec/struct.dd
+++ b/spec/struct.dd
@@ -141,9 +141,9 @@ struct S;
 union U;
 ---
 
-	$(P The members are completely hidden to the user, and so the only operations
-	on those types are ones that do not require any knowledge of the contents
-	of those types. For example:)
+        $(P The members are completely hidden to the user, and so the only operations
+        on those types are ones that do not require any knowledge of the contents
+        of those types. For example:)
 
 ---
 struct S;
@@ -152,8 +152,8 @@ S s;      // error, cannot initialize unknown contents
 S* p;     // ok, knowledge of members is not necessary
 ---
 
-	$(P They can be used to implement the
-	$(LINK2 https://en.wikipedia.org/wiki/Opaque_pointer, PIMPL idiom).)
+        $(P They can be used to implement the
+        $(LINK2 https://en.wikipedia.org/wiki/Opaque_pointer, PIMPL idiom).)
 
 
 $(H3 $(LNAME2 static_struct_init, Static Initialization of Structs))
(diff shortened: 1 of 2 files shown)


spec/struct.dd
```diff
@@ -521,8 +521,20 @@ $(SECTION3 $(LEGACY_LNAME2 AssignOverload, assign-overload, Identity Assignment
...
+	$(LI it has a $(LINK2 #struct-destructor, destructor))
+	$(LI it has a $(LINK2 #struct-postblit, postblit))
+	$(LI it has a field with an identity assignment overload)
+	)
+
+	$(P If an identity assignment overload is required and does not
+	exist, an identity assignment overload function of the type
+	$(CODE ref S opAssign(ref S))  will be automatically generated.)
+
+        $(P A user-defined one can implement the equivalent semantics, but can
```
mihails-strasuns-sociomantic: Indentation seems to be not in sync.

mihails-strasuns-sociomantic: Also "but can one must" doesn't seem to be a well-formed English to me :)

mihails-strasuns-sociomantic:
Thanks!

spec/struct.dd
```diff
@@ -141,9 +141,9 @@ struct S;
 union U;
 ---
 
-	$(P The members are completely hidden to the user, and so the only operations
-	on those types are ones that do not require any knowledge of the contents
-	of those types. For example:)
+        $(P The members are completely hidden to the user, and so the only operations
```
andralex: Are these hard tabs?

CyberShadow: No. Trivial test: if you can select individual spaces (here, on this webpage even), then they're not hard tabs.

spec/struct.dd
```diff
@@ -512,7 +512,7 @@ $(SECTION3 $(LEGACY_LNAME2 AssignOverload, assign-overload, Identity Assignment
         $(P where $(CODE opAssign) is a member function of S:)
 
         ---
-        ref S opAssign(S s)
+        ref S opAssign(ref S s)
```
9rnsr: This is incorrect change. The assigned lvalue needs to be copied before the assignment.

ZombineDev: @9rnsr I think you meant assignment from **rvalue**. In that case, is `ref S opAssign()(auto ref S s)` a more correct representation?

PR #1286, opened 2016-04-29
replace usage of warnings for deprecations

MartinNowak:


deprecate.dd
```diff
@@ -9,47 +9,45 @@ $(SPEC_S Deprecated Features,
...
     )
 
     $(DL
         $(DT Spec)
         $(DD Removal from the Specification)
-        $(DT Warn)
-        $(DD The compiler emits a warning if the feature is used)
         $(DT Dep)
-        $(DD The compiler issues an error if not compiled with the -d switch)
+        $(DD The compiler warns by default, issues an error with the -de switch, and can be silenced with the -d switch)
```
mathias-lang-sociomantic: Maybe we could use another adjective than "warns" so it doesn't get confused with warnings. I would just say "The compiler emits a 'Deprecation' message on usage by default, ...".

A bit OT:
I've always found DMD usage (`  -dw            show use of deprecated features as warnings (default)`) confusing, as I expect `dmd -dw -w` to have the same effect as `dmd -de` for deprecations.

mathias-lang-sociomantic:
Besides the wording suggestion, LGTM.

andralex:
guess I'll pull this, feel free to adjust the wording later

CyberShadow:
Wasn't this blocked by the dmd pull?

deprecate.dd
MartinNowak: Not sure what the use-case for `-dw` is. If you have a better wording, just make a PR.

PR #1288, opened 2016-05-02
add anchors to declarations

wilzbach:
I was missing anchors at the declarations for a long time. Turns out, while it's pretty difficult to do with DDoc, it's very easy to do so in JS :)

It will add an anchor at the top right.

image

### DIFF
--- a/js/listanchors.js
+++ b/js/listanchors.js
@@ -13,6 +13,23 @@ function lastName(a) {
     return a.slice(pos + 1);
 }
 
+// adds a anchor link to every documented declaration
+function addAnchors()
+{
+    var items = document.getElementsByClassName('d_decl');
+    if(!items) return;
+    for (var i = 0; i < items.length; i++)
+    {
+        // we link to the first children
+        var a = items[i].querySelector('a');
+        if(!a) continue;
+        var permLink = document.createElement("a");
+        permLink.setAttribute('href', '#' + a.name);
+        permLink.className = "fa fa-anchor decl_anchor";
+        items[i].insertBefore(permLink, items[i].firstChild);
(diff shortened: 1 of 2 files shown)


js/listanchors.js
```diff
@@ -13,6 +13,27 @@ function lastName(a) {
...
+        // we link to the first children
+        var a = items[i].querySelector('a');
+        if(!a) continue;
+        var permLink = document.createElement("a");
+        permLink.setAttribute('href', '#' + a.name);
+        permLink.className = "fa fa-anchor";
+        permLink.style.float = "right";
+        permLink.style.marginTop = "5px";
+        permLink.style.marginRight = "5px";
+        permLink.style.textDecoration = "none";
```
CyberShadow: Perhaps these should be in `style.css`

CyberShadow:
Nice.

What do you think of e.g. `.anchor { visibility: hidden; } dt.ddecl:hover .anchor { visibility: visible; }`

wilzbach:
> Perhaps these should be in style.css

Done.

> What do you think of e.g. .anchor { visibility: hidden; } dt.d_decl:hover .anchor { visibility: visible; }

Like it - done, but used `anchor_decl`.

JackStouffer:
Looks great; not to bikeshed to hard, but the symbol typically used for this is the ¶ symbol.

wilzbach:
> Looks great; not to bikeshed to hard, but the symbol typically used for this is the ¶ symbol.

I like the anchor, but happy to change. @CyberShadow what do you think?

CyberShadow:
I don't mind it either way.

PR #1289, opened 2016-05-03
Fix HTML link to "Nested Functions" section.

JohanEngelen:
This should fix the link from the contents mini-menu at the top.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -1558,7 +1558,7 @@ void main()
 }
 ---
 
-$(H3 $(LNAME2 variadicnested, Nested Functions))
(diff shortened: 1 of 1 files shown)


CyberShadow:
Why change the anchor instead of fixing the link?

JohanEngelen:
Because the anchor doesn't make sense? :)

schuetzm:
It seem the anchor was inadvertantly changed by c&p:
https://github.com/dlang/dlang.org/commit/ed19b56382269fd6698e42c9efb9886da34611ad#diff-d276404d05b7cf227be9ff33a44e73a2L1218

PR #1290, opened 2016-05-03
Fix erroneous function spec statement

JackStouffer:
AFAIK this has never been true. And for why it shouldn't ever become true, see the discussions on the mentioned bugs.


Hackerpilot:
:white_check_mark:

JackStouffer:
Thanks!

PR #1291, opened 2016-05-04
Fix runable examples when running dlang via https

nazriel:
Currently if you try to evaluate code examples on the https://dlang.org the xhr request goes via plain http to dpaste - causing fails on some browsers like safari

DPaste supports https.

### DIFF
--- a/js/run.js
+++ b/js/run.js
@@ -491,7 +491,7 @@ $(document).ready(function()
            
             $.ajax({
                 type: 'POST',
-                url: "http://dpaste.dzfl.pl/request/",
+                url: "//dpaste.dzfl.pl/request/",
                 dataType: "json",
                 data: 
                 {


wilzbach:
On a related matter, how about setting dlang.org by default to https? 
Google actually includes this in the page ranking...

CyberShadow:
We couldn't do this a while ago because it broke the widgets on the front page, among other things. It might be fine now.

CyberShadow:
https://issues.dlang.org/show_bug.cgi?id=15378

Looks like this one was the last blocker?

CyberShadow:
Deployed. Works on https://dlang.org/ (needs a full refresh)

wilzbach:
Now we only need to fix the performance of dpaste - maybe we can add some simple caching?
Have a look at the waiting time:

image

wilzbach:
btw the [GoLang tour](https://tour.golang.org/welcome/4) apparently also uses a cache - it makes it in 600ms (with network latency).

> In the playground the time begins at 2009-11-10 23:00:00 UTC (determining the significance of this date is an exercise for the reader). This makes it easier to cache programs by giving them deterministic output.

nazriel:
@wilzbach my hosting provider screw up something and dpaste itself works horribly slow.
Hopefully I can catch-up with @CyberShadow on dconf and we can discuss full migration of dpaste to his infrastructure (if he agrees in the first place of course :))

@CyberShadow thanks for marging.

PR #1296, opened 2016-05-11
Add note to range based for about implicit type attribute.

GilesBathgate:
Coming from a background in C++ its natural to expect that the foreach loop within D should take an auto specifier. Because D doesn't need these specifiers a simple note in the documentation can help people understand the difference.

### DIFF
--- a/spec/statement.dd
+++ b/spec/statement.dd
@@ -491,6 +491,21 @@ foreach (int i, char c; a)
         order.
         )
 
+        $(P $(B Note:) The $(I ForeachTypeAttribute) is implicit, and when a
+        type is not specified, it is inferred. In that case, $(D auto) is
+        implied, and it is not necessary (and actually forbidden) to use it.
+        )
+
+--------------
+int[] arr;
+...
+foreach (n; arr) // ok, n is an int
+    writeln(n);
+
+foreach (auto n; arr) // error, auto is redundant
+    writeln(n);
+--------------
+
 $(H4 Foreach over Arrays of Characters)
 
         $(P If the aggregate expression is a static or dynamic array of


spec/statement.dd
```diff
@@ -459,6 +459,32 @@ $(P
...
+
+$(P
+        D is slightly different in that the $(I ForeachTypeList) does not need
+        to specify auto. The following code would be equivalent.
+
+------
+
+foreach(i; v) // The type of i is int
+    write(i,' ');
+writeln();
```
wilzbach: with UCFS you don't need parentheses ;-)

spec/statement.dd
```diff
@@ -459,6 +459,32 @@ $(P
...
+
+$(CPPCODE
+for(auto i: v) // The type of i is int
+    std::cout << i << ' ';
+std::cout << '\n';
+)
+
+$(P
+        D is slightly different in that the $(I ForeachTypeList) does not need
+        to specify auto. The following code would be equivalent.
```
wilzbach: you need to close P

wilzbach:
Thanks for helping to make the docs better :)

I personally think that we should just add an example without type and not the C++ example.
The page is already pretty full and the main problem here is that isn't not clear to the reader that `foreach` uses auto by default. Moreover imho we should be careful with providing such language comparisons in the standard documentation, because there's a "similar" foreach in Java, Python, ...

GilesBathgate:
@wilzbach Yes I agree, perhaps this would be more useful in the ```cpptod.dd``` This was really just to open up a discussion as this was a sticking point when I first started trying to use foreach.

JackStouffer:
An example isn't something that belongs in the language specification.

CyberShadow:
Yep, it's better suited for http://dlang.org/cpptod.html

GilesBathgate:
@CyberShadow @JackStouffer What can we put in the spec that will make this easy to grasp. I wouldn't have thought to look in ```cpptod.dd```, really I just want to say in the spec in big bold letters __"look ma no auto"__

wilzbach:
Yep I agree that's one of the cool features :)
How about changing the foreach examples and removing the type but of one? Imho I prefer the way to say we have auto, but you can still explicitly state the type. Maybe it's even enough to put that in the text?

GilesBathgate:
@wilzbach No you can't put auto specifier in D's foreach, i.e this does not compile:

```d
foreach(auto a; v)
   writeln(a);
```

However this is fine:

```d
foreach(immutable a; v)
   writeln(a);
```

JackStouffer:
> What can we put in the spec that will make this easy to grasp

As I said, this really doesn't belong in the spec. The spec is a very specific document for a specific purpose, and that doesn't include comparing D to other languages.

Plus, the spec already states that `auto` isn't allowed:

```
ForeachTypeAttribute:
    ref
    TypeCtor
```

This is the purview of something like a tutorial.

GilesBathgate:
@JackStouffer The language spec is full of examples in D, regarding what is allowed and not allowed. While I appreciate that the context free grammar is clearly there, its not easy to read and can easily be missed. I don't see any harm in having:

```d
foreach(auto a; v)   //Not OK
   writeln(a);
```

As these __not ok__ examples are all over the spec elsewhere.

CyberShadow:
I suggest shortening this patch to a simple mention along the lines of "Note: the variable declaration is implicit, and when a type is not specified, it is inferred. In that case, `auto` is implied, and it is not necessary (and actually forbidden) to use it." OSLT

wilzbach:
> No you can't put auto specifier in D's foreach, i.e this does not compile:

@GilesBathgate I meant auto type inferring.

> I suggest shortening this patch to a simple mention along the lines

Great idea and as said maybe even removing the type for one of the  `foreach` examples, s.t. it gets more obvious ;-)

GilesBathgate:
@CyberShadow I have updated the documentation with a note as you suggested, and added D examples which illustrate what is meant.

spec/statement.dd
```diff
@@ -491,6 +491,19 @@ foreach (int i, char c; a)
...
+        )
+
+--------------
+char[] a;
+...
+foreach(c; a) // ok the type of c is char
+    write(c,' ');
+
+foreach(auto c; a) // error
+    write(c,' ');
```
wilzbach: I am pretty opinionated, but I would remove the "error" example - the other one makes it already clear.

GilesBathgate: In my opinion I think this is worthwhile. For example if I was searching the document for an example I  might search for "foreach(auto" and this would show up. It could also show up in Google searches. ;)

spec/statement.dd
```diff
@@ -491,6 +491,19 @@ foreach (int i, char c; a)
         order.
         )
 
+        $(P $(B Note:) The $(I ForeachTypeAttribute) is implicit, and when a type is not specified, it is inferred. In that case, $(D auto) is implied, and it is not necessary (and actually forbidden) to use it.
```
CyberShadow: Line length

spec/statement.dd
```diff
@@ -491,6 +491,19 @@ foreach (int i, char c; a)
...
+        )
+
+--------------
+char[] a;
+...
+foreach(c; a) // ok the type of c is char
```
CyberShadow: Phobos style please (space between `foreach` and `(`)

spec/statement.dd
```diff
@@ -491,6 +491,19 @@ foreach (int i, char c; a)
...
+    write(c,' ');
```
CyberShadow: Ditto, space after comma

CyberShadow:
I suggest these changes to your example code:

```
int[] arr;

foreach (n; arr) // OK (c is an int)
    writeln(n);

foreach (auto n; arr) // NG, auto is redundant
    writeln(n);
```

* Use consistent wording for broken examples
* Capitalization
* Phobos style
* Iterate over an `int` array, to avoid the distraction of the iteration unit of strings (see dchar over string iteration)
* Use `writeln` for less syntactic noise

GilesBathgate:
@CyberShadow Agreed and amended, except ```ok``` is not  capitalised elsewhere in the spec, and there is no __NG__, only ```error``` or ```illegal```

CyberShadow:
https://www.google.com/search?q=ng+ok++site%3Adlang.org%2Fspec

GilesBathgate:
@CyberShadow Well what's the right way then sometimes is ```ok``` others ```OK``` sometimes ```NG``` or ```illegal``` or ```error```?!  (I presume NG means no good)

CyberShadow:
If it's not consistent, then I guess it doesn't really matter :)

GilesBathgate:
@CyberShadow  Oh its already merged....Well it was in lower case. for anyone who cares they can amend it.

PR #1297, opened 2016-05-13
Edit & run examples from unittests

wilzbach:
Hey all,

I wanted to be able to execute the examples directly from the documentation for a long time. It turns out, it's not that hard ;-)

The current module and `write,writeln` are imported. It might be that a couple of tests depend on the private imports in a module, but this is discouraged anyways as the public documented unittests should really only need the current module imported. I tested it randomly with a couple of modules and so far everything seems to work well.

This feature is experimental and I would love feedback on it. 
As dpaste requires CORS and the origin must be dlang.org, you have to trick a bit to test it.
If you happen to use pdns, it as easy as `pdnsd-ctl add a 127.0.0.1 dlang.org` though.

Therefore here are some images as impressions:

image

image

image

writeln works too:

image

@CyberShadow: maybe we can even send the D version to DPaste so we never run into troubles of library mismatches?

Edit: maybe @CyberShadow can add DAutoTest to the origin whitelist of DPaste?


css/style.css
```diff
@@ -2018,3 +2018,8 @@ the home page, intro pitch and your-code-here layed out vertically */
 dt.d_decl:hover .decl_anchor {
     visibility: visible;
 }
+
+/* Style for the example run buttons on the Phobos library documentation */
+.d_example_buttons {
+    text-align: right;
+}
```
JackStouffer: new line here please

js/run.js
```diff
@@ -254,7 +254,7 @@ var MD5 = function (string) {
     return temp.toLowerCase();
 }
 
-String.prototype.nl2br = function()
+var nl2br = function()
```
JackStouffer: Ah, PHP style naming, my favorite.

Maybe we should start enforcing Phobos style on the site JS.

CyberShadow:
Why is there so much new JS code? Can't the front page code be reused?

CyberShadow:
Should this not be enabled only for Phobos pages?

CyberShadow:
As I mentioned at DConf, this should only be enabled for documented unittests (runnable examples). This may need a small compiler patch to allow distinguishing the two. A lot of the non-unittest examples are nonsensical out of context (see e.g. std.file.timeLastModified).

js/run_examples.js
```diff
@@ -0,0 +1,159 @@
+$(document).ready(function()
+{
+    if ($('body')[0].id == "Home")
```
JackStouffer: `document.getElementsByTagName('body')[0].id == "Home"`

I'm all for jQuery for when it abstracts differences between browsers, but for the simple stuff let's stick with plain JS.

CyberShadow: If jQuery is already in use then there is no reason not to use all of it.

JackStouffer:
> If jQuery is already in use then there is no reason not to use all of it.

@CyberShadow We really shouldn't. I know I added some code to the site that used JQuery, but I recently changed my mind on this.

Just run this code in your console on the dlang.org home and look at the performance:

```
function test1() {
    return document.getElementsByTagName('body')[0].id == "Home";
}
function test2() {
    return $('body').id == "Home";
}
```

My results
```
> console.time("test"); for (var i = 0; i < 100000; i++) { test1(); } console.timeEnd("test");
test: 8928.709ms
> console.time("test"); for (var i = 0; i < 100000; i++) { test2(); } console.timeEnd("test");
test: 26968.555ms
```

For a few extra keystrokes you suddenly get 3x faster.

CyberShadow:
Last I checked we had no pages with hundreds of thousands of runnable examples on them.

Profile **then** optimize. What you are suggesting is premature optimization.

js/run_examples.js
```diff
@@ -0,0 +1,159 @@
...
+        return;
+
+    // only for std at the moment
+    if(!$('body').hasClass("std"))
+        return;
```
wilzbach: This enables the tests only on phobos

adamdruppe:
On Fri, May 13, 2016 at 07:53:06AM -0700, Vladimir Panteleev wrote:
> Profile **then** optimize. What you are suggesting is premature optimization.

If it is trivial to use the better way, you might as well just do it
right the first time....

js/run_examples.js
CyberShadow: Would be nice to avoid including the JS file at all on other pages, so there's no useless impact on load speed.

JackStouffer:
> Last I checked we had no pages with hundreds of thousands of runnable examples on them.

No, we don't. But that doesn't mean that we should be consciously using code that is slower in order to save like, ten characters.

> Profile then optimize. What you are suggesting is premature optimization.

These things add up.

CyberShadow:
I'm going to have to ask you to kindly stop it with the bullshit.

1. The particular line of code you have clinged yourself to criticizing runs **once** per page load.

2. The code is pre-existing, the line has been simply copied from existing code in `run.js`.

If you want to make a positive contribution, you could try rewriting `run.js` to not require jQuery. Show some readable, maintainable code as the result; show a substantial, repeatable and practically significant improvement in performance, and we can discuss merging that.

Otherwise, I suggest you refrain from such future criticism, because the only impact you're making is aggravating contributors who are actually trying to make a positive change.

JackStouffer:
Dude,

`$('body')[0].id == "Home"` to `document.getElementsByTagName('body')[0].id == "Home"`

`$('pre[class~=d_code]')` to `document.querySelectorAll('pre[class~=d_code]')`

I'm not asking for much here. I'm asking that thirty seconds be taken to use the faster version. And if you disagree with me, you're the one with merge rights at the end of the day.

CyberShadow:
I see my point did not get through.

It is not at all faster, in this case. The only effect such a change would have is make the code longer and more inconsistent.

wilzbach:
> As I mentioned at DConf, this should only be enabled for documented unittests (runnable examples). This may need a small compiler patch to allow distinguishing the two. A lot of the non-unittest examples are nonsensical out of context (see e.g. std.file.timeLastModified).

That would be great to have. Can we just have two separate css classes maybe?

schuetzm:
I agree with CyberShadow. Getting rid of jQuery altogether can be worthwhile, but only because it would be one dependency fewer, not because `$()` adds a quarter of a millisecond to the page load time. Especially if we already have existing code that does what you want, just reuse that instead of adding another copy and letting the copies diverge.

js/run_examples.js
```diff
@@ -0,0 +1,159 @@
...
+                    + '<div class="d_code">'
+                        + '<pre class="d_code">'+orig+'</pre>'
+                    + '</div>'
+                    + '<div class="d_run_code">'
+                        + '<textarea class="d_code" style="display: none;"></textarea>'
+                        + '<div class="d_code_output"><span class="d_code_title">Application output</span><br><textarea class="d_code_output" readonly>Running...</textarea>'
+                    + '</div>'
+                + '</div>'
+        );
+        console.log(orig);
```
nazriel: Won't it break on IE?
IIRC IE doesn't like console. too much ;)

js/run_examples.js
```diff
@@ -0,0 +1,159 @@
...
+            hideAllWindows();
+            output.css('height', height(31));
+            outputDiv.css('display', 'block');
+            outputTitle.text("Application output");
+            output.html("Running...");
+            output.focus();
+
+            var currentPackage = $('body')[0].id;
+            var codeOut = "void main(){ ";
+            codeOut += "import " + currentPackage + "; ";
```
nazriel: Clever! 

I had a spin with this some time ago.

From what I recall some modules, like std.base64 depend on 3rd party modules like std.stdio and of course do not include the imports in their unittests/examples. This is something we should also consider hm

nazriel:
> @CyberShadow: maybe we can even send the D version to DPaste so we never run into troubles of library mismatches?

What do you mean by this?

wilzbach:
> What do you mean by this?

s.t. like

```
compiler: {
  name: "dmd",
  version: "2.071"
}
```

js/run_examples.js
wilzbach: > Won't it break on IE?
> IIRC IE doesn't like console. too much ;)

Oh that was just for testing - will be removed soon. Sorry about that.

nazriel:
Hm - dpaste uses the newest stable DMD for any request from dlang.org

How would you annotate specific unittest/example to run with different compiler version? 
(actually I am curious - you have some clever ideas - maybe we could get rid of this MD5 madness on main page as well).

wilzbach:
> Hm - dpaste uses the newest stable DMD for any request from dlang.org

Oh I didn't know that. That's great!

> How would you annotate specific unittest/example to run with different compiler version? 
(actually I am curious - you have some clever ideas - maybe we could get rid of this MD5 madness on main page as well).

Have you tried user-defined attributes.
This is probably not useful for Phobos documentation, but feel free to ping me at your Dpaste repo, maybe we can continue this discussion there?

CyberShadow:
> How would you annotate specific unittest/example to run with different compiler version?

That is not right. If it's a Phobos function, then use the version from which that code comes from, no?

nazriel:
> That is not right. If it's a Phobos function, then use the version from which that code comes from, no?

Yeah - as I said dpaste uses newest-dmd available for dlang.org's examples. 

Same goes for phobos. The only thing worth changing probably would be phobos' prerelease docs - there we probably should use nightly builds of DMD @ Dpaste.

I was asking mostly out of curiosity - I though that maybe @wilzbach has got an idea how we could annotate code examples somehow - so for example we could get rid of MD5 sums co-related with examples on the main page.

I had an idea that maybe we could use comments in code examples.
Like:

```d
// @input: Mercury\nVenus\nEarth\nMars\nJupiter\nSaturn\nUranus\nNeptune
// @args: foo bar if needed
// Sort lines
import std.stdio, std.array, std.algorithm;

void main()
{
    stdin
        .byLineCopy
        .array
        .sort!((a, b) => a > b) // descending order
        .each!writeln;
}
```

Anyways - sorry for the off-topic.

Back to the case:
In overall this Pull Request is something that some time ago we were talking about with Andrei.
He wanted to have all examples in phobos documentation runnable.
Back in the day examples were heavily dependant on external modules being imported (and of course there were no import statements in examples), external input from stdin or arguments and so on.

From what I can see now mostly there are self-contained assertions and so on and the missing bits like std.base64 should be just corrected to either provide imports, stdin etc or re-arrange example entirely.

MartinNowak:
Great!
I think the biggest problem is that unittest examples don't currently produce output, so running them is sort of boring.

wilzbach:
Okay I found a solution to distinguish between text examples and ddoced unittest without patching the compiler :)
Turns out that `DDOC_EXAMPLES` only used for ddoced unittests, how the in-text examples get this header is deep ddoc magic.

> I think the biggest problem is that unittest examples don't currently produce output, so running them is sort of boring.

I added a simple regex that replaces all basic `assert` with writeln's - is that's more interesting?

> I had an idea that maybe we could use comments in code examples.

Yep that would be pretty cool. it is not off-topic, it would actually be needed for most of `std.stdio`.

> Back in the day examples were heavily dependant on external modules being imported (and of course there were no import statements in examples), external input from stdin or arguments and so on.

We could just run all ddoced unittest in Phobos with this simple wrapper and see which fail and fix those. How does this sound? Has anyone a better idea to fix the missing external imports?

andralex:
Great work. I think the right approach here is to just add the missing imports. Hopefully we can crowdsource that a bit - or at least gangsource :o). What percentage of examples don't currently work?

I wouldn't want to launch this feature with many failures, it could cause more harm than good. We can only launch once, so let's make it count. @wilzbach please advise

wilzbach:
>  What percentage of examples don't currently work?

I don't know, but I will write a small script to test.

> I wouldn't want to launch this feature with many failures, it could cause more harm than good. We can only launch once, so let's make it count

I agree.

wilzbach:
> What percentage of examples don't currently work?

You get 1100 errors for 3849 unittests (29%) - as error don't come alone the number of not working examples might be a bit lower, but still very high.

- [Entire log output](http://sprunge.us/eYBO).
- [Script](https://github.com/wilzbach/d-scripts/blob/master/check_phobos_tests.d) (needs the newest dub)

It uses libdparse to separate out all unittests. For each tests it adds the import to the current module and then runs the tests with rdmd.

> Hopefully we can crowdsource that a bit - or at least gangsource

Or has someone an idea how we could automate adding the imports? Ping @joakim-noah

wilzbach:
Edit: I forgot to filter only for ddoced unittests, so forgot the numbers from above.

Tests: 1069
Errors : 232 (21%)
Deprecation: 56

This [output](http://sprunge.us/Xggf) looks a lot more reasonable to fix :)

joakim-noah:
I use dmd to dump all symbols from working code, ie imports manually added, whereas this is searching for what imports to add.  Maybe Dscanner's symbol finder can be used for this.  An alternative would be to import all Phobos modules, then use dmd to dump which module was actually used, though you may have symbol collisions that way.

wilzbach:
> Why is there so much new JS code? Can't the front page code be reused?

Done.

> I wouldn't want to launch this feature with many failures, it could cause more harm than good.

It's now only enabled on the `prerelease` pages as I assume that most users won't check those pages and we have a safe playground to test this feature?
The next plan would be to enable it module per module, e.g. for `std.algorithm` I submitted #4431.

andralex:
@wilzbach any plan of action on this? Let me know if I can help.

wilzbach:
> any plan of action on this? Let me know if I can help.

See above:

It's now only enabled on the prerelease pages as I assume that most users won't check those pages and we have a safe playground to test this feature?
The next plan would be to enable it module per module, e.g. for `std.algorithm` I submitted https://github.com/dlang/phobos/pull/4431
... and eventually we should add more automation (see the other PR for details).

css/style.css
wilzbach: may I ask for the motivation?

wilzbach:
> It's now only enabled on the prerelease pages as I assume that most users won't check those pages and we have a safe playground to test this feature?

Would this strategy work? (I am bit spare on time atm)
At least it should be easier for others to contribute ;-)

wilzbach:
Okay I managed to find some time to revive this PR. As discussed above to ensure that we won't introduce regressions and verify the runnability of existing snippets, we do need to run them :)
I moved the text extraction script to dlang/tools (https://github.com/dlang/tools/pull/203) and opened a PR at Phobos https://github.com/dlang/phobos/pull/4943 which will enable CI checking for the public unittests.

wilzbach:
Okay with https://github.com/dlang/phobos/issues/4943 I this is finally in a usuable state :)
Two remarks:

1) It's only enabled for a subset of modules (= those that are automatically checked at Phobos)
2) It's only enabled for the pre-release pages for now as the missing import changes need to propagate to the next release. Imho the pre-release pages don't have many visitors, so we should have a safe playground to test this

andralex:
Great. About 1 - where is that list of modules for which this is enabled? And what to do to expand it?

wilzbach:
> Great. About 1 - where is that list of modules for which this is enabled?

There's a hard-coded exclusion list:

https://github.com/dlang/phobos/blob/master/posix.mak#L523

This PR contains a copy of it

https://github.com/dlang/dlang.org/pull/1297/files#diff-034369de775110cb179ed1e925b62a0fR53

>  And what to do to expand it?

Simple: remove a module from the exclusion list, have a look at the missing imports, make a PR.
It shouldn't take that long - maybe I can fit it into it into this weekend ;-)

wilzbach:
> Simple: remove a module from the exclusion list, have a look at the missing imports, make a PR.
It shouldn't take that long - maybe I can fit it into it into this weekend ;-)

e.g. https://github.com/dlang/phobos/pull/4966

@andralex Unfortunately I can't merge stuff here. Is anything blocking this experiment?

andralex:
cool cool cool thanks @wilzbach

CyberShadow:
> FWIW @CyberShadow I saw a confusing message in the log of your doc builder:

Yep, those are normal. Digger checks whether a cache entry exists by asking git to describe it, and git prints that to stderr when it doesn't (i.e. on a cache miss).

andralex:
@wilzbach saw the list thx, if you work on it great, if you stop at a point please put the balance of it in bugzilla. Thanks!

wilzbach:
> cool cool cool

https://dlang.org/phobos-prerelease/std_algorithm_iteration.html#cache

Yeah it's online :)

> if you stop at a point please put the balance of it in bugzilla. Thanks!

https://issues.dlang.org/show_bug.cgi?id=16984 (extend list of runnable modules on dlang.org)
https://issues.dlang.org/show_bug.cgi?id=16985 (enabling the unittests after 2.073)

andralex:
Cool! Just tried one, it works!! How about enabling everything for ddox e.g. https://dlang.org/library-prerelease/std/algorithm/iteration/cache.html, is this easy? cc @s-ludwig

wilzbach:
>  How about enabling everything for ddox e.g. https://dlang.org/library-prerelease/std/algorithm/iteration/cache.html, is this easy? cc @s-ludwig

Started a discussion:
https://github.com/rejectedsoftware/ddox/issues/137

wilzbach:
> Hm - dpaste uses the newest stable DMD for any request from dlang.org

@nazriel do you also maintain a copy of the latest nightly release?
This would allow to ensure that the examples on the `prerelease` pages are really runnable. Some of them are not as of now :/
Any way I can help you with this?

PR #1299, opened 2016-05-13
Update faq.dd

jmh530:
Fixing grammatical error. publically -> publicly


CyberShadow:
Not an error? http://www.thefreedictionary.com/publically

jmh530:
@CyberShadow Note that your link says that it is non-standard English. Perhaps I was hasty in calling it an error, but as a native English speaker I would consider it a typo given how rare publically is. I did some research and publicly is not only the more common spelling, but it is the older spelling as well, dating back to the 1500s. 

Moreover, the FAQ also uses publicly elsewhere. Even if it is not an error in the most grammatical sense, it is inconsistent. I consider that sufficient grounds for changing it.

PR #1301, opened 2016-05-15
Give some examples of " used in backquoted strings.

GilesBathgate:
This just extends the existing examples to show how " can be used in backquoted strings.

### DIFF
--- a/spec/lex.dd
+++ b/spec/lex.dd
@@ -397,8 +397,9 @@ r"ab\n" // string is 4 characters,
 ---------------
 `hello`
 `c:\root\foo.exe`
-`ab\n`  // string is 4 characters,
-        // 'a', 'b', '\', 'n'
+`The "lazy" dog`
+`a"b\n`  // string is 5 characters,
+        // 'a', '"', 'b', '\', 'n'
 ---------------
 
 $(H4 Double Quoted Strings)


spec/lex.dd
```diff
@@ -397,8 +397,9 @@ r"ab\n" // string is 4 characters,
 ---------------
 `hello`
 `c:\root\foo.exe`
-`ab\n`  // string is 4 characters,
-        // 'a', 'b', '\', 'n'
+`The "lazy" dog`
+`a"b\n`  // string is 5 characters,
+        // 'a', '"', 'b', '\', 'n'
 ---------------
```
schuetzm: Please change `'\'` to `'\\'`.

GilesBathgate: Then it would be incorrect!

```d
import std.stdio;

void main()
{
  writeln(`ab\n`);
  writeln(`ab\\n`);
}
```

```bash
giles@tikka:~$ ./test
ab\n
ab\\n
```

GilesBathgate: Or if you mean in the comment, then it would be inconsistent with the previous example:

https://github.com/GilesBathgate/dlang.org/blob/7fc048efeb02a282a1da280d818416c645ae2e0b/spec/lex.dd#L386

schuetzm: Yes, I mean in the comment. Didn't see the other examples. I guess we can interpret the `''` here as orthographical quotation marks, not character literals, so it can stay as it is. Thanks!

PR #1303, opened 2016-05-17
remove menu link to review queue and digital mars

wilzbach:
The digital mars page was last updated on "Tue May 28 11:53:30 2013" and I hope we can do better.
The review queue is also usually hopelessly outdated and unmaintained. Imho it's besser if we just remove those two links and let the reader explore the wiki.


schuetzm:
I agree. In the unlikely event that these pages get up-to-date content, we can easily reinstantiate the links. I'm going to wait a bit for others to comment, though.

JackStouffer:
👍

CyberShadow:
Yes for digitalmars link.

The review queue is already linked to from the wiki front page, so OK.

PR #1306, opened 2016-05-19
Update obsolete varargs example that uses `__va_argsave`

yebblies:
The portable syntax has been supported on the main platforms since 2.067 or earlier.


yebblies:
Go go go?!

CyberShadow:
OK :)

yebblies:
Thanks!

PR #1307, opened 2016-05-22
show list of contributors per module

wilzbach:
This adds a list of all contributors to a module at the end of the documentation.

image

The [backend](https://github.com/wilzbach/phobos-contribs) queries the Github API for commits to that file / folder and reduces this to Github user + their avatars. Every query is cached for 24h.
That being said yes - the backend could maintain [it's own git repo](https://github.com/wilzbach/phobos-contribs/issues/1) and map the commits to Github avatars, but so far this approach is also working.

The contributor avatars are sorted alphabetically and only enabled for "std" modules.

### DIFF
--- a/js/show_contributors.js
+++ b/js/show_contributors.js
@@ -1,0 +1,45 @@
+/**
+Shows a list of the contributors to a Phobos module
+It queries the backend contribs.dlang.io
+Backend source: https://github.com/wilzbach/phobos-contribs
+
+License:   http://boost.org/LICENSE_1_0.txt, Boost License 1.0
+
+Authors:   Sebastian Wilzbach
+*/
+
+// find out whether the current page is a module or package
+function isPackage()
+{
+    return $('.tip.smallprint a[href*="package.d"]').length > 0;
+}
+
+$(document).ready(function()
+{
+    // only for std at the moment
+    if (!$('body').hasClass("std"))
+        return;
+
+    // for index modules (package.d), we want to display all contributors of the package
+    var modulePath = document.body.id.replace(/[.]/g, '/');
+    if (!isPackage())
+       modulePath += '.d';
+
+    $.getJSON( "https://contribs.dlang.io/contributors/file/dlang/phobos?file=" + modulePath, function(contributors) {
+
+        var posToInsert = $('#copyright');
+        var contentNode = $("<div id='contributors-github'></div>");
+
+        var totalContributors = contributors.length;
+        contentNode.append("<h3>" + totalContributors + " Contributors</h3>");
+
+        // list contributors with github avatar
+        $.each(contributors, function(i, contributor){
+            var contributorDiv = '<a href=' + contributor.html_url +' target="_blank">';
+            contributorDiv += '<img src="'+ contributor.avatar_url +'&size=40" height="40" width=40" alt="' + contributor.login + '"/>';
+            contributorDiv += "</a>";
+            contentNode.append(contributorDiv);
+        });
+        contentNode.insertBefore(posToInsert);
+    });
+});
--- a/css/style.css
+++ b/css/style.css
@@ -1463,7 +1463,7 @@ body#Menu #content li.expand-container ul
 
 div#copyright
 {
-    margin-top: 5em;
+    margin-top: 1.5em;
     text-align: center;
 }
 
@@ -1478,6 +1478,10 @@ div#copyright
     color: #555;
 }
 
+#contributors-github img
+{
+    padding-right: 5px;
+}
 
 /* These are for the changelogs */
 div.version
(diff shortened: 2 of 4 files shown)


js/show_contributors.js
```diff
@@ -0,0 +1,50 @@
...
+License:   http://boost.org/LICENSE_1_0.txt, Boost License 1.0
+
+Authors:   Sebastian Wilzbach
+*/
+
+// find out whether the current page is a module or package
+function isPackage()
+{
+    var ret = false;
+    $('.tip.smallprint').find('a').each(function(i, el) {
```
JackStouffer: I think

```
return $('.tip.smallprint a[href*="package.d"]').length > 0 ? true : false;
```

is slightly faster and cleaner to boot.

js/show_contributors.js
```diff
@@ -0,0 +1,50 @@
...
+        if (el.href.indexOf("package.d") >= 0)
+            ret = true;
+    });
+    return ret;
+}
+
+$(document).ready(function()
```
JackStouffer: No need to be explicit here, 

```
$(function () {
    ...
})
```

is the same thing.

JackStouffer:
This is a cool idea. LGTM sans comments.

CyberShadow:
> This is a cool idea.

I agree.

Some thoughts:

- It should not be present in the offline versions of the documentation.

- Is there any way we could get this information during build time instead, i.e. extract it directly from the git history?

- There is an "Authors" section at the top of each module, perhaps it should be integrated there somehow.

- I'm not sure about loading all those images for every page load. It's one more website that can track users if they wanted to, and it will affect total page load time.

wilzbach:
> It should not be present in the offline versions of the documentation.

The request will error and we can silently catch that. If we start to split into an offline version, this means yet another build to maintain, but I am more than open to suggestions. 

> Is there any way we could get this information during build time instead, i.e. extract it directly from the git history?

git log <path> and then applying some reductions. The two big problems are
1) to get a mapping between email address and github username/avatar (this requires the github api, an Internet connection and is rate-limited in the anonymous form) 
2) Getting it as part of ddoc as this needs to be generated for every file 

> There is an "Authors" section at the top of each module, perhaps it should be integrated there somehow.

I agree, my main reason to list the contributors was that the author seems to be inadequate for many modules. 
Btw the authors section is optional and afaik some modules don't have it. 

> I'm not sure about loading all those images for every page load. It's one more website that can track users if they wanted to, and it will affect total page load time.

It's async and after page load, so it won't affect the page load. However I agree that we should keep the dependencies low (it's actually two websites contribs.dlang.io and github.com). 
On mobile one might be interested in saving bytes, thus loading only the first row might be an option there. 
As said before, I couldn't find a way to integrate this with Ddoc, but this doesn't mean we couldn't download all needed avatars. However then we would need to have this mapping file  somewhere.

js/show_contributors.js
wilzbach: that's nice. thanks!

js/show_contributors.js
wilzbach: I think it's good to be explicit here as people not so familiar with jQuery can immediately grasp that this callback waits for the page to be ready - other dlang javascripts also use this style (e.g. run.js)

WalterBright:
I think it's a great idea, but am concerned about bandwidth consumption, and of course loading and dependency issues. One nice thing about dlang pages is they are tight and fast. A compromise: provide a `contributors` clickable thing that invokes a bit of javascript that then goes and loads the avatars and inserts them into the web page?

Another idea just for fun is to sort the avatars by number of PRs. Please, not by lines of code :-)

jacob-carlborg:
Please no. There's enough information in the docs already.

ZombineDev:
I also think that it's a great idea. Another option is to move this list to the bottom of the page. That way it shouldn't bother people that much (I personally won't mind).

wilzbach:
> Another option is to move this list to the bottom of the page

It's already at the bottom ;-)

> I think it's a great idea, but am concerned about bandwidth consumption, and of course loading and dependency issues.

It's async and the request and loading happens after the page was loaded & rendered. The only concern I have is that on mobile the avatar's take some traffic. However they between 1-5kb. That's about 100kb in total.
Additionally this is shipped compressed and will be heavily cached.(Github shows the avatar everywhere).
If you compare this with codemirror (150kb) which is loaded on every page without being needed (yet, which will hopefully change soon - see #1297).

> A compromise: provide a contributors clickable thing that invokes a bit of javascript that then goes and loads the avatars and inserts them into the web page?

Hmm I actually think the number of contributor for a module is something we can be proud of, so I would at least display this. I don't mind about the avatars being optional - my point was just that we should start to give credits to people and this is something moderately easy & free. Also for most modules just stating the "original author" is probably not the best way.

ZombineDev:
Great, let's move forward with this!

wilzbach:
> Great, let's move forward with this!

So what's the consenus? @WalterBright @andralex

wilzbach:
So how do we move forward?

wilzbach:
Idea: we remove the unused comments section (aka "feature") at the documentation and replace it with the contributors list?

andralex:
Let's try it out!

PR #1309, opened 2016-05-26
make the how-to-promote guide a bit more modern

wilzbach:


howto-promote.dd
```diff
@@ -94,4 +91,3 @@ $(H3 Promoting the Project)
 Macros:
 	TITLE=Promoting D Projects (or Internet Marketing 101)
 	WIKI=PromotingDProjects
```
wilzbach: for what is this macro used?

CyberShadow: It's no longer used. Previously each page had a "page wiki" link, which lead to an associated page on the wiki, where users could add notes.

wilzbach:
Rebased after the wiki changes. In my second pass I also took the courage to modify the list of old links and websites.

CyberShadow:
There's some DDoc peeking out:

```
 $(D_S Promoting D Projects (or Internet Marketing 101), 
```

wilzbach:
> There's some DDoc peeking out:

It was a missing parenthesis somewhere :D

wilzbach:
> I didn't even know this page existed, but it's certainly an improvement. ;)

I also found it by accident, but the plan is to show a link to this when one creates a new announce thread.
See also https://github.com/CyberShadow/DFeed/issues/69

PR #1310, opened 2016-05-26
remove deprecated WIKI macro

wilzbach:
If this wiki macro is no longer used, let's remove it!

```
sed '/^\s*WIKI=/d' -i **/*.dd
```

dlang.org.ddoc was manually modified.


aG0aep6G:
Deletions looks good to me, but I don't know how dead WIKI is exactly.

CyberShadow:
Yep, it's dead.

These tags are back from the days when the official D wiki was http://www.prowiki.org/wiki4d/wiki.cgi.

aG0aep6G:
@wilzbach, now do phobos and druntime :)

PR #1311, opened 2016-05-26
add D Language Foundation page

wilzbach:
AFAIK it's already registered and official - e.g. the D foundation is hosting this year's GSoC.

Unfortunately I don't know much about the D foundation (that's the reason why such a page should exist), but I tried to put all the information in here.

Ping @WalterBright @andralex @CyberShadow @MartinNowak 

1) How far are you with applying for non-profit status
2) Do you already have a bank acocunt?
3) Is there more known?

Other pages:

https://www.python.org/psf-landing/
https://www.mozilla.org/en-US/foundation/
http://www.linuxfoundation.org/
http://www.fsf.org/

### DIFF
--- a/foundation.dd
+++ b/foundation.dd
@@ -1,0 +1,64 @@
+Ddoc
+
+$(D_S $(TITLE),
+
+$(H4 The D Language Foundation is an organization devoted to advancing
+     open source technology related to the D programming language.
+)
+$(H3 History)
+
+$(P The D programming language was started by Walter Bright, as a 'better C++'
+and has grown slowly, but steadily, for more than 15 years. Recent improvements
+to the D development process have accelerated development of the language and
+shown the need for a governing organization that overlooks and arranges the
+future development of the D programming language.)
+
+$(H3 Current projects)
+
+The D foundation
+
+$(UL
+
+    $(LI is providing the various resources for the D community
+        (e.g. this website, dmd downloads, the D tour and more))
+    $(LI is an accepted
+        $(HTTPS summerofcode.withgoogle.com/organizations/5078256051027968/,
+        Google Summer of Code organization) and hosts four projects during
+        the summer of 2016.
+    )
+    $(LI is organizing the yearly $(HTTP dconf.org, D conference))
+)
+
+$(H3 Committee)
+
+The officers of the D Language Foundation are:
+
+$(UL
+    $(LI $(HTTP www.walterbright.com, Walter Bright), President)
+    $(LI $(HTTPS erdani.com/index.php/contact, Tudor Andrei Cristian Alexandrescu), Vice President and Treasurer)
+    $(LI $(HTTP http://acehreli.org, Ali Çehreli), Secretary)
+)
+
+$(H3 Becoming a member)
+
+Individual and corporate membership packages will be announced in the future.
+
+$(H3 Sponsoring D)
+
+$(P If you want to help the D Language to advance further and have resources
+    that you can spend, don't hesitate to contact
+    $(HTTPS erdani.com/index.php/contact, Andrei) and
+    $(HTTP www.walterbright.com, Walter).
+    $(BR)
+    Your help is highly appreciated!
+)
+
+$(H3 Official contacts)
+
+$(P The D Language Foundation has been incorporated with the state of Washington, USA.
+    The employer ID is 47-5352856.)
+
+)
+
+Macros:
+	TITLE=The D Language Foundation
--- a/dlang.org.ddoc
+++ b/dlang.org.ddoc
@@ -40,7 +40,7 @@ COMMON_SCRIPTS_DLANG =
 _=
 
 CONSOLE=$(TC pre, console notranslate, $0)
-COPYRIGHT=Copyright &copy; 1999-$(YEAR) by Digital Mars $(REG), All Rights Reserved
+COPYRIGHT=Copyright &copy; 1999-$(YEAR) by the $(LINK2 $(ROOT)/foundation.html, D Language Foundation)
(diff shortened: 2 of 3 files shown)


foundation.dd
```diff
@@ -0,0 +1,58 @@
+Ddoc
+
+$(D_S $(TITLE),
+
+$(H4 The D Language Foundation is an organization $(B devoted) to advancing
```
andralex: no need for bold text here

foundation.dd
```diff
@@ -0,0 +1,58 @@
...
+open source technology related to the $(B D programming language).)
```
andralex: no need for bold text here

foundation.dd
```diff
@@ -0,0 +1,58 @@
...
+future development of the D programming language.)
+
+$(H3 Current projects)
+
+The D foundation
+
+$(UL
+
+    $(LI is providing the various resources for the D community
+        (e.g. this website, dmd downloads, the D tour and a lot more))
```
andralex: remove "a lot" as too colloquial

foundation.dd
```diff
@@ -0,0 +1,58 @@
...
+    $(LI is an accepted
+        $(LINK2 https://summerofcode.withgoogle.com/organizations/5078256051027968/,
```
andralex: use WEB for lnks (http will forward to https)

foundation.dd
```diff
@@ -0,0 +1,58 @@
...
+        Google Summer of Code organization) and hosts four projects during
+        the summer of 2016.
+    )
+    $(LI is organizing the yearly $(LINK2 dconf.org, D conference))
```
andralex: use WEB, otherwise this link won't even work

foundation.dd
```diff
@@ -0,0 +1,58 @@
...
+)
+
+$(H3 Becoming a member)
+
+To be announced.
```
andralex: "Individual and corporate membership packages will be announced in the future."

foundation.dd
```diff
@@ -0,0 +1,58 @@
...
+
+$(H3 Sponsoring D)
+
+$(P If you want to help the D Language to advance further and have resources
+    that you can spend, don't hesitate to contact
+    $(LINK2 http://erdani.com/index.php/contact, Andrei) and
```
andralex: WEB

foundation.dd
```diff
@@ -0,0 +1,58 @@
...
+    $(LINK2 http://www.walterbright.com/, Walter).
```
andralex: WEB

foundation.dd
```diff
@@ -0,0 +1,58 @@
...
+    $(BR)
```
andralex: why?

foundation.dd
```diff
@@ -0,0 +1,58 @@
...
+    Your help is highly appreciated!
+)
+
+$(H3 Official contacts)
+
+$(P The D Language Foundation has been incorporated with the state of Washington, USA.
+    The employer ID is 47-5352856.)
+
+$(H3 Bank account)
```
andralex: may delete this for now

andralex:
Let's make one more pass after these nits.

wilzbach:
1) About using boldness or not - I don't care much, but here are both versions in comparison

image

image

(boldness is removed now)

2) I changed to `HTTP`/`HTTPS` - imho if a site has SSL you always want to link to that. So it's better to be explicit.

A bit off-topic:
3) `$(BR)` is needed because Ddoc doesn't insert newlines in `$(P` macros, however it does when one doesn't want it to, that's why I resort to the hack to put the ending parenthesis in a new-line to avoid a empty line which will result in two `<br>`'s inserted by Ddoc. I tried to fix those formatting errors in Phobos (e.g. https://github.com/dlang/phobos/pull/4357, https://github.com/dlang/phobos/pull/4356 and a couple more), but we really should make Ddoc more user-friendly. It's not a pleasant experience.

> Let's make one more pass after these nits.

All other nits have been adressed ;-)

foundation.dd
CyberShadow: This is an antipattern, doing so defeats half the point of HTTPS (as the HTTP redirect can be intercepted). We should go the other way.

andralex:
Great, thanks. Please add a section with the following information.

The officers of the D Language Foundation are:

* Walter Bright, President
* Tudor Andrei Cristian Alexandrescu, Vice President and Treasurer
* Ali Çehreli, Secretary

Please make sure you get the cedilla right :o).

With that we should be ready to go!

wilzbach:
> With that we should be ready to go!

Added.

> Please make sure you get the cedilla right :o).

Do you mean "Çehreli"?

foundation.dd
```diff
@@ -0,0 +1,64 @@
...
+)
+
+$(H3 Committee)
+
+The officers of the D Language Foundation are:
+
+$(UL
+    $(LI $(HTTP www.walterbright.com, Walter Bright), President)
+    $(LI $(HTTPS erdani.com/index.php/contact, Tudor Andrei Cristian Alexandrescu), Vice President and Treasurer)
+    $(LI $(HTTP http://acehreli.org, Ali Çehreli), Secretary)
```
andralex: Yah, that was the one. I think this works in most browsers but `&Ccedil; ` would put us on the safe side. I'll pull anyway.

adamdruppe: On Fri, May 27, 2016 at 06:47:00PM -0700, Andrei Alexandrescu wrote:
> Yah, that was the one. I think this works in most browsers but `&Ccedil; ` would put us on the safe side. I'll pull anyway.

If you are properly declaring the content type as ut-8, you are fine either way.

PR #1312, opened 2016-05-27
Add Organizations using D

wilzbach:
Idea: If someone asks you about companies using D, you redirect him to this page.
Inspired by this [forum thread](http://forum.dlang.org/post/elqwblvbafcgodsvwcdz@forum.dlang.org)

I hope this looks a lot more visually appealing than the [wiki](http://wiki.dlang.org/Current_D_Use) ;-)

I just started by adding companies from the current D uses and I hope in the future we can be more selective on dlang.org and highlight key players - or when the D foundation progresses: sponsors and members.

Question: @CyberShadow what is the best place to link to this? 
1) Menu "Community"
2) Community text
3) ...

It is responsive too.

image

Medium size:

image

Mobile:

image


CyberShadow:
This looks very nice but I don't know the policy for such things, do we ask companies permission first, or do we mention them and only remove them if they ask us to?

CyberShadow:
I think that if we go forward with this, it should be in a terser form on the front page too.

wilzbach:
> This looks very nice but I don't know the policy for such things, do we ask companies permission first, or do we mention them and only remove them if they ask us to?

Disclaimer: I am not an expert on legal things, but this is Wikipedia's has policy:

> Note that it is not necessary to seek formal permission from the owner in advance of using its logo, so long as the usage is fair use, does not create any impression that the logo is associated with Wikipedia or endorses either Wikipedia or the article in which the logo appears, and does not create any reasonable grounds for complaint by the owner.
...
> This is a logo of an organization, item, or event, and is protected by copyright and possibly trademark. It is believed that the use of low-resolution images on the English-language Wikipedia, hosted on servers in the United States by the non-profit Wikimedia Foundation, of logos for certain uses involving identification and critical commentary may qualify as fair use under United States copyright law.

https://en.wikipedia.org/wiki/Wikipedia:Logos

Imho we clearly state differentiate by saying: "Organizations running D in production.".

wilzbach:
> I think that if we go forward with this, it should be in a terser form on the front page too.

Great idea - is there anywhere space for this?

CyberShadow:
Okay, but there is a difference between what is legal and what is nice.

CyberShadow:
> Great idea - is there anywhere space for this?

Subjective but I would put it right above "Why D?", perhaps as a banner collage of the companies' logos.

wilzbach:
> Subjective but I would put it right above "Why D?", perhaps as a banner collage of the companies' logos.

I put it at the bottom of the page under "Industry-proven" (is there a better name?). Looks still great:

image

tsbockman:
> Note that it is not necessary to seek formal permission from the owner in advance of using its logo, so long as the usage is fair use, does not create any impression that the logo is associated with Wikipedia or endorses either Wikipedia or the article in which the logo appears, and does not create any reasonable grounds for complaint by the owner.

Titling the page "*Friends of D*" implies endorsement, as does "*Industry-proven*". I really think you need permission from the companies to use their trademarks in this way.

wilzbach:
> I really think you need permission from the companies to use their trademarks in this way.

Ok I tried to ping a Dlang developer for each company. Let's hope they all approve being listed on "Friends of D" and allow the use of their logo :)

Facebook: @deadalnix
Sociomantic: @leandro-lucarella-sociomantic
Remedy Games: @GooberMan
Weka: [@liranzvibel](https://twitter.com/liranzvibel)
Funkwerk AG: @lindt
eBay: @jondegenhardt
RandomStorm: @ethicalhack3r
EMSI: @Hackerpilot 
SR Labs: @pinver
Infognition: [thedeemon](https://bitbucket.org/thedeemon/)

PS: I hope you don't mind being listed here as contact person for the company, please let us know if I found the wrong person.

yannick:
Tamedia Digital: https://raw.githubusercontent.com/tamediadigital/tamedia-logos/master/tamedia-digital/logo_tam_digital_pos.jpg 
we're using D and Vibe.d for processing data streams and for API's

thedeemon:
As the head of Infognition Co. Ltd. I'd be happy for it to be included in such list.
Logo: http://www.infognition.com/images/logo.svg
http://www.infognition.com/images/infognition_logo_650_500.png

GooberMan:
I'm running it by the higher-ups, will let y'all know when I have confirmation.

liranz:
I think it's a great idea, and would love to help push D to becoming a more mainstream option.

jondegenhardt:
I need to check on company policy, will respond when I have more info.

Two things I'm expecting to be asked: First: what is the entity making the request. Is it the D Foundation? Second is simply the specific image used for the logo. There is often sensitivity about this.

andralex:
I'm not in love with "Frends of D". (1) Didn't like "Friends of Rust" one bit, and the fact we're copying it doesn't help. Brands and products don't have friends. People do. (2) Does this open the phrase to "the D" jokes?

How about simply "Organizations Using the D Language"?

andralex:
Forgot to mention - the notion itself is awesome.

wilzbach:
> How about simply "Organizations Using the D Language"?

Fair point - renamed it.
Btw I also tuned the UI a bit - that's how it looks at the moment.

image

andralex:
Great. I know next to nothing about css, but would it make sense for that page to have its own instead of adding to the global one?

wilzbach:
> Great. I know next to nothing about css, but would it make sense for that page to have its own instead of adding to the global one?

It's it own - @CyberShadow just proposed to add a selection to the frontpage.
Have a look yourself:

http://dtest.thecybershadow.net/artifact/website-1505d4939c110b1136ed2a1d0bf6eff5cf809148-fd1a65e9eae8059f93500937e6cff67b/web/index.html
http://dtest.thecybershadow.net/artifact/website-1505d4939c110b1136ed2a1d0bf6eff5cf809148-fd1a65e9eae8059f93500937e6cff67b/web/orgs-using-d.html

andralex:
@jondegenhardt yah the site's material is by the D Language Foundation so the request comes from it. Thanks for looking into this!

andralex:
@wilzbach cool then, I approve this pending @jondegenhardt 's response.

andralex:
THANKS!

jondegenhardt:
@andralex, @wilzbach - Thanks for confirming that it's the D Foundation. And for changing the language. "Organizations using D" is less likely to require additional explanation or discussion.

For the same reason, you might consider dropping the "Organizations using D in production" sub-title as well. At the moment there's only one grouping of corporations, right? And "production" has specific connotations to many people.

It's now late Friday, and Monday is a company holiday, so I think it'll be a few days before I have a response.

andralex:
yes, "production" is too specific

wilzbach:
> For the same reason, you might consider dropping the "Organizations using D in production" sub-title as well.

Agreed & removed ;-)

ethicalhack3r:
I may be wrong, but I'm not sure if RandomStorm has ever used D. Can you link me to somewhere that said they did?

RandomStorm we through a bunch of acquisitions last year. I'm not certain what the new owners plans are with the brand.

wilzbach:
> I may be wrong, but I'm not sure if RandomStorm has ever used D. Can you link me to somewhere that said they did?

I took all information from the D wiki:

https://wiki.dlang.org/Current_D_Use

Hackerpilot:
200x100 with alpha: emsilogo

JackStouffer:
Adroll also uses D: http://tech.adroll.com/blog/data/2014/11/17/d-is-for-data-science.html

JackStouffer:
Did a little more research, not only is Adroll using D, they [built integral](http://tech.adroll.com/blog/data-science/2015/12/08/data-science_event_processing.html) parts of [their business on it](http://tech.adroll.com/blog/data-science/2015/08/25/factorization-machines.html).

So, two of the largest ad companies use D, who knew ¯\\\_(ツ)\_/¯.

andralex:
@wilzbach I think until @jondegenhardt gets back with a response we should push this without eBay (but with EMSI). Could you please make that change? I preapprove this.

CyberShadow:
This looks a bit crooked:

index.dd
```diff
@@ -366,6 +367,23 @@ consistency. $(LINK2 spec/memory-safe-d.html, Read more).)
...
+    $(FRONTPAGE_ORG Facebook, https://www.facebook.com, facebook.svg,
+    )
+    $(FRONTPAGE_ORG eBay, https://ebay.com, ebay.svg,
+    )
+    $(FRONTPAGE_ORG Remedy Games, http://remedygames.com, remedy.svg,
+    )
+    $(FRONTPAGE_ORG Sociomantic, https://www.sociomantic.com, sociomantic.png,
+    )
+    <div id="frontpage-orgs-more">
+        $(LINK2 $(ROOT_DIR)/orgs-using-d.html, Show more organizations using D)
```
CyberShadow: `$(ROOT_DIR)orgs-using-d.html` (or `$(ROOT)/orgs-using-d.html`), not `$(ROOT_DIR)/orgs-using-d.html`

Also, no "Show" I think, just "More organizations using D". Technically most links are used to show something else.

CyberShadow:
The new page looks weird at certain widths:

aG0aep6G:
> The new page looks weird at certain widths:

That's `text-align: justify`. `center` would be better here.

css/style.css
```diff
@@ -2018,3 +2018,84 @@ the home page, intro pitch and your-code-here layed out vertically */
 dt.d_decl:hover .decl_anchor {
     visibility: visible;
 }
+
+/* Friends of D */
+.orgs-using-d {
+    width: 100%;
+    padding-left: 20px;
```
aG0aep6G: I think this padding should not be here. It's only on one side, and it gets added to the 100% width, resulting in an element that's more than 100% wide.

css/style.css
```diff
@@ -2018,3 +2018,84 @@ the home page, intro pitch and your-code-here layed out vertically */
...
+}
+
+.orgs-using-d-cell {
+    margin-left: auto;
+    margin-right: auto;
```
aG0aep6G: `auto` is the default value for `margin-left` and `margin-right`. No need to specify it here, as far as I see.

css/style.css
```diff
@@ -2018,3 +2018,84 @@ the home page, intro pitch and your-code-here layed out vertically */
...
+.orgs-using-d img {
+    max-width: 200px;
+    max-height: 100px;
+}
+
+.orgs-using-d h3 , .frontpage-orgs h3 {
+   margin: 0.4em;
+   margin-left: 0;
+   margin-right: 0;
+   color: black;
```
aG0aep6G: I'd leave the color at the default (very dark grey). I don't think the company names need the extra emphasis. This is purely subjective of course.

orgs-using-d.dd
```diff
@@ -0,0 +1,61 @@
...
+please open a
+$(LINK2 https://github.com/dlang/$(PROJECT)/edit/master/$(SRCFILENAME), PR) or
+$(LINK2 https://issues.dlang.org/enter_bug.cgi?component=dlang.org&short_desc=Add%20(insert%20your%20organization)%20to%20the%20DORGs%20of%20D, issue)
+
+)
+
+Macros:
+	TITLE=Organizations using the D Language
+	DORGIMG=$(DIVC orgs-using-d-img-wrapper, <img src="$(ROOT_DIR)images/orgs-using-d/$0" />)
+    DORG=$(DIVC orgs-using-d-cell, $(LINK2 $2, $(DORGIMG $3)) $(H3 $1) $(P $4))
```
aG0aep6G: Call me old-fashioned, but I don't like `<div>`s in my `<a>`s. Maybe turn it around:

    DORG=$(DIVC orgs-using-d-cell, $(DIVC orgs-using-d-img-wrapper, $(LINK2 $2, )) ...)

Or since we're using raw HTML already:

    DORG=$(DIVC orgs-using-d-cell, <a href="$2" class="orgs-using-d-img-wrapper"></a> ...)

css/style.css
```diff
@@ -2018,3 +2018,84 @@ the home page, intro pitch and your-code-here layed out vertically */
...
+   text-align: center;
+}
+
+.orgs-using-d img , .frontpage-orgs img {
+    /* vertical centering in CSS - cross-platform way */
+    position: relative;
+    top: 50%;
+    transform: translateY(-50%);
```
aG0aep6G: `transform` is relatively new. Maybe use one of the older tricks, like `line-height: 100px` on the parent, and `vertical-align: middle` here.

css/style.css
```diff
@@ -2018,3 +2018,84 @@ the home page, intro pitch and your-code-here layed out vertically */
...
+}
+
+/* Front page organizations */
+.frontpage-orgs {
+    width: 100%;
+    padding-left: 20px;
```
aG0aep6G: Same suspicious padding as above.

css/style.css
```diff
@@ -2018,3 +2018,84 @@ the home page, intro pitch and your-code-here layed out vertically */
 dt.d_decl:hover .decl_anchor {
     visibility: visible;
 }
+
+/* Friends of D */
+.orgs-using-d {
+    width: 100%;
```
aG0aep6G: `width: 100%` shouldn't be needed. It's a block element, right?

css/style.css
aG0aep6G: > auto is the default value for margin-left and margin-right.

Actually, the default is `0`, but `auto` has the same effect here, because of `display: inline-block`.

css/style.css
```diff
@@ -2018,3 +2018,84 @@ the home page, intro pitch and your-code-here layed out vertically */
...
+}
+
+.frontpage-orgs img {
+    max-width: 150px;
+    max-height: 75px;
+}
+
+.frontpage-orgs-cell {
+    margin-left: auto;
+    margin-right: auto;
```
aG0aep6G: These two have no effect. Same as above.

css/style.css
```diff
@@ -2018,3 +2018,84 @@ the home page, intro pitch and your-code-here layed out vertically */
...
+    display: inline-block;
+    vertical-align: top;
+    margin-bottom: 0.5em;
+    padding-right: 15px;
+}
+
+#frontpage-orgs-more {
+    text-align: right;
+    padding-right: 10px;
+    margin-bottom: -7px;
```
aG0aep6G: The surrounding padding is in `em`, so the negative margin here should be in `em`, too.

Or maybe just throw out the margin and padding declarations here. I don't think they add visual appeal.

orgs-using-d.dd
```diff
@@ -0,0 +1,61 @@
+Ddoc
+
+$(D_S $(TITLE),
+
+$(DDOC_COMMENT
+    - Put your log (png/svg) to images/DORGs.
+    - max. dims: 200x100 (resize with e.g. `mogrify -resize 200x100 image.png`)
+    - if you need commas, use a paragraph `$(P ...)`
```
aG0aep6G: No, don't use `$(P ...)` for commas here. `DORG` already generates a `$(P ...)`. You end up with `<p><p>...</p></p>` then, which is not valid HTML. You can replace commas with `$(COMMA)` instead.

orgs-using-d.dd
```diff
@@ -0,0 +1,61 @@
...
+    )
+    $(DORG Tamedia digital, http://tamediadigital.ch/?lang=en, tamedia_digital.png,
+         Largest media company in Switzerland. $(LINK2 https://github.com/tamediadigital, Github)
+    )
+)
+
+If your organization uses D in production and you want it listed on this page,
+please open a
+$(LINK2 https://github.com/dlang/$(PROJECT)/edit/master/$(SRCFILENAME), PR) or
+$(LINK2 https://issues.dlang.org/enter_bug.cgi?component=dlang.org&short_desc=Add%20(insert%20your%20organization)%20to%20the%20DORGs%20of%20D, issue)
```
aG0aep6G: Need to escape the `&` here as `&amp;`.

aG0aep6G:
Some of the descriptions tell how D is being used (e.g. Facebook), and others tell what the company does in general (e.g. eBay). It's not clear which is which for all of them.

jondegenhardt:
@andralex, @wilzbach - This is being pursued internally. No reason to expect an issue, but at a big company things like this go into a queue of already busy people. So, you may indeed want to push it out without eBay, with the intent of adding eBay when I've got a response.

wilzbach:
Thanks a lot @aG0aep6G for your feedback regarding the styling - I made a couple of edits and I now looks like this:

image

Remaining permissions

### Facebook

Afaict we are good as long as we don't modify the logo:
https://www.facebookbrand.com/guidelines?asset=2&media=1

Information about Facebook using D is luckily public.

### Sociomantic

I haven't got a final confirmation from @leandro-lucarella-sociomantic, but considering that they were the main sponsor of this year's DConf I hardly doubt that deny the fact that they use D.
We use their Sociomantic's logo on dconf.org too ;-)

Outstanding
----------------

Remedy Games, Adroll & eBay

@GooberMan, @jondegenhardt, @appascoe - please give me a quick ping once you got the ok from higher up and I make a follow-up PR ;-)

I removed the logo for Remedy Games, Adroll & eBay and commented them out, so that we can publish an initial version.

Addressing the review
-----------------------------

I tried to address all points of @aG0aep6G review. However about:

> Need to escape the & here as &amp;.

Are you sure? The link works fine for me.

https://issues.dlang.org/enter_bug.cgi?component=dlang.org&short_desc=Add%20(insert%20your%20organization)%20to%20organizations%20using%20D

aG0aep6G:
> Are you sure? The link works fine for me.
> 
> https://issues.dlang.org/enter_bug.cgi?component=dlang.org&short_desc=Add%20(insert%20your%20organization)%20to%20organizations%20using%20D

The URL itself is fine with '&'. But to put a '&' in HTML it needs to be written as `&amp;`. Like you have to write `\"` to get a double quote in a double quoted string: `"\""`.

Or rather, that was needed in older versions of HTML. Looks like HTML5 relaxed the rules there. Still, writing it as `&amp;` is not wrong, and we generally do it on dlang.org. GitHub does it too. If you check the source, you can see that the link in your comment is written with `&amp;` in the source.

aG0aep6G:
By the way, as a reviewer I don't like amended commits. Now I can't easily see what changed from the last revision, so I'd have to review everything again to make sure nothing slips through.

I'd prefer multiple commits. They can still be squashed into one at the end.

orgs-using-d.dd
```diff
@@ -0,0 +1,123 @@
+Ddoc
+
+$(D_S $(TITLE),
+
+$(COMMENT
+    - Put your logo (png/svg) to images/orgs-using-d
+    - max. dims: 200x100 (resize with e.g. `mogrify -resize 200x100 image.png`)
+    - if you need commas, use `$(COMMA`)
```
aG0aep6G: misplaced ` here

orgs-using-d.dd
```diff
@@ -0,0 +1,123 @@
...
+)
+$(DIVC orgs-using-d center,
+
+    $(DORG Facebook, https://www.facebook.com, facebook.svg,
+        Online social networking service,
+        $(HTTPS https://github.com/facebookarchive/warp, C Preprocessor warp) and $(HTTP www.drdobbs.com/mobile/facebook-adopts-d-language/240162694, more infrastructure tools).
```
aG0aep6G: Don't put `https://` in `$(HTTPS ...)`.

orgs-using-d.dd
```diff
@@ -0,0 +1,123 @@
...
+        $(LINK_ROW
+            $(FA_TALK http://dconf.org/2014/talks/simpkins.html, DConf talk)
```
aG0aep6G: `FA_TALK` forwards to `WEB` which forwards to `HTTP` which prepends `http://`. So no `http://` here, please.

orgs-using-d.dd
```diff
@@ -0,0 +1,123 @@
...
+        )
+    )
+    $(DORG Sociomantic, https://www.sociomantic.com, sociomantic.png,
+        Real-time bidding for eCommerce,
+        Key infrastructure in D - main sponsor of D conference 2016.
+        $(LINK_ROW
+            $(FA_TALK http://dconf.org/2013/talks/clugston.html, DConf talk) $(FA_SEPARATOR)
```
aG0aep6G: Ditto. Bad `http://`.

orgs-using-d.dd
```diff
@@ -0,0 +1,123 @@
...
+    text-align: center;
+}
+)
+
+)
+
+Macros:
+	TITLE=Organizations using the D Language
+	FA_ICON=<i class="fa fa-$1" aria-hidden="true"></i>
+	FA_HIRING=$(FA_ICON rocket) $(WEB 1, hiring)
```
aG0aep6G: `1` -> `$1`

orgs-using-d.dd
```diff
@@ -0,0 +1,123 @@
...
+	FA_GITHUB=$(FA_ICON github) $(HTTPS https://github.com/$1, Github)
```
aG0aep6G: Another wrong `https://`.

orgs-using-d.dd
```diff
@@ -0,0 +1,123 @@
...
+    please open a
+    $(LINK2 https://github.com/dlang/$(PROJECT)/edit/master/$(SRCFILENAME), PR) or
+    $(LINK2 https://issues.dlang.org/enter_bug.cgi?component=dlang.org&short_desc=Add%20(insert%20your%20organization)%20to%20organizations%20using%20D, issue)
+)
+
+$(T style,
+#content, #content h1 {
+    text-align: center;
+}
+)
```
aG0aep6G: I don't like putting a `<style>` block in the document body like this. I understand it's because you can't get a hold of the `<h1>` otherwise.

The body id is supposed to help here, but we're generating an invalid one with spaces in it. However, I think you can override `BODYCLASS` here in the *.dd file:

    BODYCLASS=doc organizations-using-d

Then select by it in CSS:

    body.organizations-using-d #content h1 { text-align: center; }

aG0aep6G: ... or just put the `center` class there

wilzbach:
> By the way, as a reviewer I don't like amended commits. 

Sorry - last change was rather huge. Used small commits for now.

> I don't like putting a <style> block in the document body like this.

I thought about this and I moved all custom css of the page in the a header style block. `style.css` is used for many other subsites to achieve the branding, so we shouldn't bloat this.

All comments are addressed.

wilzbach:
> Based on their website, "Tamedia digital" should be capitalised, and "Swiss" too.

Oh thanks for noticing - silly mistake :S.
Fixed.

wilzbach:
> The "PR" link is broken 

This file hasn't been merged yet, so you get a 404 from Github ;-)

> (and I'd probably spell out "pull request" for people less familiar with the terminology).

fair point - done ;-)

more nitpicks before we publish the initial version?

wilzbach:
> Do we have approval from all the parties yet? If so, I'd say go for it.

Yes except for Sociomantic (dconf key sponsor + logo on dconf.org) and Facebook (2013 dconf sponsor). Should I remove them too or do you think that being a dconf sponsor is enough?

index.dd
```diff
@@ -408,3 +408,34 @@ Macros:
     _=
 	FRONTPAGE_ORG_IMG=$(DIVC frontpage-orgs-img-wrapper vcontainer-box, <img class="vcontainer-element" src="$(ROOT_DIR)images/orgs-using-d/$0" />)
     FRONTPAGE_ORG=$(DIVC frontpage-orgs-cell dont-highlight-link, $(LINK2 $2, $(FRONTPAGE_ORG_IMG $3)) $(H3 $1) $(P $4))
+    C_A=$1:$2;
+    EXTRA_HEADERS=$(T style,
+        .frontpage-orgs h3 {
+           $(C_A margin, 0.4em)
```
aG0aep6G: What's the point of the `C_A` macro?

wilzbach: Ddoc!
It recognizes `.*:` as title (like `title:` or `params:`) and behaves weirdly.
I should probably add a comment about this.

aG0aep6G:
> I thought about this and I moved all custom css of the page in the a header style block. style.css is used for many other subsites to achieve the branding, so we shouldn't bloat this.

I'd prefer having it in style.css or another css/.css file, so that all CSS is in one place. But that'ss just my personal preference.

index.dd
aG0aep6G: > Ddoc!
> It recognizes .*: as header macro (like title: or params:) and behaves weirdly.

Ouch.

> I should probably add a comment about this.

Or just move to a *.css file :P

aG0aep6G:
> @aG0aep6G: Keeping per-page CSS inline is actually best practice, as it saves on loading times.

I think it's more important that we easily find it, but I could live with it being inline. However, as @wilzbach has discovered, writing CSS in Ddoc is quite a pain. I don't think it's worth the hassle.

wilzbach:
> > @aG0aep6G: Keeping per-page CSS inline is actually best practice, as it saves on loading times.
> I think it's more important that we easily find it, but I could live with it being inline.

Imho it's better if we keep it inline - it's faster, cleaner and easier to maintain. If someone wants to modify this page, he basically just needs to scroll down for the CSS.

> However as @wilzbach has discovered, writing CSS in Ddoc is quite a pain. I don't think it's worth the hassle.

It's not that horrible - compared to all the pain writing CSS in general ;-)

> Facebook I can't really judge; for example, their logo wasn't used on the DConf website either.

I asked Andrei about Facebook and Sociomantic:

> Feel free to list both, I'll take the heat if anything happens. -- Andrei

Should I squash the commits?

aG0aep6G:
> Imho it's better if we keep it inline - it's faster, cleaner and easier to maintain. If someone wants to modify this page, he basically just needs to scroll down for the CSS.

I don't think it's obvious that the origin of the CSS block is the *.dd file.

> It's not that horrible - compared to all the pain writing CSS in general ;-)

I disagree. I don't think having it in a *.css is bad enough to put up with the weirdness.

wilzbach:
> Yes, please.

Done and added a comment about inline CSS.
Be a bit careful with merging - thanks to @CyberShadow it's now auto-deployed ;-)

andralex:
Noice

liranz:
Didn't view for some time.
You specify "Entire infrastructure and network stack", which may lead people to think most of the product is not D, but the whole product (complete distributed storage solution) is in D, not only the infrastructure and network stack.
So, while the statement is true, you can claim a stronger message.

wilzbach:
> Didn't view for some time.

The page was improved quite a bit during the life of this PR ;-)

> So, while the statement is true, you can claim a stronger message.

Thanks! See #1334

leandro-lucarella-sociomantic:
> Sociomantic
> 
> I haven't got a final confirmation from @leandro-lucarella-sociomantic, but considering that they were the main sponsor of this year's DConf I hardly doubt that deny the fact that they use D.
> We use their Sociomantic's logo on dconf.org too ;-)

I've got the final official OK, but I was asked to change the *description* of the company to "Real-time personalized advertising". I hope that's OK.

Thanks!

leandro-lucarella-sociomantic:
Also, there are 2 other DConf talks that might be relevant to describe what the company do or its relationshipt to D: http://dconf.org/2014/talks/clugston.html and http://dconf.org/2016/talks/lucarella.html. I'm not sure if it does make any sense to include them though.

wilzbach:
> I've got the final official OK,

Great to hear & sorry that I pushed this a bit too much!

>  but I was asked to change the description of the company to "Real-time personalized advertising". I hope that's OK.

Do you want to quickly submit the PR yourself?
Just edit this file: https://github.com/dlang/dlang.org/blob/master/orgs-using-d.dd

> Also, there are 2 other DConf talks that might be relevant to describe what the company do or its relationshipt to D

Again feel free to change the description, but we probably can only put one DConf link there :/
So just take the newest?

CyberShadow:
@wilzbach BTW, I noticed that there are a lot more entries on https://wiki.dlang.org/Current_D_Use than mentioned here. What's the story with the other ones?

wilzbach:
>  BTW, I noticed that there are a lot more entries on https://wiki.dlang.org/Current_D_Use than mentioned here. What's the story with the other ones?

I pinged them via mail, but didn't get any reply back (yet) :/

deadalnix:
Ping @markisaa who'll be able to provide more detailed infos than I.

CyberShadow:
> I pinged them via mail, but didn't get any reply back (yet) :/

Perhaps if they added themselves to the wiki page, they won't mind if we list them on the site either. Of course we'd need to also be sure that the person who edited the wiki page is from the respective organization.

wilzbach:
> Perhaps if they added themselves to the wiki page, they won't mind if we list them on the site either. Of course we'd need to also be sure that the person who edited the wiki page is from the respective organization.

I am pretty sure they won't mind, but iirc
1) we have created this huge thread because we need the permission to copy & display the proprietary logo?
2) we don't know exactly for what they use D (and whether they even still use D) - for one company (called "Large hedge fund" in the wiki) they responded they currently reevaluated their tech stack and can't tell about the future.
3) We don't have any specific links - all orgs currently listed have at least one link where the reader can read more about the company's involvement with D

That being said, I will try to ping them again next week ;-)

PR #1315, opened 2016-05-29
quickindex: hide empty caption and use auto-width

wilzbach:
before:

image

after:

image

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -930,6 +930,11 @@ table td, table th, table caption
     padding: 0.3em;
 }
 
+.quickindex .book caption:empty
+{
+    padding: 0px;
+}
+
 table td
 {
     border: none;


aG0aep6G:
This has a couple issues.

It adds whitespace below the module heading when there is no default quickindex (before/after):

diff-a-after1

This next one isn't so bad, but you didn't mention it, so it seems to be unintentional. Default quickindexes get a visual padding (before/after):

diff-a-after1

Quickindexes for aggregates are broken:

diff-a-after1

wilzbach:
> so it seems to be unintentional.

Oh yes it is :/
I rebased the PR, s.t. it only sets the caption to for booktable captions in quickindexes. 

I don't know how to fix the other problem of the grey area, see e.g.:
https://dlang.org/phobos/std_digest_crc.html

The only fix that comes to my mind is to remove the ddoc macro/css class `quickindex` from those pages.

aG0aep6G:
> I don't know how to fix the other problem of the grey area, see e.g.:
> https://dlang.org/phobos/std_digest_crc.html

How about this: Go through phobos and change `$(DIVC quickindex, $(BOOKTABLE , ...))` to `$(TC table, quickindex, ...)`. I.e., get rid of the extra `<div>`s around the tables.

wilzbach:
> How about this: Go through phobos and change $(DIVC quickindex, $(BOOKTABLE , ...)) to $(TC table, quickindex, ...). I.e., get rid of the extra <div>s around the tables.

Thanks for pushing this too @aG0aep6G!

PR #1316, opened 2016-05-29
add whitespace in selective imports to dstyle

wilzbach:
Having a space before and after the colon for selective imports seems to be an informal rule and has been critized in many PR's. 
I personally prefer `import std.range: zip` instead of `std.range : zip`, but the point is to make it a rule and standardize it, so we can automatically check it & no further time is wasted about such nitpicks.

### DIFF
--- a/dstyle.dd
+++ b/dstyle.dd
@@ -294,6 +294,8 @@ else if (…)
     …
 }
 -------------------------------
+    $(LI Selective imports should have a space before and after the colon (`:`) like
+    `import std.range : zip`)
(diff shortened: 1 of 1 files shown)


joakim-noah:
Change "require" to "should have," otherwise some will get confused.  Other than that, LGTM.

wilzbach:
> Change "require" to "should have," otherwise some will get confused

Rebased.

DmitryOlshansky:
LGTM

PR #1317, opened 2016-05-30
use <div> for DDOC_BLANKLINE

aG0aep6G:
The old `<br><br>` would result in more whitespace after block elements
than after plain text. A `<div>` doesn't have that problem.

Exemplary before/after shot:

diff-a-after1


wilzbach:
Thanks!

PR #1318, opened 2016-05-31
splitup css into files

wilzbach:
The current `style.css` is a huge mess and very hard to maintain, e.g. the menu is specified at four different places.

It's also very difficult to propagate the branding to other sites if only parts are wanted.
This is sepates the `style.css` in a couple of minor files after their function - e.g. `menu.css` contains all css classes related to the menu. Moreover they are grouped into folders.

It is also a preparation for hopefully proper CSS (Sass?) tooling to come, s.t. we can get rid of such inflexible elements.

```
#top > .helper,
body > .container,
body#Home #content #tools > div,
body#Home #content > .intro > div,
body#Home #content > .more
{
    margin: auto;
    max-width: 76em;
    padding: 0 1em;
}
```

While doing this, it might be pretty hard to review, so better ideas how to do this in a reviewable are welcome ;-)
My naive attempt to sort the lines and diff didn't work entirely, because I added some comments and splitted off the media sections.

I also left a `todo.css` with harder to split/group classes that would burst the scope of this PR entirely.


aG0aep6G:
> better ideas how to do this in a reviewable are welcome ;-)

Is it feasible to split this into multiple commits?

Like:

* move stuff around in style.css,
* split style.css to separate files,
* add comments.

CyberShadow:
I am skeptical that this is needed considering that I successfully reused `style.css` for the forum without any issues :)

wilzbach:
> I am skeptical that this is needed considering that I successfully reused style.css for the forum without any issues :)

- For code.dlang.org there's a huge CSS file with overwrites for `style.css`
- For tour.dlang.org we had to separate the menu (https://github.com/stonemaster/dlang-tour/pull/216) and other common styles (https://github.com/stonemaster/dlang-tour/pull/253) manually

hence the idea to split `style.css` ;-)

wilzbach:
I will just close this as currently I am low on resources and I don't want to block anyone else who wants to have the fun of splitting up `style.css` ;-)

PR #1319, opened 2016-06-02
Upgrade to DDOX 0.15.4 - Fixes the DDOX part of issue 14608.

s-ludwig:
- Removes per-enum-member pages and instead displays the full enum member documentation inside of a single table
- Adds template arguments and constraints display for composite types and enums


andralex:
thx!

PR #1320, opened 2016-06-02
blacklist /library/

aG0aep6G:
The DDOX pages are not ready yet.

See also: http://forum.dlang.org/post/575079DE.2040508@erdani.org

### DIFF
--- a/robots.txt
+++ b/robots.txt
@@ -1,4 +1,5 @@
 User-agent: *
 Disallow: /phobos-prerelease/
 Disallow: /library-prerelease/
+Disallow: /library/
 Disallow: /cutting-edge/


CyberShadow:
This is probably the fifth proposal to remove DDox from search indexing.

The arguments in favor of doing this have been as before, except that DDox output has had **a lot** more problems in the past.

The arguments against this change have been that if DDox is not indexed, no one will use it, thus no one will report bugs, thus it will never be improved.

The first argument has of course weaked over time as DDox has improved. I think DDox is in good enough shape right now to be indexed right now. And as has been posted in the thread, the main problem pointed out has already been partially fixed in master, with a complete fix coming in later.

Let's just fix it instead.

PR #1321, opened 2016-06-02
Use stable DUB version for building dpl-docs.

s-ludwig:
Should fix "Non-optional dependency vibe-d:web of ddox
not found in dependency tree!?." errors.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -25,7 +25,7 @@ REMOTE_DIR=d-programming@digitalmars.com:data
 GENERATED=.generated
 
 # stable dub and dmd versions used to build dpl-docs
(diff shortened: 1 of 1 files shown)


CyberShadow:
Thanks! I wonder how this got past the autotester. I guess it depends on ~`/.dub`.

Do you think the autotester should nuke it between builds? It would guarantee checking that a completely clean build succeeds, but of course means downloading all needed packages for every build.

s-ludwig:
I think it wouldn't hurt to clean everything. Considering that any Travis-CI based project already re-downloads all packages for each commit, the added traffic should be negligible (although traffic in general has seen a steep rise recently).

PR #1322, opened 2016-06-02
Fix formatting of (M)REF_ALTTEXT. See issue 14403.

s-ludwig:
Uses the special DDOX_NAMED_REF built-in macro to achieve the desired effect.


CyberShadow:
```
core.exception.AssertError@../../home/.dub/packages/ddox-0.15.5/ddox/source/ddox/api.d(68): Missing identifier after dot!
```

s-ludwig:
Need to reproduce that assertion failure locally...

aG0aep6G:
Sweeto. Looks good to me.

PR #1326, opened 2016-06-04
Add Official D Blog to menu

wilzbach:
Is there any special sorting mechanism of the menu? Should I sort the links alphabetically?

Ping @JackStouffer @mdparker

Btw
1) the font-size for the menu on http://dlang.org/blog/ should be `15px` and your menu seems to be outdated (e.g. #1303 is missing)
2) have you tried the blog on your phone yet? The D menu doesn't seem to be responsive anymore :/

### DIFF
--- a/dlang.org.ddoc
+++ b/dlang.org.ddoc
@@ -224,6 +224,7 @@ $(DIVID cssmenu, $(UL
 NAVIGATION_COMMUNITY=
 $(SUBMENU
     $(ROOT_DIR)bugstats.php, Bug Tracker,
+    https://dlang.org/blog, Blog,
     https://forum.dlang.org, Forums,
     irc://irc.freenode.net/d, IRC,
     https://github.com/dlang, D on GitHub,


mdparker:
The menu on the blog is up to date and is now 15px. Not sure yet where the problem is with the menu responsiveness, but working on it. Thanks, Seb!

mdparker:
Thanks, David. I'll dig into it. Or maybe @JackStouffer  can figure it out first.

wilzbach:
@mdparker @JackStouffer again a bit off-topic, but I also have a few more nitpicks and ideas:

- the fonts used seem a bit different - at least on dlang.org nearly everything is thin
- how about adding to the li item "Community" the css class `active`?
- instead of word press search box you could use the Dlang search bar (you might fix the menu shift with this)
- there are currently two links to WordPress. Afaik [it's not required to mention that](https://wordpress.org/support/topic/is-it-allowed-and-possible-to-remove-proudly-powered-by-wordpress), so in terms of hiding a bit that we use PHP, you might consider to remove that ;-)

JackStouffer:
> the fonts used seem a bit different - at least on dlang.org nearly everything is thin

The headings are the same font, the body text is intentionally different, as to distinguish the body and the headers.

> how about adding to the li item "Community" the css class active?

Simple enough.

> instead of word press search box you could use the Dlang search bar (you might fix the menu shift with this)

Hmm, I think this could be possible. I could add a "only search the blog" dropdown.

> there are currently two links to WordPress. Afaik it's not required to mention that, so in terms of hiding a bit that we use PHP, you might consider to remove that ;-)

>you probably want to hide the login link too, it just makes script kiddies more aware of your backend

These are up to @mdparker

mdparker:
> there are currently two links to WordPress. Afaik it's not required to mention that, so in terms of hiding a bit that we use PHP, you might consider to remove that ;-)

Ugh. The one in the footer wasn't supposed to be there. The function generating it was hiding the Copyright text I had there. That bit is sorted now. As for the one in the meta box...

> you probably want to hide the login link too, it just makes script kiddies more aware of your backend

I honestly don't think it makes a difference. It's easy enough to see what we're running just by typeing in a WP specific URL. I always have the meta widget turned off on my personal blogs, simply because I don't allow multiple accounts. My initial thoughts here was that other people will eventually have accounts and so it will be convenient to have a log in link on the front page.

That said, there's no way to edit the Meta widget to remove the WP link without modifying the core Wordpress files. I looked into that when I first set the blog up. And now that you've mentioned it here, I realize I really don't like having the WP link there either. So, I've removed the meta box (log- in link and all, which I may add back in later) and replaced it with links to the RSS feeds.

@WalterBright  @andralex Is there anyway we can get the repo @JackStouffer created for the blog theme transferred to the dlang organization? I keep getting issues reported all over the place.

JackStouffer:
> @WalterBright @andralex Is there anyway we can get the repo @JackStouffer created for the blog theme transferred to the dlang organization? I keep getting issues reported all over the place.

I created a news group thread for this.

PR #1328, opened 2016-06-04
modernize macros in intro-to-datetime

aG0aep6G:
Replacing the custom macros XREF and CXREF with a custom REF_SHORT which
has the same syntax as REF and may prove useful generally.

The custom macro FULL_CXREF is replaced with REF.

Replacing raw HTML with MREF in definitions of CORE_TIME, STD_DATETIME, and
STD_FILE.

FULL_XREF does not have a custom definition here. It's going to be replaced
in a repository-wide purge.

sed commands:

XREF -> REF_SHORT:

    arg='\s*([^(),]*)'
    from='\$\(XREF\s'$arg','$arg'\)'
    to='$(REF_SHORT \2, std,\1)'
    sed -i -r "s/$from/$to/g" intro-to-datetime.dd

CXREF -> REF_SHORT:

    arg='\s*([^(),]*)'
    from='\$\(CXREF\s'$arg','$arg'\)'
    to='$(REF_SHORT \2, core,\1)'
    sed -i -r "s/$from/$to/g" intro-to-datetime.dd

FULL_CXREF -> REF:

    arg='\s*([^(),]*)'
    from='\$\(FULL_CXREF\s'$arg','$arg'\)'
    to='$(REF \2, core,\1)'
    sed -i -r "s/$from/$to/g" intro-to-datetime.dd

### DIFF
--- a/intro-to-datetime.dd
+++ b/intro-to-datetime.dd
@@ -64,9 +64,9 @@ $(D_S $(TITLE),
 
     $(P The duration types can actually be found in $(CORE_TIME). They are
         $(DURATION) and $(TICK_DURATION). $(TICK_DURATION) is intended for
-        precision timing and is used primarily with $(XREF datetime, StopWatch)
+        precision timing and is used primarily with $(REF_SHORT StopWatch, std,datetime)
         and the benchmarking functions found in $(STD_DATETIME) - such as
-        $(XREF datetime, benchmark) - and you're unlikely to use it outside of
+        $(REF_SHORT benchmark, std,datetime) - and you're unlikely to use it outside of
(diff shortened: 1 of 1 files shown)


CyberShadow:
Looking at the diff, the links should still be in monospace font, I think, to be consistent with the unlinked identifiers.

aG0aep6G:
> Looking at the diff, the links should still be in monospace font, I think, to be consistent with the unlinked identifiers.

https://github.com/dlang/dlang.org/pull/1327 takes care of that.

CyberShadow:
I see, thanks.

PR #1330, opened 2016-06-04
Add clipboard button to Phobos documentation

wilzbach:
Idea: http://forum.dlang.org/thread/niuv0a$2giv$1@digitalmars.com

Actually was easy to do. I renamed `listanchors.js`, because it contains now more code and thus `phobos.js` is a better name. Also jQuery wait until page load can be done within the script ;-)

CC @MartinNowak


js/phobos.js
```diff
@@ -23,14 +23,28 @@ function addAnchors()
         // we link to the first children
         var a = items[i].querySelector('a');
         if(!a) continue;
+
+        // copy-to-clipboard
+        var el = document.createElement("i");
+        var symbol = a.name.replace(".", "");
```
MartinNowak: Just substring to strip the leading dot?

js/phobos.js
```diff
@@ -23,14 +23,28 @@ function addAnchors()
...
+        var text = "import " + document.body.id + " : " + symbol + ";";
+        el.setAttribute("data-clipboard-text", text);
+        el.setAttribute("aria-hidden", true);
+        el.className = "fa fa-clipboard decl_relement";
```
MartinNowak: Should at least get a title, saying sth. like copy as import.

js/phobos.js
wilzbach: This was the "safer" choice - it will continue to work if the naming of the anchor link is changed and I do imagine that we want to kill the (unneeded) dot in the future ;-)

js/phobos.js
```diff
@@ -23,14 +23,28 @@ function addAnchors()
...
+        items[i].insertBefore(el, items[i].firstChild);
+
+        // instantiate clipboard
+        var clipboard = new Clipboard(el);
```
MartinNowak: Same issue as on code.dlang.org, there is no indication that something was copied.

js/phobos.js
MartinNowak: Should get the same hover color as the anchor.

wilzbach: Good point - I will try "smallprints" as known from the "Improve this page" link at the top right.

MartinNowak:
Let's wait for more opinions before continuing with this.

js/phobos.js
MartinNowak: There is another anchor w/o the dot, on some other element. There are too many links in the internet to easily change this.

MartinNowak: BTW, doesn't work for methods, e.g. `import std.math : FloatingPointControl.severeExceptions;`.

wilzbach:
> BTW, doesn't work for methods, 

Fixed.

> Same issue as on code.dlang.org, there is no indication that something was copied.

As you can see it's easy to listen for the event and use a tooltip button, but currently that requires a bit of further styling which I will do if we decide to go this road.

JackStouffer:
I'm not really sold on the utility of this. Can't you use DCD and get the result just as fast?

MartinNowak:
I gave DCD another shot, auto-completion with both ac-dcd and company-dcd is annoying as usual, but the symbol search (<kbd>C-c s</kbd>) is fairly usable.
Reading documentation in source code doesn't work though, and getting docs from DCD <kbd>C-c ?</kbd> also just spits out the doc comments of all overloads.
And even if I could get easily readable documentation with examples, I'd still have to type `ìmport std.algorithm.sorting : partialSort;`. Not to mention, that this doesn't cover the browsing documentation workflow.

js/phobos.js
```diff
@@ -13,6 +13,48 @@ function lastName(a) {
     return a.slice(pos + 1);
 }
 
+function showTooltip(elem, msg){
+    elem.setAttribute('class','btn tooltipped tooltipped-s');
+    elem.setAttribute('aria-label',msg);
+    elem.addEventListener('mouseleave',function(e){
+        //e.currentTarget.setAttribute('class','btn');
```
JackStouffer: No reason to include commented out code

js/phobos.js
```diff
@@ -13,6 +13,48 @@ function lastName(a) {
...
+    // name start with .
+    var symbol = a.name.split(".")[1];
+    var text = "import " + document.body.id + " : " + symbol + ";";
+    el.setAttribute("data-clipboard-text", text);
+    el.setAttribute("aria-hidden", true);
+    el.className = "fa fa-clipboard";
+
+    var container =  document.createElement("span");
+    container.className = "tip smallprint decl_relement";
+    container.setAttribute("width", "auto");
```
JackStouffer: It's a good idea to avoid putting style rules in JS.

wilzbach: This was only a test - as you can [see](http://dtest.thecybershadow.net/artifact/website-94acb02e4bb4c63b7b53c1fa2a531d29496be7ae-77a9a64d4957722ede41343ea2fdc28d/web/phobos-prerelease/std_algorithm_searching.html) we still need to fix the width & alignment of the tooltip.

js/phobos.js
wilzbach: removed the entire function ;-)

JackStouffer:
The anchor buttons seem to be broken; they don't create an id link and they only appear on some symbol headers and not others.

One style nitpick I also noticed is that these buttons aren't flush with either the "Jump to X" link or the symbol header and instead create a new row when they could just be floated to the right.

CyberShadow:
We still aren't sure if we want this, so maybe it's not yet worth spending time on this.

MartinNowak:
Any new opinions on that? I tried dcd for a while and find it rather annoying. Takes way too long to come up with completion suggestions, and once the suggestion box pops up, it breaks my normal typing/[hippie-expand](https://www.emacswiki.org/emacs/HippieExpand) workflow.
Also it doesn't really solve my search/browse documentation workflow, b/c while dcd has a well working symbol search, I still don't get usable documentation from that and still have to type the import.

CyberShadow:
Well, it seems that you are still the only one who has expressed interest in this.

> Takes way too long to come up with completion suggestions, and once the suggestion box pops up, it breaks my normal typing/hippie-expand workflow.

Is it really slower than going to a web browser, doing a web search, finding the right symbol, and using *the mouse* to click on a "copy to clipboard" button?

> I still don't get usable documentation from that

That should be improved. It's been on my list. It's fairly straight-forward:

1. Get dcd to stop duplicating the documentation for overloads (this is probably just a bug)
2. Filter it through `dmd` with a custom set of macros
3. Parse it through `eww` and put it in a documentation window.

> and still have to type the import.

That still seems rather specific to your workflow and code style.

My suggestion is to write a userscript.

CyberShadow:
How about the following proposal:

1. Keep the clipboard icon near the anchor as it is now.
2. Change the hover popup to use the `title` attribute. It's what it's for. The current one is too annoying when you move the mouse over it accidentally.
3. When you click it, it should load the clipboard JavaScript library, and open a small popup window (see add reaction to GitHub comments as a model example) which has the options to copy the symbol as:
    - selective import
    - fully qualified name (perhaps even with placeholders for arguments in the case of functions/templates)
  Just show what would be copied as the options.
4. The JavaScript clipboard library should only be loaded when the user clicks the popup, not before. I don't think this justifies increasing page load time.

If someone is willing to implement the above then that would be a sensible improvement I wouldn't mind merging.

BTW, I expected this to use the clipboard icon that GitHub uses for its clone URLs. This one looks different.

CyberShadow:
No takers?

PR #1331, opened 2016-06-04
DDOX: update top menu

aG0aep6G:
It had gone out of sync with the Ddoc one.

### DIFF
--- a/dpl-docs/views/layout.dt
+++ b/dpl-docs/views/layout.dt
@@ -41,7 +41,7 @@ html(lang='en-US')
           #cssmenu
             ul
               li
-                a(href="#{root_dir}getstarted.html")
+                a(href="http://tour.dlang.org")
                   span Learn
               li.expand-container
                 a.expand-toggle(href="#{root_dir}documentation.html")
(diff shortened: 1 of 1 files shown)


aG0aep6G:
@CyberShadow and @rainers: The menus on forum.dlang.org and http://rainers.github.io/visuald/visuald/StartPage.html are outdated as well.

rainers:
> The menus on http://rainers.github.io/visuald/visuald/StartPage.html are outdated as well.

Updated.

wilzbach:
> It had gone out of sync with the Ddoc one.

As the one who forgot to update the ddox menu. I am sorry (at the beginning having ddoc and ddox was very confusing) and thanks a lot for noticing and fixing!
Hence  obviously LGTM ;-)

wilzbach:
Sorry for the a bit off-topic comment:
@rainers How about using a CNAME for the Github Pages like `visuald.dlang.org`?
It's rather easy: https://help.github.com/articles/using-a-custom-domain-with-github-pages/
and while I already ask - why don't you use the Github Pages at https://github.com/dlang/visuald?

CyberShadow:
There are a lot of site navigation changes lately FWIW.

rainers:
> @rainers How about using a CNAME for the Github Pages like visuald.dlang.org?
> It's rather easy: https://help.github.com/articles/using-a-custom-domain-with-github-pages/

If I understand correctly, that needs to be done by someone with admin rights for dlang.org. I'm not sure who would be the one to add redirects.

> and while I already ask - why don't you use the Github Pages at https://github.com/dlang/visuald?

Good question. I remember we didn't host it on dlang.org so that I would not have to nag someone whenever I want to update the documentation. Using the "official" github pages seems natural, though.

PR #1333, opened 2016-06-05
Update community section on frontpage

wilzbach:
1) adds a link to the official dlang blog CC @mdparker 
2) removes the link to the empty/outdated review queue

The best replacement I could find for 2) were the DIPs, better ideas are welcome.

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -95,12 +95,13 @@ $(DIVC boxes,
         )
         $(TOUR comments, Community,
             $(P Discuss D on the $(LINK2 https://forum.dlang.org/, forums), join
-                the $(LINK2 irc://irc.freenode.net/d, IRC channel), or follow
-                us on $(LINK2 https://twitter.com/D_Programming, Twitter).)
+                the $(LINK2 irc://irc.freenode.net/d, IRC channel), read our
+                $(LINK2 https://dlang.org/blog, official Blog), or follow us
+                on $(LINK2 https://twitter.com/D_Programming, Twitter).)
             $(P Browse the $(LINK2 https://wiki.dlang.org, wiki) where among
                 other things you can find the
-                $(LINK2 https://wiki.dlang.org/Review_Queue, review queue) for
-                major additions to the D project.)
+                $(LINK2 https://wiki.dlang.org/Vision/2016H1, high-level vision)
+                of the D foundation.)
         )
     )
     $(DIVC row,


index.dd
```diff
@@ -95,12 +95,13 @@ $(DIVC boxes,
         )
         $(TOUR comments, Community,
             $(P Discuss D on the $(LINK2 https://forum.dlang.org/, forums), join
-                the $(LINK2 irc://irc.freenode.net/d, IRC channel), or follow
-                us on $(LINK2 https://twitter.com/D_Programming, Twitter).)
+                the $(LINK2 irc://irc.freenode.net/d, IRC channel), read our
+                $(WEB https://dlang.org/blog, official Blog) or follow us
```
aG0aep6G: Link is broken (`WEB` implies "http://").

I don't know if we demand the Oxford comma, but if we do, one's missing here before "or".

wilzbach:
> Link is broken (WEB implies "http://").

Argh - thanks. Will I ever learn those Ddoc macros? :S

> I don't know if we demand the Oxford comma, but if we do, one's missing here before "or".

Fair point - seems like the Oxford comma style is prevalent on dlang.org.
Is there a good place to document this?

aG0aep6G:
> Argh - thanks. Will I ever learn those Ddoc macros? :S

I for one would be in favor of ditching `WEB` (foremost), and `HTTP` and `HTTPS` (while we're at it).

aG0aep6G:
Looks good to me.

wilzbach:
> I for one would be in favor of ditching WEB (foremost), and HTTP and HTTPS (while we're at it).

Yes please and even Andrei is in favor of this:

> Yah, I wouldn't disagree things can be vastly improved. My point there is to make sure we distinguish what's trivially fixable (e.g. redundant macro definitions) from larger issues. Again, if you find identical macros across dd/ddoc files, or redundant macros that do the same thing in just slightly different ways, please fix or file. Thanks!
>
> Andrei

http://forum.dlang.org/post/ni4b13$21go$1@digitalmars.com

Btw ideally Ddoc should at least partially support Markdown - I don't see any reason why `[text](link)` should be impossible to parse ;-)

CyberShadow:
> The best replacement I could find for 2) were the DIPs, better ideas are welcome.

What about the vision pages?

E.g. https://wiki.dlang.org/Vision/2016H2_%28Draft%29 (still a draft)

wilzbach:
> What about the vision pages?

Nice idea - how about linking to both DIPs and vision pages?

For the meantime we could just use the current one:
https://wiki.dlang.org/Vision/2016H1

CyberShadow:
Not sure the DIP page is a good idea considering the vast majority of DIPs are abandoned or rejected.

wilzbach:
> Not sure the DIP page is a good idea considering the majority of DIPs are abandoned or rejected.

Fair enough - something that we should fix & work on that it doesn't happen (so often) in the future!
Rebased.

index.dd
```diff
@@ -95,12 +95,13 @@ $(DIVC boxes,
         )
         $(TOUR comments, Community,
             $(P Discuss D on the $(LINK2 https://forum.dlang.org/, forums), join
-                the $(LINK2 irc://irc.freenode.net/d, IRC channel), or follow
-                us on $(LINK2 https://twitter.com/D_Programming, Twitter).)
+                the $(LINK2 irc://irc.freenode.net/d, IRC channel), read our
+                $(LINK2 https://dlang.org/blog, official Blog), or follow us
```
CyberShadow: Nit, should have a trailing slash to avoid redirect

PR #1334, opened 2016-06-05
orgs using d - update weka's description

wilzbach:
Follow-up to #1312

CC @liranz

Is the size of Weka's codebase something that should be mentioned or should I better remove that?

### DIFF
--- a/orgs-using-d.dd
+++ b/orgs-using-d.dd
@@ -35,7 +35,7 @@ $(DIVC orgs-using-d center,
     )
     $(DORG Weka.IO, http://www.weka.io, weka.png,
         Software defined storage,
-        Entire infrastructure and network stack (over 200K LoC).
+        Complete distributed storage solution (over 200K LoC).
         $(LINK_ROW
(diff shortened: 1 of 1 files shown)


liranz:
You can use our current code base size. It won't stop us from writing new code :)
The impressive part of the figure is that comparable projects are much larger, so it's a tribute to D's "efficiency" as a programming language.

wilzbach:
> You can use our current code base size. It won't stop us from writing new code :)
The impressive part of the figure is that comparable projects are much larger, so it's a tribute to D's "efficiency" as a programming language.

Good point - should I change it to "only"?

JackStouffer:
> Good point - should I change it to "only"?

If you do so, be sure to include the sizes of comparable projects as a reference, as 200K sounds huge at first glance.

wilzbach:
> If you do so, be sure to include the sizes of comparable projects as a reference, as 200K sounds huge at first glance.

Hmm I am limited to only two lines, so I wouldn't know how.  Do you have an idea?

JackStouffer:
I'd just leave the part about "only" out then.

PR #1336, opened 2016-06-05
Remove carriage-returns from sitemap

wilzbach:
Somewhere is a carriage-return hidden. The sitemap `sed` puts that into the HTML.
This PR avoids matching the carriage return symbol:

Before:

http://dlang.org/sitemap.html

image

After:

image


JackStouffer:
👍

PR #1337, opened 2016-06-05
Fix bad host.

veelo:


aG0aep6G:
Looks good to me.

wilzbach:
Thanks for fixing!
@aG0aep6G seems like we really need to get rid of all these macros a la WEB/HTTP/HTTPS, ...

How about:

```
sed 's/(HTTPS /(LINK2 https:\/\//' -i **/*.d
sed 's/(HTTP /(LINK2 http:\/\//' -i **/*.d
sed 's/(WEB /(LINK2 http:\/\//' -i **/*.d
```

... and they are all gone?

CyberShadow:
Sorry, misclicked :)

aG0aep6G:
> How about:
> 
> sed 's/(HTTPS /(LINK2 https:\/\//' -i **/*.d
> sed 's/(HTTP /(LINK2 http:\/\//' -i **/*.d
> sed 's/(WEB /(LINK2 http:\/\//' -i **/*.d
> 
> ... and they are all gone?

Sure. Need to do phobos, druntime, and dmd as well. Also look for cases where the macro name is at the end of a line. Though they would probably make broken links. In any case, when done, check with a simpler grep (e.g. `grep -r HTTPS`) that you got everything. And make sure that latex.ddoc or some such don't do anything funky with the macros.

Myself, I'd like to finish the XREF thing before I start the next `sed` adventure.

PR #1338, opened 2016-06-05
use <span id="foo"> for anchors instead of <a name="foo">

aG0aep6G:
The old method would mess with surrounding links.

Hit recently in <https://github.com/dlang/phobos/pull/4410>.

### DIFF
--- a/html.ddoc
+++ b/html.ddoc
@@ -4,7 +4,7 @@ here for completeness. Macros defer wherever possible to style classes.
 _=Simple tags, ordered alphabetically
 
 A = <a href="$1">$+</a>
-ADEF = <a name="$0"></a>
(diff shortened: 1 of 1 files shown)


CyberShadow:
`<span id=".isLoggingActiveAt">`

Is this valid HTML?

adamdruppe:
On Sun, Jun 05, 2016 at 08:52:28PM -0700, Vladimir Panteleev wrote:
> `<span id=".isLoggingActiveAt">`
> 
> Is this valid HTML?

Yes, valid HTML5. The leading period is illegal in HTML4 id attributes... but so is the leading period in the name attribute.

This will function identically to the existing `<a name` regardless.

PR #1341, opened 2016-06-06
uncomment dman against bugs

wilzbach:
I just found this by accident:

image

and I really like the our D man, see also

- http://dlang.org/overview.html
- https://dlang.org/spec/iasm.html

Unfortunately the git history doesn't tell why it was removed, maybe @aG0aep6G or @CyberShadow know?
We need more of such great artwork

Btw I don't know the author, but he seems to tweet a bit: https://twitter.com/DlangGuy

### DIFF
--- a/spec/contracts.dd
+++ b/spec/contracts.dd
@@ -20,7 +20,7 @@ $(SPEC_S Contract Programming,
 	$(LI handling of contract inheritance)
 	)
 
-$(COMMENT <img src="../images/d4.gif" alt="Contracts make D bug resistant" border="0">)
+<img src="../images/d4.gif" alt="Contracts make D bug resistant" border="0">
 
 	$(P The idea of a contract is simple - it's just an expression that must evaluate to true.
 	If it does not, the contract is broken, and by definition, the program has a bug in it.


CyberShadow:
Not sure it's the same author.

Also, the image on http://dlang.org/overview.html is not the original - it's from when someone traced/vectorized the old one.

This drawing does seem a little crude / pixelated, maybe that's why it was commented?

wilzbach:
> Not sure it's the same author.

Maybe @WalterBright knows more?

> This drawing does seem a little crude / pixelated, maybe that's why it was commented?

We could just make it smaller?

WalterBright:
I drew the original D-Man, and the one here. They were done on a scrap of paper, then scanned in, then 'cleaned up' a bit with my extremely limited photoshop skillz. Someone converted the original D-Man which was a bit pixelated as well into a smooth svg rendering, but obviously that didn't happen to the cannon one.

I think it was removed at some point to lend a more 'professional' look to the web pages, but a bit of whimsy like these cartoons is appropriate as far as I'm concerned. After all, there is the Go gopher, Tux, the FreeBSD demon, etc., and they're all fun and good.

WalterBright:
I quite enjoy the D-Man comic strip and cartoons done by others.

PR #1342, opened 2016-06-06
Orgs using D: add Tripaneer.com

wilzbach:
Adds Tripaneer.com to the organizations using D.

CC @mario-silva (from Tripaneer.com)


orgs-using-d.dd
```diff
@@ -107,15 +119,16 @@ $(P
...
-	LINK_ROW=$(DIVC orgs-using-d-linkrow, $1)
-	DORGIMG=<a class="orgs-using-d-img-wrapper vcontainer-box" href="$1"><img class="vcontainer-element" src="$(ROOT_DIR)images/orgs-using-d/$2" /></a>
+    TITLE=Organizations using the D Language
+    FA_ICON=<i class="fa fa-$1" aria-hidden="true"></i>
+    FA_HIRING=$(FA_ICON rocket) $(WEB $1, Hiring)
+    FA_GITHUB=$(FA_ICON github) $(HTTPS github.com/$1, Github)
+    FA_TALK=$(FA_ICON youtube-play) $(WEB $1, $2)
+    FA_SEPARATOR=&nbsp;
+    FA_TESTIMONIAL=$(FA_ICON heart)
+    FA_QUOTE=$(FA_ICON quote-left) $1 $(FA_ICON quote-right)
```
CyberShadow: I think if you use `$0` instead of `$1` here, you won't need to escape commas with `$(COMMA)` when using this macro.

CyberShadow:
Do we have their OK?

wilzbach:
> I think if you use $0 instead of $1 here, you won't need to escape commas with $(COMMA) when using this macro.

Nice idea - done :)

> Do we have their OK?

> Yes please 
> Do you need any more details from us?
> Thanks,
> Mario

...

>   a) For which applications/products you use D?
Whole infrastructure in D

>  b) The reason to use D?
Very good ratio on Performance/Productivity

>  c) The features you like about D?
Succinct syntax; Introspection; Native, Typed; Good compilation times with
DMD

orgs-using-d.dd
```diff
@@ -70,7 +70,19 @@ $(DIVC orgs-using-d center,
...
+            $(FA_TESTIMONIAL) $(HTTP www.infognition.com/blog/2014/why_d.html, Testimonial)
+        )
+    )
+    $(DORG Tripaneer.com, https://www.tripaneer.com, tripaneer.png,
+        Themed vacation websites,
+        $(FA_QUOTE
+            Very good ratio on performance/productivity
+        )
+        $(LINK_ROW
+            $(FA_GITHUB ebookingservices)
```
CyberShadow: I think this is missing a `$(FA_SEPARATOR)`?

wilzbach:
> I think this is missing a $(FA_SEPARATOR)?

Yep how silly - thanks for catching :)

mario-silva:
Being connected with us (Tripaneer.com) is all good :+1:

PR #1344, opened 2016-06-06
Orgs using D: update sociomantic's description

wilzbach:
follow-up to #1312

CC @leandro-lucarella-sociomantic

> but I was asked to change the description of the company to "Real-time personalized advertising".  I hope that's OK.

Sure - your only enemy is the font. If I add "personalized", it will wrap to two lines. Any ideas?
(I changed it now to "Real-time advertising")

> Also, there are 2 other DConf talks that might be relevant to describe what the company do or its relationshipt to D:

Updated to the latest as it is a summary of all the great stuff Sociomantic has been doing for/with D.

### DIFF
--- a/orgs-using-d.dd
+++ b/orgs-using-d.dd
@@ -59,10 +59,10 @@ $(DIVC orgs-using-d center,
     )
     )
     $(DORG Sociomantic, https://www.sociomantic.com, sociomantic.png,
-        Real-time bidding for eCommerce,
+        Real-time personalized advertising,
         Key infrastructure in D - main sponsor of D conference 2016.
         $(LINK_ROW
-            $(FA_TALK dconf.org/2013/talks/clugston.html, DConf talk) $(FA_SEPARATOR)
+            $(FA_TALK dconf.org/2016/talks/lucarella.html, DConf talk) $(FA_SEPARATOR)
             $(FA_HIRING www.sociomantic.com/jobs/d-software-developer)
         )
     )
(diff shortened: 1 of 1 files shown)


CyberShadow:
> Sure - your only enemy is the font. If I add "personalized", it will wrap to two lines. Any ideas?

There's lots of space on the sides, what's the problem? Add some negative margins or something.

CyberShadow:
Think outside the box ;) ;) ;)

(couldn't help myself)

wilzbach:
> There's lots of space on the sides, what's the problem? Add some negative margins or something.

The idea was to limit the width of a cell to `250px`, so we shouldn't cheat there. That being said either increasing the `width` to `255px` (or even more) or a margin hack work:

```css
margin-left: -1px;
margin-right: -1px;
```

The `container` has a maximal width of `1140px` and `250 * 50 (padding) * 3 (# cells) = 900`, so we still have "space".

CyberShadow:
> The idea was to limit the width of a cell to 250px, so we shouldn't cheat there.

Why?

wilzbach:
Rebased after the alphabetical sorting and added the margin "hack":

image

leandro-lucarella-sociomantic:
Thanks!

leandro-lucarella-sociomantic:
Any reason not to use the same style in the home page than what's show in here / the full http://dlang.org/orgs-using-d.html page?

wilzbach:
> Any reason not to use the same style in the home page than what's show in here / the full

Not in particular - in the beginning the style was rather lame compared to now, so then I just thought we save a bit of space.

-> I will give it a try - thanks for the feedback!

PR #1345, opened 2016-06-06
Visual improvements to orgs using D

wilzbach:
- dont hyphenate the uses
- removed duplicate "hiring"

### DIFF
--- a/orgs-using-d.dd
+++ b/orgs-using-d.dd
@@ -3,9 +3,11 @@ Ddoc
 $(D_S $(TITLE),
 
 $(COMMENT
-    - Put your logo (png/svg) to images/orgs-using-d
-    - max. dims: 200x100 (resize with e.g. `mogrify -resize 200x100 image.png`)
-    - if you need commas, use `$(COMMA)`
+    - Put your logo (svg or transparent png) to images/orgs-using-d
+    - At the moment, we use 150x75 as maximal dimensions for the displayed logos
+       (resize with e.g. `mogrify -resize 150x75 logo.png`)
+    - If you have a better quality logo, please name it `<org-name>_hq.<svg|png>`
+    - If you need commas, use `$(COMMA)`
 )
 $(DIVC orgs-using-d center,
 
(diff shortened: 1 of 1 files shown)


CyberShadow:
Can we change the text at the bottom?

"Is your organization missing from this page? Please [create an issue](#) or [open a pull request](#)."

Also, what order are the companies in? E.g. Rust goes with alphabetical.

CyberShadow:
> ```
    - max. dims: 200x100 (resize with e.g. `mogrify -resize 200x100 image.png`)
```

Why? On high-DPI screens (mobile devices) the extra resolution will not go to waste.

Most logos are simple shapes which do not suffer in file size much at bigger resolutions.

**Comment by CyberShadow** [comment]
> ```
    - if you need commas, use `$(COMMA)`
```

Is this still needed after the move to `$0`?

wilzbach:
> Can we change the text at the bottom?

Sure. Thanks & done

> Also, what order are the companies in? E.g. Rust goes with alphabetical.

Initially it was based on where I had content with similar height, but yes sorting alphabetically is fair to all. I will submit another PR ;-)

> Why? On high-DPI screens (mobile devices) the extra resolution will not go to waste.

It sounded like a reasonable cut-off for me, but we can of course change that.
- Currently the same images are used for the front page too.
- the "real" size is 150x75 at the moment

> if you need commas, use `$(COMMA)`

Yes, we use `$0` only for the FA_QUOTES.

CyberShadow:
Yep, there should not be restrictions on logo size. In any case we should have the original in the repo, and maybe a scaled down version if necessary (see the OS logos for the download page).

In addition to benefiting high-DPI displays, a future site redesign might require logos of different resolution.

wilzbach:
> Yep, there should not be restrictions on logo size. In any case we should have the original in the repo, and maybe a scaled down version if necessary 

Okay I added a better explanation.

> (see the OS logos for the download page).

Aren't they also all less than `50x50` at the moment?

CyberShadow:
> Aren't they also all less than 50x50 at the moment?

Ah yes, the higher res ones + conversion commands are in the git history.

PR #1347, opened 2016-06-06
Add a Continuous Integration guide

wilzbach:
I have explored testing with AppVeyor for  [mir](https://github.com/libmir/mir) quite a while ago and it seems to me that people luckily found our setup and copied it for their use case e.g. https://github.com/etcimon/libasync/pull/65. Thanks @Calrama!

Imho we should make this information a lot easier to access and provide simple instruction on how one can add CI providers if we want to see wide-spread use and thus a better D ecosystem.
Btw thanks to @MartinNowak's script this is already pretty awesome for Linux!

This CI guide is by no means complete, but I hope it's a good base to start for a discussion.

CC @MartinNowak

### DIFF
--- a/ci.dd
+++ b/ci.dd
@@ -1,0 +1,192 @@
+Ddoc
+
+$(D_S $(TITLE),
+
+Continuously testing your project is ensures a conflict-free shipping.
+For all the listed Continuous Integration provides, you will need a free account
+to enable building.
+
+$(H3 Travis)
+
+Travis has built-in support for D and building your project with D is as easy as
+adding to your `.travis.yml` the compilers to be tested :
+
+----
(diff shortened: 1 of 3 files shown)


CyberShadow:
Nice but I think the wiki would be a better place for this.

wilzbach:
> Nice but I think the wiki would be a better place for this.

Fair enough. If we want to enable [CI testing for dub](https://github.com/dlang/dub-registry/issues/112), we might want to spot-light the resulting document in both cases.

CC @s-ludwig

wilzbach:
> Nice but I think the wiki would be a better place for this.

For reference the page moved over here:

https://wiki.dlang.org/Continuous_Integration

PR #1350, opened 2016-06-07
Orgs using D: nitpick in Tripaneer quote

wilzbach:
Apparentely it's well defined how to express a ratio:
https://en.wikipedia.org/wiki/Ratio#Notation_and_terminology

Also ends an ending dot.

### DIFF
--- a/orgs-using-d.dd
+++ b/orgs-using-d.dd
@@ -73,7 +73,7 @@ $(DIVC orgs-using-d center,
     $(DORG Tripaneer.com, https://www.tripaneer.com, tripaneer.png,
         Themed vacation websites,
         $(FA_QUOTE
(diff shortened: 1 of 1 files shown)


CyberShadow:
If that's an actual, word-by-word quote, then perhaps we shouldn't be putting words in their mouths?

wilzbach:
> we shouldn't be putting words in their mouths?

CC @mario-silva - are you ok with changing your quote from "Very good ratio on performance/productivity." to "Very good ratio of performance to productivity."?

See: https://en.wikipedia.org/wiki/Ratio#Notation_and_terminology for the rationale

mario-silva:
It's ok by me. No problem at all
Thanks

PR #1351, opened 2016-06-07
Shorten TWiD frontpage text

wilzbach:
At the moment TWiD uses four lines:

image

this shortens it to three lines so that's more similar to the height of the Learn section:

image

Btw Both links redirect (This week in D, "June 5") redirect to the same page - so we could remove one?

CC @adamdruppe 

Btw @adamdruppe now that @CyberShadow's auto-deploy is there, you could think about dropping your custom js ;-)

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -89,8 +89,8 @@ $(DIVC boxes,
             )
             $(P Stay updated with
                 $(LINK2 http://arsdnet.net/this-week-in-d,
-                    <cite>This Week in D</cite>),
-                Adam D. Ruppe's weekly newsletter. $(SPANC twid)
+                    <cite>This Week in D</cite>) -
+                $(SPANC twid)
             )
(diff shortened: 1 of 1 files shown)


adamdruppe:
On Tue, Jun 07, 2016 at 06:08:24AM -0700, Sebastian Wilzbach wrote:
> this shortens it to three lines so that's more similar to the height of the Learn section:

looks ok to me.

> Btw Both links redirect (This week in D, "June 5") redirect to the same page - so we could remove one?

Actually, I think I might make the title go to the RSS and the
date go to the current page.

> Btw @adamdruppe now that @CyberShadow's auto-deploy is there, you could think about dropping your custom js ;-)

Eh, it is more about the ease of cross-domain updating than anything else. With the JS, I just edit my file on my website and then it just works.

PR #1352, opened 2016-06-07
Add description and dconf talk to frontpage companies

wilzbach:
as @leandro-lucarella-sociomantic suggested, this PR puts the short description of the companies and their DConf Link onto the frontpage.

before:

image

after:

image

Unfortunately imho adding the quote and more links makes it to messy for the frontpage, see:

After experimenting with this, I am not sure anymore whether it's a good idea to add Links
@aG0aep6G @CyberShadow - what do you think?

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -386,7 +386,7 @@ $(DIVC frontpage-orgs center,
     $(FRONTPAGE_ORG Weka.IO, http://weka.io, weka.png,
         Software defined storage
     )
-    $(FRONTPAGE_ORG Funkwerk AG, http://remedygames.com, funkwerk.png,
+    $(FRONTPAGE_ORG Funkwerk AG, http://www.funkwerk.com/en, funkwerk.png,
         Passenger information systems
     )
 )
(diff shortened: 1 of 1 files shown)


wilzbach:
Another idea would be no just show the play icon - aligned to the right, roughly like this:

image

leandro-lucarella-sociomantic:
I think is enough with the company description, if you want to add the DConf talk link too, for me both options loop fine (although I prefer the first one).

CyberShadow:
Yep, the whole package looks too busy for front page. Just descriptions should be enough

wilzbach:
> Yep, the whole package looks too busy for front page. Just descriptions should be enough

Great - sounds like a consensus :)
Rebased.

(I also just realized that the link to Funkwerk AG was wrong)

CyberShadow:
BTW, at certain widths the wrapping is strange:

https://dump.thecybershadow.net/ae9bb36f28364159ff3c95f033c0fc37/1465327407.png

I think you can put each pair in an `inline-block` block, that'll cause them to wrap in the middle but won't prevent them from wrapping into one element per line on narrow viewports.

wilzbach:
> I think you can put each pair in an inline-block block, that'll cause them to wrap in the middle but won't prevent them from wrapping into one element per line on narrow viewports.

Yes we could use `inline-block` elements,  to achieve something similar to, but then we would need more viewport css to toggle the `inline-block` off, s.t. we can put all four off them in one line. Does this really look better?

image

PR #1353, opened 2016-06-08
Orgs using D: Add putao

wilzbach:
@zoujiaqing and @dushibaiyu - based on the information I could find on the D wiki I created this entry. Please let me know what I should change. If you happen to have more links that connect your company with D (e.g. blog posts, conference videos, job offers, ...), I would be happy to link to them too ;-)

image

but it also looks nice with your full logo:

image


zoujiaqing:
Thanks @wilzbach , use the first style :)

PR #1354, opened 2016-06-08
Orgs using D: Add ALEX

wilzbach:
@irixi Please review and let me know what you would like to be changed. Especially for the description I couldn't find much.

As you can see on http://dlang.org/orgs-using-d.html, it's also nice to give a reason why you decided to D or why you like it.

Preview:

image


irixi:
@wilzbach cheers, please use our logo as shown on https://twitter.com/arexeu as the avatar, as we never use the origami on its own. The logo comprised both the origami and the text.

The description could be a quote "D is a superb language for doing high volume and low latency order matching."

wilzbach:
> the logo comprised both the origami and the text.

Oh sorry (I removed it because there was ALEX written twice)

> The description could be a quote "D is a superb language for doing high volume and low latency order matching."

Sounds great! It's a bit too large for our current system, 

image

@CyberShadow @aG0aep6G - I could increase the width from `250px` to `280px`, then it would like this:

image

Better ideas?

aG0aep6G:
> @CyberShadow @aG0aep6G - I could increase the width from 250px to 280px, then it would like this:
[...]
> Better ideas?

Both images look fine to me. I don't think it's too jarring when a quote needs a line more. The slightly wider boxes would be ok, too.

PR #1358, opened 2016-06-08
Fix issue 16141 - Organizations page should have a menu item

wilzbach:
Thanks @qznc for pointing this out. We have to be a bit careful about not adding too much to the menu, but I recently purged to items from the menu, so I guess we are good.
Moreover I took the liberty to sort this alphabetically (it's easier to parse for the human eye).

Btw there are many other great pages that are currently hard to find, see e.g. http://dlang.org/sitemap.html

PR #1359, opened 2016-06-08
Menu: Sort resources alphabetically

wilzbach:
follow-up to #1358

- I removed "The" because it seemed rather foreign to the prevalent style (e.g. it isn't"The Sitemap") and it destroy the point of sorting
- there was a bit of grouping, but this shows us that "Acknowledgment", "Sitemap" don't fit entirely into this menu?

### DIFF
--- a/dlang.org.ddoc
+++ b/dlang.org.ddoc
@@ -242,17 +242,17 @@ $(SUBMENU
     $(ROOT_DIR)articles.html, Articles
 )
 NAVIGATION_RESOURCES=
-$(SUBMENU
-    $(ROOT_DIR)tools.html, D-Specific Tools,
-    $(VISUALD), Visual D,
-    https://wiki.dlang.org/Editors, Editors,
-    https://wiki.dlang.org/IDEs, IDEs,
-    https://wiki.dlang.org/Tutorials, Tutorials,
-    https://wiki.dlang.org/Books, Books,
-    $(ROOT_DIR)dstyle.html, The D Style,
-    $(ROOT_DIR)glossary.html, Glossary,
-    $(ROOT_DIR)acknowledgements.html, Acknowledgments,
-    $(ROOT_DIR)sitemap.html, Sitemap
+$(SUBMENU_MANUAL
+    $(SUBMENU_LINK https://wiki.dlang.org/Books, Books)
+    $(SUBMENU_LINK https://wiki.dlang.org/Tutorials, Tutorials)
+    $(SUBMENU_LINK_DIVIDER $(ROOT_DIR)tools.html, D-Specific Tools)
+    $(SUBMENU_LINK https://wiki.dlang.org/Editors, Editors)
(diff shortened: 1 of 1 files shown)


dlang.org.ddoc
```diff
@@ -242,16 +242,16 @@ $(SUBMENU
 )
 NAVIGATION_RESOURCES=
 $(SUBMENU
+    $(ROOT_DIR)acknowledgements.html, Acknowledgments,
+    https://wiki.dlang.org/Books, Books,
     $(ROOT_DIR)tools.html, D-Specific Tools,
```
wilzbach: I would just write Tools, we don't write "Books about D" or "Tutorials about D" either

CyberShadow:
> there was a bit of grouping, but this shows us that "Acknowledgment", "Sitemap" don't fit entirely into this menu?

Yep, "Resources" is essentially a wild card name for "everything else" ;)

CyberShadow:
Not sure about this. Maybe IDEs and editors should be next to each other.

dlang.org.ddoc
aG0aep6G: > I would just write Tools, we don't write "Books about D" or "Tutorials about D" either

The idea is that these tools were written specifically for the D language. They're not general purpose tools with D support. The latter could be expected with just "Tools" as a title, and then we disappoint people with just these four weird little programs.

CyberShadow: That page needs to be dealt with, I'm just not clear on how.

wilzbach:
> Not sure about this. Maybe IDEs and editors should be next to each other.

Maybe we can have horizontal dividers if we want to continue with the grouping?

dlang.org.ddoc
wilzbach: The problem is that dfix and dfmt aren't official tools (yet) and htod has to be downloaded separately. Some ideas that come to my mind:
1) We move the rdmd page over to the "DMD Manual" page and deprecate in favor of the wiki (the htod.html page looks very nice though)
2) We start to extend and update the tools page

aG0aep6G: > That page needs to be dealt with, I'm just not clear on how.

Maybe we should join it with Editors and IDEs (and build tools and debuggers etc). Would add some meat to the page, and slim down the menu. Having links to the Wiki directly in the Menu is a bit confusing anyway.

CyberShadow:
Sure why not.

dlang.org.ddoc
CyberShadow: Probably a better Wiki page would be https://wiki.dlang.org/Development_tools .

Anyway, the other question is that the pages for rdmd and htod should still be reachable on dlang.org (without having to go through the wiki).

wilzbach:
Added the same dividers to resources - I grouped them into three categories:

- education materials 
- tools
- other

Within the group, they are sorted alphabetically.

image

PR #1360, opened 2016-06-08
index.dd: Fix stray )

CyberShadow:


wilzbach:
As the one who messed up the "Lispian" system, it obviously LGTM ;-)
Sorry!

PR #1365, opened 2016-06-10
Rename tools to CLI-reference and add DMD

wilzbach:
As said, having a CLI-reference page that includes the documentation for all shipped CLI tools makes sense - at least instead of linking to "DMD for Windows" by default.

The renamed CLI-reference page now includes all four versions of the DMD compiler :)

image

(first commit is separate, s.t. the git recognizes the commit and the git history is preserved)


CyberShadow:
I don't think the rename is necessary TBH

wilzbach:
> I don't think the rename is necessary TBH

My motivation was that if I click on "Command-like reference" it's a bit awkward when I end up at "tools" ;-)

CyberShadow:
IDK, I don't see anything wrong that a page titled "Command-like reference" to have the URL "tools.html"

wilzbach:
> IDK, I don't see anything wrong that a page titled "Command-like reference" to have the URL "tools.html"

Alrighty - rebased without the renaming.

dlang.org.ddoc
```diff
@@ -438,8 +438,12 @@ _=
...
 $(SUBNAV_TEMPLATE
-    $(SUBNAV_HEAD D-Specific Tools, $(ROOT_DIR)tools.html, overview)
+    $(SUBNAV_HEAD Command-line Reference, $(ROOT_DIR)tools.html, overview)
     $(UL $(SUBMENU2
+        $(ROOT_DIR)dmd-windows.html, dmd (Windows),
+        $(ROOT_DIR)dmd-linux.html, dmd (Linux),
+        $(ROOT_DIR)dmd-osx.html, dmd (OSX),
+        $(ROOT_DIR)dmd-freebsd.html, dmd (FreeBSD),
         https://github.com/Hackerpilot/dfix, dfix,
         https://github.com/Hackerpilot/dfmt, dfmt,
```
CyberShadow: How about replacing these with a link to the tools page on the wiki, at the bottom of the list?

dlang.org.ddoc
```diff
@@ -237,16 +237,15 @@ $(SUBMENU_MANUAL
...
     $(SUBMENU_LINK_DIVIDER $(ROOT_DIR)comparison.html, Feature Overview)
     $(SUBMENU_LINK $(ROOT_DIR)articles.html, Articles)
 )
 NAVIGATION_RESOURCES=
 $(SUBMENU_MANUAL
     $(SUBMENU_LINK https://wiki.dlang.org/Books, Books)
     $(SUBMENU_LINK https://wiki.dlang.org/Tutorials, Tutorials)
-    $(SUBMENU_LINK_DIVIDER $(ROOT_DIR)tools.html, D-Specific Tools)
-    $(SUBMENU_LINK https://wiki.dlang.org/Editors, Editors)
+    $(SUBMENU_LINK_DIVIDER https://wiki.dlang.org/Editors, Editors)
```
wilzbach: should we keep the D Tools link?

CyberShadow: Hmm, either kill it or link to the wiki page. I guess wouldn't hurt making the wiki page more discoverable

wilzbach:
> How about replacing these with a link to the tools page on the wiki, at the bottom of the list?

Ok as discussed the PR is now changed too:

- remove `tools.html` in favor of the wiki (adds redirect)
- add (sub)navigation to the dmd CLI pages
- add OS detection for CLI pages with `.htaccess` (soft-fail to windows)
(I tested it on an Apache instance of mine, but you might want to give it a try to be sure)

dlang.org.ddoc
wilzbach: Hmm after having had a look at the wiki and https://github.com/zhaopuming/awesome-d, I am not so sure how about to date the wiki is.

dlang.org.ddoc
```diff
@@ -436,12 +435,14 @@ SUBNAV_HEAD=
 SUBNAV_TEMPLATE=$(DIVC subnav-helper) $(DIVC subnav, $0)
 _=
 
-SUBNAV_TOOLS=
+SUBNAV_CLI_REFERENCE=
 $(SUBNAV_TEMPLATE
-    $(SUBNAV_HEAD D-Specific Tools, $(ROOT_DIR)tools.html, overview)
+    $(SUBNAV_HEAD Command-line Reference, http://wiki.dlang.org/Development_tools, overview)
```
CyberShadow: Wiki page is less of an overview and more of a "see more" thing. Perhaps it could be moved at the bottom as a "more tools" link OSLT?

htod.dd
```diff
@@ -230,4 +230,4 @@ $(OL
 
 Macros:
 	TITLE=htod
-	SUBNAV=$(SUBNAV_TOOLS)
+	SUBNAV=$(SUBNAV_CLI_REFERENCE)
```
CyberShadow: This rename seems gratuitous but whatever

CyberShadow:
LGTM sans nits, merge or do you want to fix them?

dlang.org.ddoc
CyberShadow: Then we need to update the wiki! It's what it's there for.

dlang.org.ddoc
wilzbach: The subnav macro is rather strict, I would need to break it and I am not sure whether that's worth it, e.g. the macro ensures that the submenu is responsive too.

image

CyberShadow: OK

wilzbach:
> LGTM sans nits, merge or do you want to fix them?

Addressed ;-)

PR #1366, opened 2016-06-11
Remove 'New Library Reference' from menu

wilzbach:
Imho it's very confusing to have "Library Reference" and Library Reference Preview" - how about linking it directly from the docs?

stable:

image

pre-release:

image

(for pre-release I thought people are familar with D and know about the difference between ddoc and ddox)

### DIFF
--- a/std_navbar-prerelease.ddoc
+++ b/std_navbar-prerelease.ddoc
@@ -3,7 +3,10 @@ $(SUBNAV_TEMPLATE
     $(DIVC head,
         $(H2 Library Reference)
         $(P $(SPANC smallprint, pre-release version $(SPANC separator, $(BR))
-            $(LINK2 ../phobos/index.html, switch to version $(LATEST))))
+            switch to $(LINK2 ../phobos/index.html, $(LATEST)). $(BR)
+            $(LINK2 ../library-prerelease/index.html, preview) beta library reference.
+            )
+        )
         $(P $(LINK2 index.html, overview))
     )
     $(UL $(MODULE_MENU))
--- a/dlang.org.ddoc
+++ b/dlang.org.ddoc
@@ -236,7 +236,6 @@ NAVIGATION_DOCUMENTATION=
 $(SUBMENU_MANUAL
     $(SUBMENU_LINK $(ROOT_DIR)spec/spec.html, Language Reference)
     $(SUBMENU_LINK $(ROOT_DIR)phobos/index.html, Library Reference)
-    $(SUBMENU_LINK $(ROOT_DIR)library/index.html, Library Reference Preview)
     $(SUBMENU_LINK $(ROOT_DIR)dmd.html, Command-line Reference)
     $(SUBMENU_LINK_DIVIDER $(ROOT_DIR)comparison.html, Feature Overview)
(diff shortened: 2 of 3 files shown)


std_navbar-prerelease.ddoc
```diff
@@ -3,7 +3,11 @@ $(SUBNAV_TEMPLATE
     $(DIVC head,
         $(H2 Library Reference)
         $(P $(SPANC smallprint, pre-release version $(SPANC separator, $(BR))
-            $(LINK2 ../phobos/index.html, switch to version $(LATEST))))
+            switch to $(LATEST):
+            $(LINK2 ../phobos/index.html, ddoc) or
+            $(LINK2 ../library/index.html, ddox).
```
CyberShadow: Why not the same text as below?

wilzbach:
> Why not the same text as on release?

like this? (rebased)

image

(it links to library-prerelease)

CyberShadow:
"preview the new doc engine" sounds weird but I can't think of an obvious improvement :/

"doc engine" is something internal, and the user won't be looking at the engine itself but its results.

Maybe "preview beta library reference"?

wilzbach:
> Maybe "preview beta library reference"?

renamed & rebased.

andralex:
lgtm. @CyberShadow ?

MartinNowak:
Any idea to make the existence of pre-release docs configurable, e.g. by wrapping those links in macro.
I need to upstream the dlang.org patches for release building, b/c changes here regularly break the nightly builds.
https://github.com/dlang/installer/pull/180

We'll also have to talk about ddox for the release docs (they're currently not part and they're quite big as well). Could someone get in contact with me to discuss the schedule for that?

CyberShadow:
> We'll also have to talk about ddox for the release docs (they're currently not part and they're quite big as well).

I suggest to delay that for now, i.e. keep shipping DDoc Phobos docs with DMD, even if we won't show them on the website.

CyberShadow:
> Any idea to make the existence of pre-release docs configurable, e.g. by wrapping those links in macro.

Yes.

For an example, see the `NODATETIME` `posix.mak` parameter. You can specify `NODATETIME=nodatetime.ddoc` to remove the DDoc timestamps, which makes diffs reasonable. This can be done in the same way, I think.

s-ludwig:
Just noting that we are now back to having the ddox based docs quasi invisible. I can't see how they will ever replace the ddoc ones like this, considering that nobody will use them and no issues will be found consequently.

wilzbach:
Huh? This change now displays a link to ddox on every phobos documentation page. I thought that's an improvement?

s-ludwig:
Hm, not really AFAICS. You now have to click through an additional link, which is tiny and gray-on-gray and in a place that is not commonly used at all. I actually overlooked the link when I was briefly looking for it there and had to search for the PR to find it. Maybe that's just because I was expecting only the "pre-release" link to be there, but it's definitely a lot less prominent now.

wilzbach:
> but it's definitely a lot less prominent now.

That really wasn't the intention, sorry. Do you want to add it back to the menu or have other ideas to increase exposure before 2.072?

CyberShadow:
Perhaps we could upgrade it into a nice (not too distracting) banner at the top?

PR #1368, opened 2016-06-11
remove moved pages

wilzbach:
I assume that those pages come from a time in which we didn't had a `.htaccess` redirect file?

Is there any reason to keep them?


CyberShadow:
Yes, and actually we should also have `dmd.html` etc.

Apache is an implementation detail of `dlang.org`, most people don't test their changes on a live Apache instance - so these pages are still useful for testing.

wilzbach:
> Yes, and actually we should also have dmd.html etc.

My PR initially had that :/
Ok I will submit this file soon.

> Apache is an implementation detail of dlang.org, most people don't test their changes on a live Apache instance - so these pages are still useful for testing.

Yes, but as far as I understood are those pages moved and the .htaccess links are only for compatibility with old links with dlang.org out there somewhere in the wild. So I don't understand why those empty pages are useful for testing.

CyberShadow:
If there are no internal links to these pages (incl. from Phobos), then I guess they can be removed.

comparison.dd
```diff
@@ -77,7 +77,7 @@ $(ITEMIZE
 		$(A spec/unittest.html, Unit testing),
 		$(A spec/module.html#staticorder, Static construction order),
 		$(A spec/statement.html#DeclarationStatement, Guaranteed initialization),
-		$(A memory.html#raii, RAII (automatic destructors)),
+		$(A https://wiki.dlang.org/Memory_Management#RAII_.28Resource_Acquisition_Is_Initialization.29, RAII (automatic destructors)),
```
wilzbach: I am not a huge friend of how media wiki create its headings ids. Is there a way to get a nicer URL like https://wiki.dlang.org/Memory_Management#RAII ?

spec/garbage.dd
```diff
@@ -7,7 +7,7 @@ $(SPEC_S Garbage Collection,
 	to free memory explicitly. Just allocate as needed, and the garbage collector will
 	periodically return all unused memory to the pool of available memory.
 	)
-	
+
```
wilzbach: why don't we have a trailing whitespace checker for dlang.org? ;-)

wilzbach:
> If there are no internal links to these pages (incl. from Phobos), then I guess they can be removed.

I could only find three links to `memory.html` which I replaced ;-)

win32.mak
```diff
@@ -302,8 +300,6 @@ changelog\2.058.html : $(CHANGELOG_DDOC) changelog\2.058.dd
 	$(DMD) -o- -c -D -Df$*.html $(CHANGELOG_DDOC) $*.dd
 changelog\2.059.html : $(CHANGELOG_DDOC) changelog\2.059.dd
 	$(DMD) -o- -c -D -Df$*.html $(CHANGELOG_DDOC) $*.dd
-changelog\2.060.html : $(CHANGELOG_DDOC) changelog\2.060.dd
-	$(DMD) -o- -c -D -Df$*.html $(CHANGELOG_DDOC) $*.dd
```
CyberShadow: Um. What happened here?

wilzbach:
> Um. What happened here?

No idea. Fixed.

PR #1369, opened 2016-06-12
index.dd: Pre-populate subject line for "your code here" submissions

CyberShadow:
DFeed change: https://github.com/CyberShadow/DFeed/commit/c3199d6a01887974345d3f0f72eee1f982419e9c

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -19,7 +19,7 @@ $(DIVC intro, $(DIV, $(DIV,
     )
     $(DIVID your-code-here,
         $(DIVC tip,
-            $(LINK2 https://forum.dlang.org/newpost/general, your code here)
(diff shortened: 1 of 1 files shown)


ZombineDev:
LGTM

wilzbach:
LGTM^2

wilzbach:
Any reason why we haven't rolled this out yet? 
It's a simple change & if I could, I would merge it ;-)

andralex:
New code should use `HTTP` :o)

adamdruppe:
Please kill `HTTP` and `HTTPS`, they aren't worth the cache space in the brain compared to just pasting the link into a `LINK2` (especially since copy/past from the browser URL bar will likely add the `https://` automatically anyway, so to use `HTTPS`, you'd have to manually edit it after checking for 404's, which is liable to introduce broken link errors.)

PR #1370, opened 2016-06-14
pointer subtraction yields ptrdiff_t

WalterBright:
Document the type of pointer difference.


andralex:
Don't forget - in new code backquotes are preferred to `$(D ...)`

PR #1371, opened 2016-06-14
Add icons to 'Why D'

wilzbach:
FontAwesome is amazing, so how about using it at a very important part of our frontpage?

before:

image

after: 

image

(screenshots are taken at 50%)

### DIFF
--- a/areas-of-d-usage.dd
+++ b/areas-of-d-usage.dd
@@ -29,7 +29,7 @@ $(DIVC page-contents quickindex,
 )
 
 $(DIVC start-icons,
-    $(START_ICONS_ENTRY Industry, building, industry)
+    $(START_ICONS_ENTRY Industry, industry, industry)
     $(START_ICONS_ENTRY Systems Programming, gears, systems-programming)
     $(START_ICONS_ENTRY Research, flask, cutting-edge)
     $(START_ICONS_ENTRY Academia, graduation-cap, academia)
@@ -45,7 +45,7 @@ $(DIV,
 
 $(DIV style="clear:both")
 
-$(SECTION2 $(LNAME2 industry, Industry), $(SECTION_FA_ICON building)
+$(SECTION2 $(LNAME2 industry, Industry), $(SECTION_FA_ICON industry)
 
 $(DIVC area-section2,
     $(P
--- a/dlang.org.ddoc
+++ b/dlang.org.ddoc
@@ -94,7 +94,7 @@ $(DIVC container,
 )
 $(COMMON_SCRIPTS)
 $(LAYOUT_SUFFIX)
-<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.2.0/css/font-awesome.min.css">
+<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.4.0/css/font-awesome.min.css">
 </body>
 </html>
 _=
--- a/index.dd
+++ b/index.dd
@@ -156,7 +156,7 @@ $(DIVC boxes,
 $(DIVC whyd, $(SECTION2 Why D?,
 
 $(DIVC section,
-$(DIV, $(SECTION3 Convenience,
+$(DIV, $(SECTION3 $(WHY_D_ICON magic) Convenience,
 
 $(P D allows writing large code fragments without redundantly specifying types,
 like dynamic languages do. On the other hand, static inference deduces types and other
@@ -239,7 +239,7 @@ void main()
 
 )))
 
-$(DIVC section, $(DIV, $(SECTION3 Power,
+$(DIVC section, $(DIV, $(SECTION3 $(WHY_D_ICON rocket) Power,
 
 $(P The best paradigm is to not impose something at the expense of others.
 D offers classic polymorphism, value semantics, functional
@@ -309,7 +309,7 @@ Read more).)
 
 )))
 
-$(DIVC section, $(DIV, $(SECTION3 Efficiency,
+$(DIVC section, $(DIV, $(SECTION3 $(WHY_D_ICON bolt) Efficiency,
 
 $(P D compiles naturally to efficient native code.)
 
@@ -375,7 +375,7 @@ consistency. $(LINK2 spec/memory-safe-d.html, Read more).)
 
 )))
 
-$(DIVC section, $(DIV, $(SECTION3 Industry-proven,
+$(DIVC section, $(DIV, $(SECTION3 $(WHY_D_ICON industry) Industry-proven,
 $(DIVC frontpage-orgs center,
     $(FRONTPAGE_ORG Facebook, https://www.facebook.com, facebook.svg,
         Online social networking service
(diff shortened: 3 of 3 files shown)


CyberShadow:
@andralex

wilzbach:
I still don't like the fighter icon for power, possible ideas:

- http://fontawesome.io/icon/bomb/
- http://fontawesome.io/icon/fire/
- http://fontawesome.io/icon/flask/
- http://fontawesome.io/icon/heartbeat/
- http://fontawesome.io/icon/trophy/

@CyberShadow suggested to use the new `industry` icon (since FontAwesome 4.4) and then `rocket` for power:

image

adamdruppe:
On Tue, Jun 14, 2016 at 04:27:26PM -0700, Sebastian Wilzbach wrote:
> FontAwesome is amazing

FontAwesome is a piece of crap. Just make little inline SVGs of the icons you like and use them. Better in virtually every way.

CyberShadow:
How so Adam? Doing what you suggest will greatly increase the load time and resource usage.

adamdruppe:
On Tue, Jun 14, 2016 at 09:10:25PM -0700, Vladimir Panteleev wrote:
> How so Adam? Doing what you suggest will greatly increase the load time and resource usage.

Surely a tiny handful of inline icons is going to load faster than loading ~ 70 KB of css and font assets (that's after compression!). Even font awesome themselves brag that you can get a 10x faster download by just taking the icons you actually use.

dlang.org uses, what, 9 of them? That's not a lot to just inline (and it'd save two http requests). fontawesome comes with over 600.

Moreover, even if it were all equal, svg degrades more gracefully than littering the site with private use characters. If you have webfonts disabled/blocked, they appear as those ugly boxes. svg actually renders, or can use the standard alt tag (which should prolly just ="" for the icons since they are just decoration). Using `<i class="">` just disgusts me too, that's not what the tag means! <i> has meant some variant on italic text forever (html5 calls it "alternate voice or mood", there's no standard document that calls it an icon holder.

svg also avoids strange text artifacts like antialiasing which happens with fonts.

See this list for even more: https://css-tricks.com/icon-fonts-vs-svg/

Lastly, every lazy website uses these same boring stock icons. I'd prefer to do something new, though I recognize that actually is a lot of work. But even reusing their designs as images instead of hideous html is a big step up.

CyberShadow:
> See this list for even more: https://css-tricks.com/icon-fonts-vs-svg/

That seems ... extremely opinionated. You can easily find articles that praise either.

> Surely a tiny handful of inline icons is going to load faster than loading ~ 70 KB of css and font assets (that's after compression!).

Latency can impact page load time more than the amount of data. We also load it from a CDN - anyone who's been to GitHub before should already have it in their cache.

> Using `<i class="">` just disgusts me too, that's not what the tag means!

It also does not affect the experience of dlang.org visitors in any way.

> Lastly, every lazy website uses these same boring stock icons. I'd prefer to do something new, though I recognize that actually is a lot of work. 

Well, precisely.

> But even reusing their designs as images instead of hideous html is a big step up.

Are *you* willing to do that (and do it in such a way that makes adding/changing icons easy)? But even if so, it would be churn with a very debatable trade-off.

adamdruppe:
On Tue, Jun 14, 2016 at 09:52:12PM -0700, Vladimir Panteleev wrote:
> Latency can impact page load time more than the amount of data. We also load it from a CDN - anyone who's been to GitHub before should already have it in their cache.

Yes, but replacing it with the very same thing, just in 4 KB inline instead of 80 via two external http requests doesn't even need to be cached.


> It also does not affect the experience of dlang.org visitors in any way.

Unless the font is blocked, then they get a rectangle instead of the image or alt text.

> Are *you* willing to do that (and do it in such a way that makes adding/changing icons easy)? But even if so, it would be churn with a very debatable trade-off.

Yeah, it's trivial. svg versions of the font awesome icons are easy to get, all you do is copy/paste it in minimally. I'd ideally prefer to do it as a css background-image though, which means fixing the semantic html, but it still isn't much... I'd make the TOUR thing take a descriptive class name instead of an icon name, then move the icon definition itself to the stylesheet.

To change icons in the future, you'd just paste a new one into the stylesheet.

wilzbach:
> Even font awesome themselves brag that you can get a 10x faster download by just taking the icons you actually use.

The usual practice is to create a special font via fontello or other tools, but then you loose the advantage of the having a highly-likely already cached font.

> Yes, but replacing it with the very same thing, just in 4 KB inline instead of 80 via two external http requests doesn't even need to be cached.

To be fair it's 65 KB as of 4.2, but I do get that point of inlining.

> Yeah, it's trivial.

Then I am waiting for your PR ;-)

---------------------------------------------

In any case: do we all agree that icons look nice in general? Are the ones I picked for categories matching?

adamdruppe:
On Wed, Jun 15, 2016 at 12:53:00PM -0700, Sebastian Wilzbach wrote:
> The usual practice is to create a special font via fontello or other tools, but then you loose the advantage of the having a highly-likely already cached font.

That's harder than just copy/pasting a few hundred bytes of svg!

> Then I am waiting for your PR ;-)

I'll wait until this is merged and your decisions made then do it separately, as there's existing icons on the page that can be transitioned too.

andralex:
If @CyberShadow is for it I'm for it.  `preapproved = approvedBy("@Cybershadow");`

wilzbach:
Upgraded to FontAwesome 4.4 (needed for the `industry` icon) and also applied it to the areas page.

CyberShadow:
> If @CyberShadow is for it I'm for it. preapproved = approvedBy("@Cybershadow");

I haven't actually approved adding the icons to the front page... After thinking about it, well I'm not sure.

Icons + text is a nice combination that works in a lot of cases, but perhaps we're starting to do it too much. At a certain point it passes some threshold and instead of adding visual information it becomes clutter, and becomes the one of the central elements of the page (i.e. one of the things that jumps at you when you open the page shouldn't be "Look! We're using Font Awesome everywhere we can!"). Some of the icon choices also seem forced (convenience/power).

So, I'm on the fence, and could use more opinions on this.

ZombineDev:
FWIW, I like the front page more with icons (Power - `rocket`, Efficiency - `bolt`, Industry-proven - `industry`). Icons are definitely something that captures visitors attention and makes the wall of information a bit more interesting.
The only thing I would change is the icon for Convenience. Here are a couple of alternatives: 
[diamond](http://fontawesome.io/icon/diamond/), [leaf](http://fontawesome.io/icon/leaf/), [magic](http://fontawesome.io/icon/magic/) and [paper-plane-o](http://fontawesome.io/icon/paper-plane-o/).

Python's [homepage](https://www.python.org) also uses icons extensively and I don't think it's too much.

CyberShadow:
`fa-magic` for convenience looks just right!

wilzbach:
> fa-magic for convenience looks just right!

Agreed - rebased  with that ;-)

image

CyberShadow:
Is that a screenshot for ants? :)

GitHub will automatically size images to fit into the comment's width, so there's no need to pre-resize screenshots.

CyberShadow:
Oops, please rebase again :)

wilzbach:
> Is that a screenshot for ants? :)

It's at 50% - I am currently on a laptop :/

> Anyway, yeah, doc autotester.

I thought it's nice if the image is directly included in the mail, saves a bit of time ;-)

> Oops, please rebase again :)

Should be green now.

CyberShadow:
> It's at 50% - I am currently on a laptop :/

IDK about Chrome but in Firefox you can take a screenshot of an arbitrary viewport by going into mobile development mode (Ctrl+Shift+M), resizing the viewport, and clicking on the camera button.

PR #1372, opened 2016-06-14
use fork icon instead of 'wrench' for contribute

wilzbach:
As I just studied FontAwesome a bit closer the 'wrench' icon for "Contribute" seemed a bit unlogical, how about using the fork icon?

before:

image

after:

image

Alternatives could be:

http://fontawesome.io/icon/github/ (it's a brand)
http://fontawesome.io/icon/bug/ (potentially bad, it shows that we have 'bugs')

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -138,7 +138,7 @@ $(DIVC boxes,
         )
     )
     $(DIVC row,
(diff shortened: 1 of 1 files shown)


JackStouffer:
👍

PR #1373, opened 2016-06-15
Fix issue 16152 - Fix merging of eponymous template doc comments.

s-ludwig:
This change avoids discarding the doc comment of either the parent template or the template member when merging eponymous templates to a single declaration.

### DIFF
--- a/dpl-docs/dub.selections.json
+++ b/dpl-docs/dub.selections.json
@@ -1,15 +1,13 @@
 {
 	"fileVersion": 1,
 	"versions": {
-		"backtrace-d": "~master",
-		"ddox": "0.15.6",
+		"ddox": "0.15.10",
 		"experimental_allocator": "2.70.0-b1",
 		"hyphenate": "1.1.1",
 		"libasync": "0.7.9",
 		"libdparse": "0.6.0",
 		"libev": "5.0.0+4.04",
 		"libevent": "2.0.1+2.0.16",
-		"libhttp2": "0.2.6",
 		"memutils": "0.4.5",
 		"openssl": "1.1.4+1.0.1g",
 		"vibe-d": "0.7.28"


s-ludwig:
Looks like something has gone wrong and parts of the documentation have actually been removed. Will look into that.

s-ludwig:
BTW, Ddoc also doesn't like the `Duration.split` documentation: http://dlang.org/phobos/core_time.html#.Duration.split (there is a "Ditto" text at the end)

s-ludwig:
Looks good now.

CyberShadow:
http://dtest.thecybershadow.net/artifact/website-60aaec239cc320e3d7424b1be63f0a5fac9502d9-75ae428677bcdd0e058d82547d65d8b1/web/library-prerelease/std/algorithm/iteration/cumulative_fold.html now contains the same content twice, or am I wrong?

s-ludwig:
One is the seed version and one the no-seed version: https://github.com/dlang/phobos/blob/c5bb43ce54d7feb244ba032c43aecd7aa7c9356a/std/algorithm/iteration.d#L3107

I'm currently appending the outer and the inner doc comment, because that's what Ddoc seems to do in the vanilla case. However, in this case, the eponymous members are documented as separate child entities for /phobos/. Maybe the simplest solution would be to just skip the eponymous merge whenever both, the parent and the child, are documented?

CyberShadow:
Um, I guess? I don't really understand the problem, I'm clicking around in the diff and all I see is mountains of duplicated content. As far as I see this change is purely a regression... What did I miss?

CyberShadow:
OK, I see now that it fixes `std.parallelism.TaskPool.reduce` documentation. But the gratuitous duplication of information elsewhere does seem like a regression in trade-off.

s-ludwig:
No, it actually just adds content that was missing before, just the way it's displayed is sub-optimal.

CyberShadow:
OK, so do we merge this?

s-ludwig:
Let me try how skipping the eponymous merge looks first. Maybe that's in fact a better idea.

s-ludwig:
Okay, that works better now. I've also had to fix avoiding self-cross-references. Strict self-references were already avoided, but sometimes an overload with the same name of the documented symbols still got a cross-link.

CyberShadow:
Hmm, lots of new pages. For example, why is this now a page of its own?

http://dtest.thecybershadow.net/artifact/website-d045da10bd149e09df9a78dd17d9bfa244582baf-f43d156587500c354d7041be5c603b59/web/library-prerelease/std/numeric/custom_float.to_string.html

It is a separate eponymous template, but the inner symbol is not documented.

s-ludwig:
That specific case is a bit awkward. Not sure if it makes sense to implement special case handling for it or to fix the underlying issue (https://issues.dlang.org/show_bug.cgi?id=13737). The reason it's put on a separate page is that inside the `toString` template, there is another `toString` template, which contains the actual function. If it weren't documented, that information would generally be lost.

CyberShadow:
OK, but this seems to be very far from an isolated case. I see a lot of new pages in the output diff.

s-ludwig:
There are many valid cases, where the inner template has additional template parameters or constraints. They are usually implicit, but hiding them doesn't seem like a good idea.

CyberShadow:
So do we merge this? :)

s-ludwig:
As far as I can see, I'd say yes. A possibility to make this a bit prettier would be to embed undocumented child entries on the same page, but the changes as they are now are definitely an improvement, so going step-by-step makes sense. I'll open a ticket against DDOX.

PR #1374, opened 2016-06-15
fix Issue 16167 - chm-nav.json generation is broken

MartinNowak:


MartinNowak:
Testing the fix currently.

MartinNowak:
Retry

We really should waste less time on such things as generating json using ddoc.

CyberShadow:
Sorry about that :)

MartinNowak:
> Sorry about that :)

Not your fault, but maybe we can generate chm from the html output directly? Or just ditch it?
It's 2016, who needs installed websites anyhow?

CyberShadow:
> Not your fault, but maybe we can generate chm from the html output directly? 

It's needed to generate the navigation, which is metadata.

> Or just ditch it?

Every time we ask, a few people raise their hands. The bigger advantages are the search and index. Maybe DDox will make it less necessary.

MartinNowak:
> Every time we ask, a few people raise their hands.

Sure, people don't want to change their habits, but we could do it anyhow.
If search in chm works better than on our website, than we need to fix the latter, let's ask again.

MartinNowak:
Seems to work now.

PR #1375, opened 2016-06-15
Fix broken link to rdmd

wilzbach:
It seems that we broke the link to http://dlang.org/rdmd.html, should be fixed with this.


CyberShadow:
So do we need the leading `/`? Maybe what it actually needs is '^'?

PR #1376, opened 2016-06-15
[don't merge before 2.072] Remove last standing WEB macro

wilzbach:
removes the last bit of the `WEB` macro - shouldn't be pulled before 2.072


andralex:
yay 2 dat

andralex:
please resubmit... sorry

PR #1377, opened 2016-06-17
add chm-nav.json target to posix.mak

MartinNowak:
- to check json validity @Cybershadow


MartinNowak:
Seems there is enough interest to keep them, can we integrate checking the chm-nav.json into your tester?

CyberShadow:
Thanks. This PR looks good, but auto-tester integration may need to wait a bit. The state of things is a bit half-way between this and the new thing, and now the existing one doesn't even build any more. Will try to look into this soon; even if not we'll definitely get this when the new thing is ready.

BTW, chmgen does a few more things than CHM generation, such as checking for broken internal links and generating the tags for `dman`.

PR #1382, opened 2016-06-20
fix issue 16186 - [Programming in D for C Programmers] Backticks shou…

aG0aep6G:
…ld be escaped in explanation of raw string syntax

Also changing $(B ...) with backticks for proper code styling.

https://issues.dlang.org/show_bug.cgi?id=16186


JackStouffer:
LGTM

PR #1384, opened 2016-06-21
fix anchor thingies on phobos pages

aG0aep6G:
I guess I've broken them with #1338.


wilzbach:
LGTM - Thanks for fixing :)

PR #1381, opened 2016-06-20
Orgs using D: Add Ecratum

llaine:
Hello,
As the http://dlang.org/orgs-using-d.html said I would like to add my company to website.
We are located in Berlin and we are using Dlang to run our internal web apps.

PR #1386, opened 2016-06-23
Provide infos how to add an org using d

wilzbach:
Inspired by @llaine and #1381, I thought it's worth adding easily accessible information for future org additions.

### DIFF
--- a/orgs-using-d.dd
+++ b/orgs-using-d.dd
@@ -1,15 +1,17 @@
 Ddoc
 
-$(D_S $(TITLE),
-
 $(COMMENT
     - Put your logo (svg or transparent png) to images/orgs-using-d
-    - At the moment, we use 150x75 as maximal dimensions for the displayed logos
-       (resize with e.g. `mogrify -resize 150x75 logo.png`)
-    - If you have a better quality logo, please name it `<org-name>_hq.<svg|png>`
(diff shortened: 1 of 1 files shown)


CyberShadow:
Why are we putting documentation for *one single page* into the file documenting the entire project? Are we going to start adding documentation on how to update each dlang.org page here next?

wilzbach:
> Why are we putting documentation for one single page into the file documenting the entire project? Are we going to start adding documentation on how to update each dlang.org page here next?

Fair enough - I added a couple of spaces and moved it before the title, s.t. it hopefully doesn't get ignored the next time. Should I remove the README too?

CyberShadow:
Yes.

You're taking the wrong approach to this problem.

What part of the text are potential contributors most likely to immediately start editing?

The end of the list of organizations.

That's where you need to add the explanatory text. So, either put the instructions there, or add a comment pointing them towards the instructions at the top of the file.

wilzbach:
> You're taking the wrong approach to this problem.
>
> What part of the text are potential contributors most likely to immediately start editing?
>
> The end of the list of organizations.
> That's where you need to add the explanatory text. So, either put the instructions there, or add a comment pointing them towards the instructions at the top of the file.

Alrighty - I put a pointer in there ;-)

PR #1387, opened 2016-06-23
Add D-man in rain to SafeD

wilzbach:
Found this lovely D-man drawing on [Twitter](https://twitter.com/__DaLong/status/746113909682110464)

image

@CyberShadow I used the 600x600 original image. It's just 30K - should I still save a resized version?


safed.dd
```diff
@@ -2,56 +2,58 @@ Ddoc
...
 
+<img src="images/dman-rain.jpg" border=0 alt="D-man in rain" align=right alt="D Style" height=200>
+
 	$(P
 	$(SMALL by Bartosz Milewski, a member of the D design team)
 	)
 
 	$(P
-	I've seen some very good programmers move away from C++ in favor of languages like Java or C#. Being a hard-core C++ programmer myself, I wondered why anyone would want to switch to a less powerful and less efficient language. Mind you, I could understand why a newcomer would opt for a simpler, flatter-learning-curve language but, once somebody invested the time and effort to become proficient in C++, why in the world would they want to abandon it? 
+	I've seen some very good programmers move away from C++ in favor of languages like Java or C#. Being a hard-core C++ programmer myself, I wondered why anyone would want to switch to a less powerful and less efficient language. Mind you, I could understand why a newcomer would opt for a simpler, flatter-learning-curve language but, once somebody invested the time and effort to become proficient in C++, why in the world would they want to abandon it?
```
wilzbach: seems like my VIM automatically removed the trailing whitespace. I guess that's a relict from old times. Should I remove this from this the PR?

majiang:
The author of the image, meiz_sandwich at twitter, allows the use of his drawing in dlang.org, declared in this tweet (he doesn't speak English):
https://twitter.com/meiz_sandwich/status/746121887575531522

ZombineDev:
If your worried about the image size, you may try to convert the image PNG, to see if it will make a difference. You may also try to make its background transparent.

wilzbach:
> If your worried about the image size, you may try to convert the image PNG, to see if it will make a difference. You may also try to make its background transparent.

Oh I already did. It goes up from 30K to 140K and removing the background only saves about 14K.
My question was more as "I know it's usually not a good practice to send higher than needed images to the client, but as long as we don't have an automatic resizing in the build chain and it's the sole image on the page, the 18K overhead don't matter, right?"

CyberShadow:
Converting from JPEG to PNG will never save space because then you're meticulously losslessly compressing every single JPEG compression artifact. Same as transcoding MP3 to FLAC.

safed.dd
CyberShadow: Yes please. If it was just a few lines... but this is too much noise to mix changes in one commit.

IDK about Vim but Emacs has a mode which only fixes whitespace in lines you edit.

wilzbach:
> Yes please. If it was just a few lines... but this is too much noise to mix changes in one commit.

Fair enough - fixed.
I wasn't sure how the exact policy is. You might consider making one noise commit to remove all the trailing whitespace like in the other D repositories.

CyberShadow:
Sure, feel free to.

WalterBright:
Need permission from the artist. (I like the drawing!)

wilzbach:
@WalterBright - see above:

> The author of the image, meiz_sandwich at twitter, allows the use of his drawing in dlang.org, declared in this tweet (he doesn't speak English):
> https://twitter.com/meiz_sandwich/status/746121887575531522

ZombineDev:
@CyberShadow JPG vs PNG depends on the image contents - for photos jpg is definitely the undisputed winner, however png is better for simple images with a limited number of colors such text and sometimes web comics. http://stackoverflow.com/questions/610530/what-image-type-should-i-use-when-gif-jpg-or-png

@wilzbach did you try saving the images as png-8, instead of png-24/32?

CyberShadow:
@ZombineDev How should I say this politely... I know this as well today as I knew it 15 years ago. Maybe you misunderstood what I meant or maybe I didn't express myself clearly enough, but I suggest you do a bit more research on the subject. You should never recompress JPEG as PNG.

> @wilzbach did you try saving the images as png-8, instead of png-24/32?

Might as well convert to GIF. Seriously, if your compression strategy is to restrict the palette to 256 colors, you might just as well recompress the JPEG with a higher ratio.

majiang:
@ZombineDev I ask him to (he doesn't have an GitHub account).

wilzbach:
> @wilzbach did you try saving the images as png-8, instead of png-24/32?

Nope :S - that reduces it to 57K - with transparency to 36K.

> Might as well convert to GIF. Seriously, if your compression strategy is to restrict the palette to 256 colors, you might just as well recompress the JPEG with a higher ratio.

... this wasn't my question - it was solely about whether I should add a thumbnail image too, or serving the 600x600 image to the user is OK (in my eyes it is).

ZombineDev:
I got it to 27KB: dman-rain
but @CyberShadow is right, it's not worth bothering, because JPG could easily win after a slight quality reduction and will probably provide better results in general, because there more gradients in those images than I thought at first.

majiang:
The author meiz_sandwich says that he hasn't saved original psd image (since he was drawing them just for fun) and sent @wilzbach the only jpg image(s). IMO preserving the original image is more important than the formatting issue or size difference of a few kilobytes.

ZombineDev:
@majiang I agree

WalterBright:
This one is my favorite. Will merge once it is rebased.

wilzbach:
> This one is my favorite. Will merge once it is rebased.

Rebased ;-)
It's a bit annoying because they all touch the same line in `posix.mak`.

PR #1391, opened 2016-06-24
Add twisted D-man to floating point spec

wilzbach:

image

I think this fits well to the "twisted" behavior of floating points (you can also see the D-Man as a 8), but I am open for better suggestions ;-)


WalterBright:
I don't think it works.

wilzbach:
> I don't think it works.

If someone has a better use for this graphic, please go ahead ;-)

PR #1393, opened 2016-06-24
Add D-Man looking at other languages to interface with C,C++,ObjC

wilzbach:
- I added it to C,C++,ObjC - probably one or two is enough
- We could add their custom language symbol at which the D-Man is looking at

image


CyberShadow:
This seems a little forced.

majiang:
IMO this picture, D-man with an arrow, is more suitable for Vector Extensions page.

wilzbach:
> IMO this picture, D-man with an arrow, is more suitable for Vector Extensions page.

good idea, @WalterBright what do you think?

wilzbach:
> IMO this picture, D-man with an arrow, is more suitable for Vector Extensions page.

ping @WalterBright for feedback

AndrejMitrovic:
It would be cool if there was one holding hands with C, like they're friends.

wilzbach:
> It would be cool if there was one holding hands with C, like they're friends.

@AndrejMitrovic to clarify: I am an horrible artist and all these drawings come from meiz_sandwich. I merely pushed his lovely [drawings](https://github.com/wilzbach/d-mans) to the general public ;-)

So unfortunately I can't produce new drawings and we have to decide between using it for one of the proposed pages (e.g. Vector extensions)  or closing this proposal.

wilzbach:
Closing this as I lost interest in pushing this. In case someone wants to pursue this, I started a repo with D-mans collected from all over the world: https://github.com/wilzbach/d-mans

PR #1394, opened 2016-06-24
Add winning D-Man to conditional compilation

wilzbach:

My motivation here is that CTFE is one of the features that make D a _winner_. Otherwise templates could also be used.

image


CyberShadow:
This seems a little forced

WalterBright:
While a great drawing, I don't think it matches the theme of conditional compilation.

wilzbach:
> While a great drawing, I don't think it matches the theme of conditional compilation.

Fair enough, how about
- [Why D?](http://dlang.org/overview.html) (moving the menu to the left)
- [Builtin rationale](http://dlang.org/builtin.html)
- [Feature overview](http://dlang.org/comparison.html)
- [Lazy evaluation](https://dlang.org/lazy-evaluation.html)

wilzbach:
In any case if someone has a good use for this graphic, please go ahead ;-)

PR #1395, opened 2016-06-24
Remove trailing whitespace in Ddoc pages + add check

wilzbach:
A simple `regex` ;-)

```
sed -i 's/[ \t]*$//' **/*.dd
```

To prevent this from happening again I added a Makefile target `test` that returns an error code if it can find trailing whitespace. Is it possible to execute this command with DAutoTest?


CyberShadow:
> Is it possible to execute this command with DAutoTest?

Yep. On vacation so PRs welcome :)

Needs an option in ae.sys.d and making DAutoTest enable that option.

PR #1401, opened 2016-06-29
Reworded statement

brdjns:
Reworded the initial statement under the 'Convenience' heading of the 'Why D?' section. The language flows better with the combination of a subject pronoun + verb infinitive, than with just a verb in the present participle form. Addressing the reader directly promotes a sense of inclusion in the action suggested.

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -158,7 +158,7 @@ $(DIVC whyd, $(SECTION2 Why D?,
 $(DIVC section,
 $(DIV, $(SECTION3 $(WHY_D_ICON magic) Convenience,
 
-$(P D allows writing large code fragments without redundantly specifying types,
+$(P D allows you to write large code fragments without redundantly specifying types,
 like dynamic languages do. On the other hand, static inference deduces types and other
 code properties, giving the best of both the static and the
 dynamic worlds. $(EXAMPLE 1,


CyberShadow:
Not to nitpick on such a small change and especially as someone with English as their third language, but I thought we were generally avoiding using "you" and second person in general.

wilzbach:
> Not to bikeshed on such a small change and especially as someone with English as their third language, but I thought we were generally avoiding using "you" and second person in general.

Yep I also thought that a neutral tone should be used for the specification. Even for the Dlang tour we have defined [such rules](https://github.com/stonemaster/dlang-tour/blob/master/CONTRIBUTING.md#writing-style-guide), maybe we should do the same in the `CONTRIBUTING.md` for dlang.org?

AndrejMitrovic:
That makes three of us, I think it's a step in the wrong direction. 

> Even for the Dlang tour we have defined such rules, maybe we should do the same in the CONTRIBUTING.md for dlang.org?

Sounds good. In the meantime closing this.

wilzbach:
>>> I thought we were generally avoiding using "you" and second person in general.
>> Even for the Dlang tour we have defined such rules, maybe we should do the same in the CONTRIBUTING.md for dlang.org?
> Sounds good. In the meantime closing this.

-> #1448

PR #1403, opened 2016-06-29
fix orgs using d - Arex GH link broken

BBasile:


wilzbach:
> CC @wilzbach

Thanks @BBasile!

PR #1405, opened 2016-06-30
add Auburn Sounds to orgs-using-d

wilzbach:
CC @p0nce 

(SVG is added in case we need a nigher quality image)


p0nce:
Thanks for doing this!

CyberShadow:
The gzipped SVG is smaller than the PNG. Why bother with the PNG?

wilzbach:
> The gzipped SVG is smaller than the PNG. Why bother with the PNG?

@p0nce added some specific shading to the PNG...

p0nce:
It looked bad on a white background.

CyberShadow:
I thought the _hq versions are so we can easily recreate the non-_hq versions when we switch to a different size...

p0nce:
I'll fix it if you switch to a different size.

PR #1406, opened 2016-06-30
Restructure community menu - move successful resources to the top

wilzbach:
- @p0nce suggested to move the more successful resources at the top (e.g. Blog, Orgs using D)
- I fixed the alphabetical order
- I renamed "Bug Tracker" to "Issues", because 
    - bugs has a very negative connotation
    - it's called issues.dlang.org too
    - it should be last on the list (and with "Issues" we achieve this without breaking the alphabetical order)

### DIFF
--- a/dlang.org.ddoc
+++ b/dlang.org.ddoc
@@ -223,14 +223,14 @@ $(DIVID cssmenu, $(UL
 ))
 NAVIGATION_COMMUNITY=
 $(SUBMENU_MANUAL
-    $(SUBMENU_LINK $(ROOT_DIR)bugstats.php, Bug Tracker)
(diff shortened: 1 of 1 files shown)


p0nce:
People click and see the first items in a list more (when arranged vertically and horizontally).

p0nce:
In "Organizations using D" Likewise I see that the org order is alphabetical.
That would put me in the first names, but in my opinion it's a completely wrong order. 
The orgs with the most social proof (Facebook, Sociomantic...) and well known should be put prominently.
Because it's a page built solely for social proof.

wilzbach:
@CyberShadow - any opinion on this?

PR #1407, opened 2016-06-30
Rebrand Bug Tracker to Issue Tracker

wilzbach:
As mentioned in #1406, "Bug" has a very negative connotation. Moreover about half of our "bugs" are actually enhancement requests, so I propose the rebranding to "Issues" on such a publicly visible site.

Does anyone know how I can rename the chart?

### DIFF
--- a/bugstats.php.dd
+++ b/bugstats.php.dd
@@ -1,6 +1,6 @@
 Ddoc
 
-$(D_S Bug tracker,
+$(D_S Issue tracker,
 
 $(P
     We use $(B Bugzilla) to track the issues for the D programming language.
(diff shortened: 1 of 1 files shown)


CyberShadow:
It's called Bugzilla, not Issuezilla... :P

CyberShadow:
I don't like this change.

https://en.wikipedia.org/wiki/Bug_tracking_system

https://en.wikipedia.org/wiki/Issue_tracking_system

These are two different things.

It's OK to call the things filed as "issues", that's what Bugzilla calls them too. But in the end, it's a bug tracker.

wilzbach:
> It's called Bugzilla, not Issuezilla... :P

Ok fair enough ;-)

PR #1410, opened 2016-07-02
add link to kostya's benchmarks

wilzbach:
Thanks to @ZombineDev's work D's performances is quite for [Kostya's benchmark](https://github.com/kostya/benchmarks) and as people love such benchmarks I thought it's probably a good idea to add it to a popular place.

image


CyberShadow:
As previously repeatedly discussed, I think we should avoid direct language comparisons. As soon as we officially endorse this benchmark, we open ourselves to being accused of using unfairly slow implementations in this or that language, or using outdated implementations (because of course they made component X faster in the last minor revision), etc. - and part of that backlash is going to hit the benchmark author.

I think this should only be merged with Kostya's and Andrei's OKs.

PR #1413, opened 2016-07-03
fix issue 16231 - rename std.regexp to std.regex in a code example

BBasile:


spec/statement.dd
```diff
@@ -260,21 +260,21 @@ $(GNAME ElseStatement):
...
-    writefln("[%s]", m.pre);      // prints [a]
-    writefln("[%s]", m.post);     // prints [ef]
-    writefln("[%s]", m.match(0)); // prints [bcd]
-    writefln("[%s]", m.match(1)); // prints [c]
-    writefln("[%s]", m.match(2)); // prints []
+    writefln("[%s]", m.pre);    // prints [a]
+    writefln("[%s]", m.post);   // prints [ef]
+    writefln("[%s]", m.hit[0]); // prints [b]
+    writefln("[%s]", m.hit[1]); // prints [c]
+    writefln("[%s]", m.hit[2]); // prints [d]
```
CyberShadow: The `std.regex` equivalent of `std.regexp`'s `.match` is `opIndex`, i.e. `m[1]` will be c.

BBasile: > /tmp/temp_7F029904D1B0.d(12,27): Error: no [] operator overload for type RegexMatch!(string, ThompsonMatcher)
Anyway does it really matter ? The example is compilable...the point of the example is the variable scope.

CyberShadow: > /tmp/temp_7F029904D1B0.d(12,27): Error: no [] operator overload for type RegexMatch!(string, ThompsonMatcher)

Ah, I think it should use `matchFirst`, not `matchAll`.

> Anyway does it really matter ?

Not really, no. But since if you're fixing it, why not make the example also make sense (i.e. be one that closer reflects a real-world situation)?

BBasile: The original example has a bug...

CyberShadow: Where was the bug? `match(2)` being out-of-bounds?

BBasile: I think, but this need to be verified in undead.

BBasile:
îng

JackStouffer:
LGTM

PR #1414, opened 2016-07-04
abi: add mangling for 'scope' and 'return'

WalterBright:
Also sorted the list.


WalterBright:
Blocked by what?

CyberShadow:
The DMD PR. Is it not?

Hackerpilot:
DMD pull was merged, so this is no longer blocked.

PR #1415, opened 2016-07-04
'return' is missing from the grammar

WalterBright:


ZombineDev:
I guess you should also add `scope` to the list.

Hackerpilot:
Agrees with this change to the unofficial parser: https://github.com/Hackerpilot/libdparse/commit/fc9dacfa41c0bae258814ad64e58ffaae6b961cc

PR #1417, opened 2016-07-08
Orgs using D: add eBay

wilzbach:
Wohoo we finally got the legal permission to list eBay :tada: 

CC @jondegenhardt, thanks again for your hard work. Please check if everything is correct ;-)

@CyberShadow we only have the permission for this explicit use, hence I didn't include a HQ version of the eBay logo. I still have a higher quality version from the contract, which we can use if we want to support monitors with a higher resolution.

Disclaimer @ mac users: I don't have a Retina display, so don't expect such a PR soon.


index.dd
```diff
@@ -380,15 +380,15 @@ $(DIVC frontpage-orgs center,
...
+    )
     $(FRONTPAGE_ORG Sociomantic, https://www.sociomantic.com, sociomantic.png,
         Real-time personalized advertising
     )
     $(FRONTPAGE_ORG Weka.IO, http://weka.io, weka.png,
         Software defined storage
     )
-    $(FRONTPAGE_ORG Funkwerk AG, http://www.funkwerk.com/en, funkwerk.png,
-        Passenger information systems
-    )
```
CyberShadow: ?

CyberShadow:
> Disclaimer @ mac users: I don't have a Retina display, so don't expect such a PR soon.

Just set your browser zoom to 200%?

jondegenhardt:
LGTM

PR #1409, opened 2016-07-02
Orgs using D: Add Ecratum

wilzbach:
resubmission of #1381 

CC @llaine

@CyberShadow It seems we use `150x75` for the frontpage, but here we use `200x100`. Sorry about that & corrected.

PR #1419, opened 2016-07-08
orgs-using-d: Fix Ecratum's "Hiring" link

CyberShadow:
https://github.com/dlang/dlang.org/pull/1409#issuecomment-231295391


wilzbach:
@CyberShadow should we just remove all the WEB/HTTP macros on this page? They are a lot more painful then what we gain with them...

CyberShadow:
They should be removed everywhere, not just this page.

PR #1420, opened 2016-07-09
clarify what happens with overlarge shifts

WalterBright:
Saying it is "illegal" implies it must always be diagnosed.

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -514,8 +514,9 @@ $(GNAME ShiftExpression):
         $(D >)$(D >)$(D >) is an unsigned right shift.
     )
 
-    $(P It's illegal to shift by the same or more bits than the size of the
-        quantity being shifted:
+    $(P If the quantity is shifted by the same or more bits than the size of the
+	quantity, it is implementation defined whether a diagnostic is issued
+	or the resulting value is unspecified.
 
         -------------
         int c;


spec/expression.dd
```diff
@@ -514,8 +514,9 @@ $(GNAME ShiftExpression):
         $(D >)$(D >)$(D >) is an unsigned right shift.
     )
 
-    $(P It's illegal to shift by the same or more bits than the size of the
-        quantity being shifted:
+    $(P If the quantity is shifted by the same or more bits than the size of the
+	quantity, it is implementation defined whether a diagnostic is issued
```
JackStouffer: Why not just say it's UB?

WalterBright: Because UB encompasses "launch nuclear missiles", and that makes people nervous. It's better to constrain things a bit more when possible.

andralex: "it is implementation defined whether a diagnostic is issued"

->

"the behavior is implementation defined"

because this also covers getting a silent result (such as zero), or aborting execution.

WalterBright: I wanted to constrain it to issuing a diagnostic (which includes seg faulting and aborting) or producing a random integer value. I don't want to leave open "implementation defined" as including corrupting memory.

andralex: Then say what you mean. From what you say the implementation can still do whatever it wants. It may issue a diagnostic, or it may crash the computer. Let's be precise here. What is acceptable and what isn't?

dkorpel:
Superseded by https://github.com/dlang/dlang.org/pull/2578

PR #1425, opened 2016-07-10
add dmd.html

aG0aep6G:
There's a redirect in place that detects the user's operating system and
goes to dmd-{windows,linux,...}.html. The redirect is fine and on the
online dlang.org it completely hides the dmd.html file that's being added
here. The file is needed, because dlang.org is also distributed as a bunch
of .html files as part of the dmd releases.

Not putting links to dmd-windows.html, etc into the file, because they'd
only be forgotten there.

### DIFF
--- a/dmd.dd
+++ b/dmd.dd
@@ -1,0 +1,12 @@
+Ddoc
+
+$(D_S dmd - DMD Compiler,
+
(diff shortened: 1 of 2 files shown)


wilzbach:
Makes sense, thanks!

Hackerpilot:
:white_check_mark:

PR #1427, opened 2016-07-11
Remove the 'Symbol' rule from the grammar.

Hackerpilot:
http://forum.dlang.org/thread/oaqifnqictfdaupaaptn@forum.dlang.org

The `Symbol` rule only appears as an alternative to the `Type`, `AssignExpression`, or `Expression` rules, but the `Symbol` rule is a subset of these other rules. Because of this, the grammar is ambiguous.

The difference between a type, expression, and symbol is something that can only be determined during semantic analysis, and thus should not be part of the parse grammar.

### DIFF
--- a/spec/grammar.dd
+++ b/spec/grammar.dd
@@ -762,7 +762,6 @@ $(GNAME GotoStatement):
 $(GRAMMAR
 $(GNAME WithStatement):
     $(D with) $(D $(LPAREN)) $(EXPRESSION) $(D $(RPAREN)) $(PSSCOPE)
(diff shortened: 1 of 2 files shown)


WalterBright:
But a template argument to an alias parameter can be any symbol, such as a module name. A module name is neither a type nor an expression.

(Types and Expressions can also be ambiguous.)

Hackerpilot:
I don't see what that kind of semantic information has to do with creating a parse tree, unless we're going the C++ route of requiring semantic in order to parse.

PR #1428, opened 2016-07-12
Update attribute.dd

jmh530:
Re-ordering the text to put a greater emphasis on describing abstract member functions before abstract classes. Also, fleshed out some of the details a little bit more.

### DIFF
--- a/spec/attribute.dd
+++ b/spec/attribute.dd
@@ -695,20 +695,27 @@ $(P
 $(H3 $(LNAME2 abstract, $(D abstract) Attribute))
 
 $(P
+        An abstract member function must be overridden by a derived class.
+        Only virtual member functions may be declared abstract; non-virtual 
+        member functions and free-standing functions cannot be declared 
+        abstract.
+)
+
+$(P
+        Classes become abstract if any of the virtual member functions
+        within it are declared as abstract or if they are defined within an
+        abstract attribute. 
+        Note that an abstract class may also contain non-virtual member functions.
+)
+
+$(P
         If a class is abstract, it cannot be instantiated
         directly. It can only be instantiated as a base class of
         another, non-abstract, class.
 )
+
 $(P
-        Classes become abstract if they are defined within an
-        abstract attribute, or if any of the virtual member functions
-        within it are declared as abstract.
-)
-$(P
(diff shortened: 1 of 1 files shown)


spec/attribute.dd
```diff
@@ -695,20 +695,27 @@ $(P
...
+        Only virtual member functions may be declared abstract; non-virtual 
+        member functions and free-standing functions cannot be declared 
+        abstract.
 )
+
 $(P
-        Classes become abstract if they are defined within an
-        abstract attribute, or if any of the virtual member functions
-        within it are declared as abstract.
+        Classes become abstract if any of the virtual member functions
```
BBasile: if any of **its** member function ~~within it~~ **is** declared declared abstract.

spec/attribute.dd
```diff
@@ -695,20 +695,27 @@ $(P
...
+        within it are declared as abstract or if they are defined within an
+        abstract attribute. 
+        Note that an abstract class may also contain non-virtual member functions.
 )
+
 $(P
-        Non-virtual functions cannot be declared as abstract.
+        If a class is abstract, it cannot be instantiated
```
BBasile: I would prefer this shorter, (and more correct) sentence: `As long as a class contains abstract member functions it cannot be instantiated.`

From your sentence it seems like

```D
class Foo
{
    abstract void foo1();
    abstract void foo2();
}

class Bar: Foo
{
    override void foo1(){}
}

Bar bar = new Bar;
```

is possible, which is obviously not true.

jmh530:
@BBasile I believe I have addressed your comments. I changed the wording slightly where I thought appropriate.

BBasile:
Thx, I see nothing else to say about this PR. The description is a bit better and is still compliant with what the compiler does so LGTM.

BBasile:
It might be worth waiting for resolution | official feedback about https://issues.dlang.org/show_bug.cgi?id=2525. There's a problem when an abstract class implements an interface.

jmh530:
@BBasile I'm cool with that.

AndrejMitrovic:
LGTM. I think [Issue 2525](https://issues.dlang.org/show_bug.cgi?id=2525) is fixable sometime in the near future.

PR #1429, opened 2016-07-12
Clarify spec for "+=" and friends

JohanEngelen:
See https://forum.dlang.org/post/mxvdieydnicvdxbcklrd@forum.dlang.org

(submitted through the website and I cannot make changes to this PR. Because I am cleaning up after people, if you don't like it, please submit your own PR.)

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -96,9 +96,37 @@ $(H4 Assignment Operator Expressions)
     $(P except that:)
 
     $(UL
-        $(LI operand $(D a) is only evaluated once)
+        $(LI operand $(D a) is only evaluated once, see below.)
         $(LI overloading $(I op) uses a different function than overloading $(I op)= does)
         $(LI the left operand of $(D >>>=) does not undergo integral promotions before shifting)
+    )
+    
+    $(P The expression evaluates according to
+
+        --------------
+        void doOpAssignment(ref int x, int y)
+        {
+            x = x op y;
+        }
+        doOpAssignment(a, b);
+        --------------
+
+        thus the following unittest passes
+
+        --------------
+        int sum;
+        int add7tosum_return1() { sum += 7; return 1; }
+        unittest
+        {
+            sum = 0;
+            sum += add7tosum_return1();
+            assert(sum == 8); // "eight" !
+        
+            sum = 0;
+            sum = cast(typeof(sum))(sum + add7tosum_return1());
+            assert(sum == 1); // "one" !
+        }
+        --------------
     )
 
 $(H3 Conditional Expressions)


wilzbach:
> submitted through the website and I cannot make changes to this PR.

Is you "you don't want to" or you technically can't? Editing a PR made from the website is easy:
If you don't have the repo locally, clone it:

```
git clone git@github.com/JohanEngelen/dlang.org
```

or if you already have it, update its remote info:

```
git fetch 
```

and then go into the usual flow:

```
git checkout patch-5
// make changes & commit
git push -f
```

spec/expression.dd
```diff
@@ -96,10 +96,38 @@ $(H4 Assignment Operator Expressions)
...
+        int sum;
+        int add7tosum_return1() { sum += 7; return 1; }
+        unittest
+        {
+            sum = 0;
+            sum += add7tosum_return1();
+            assert(sum == 8); // "eight" !
+        
+            sum = 0;
+            sum = cast(typeof(sum))(sum + add7tosum_return1());
```
ZombineDev: Why is the cast necessary?

JohanEngelen: It isn't. But I made it look exactly the same as the given "semantic equivalent" above.

spec/expression.dd
```diff
@@ -96,10 +96,38 @@ $(H4 Assignment Operator Expressions)
...
     )
+    
+    $(P The expression evaluates according to
+
+        --------------
+        void doOpAssignment(ref int x, int y)
+        {
+            x = x op y;
+        }
+        doOpAssignment(a, b);
```
andralex: The description needs to include the cast. I'd say a correct formulation goes as follows:

```d
$(P The expression `a @= b`, where `@` stands for any of the binary operators, is evaluated
as if it were lowered as follows:

(function ref(ref x, y) { return x = cast(typeof(x)) (x @ y); })(a, b)
```

See also (sadly) https://issues.dlang.org/show_bug.cgi?id=16271 which blocks this. Other than that this definition is really solid because defines everything by "redirecting" to function calls.

andralex: cc @WalterBright

spec/expression.dd
```diff
@@ -96,10 +96,38 @@ $(H4 Assignment Operator Expressions)
...
         $(LI overloading $(I op) uses a different function than overloading $(I op)= does)
         $(LI the left operand of $(D >>>=) does not undergo integral promotions before shifting)
     )
+    
+    $(P The expression evaluates according to
+
+        --------------
+        void doOpAssignment(ref int x, int y)
+        {
+            x = x op y;
```
deadalnix: There is a cast missing in there. I did not put it in the forum thread as I used int explicitly, but it is needed in the general case. For instance :

  byte a = 3;
  a += 3; // This is ok because of the cast, but won't work because of int promotion without).

spec/expression.dd
deadalnix: It is necessary if all isn't int.

JohanEngelen:
This needs to be extended with semantics for operations on arrays, like `~=` (for which the types of lhs and rhs are not the same, `T[]` vs `T`).
```d
int work(ref int[] array) {
	array ~= 5;
	return 0;
}

void doTildeAssignment(T)(ref T[] x, T y) {
	x = x ~ y;
}

void main(){
    int[] array;
    version(none) {
        array ~= work(array);
    } else {
        doTildeAssignment(array, work(array));
    }
    
    assert(array == [5,0]);
}
```
See https://github.com/ldc-developers/ldc/issues/2588

PR #1430, opened 2016-07-12
[Orgs using D] add Remedy

wilzbach:
CC @goberman

@CyberShadow: according to the new policy of adding a HQ source I added the SVG
version found on WikiMedia. However it's a whoppy 170kB and I am not so sure it's
worth the size. What do you think?


CyberShadow:
> However it's a whoppy 170kB and I am not so sure it's
> worth the size.

I don't understand. What exactly is the problem?

orgs-using-d.dd
```diff
@@ -108,16 +108,13 @@ $(DIVC orgs-using-d center,
             $(FA_HIRING job.oschina.net/position/792_118197_21895)
         )
     )
-
-    $(COMMENT
-    $(DORG Remedy Games, http://remedygames.com, remedy.svg,
+    $(DORG Remedy Games, http://remedygames.com, remedy.jpg,
```
CyberShadow: JPEG? For a monochrome logo? Really?

wilzbach:
> I don't understand. What exactly is the problem?

Is it worth adding the 170 kB SVG if it's freely available on wikimedia and we will probably only need 
lower-sized PNGs?

> JPEG? For a monochrome logo? Really?

Done. Btw I changed the name of Remedy to Remedy Entertainment, their [official name](https://en.wikipedia.org/wiki/Remedy_Entertainment). For the front page this yields us two lines for the title:

image

Ideas:

1) Add padding to the other titles too
2) Just use Remedy
3) Your idea(s)

CyberShadow:
> Is it worth adding the 170 kB SVG if it's freely available on wikimedia and we will probably only need lower-sized PNGs?

Yes, absolutely. No question about it. There is no cost to this except a slight increase in cloning time, and even then that's the size before compression. Whether it's available elsewhere is not of much importance, since that source might change or disappear, and might be difficult to track down years from now, and would make future re-conversions more onerous than necessary.

> Done. Btw I changed the name of Remedy to Remedy Entertainment, their official name. For the front page this yields us two lines for the title:

Well... crazy idea, why not just change it back to "Remedy Games"? That's the name they use for all their social media.

orgs-using-d.dd
```diff
@@ -108,16 +108,13 @@ $(DIVC orgs-using-d center,
             $(FA_HIRING job.oschina.net/position/792_118197_21895)
         )
     )
-
-    $(COMMENT
-    $(DORG Remedy Games, http://remedygames.com, remedy.svg,
+    $(DORG Remedy Entertainment, http://remedygames.com, remedy.png,
```
wilzbach: we have more space here, so the long name would fit. However should I use "Remedy Games" for consistency?

wilzbach:
> Well... crazy idea, why not just change it back to "Remedy Games"? That's the name they use for all their social media.

nice idea - works superb :)

wilzbach:
Yes we do :)

orgs-using-d.dd
klickverbot: I'd use Remedy Games here too.

orgs-using-d.dd
```diff
@@ -108,16 +108,13 @@ $(DIVC orgs-using-d center,
...
         Triple-A game development,
         First Triple-A game $(I $(HTTP quantumbreak.com, Quantum break)) with D code.
```
klickverbot: Should probably read Quantum Break as it's a proper name (also stylised that way on their official web page).

wilzbach:
@klickverbot thanks. addressed.

wilzbach:
> CC @goberman

@GooberMan sorry about the typo & Remedy Games is now on dlang.org ;-)
Please let us know when you do more amazing things with D.

PR #1432, opened 2016-07-27
Added doc for __FILE_FULL_PATH__

marler8997:
https://issues.dlang.org/show_bug.cgi?id=16441

### DIFF
--- a/spec/traits.dd
+++ b/spec/traits.dd
@@ -71,6 +71,7 @@ $(P Additionally special keywords are provided for debugging purposes:)
 $(GRAMMAR
 $(GNAME SpecialKeyword):
     $(D $(RELATIVE_LINK2 specialkeywords, __FILE__))
+    $(D $(RELATIVE_LINK2 specialkeywords, __FILE_FULL_PATH__))
     $(D $(RELATIVE_LINK2 specialkeywords, __MODULE__))
     $(D $(RELATIVE_LINK2 specialkeywords, __LINE__))
     $(D $(RELATIVE_LINK2 specialkeywords, __FUNCTION__))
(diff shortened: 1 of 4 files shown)


spec/traits.dd
```diff
@@ -939,6 +940,9 @@ $(H2 $(LNAME2 specialkeywords, Special Keywords))
     $(P $(CODE __FILE__) and $(CODE __LINE__) expand to the source
```
BBasile: For `__FILE__` I think that a short sentence could explain that it _expands to the source file name, as passed to the compiler_.

marler8997: This doesn't appear to be the case though (I'm only testing on windows, may be different on Linux).  I can go to another directory and compile a source file with a relative path, and the relative path I pass to the compiler isn't maintained:

```
/somedir/test.d

/somedir/build> dmd ../test.d

// __FILE__ will expand to test.d (not ../test.d)
```

BBasile: I meant that if you pass relative it expands to relative and vice versa...not well formulated. So don't bother and ignore this comment, unless you manage to explain accurately the behavior of this macro.

marler8997: I figured the documentation was vague because the behavior was meant to be undefined.  It lets the compiler do whatever it wants, which is ok IMO.  That's part of the reason why I thought adding `__FILE_FULL_PATH__` was a good idea.  Let `__FILE__` remain vague so the compiler can decide what's best to do, but require that `__FILE_FULL_PATH__` always be absolute in case the application needs a guaranteed behavior.

AndrejMitrovic:
I've created and closed https://issues.dlang.org/show_bug.cgi?id=16441. In the future please make issues in bugzilla and link to them (via `Fix Issue 16441` in the commit message) when making the DMD pulls. :)

AndrejMitrovic:
LGTM, thanks!

PR #1435, opened 2016-07-31
Fix dup/idup documentation

schveiguy:
The documentation didn't clearly specify that `dup` removes immutable/const, and that if the operation would violate const/immutable guarantees, it doesn't compile.

There was also a needless note that `idup` is D2.0 only.

### DIFF
--- a/spec/arrays.dd
+++ b/spec/arrays.dd
@@ -424,8 +424,8 @@ Returns an array literal with each element of the literal being the $(D .init) p
         $(TROW $(D .length), Returns the number of elements in the array.
         This is a fixed quantity for static arrays. It is of type $(D size_t).)
         $(TROW $(D .ptr), Returns a pointer to the first element of the array.)
-        $(TROW $(D .dup), Create a dynamic array of the same size and copy the contents of the array into it.)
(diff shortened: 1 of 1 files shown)


schveiguy:
See forum thread for reference: https://forum.dlang.org/post/cdxdcxikdcbsbygwmhpc@forum.dlang.org

schveiguy:
See forum thread for reference: https://forum.dlang.org/post/cdxdcxikdcbsbygwmhpc@forum.dlang.org

schveiguy:
@CyberShadow I can't tell what's wrong with this, why is the build failing?

CyberShadow:
Sorry, I don't know anything beyond that the PDF build is failing with the error messages indicated in the log.

schveiguy:
TBH, I don't even know where the error is. I see a lot of warnings and a lot of things. And then "No PDF file produced". Why not? 😠

schveiguy:
Who knows how the PDF build works?

schveiguy:
I think I know what it is -- the commas. `TROW` is probably adding extra `TD`s...

schveiguy:
OK, building now.

PR #1437, opened 2016-08-06
add platform dependent download buttons

MartinNowak:


index.dd
```diff
@@ -415,6 +415,7 @@ Macros:
     EXTRA_EXAMPLE=<div class="your-code-here-extra" style="display:none">$0</div>
     LAYOUT_PREFIX=
     LAYOUT_SUFFIX=
+    $(SCRIPTLOADA $(ROOT_DIR)js/platform-downloads.js, data-latest="$(LATEST)")
```
wilzbach: typo here ;-)

MartinNowak: A for SCRIPTLOAD w/ attributes, similar to DIVC for div with classes.
Just extending the existing SCRIPTLOAD runs into some annoying spacing issues without an argument.

MartinNowak: Fixed, I can actually add the space at the argument.

js/platform-downloads.js
```diff
@@ -0,0 +1,45 @@
+(function() {
+    var script = document.currentScript;
+    if (!script) { // polyfill for IE
+        var scripts = document.getElementsByTag('script');
+        script = scripts[scripts.length - 1]; // depends on synchronous load
+    }
+    var latest = script.dataset.latest,
+        platform = navigator.platform.toLowerCase();
```
wilzbach: this is just `linux` on my Chrome

MartinNowak:
Making ldc the default compiler I guess ;).
Seriously, one output of our first five minute review was that the download page, trying to present all compilers, overwhelms people with too much information.
The actual downloads being already below the fold. A 1-click download button that offers a complex choice misses it's purpose.
But we should forward people to take the tour and early on explain that ldc is a better optimizing compiler.

js/platform-downloads.js
MartinNowak: Oh, then we'll default to 64.

MartinNowak:
Made 64-bit downloads the default and aligned the buttons.
freebsd_down
linux_down
osx_down
win_down

Not enough space for the curl installer script though, that might only work as clipboard button.
installer

css/style.css
```diff
@@ -1309,9 +1333,20 @@ body#Home #content > .intro .download
...
+{
+    border-color: #98312A;
 }
 
+body#Home #content > .intro .download .btn.action:hover
+{
+    border-color: #943029;
+}
+
+
```
wilzbach: FYI I think a couple of PRs ago we agreed that inlining page-specific CSS into the page makes sense, because (a) it helps to cleanup the massive `style.css` and (b) it's also faster for clients and (c) the `style.css` is now used at many different places (main page, docs, forum, code.dlang.org, tour.dlang.org, ...) s.t. every addition adds to size that needs to be downloaded by all those pages.

js/platform-downloads.js
wilzbach: but you could use `navigator.userAgent` (I know that it's easy to fake it), but the majority doesn't. E.g. my Chrome sends this funny string to the world, 

```
navigator.userAgent
"Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/50.0.2661.94 Safari/537.36"
```

(OT: the invention "feature detection" never made it to the browser world)

wilzbach:
@MartinNowak it looks really good :)

> not enough space for the curl installer script though, that might only work as clipboard button.

Well we could still map `i.dlang.io` - it saves you 9 chars ;-)

```
curl -fsS https://i.dlang.io | bash -s dmd
```

if we use redirects to SSL it would be even 7 chars more (=16 in total).

```
curl -fsSL i.dlang.io | bash -s dmd
```

> Leaving that for another PR.

Looking forward to it ;-)

MartinNowak:
In fact installing dmd is the default action, so `curl i.dlang.io | bash` should work as well.

MartinNowak:
Exactly, we should find a better place for the full command, could be a clipboard button with explanation.
Anyhow OT for this PR.

wilzbach:
> In fact installing dmd is the default action, so curl i.dlang.io | bash should work as well.

If no command is given, it will currently just print the help page - see https://github.com/dlang/installer/blob/master/script/install.sh#L271

Btw when looking at other languages's installer, I saw that the Rust team has a pretty [clean page](https://www.rust-lang.org/en-US/downloads.html).

> Anyhow OT for this PR.

@MartinNowak did you see my comment about putting the CSS to `index.dd`? Otherwise the PR looks good to ship for me. Using 64-bit as default seems a very reasonable choice to me (and if wanted someone can still come up with a better detection method).

wilzbach:
> In fact installing dmd is the default action, so curl i.dlang.io | bash should work as well.

As a final response to our OT discussion, I quickly set `i.dlang.io` to return the `install.sh`. As `curl` doesn't allow redirections by default, `install.sh` is downloaded every six hours.
Of course if we ever announce this, the `install.sh` shouldn't be served from my server and shouldn't be downloaded automatically from an untrusted third-party (Github), however for the lazy people around here it's now possible to:

```
curl i.dlang.io | bash -s dmd
curl https://i.dlang.io | bash -s dmd  
```

... and if https://github.com/dlang/installer/pull/189 (`dmd` by default) get's accepted, we would save even 7 chars more :tada:

MartinNowak:
On the topic.
https://sandstorm.io/news/2015-09-24-is-curl-bash-insecure-pgp-verified-install
Interestingly enough, they have exactly the same setup, i.e. we could copy their manual https://docs.sandstorm.io/en/latest/install/.

andralex:
I'll merge this, feel free to iterate going forward.

css/style.css
MartinNowak: It's a bit problematic b/c of synchronizing the colors, but at least separating out all that `body#Home` stuff would make sense.

PR #1441, opened 2016-08-19
default opEquals does not compare object contents

Cauterite:
this documents the behaviour more clearly, addressing issue 10549


AndrejMitrovic:
LGTM. But I believe we need a Druntime PR to fix opEquals docs as well.

PR #1445, opened 2016-08-25
Fix relative links in deprecate page

wilzbach:
I was now annoyed a couple of times by the lack to be able to link directly to a deprecation page (or only with manually replacing spaces with `%20`, so here is a fix for it.

It's a shame DDoc doesn't allow function execution / CTFE :/

### DIFF
--- a/deprecate.dd
+++ b/deprecate.dd
@@ -9,40 +9,40 @@ $(SPEC_S Deprecated Features,
     )
 
     $(TABLE2 Deprecated Features,
-        $(THEAD Feature,                                                          Spec,  Dep,    Error,  Gone)
-        $(TROW $(DEPLINK Using the result of a comma expression),                 future, &nbsp;, &nbsp;, &nbsp;)
-        $(TROW $(DEPLINK delete),                                                 future, &nbsp;, &nbsp;, &nbsp;)
-        $(TROW $(DEPLINK scope for allocating classes on the stack),              future, &nbsp;, &nbsp;, &nbsp;)
-        $(TROW $(DEPLINK Imaginary and complex types),                            future, &nbsp;, &nbsp;, &nbsp;)
-        $(TROW $(DEPLINK Implicit catch statement),                               2.072, 2.072,  future, future )
-        $(TROW $(DEPLINK .sort and .reverse properties for arrays),               ?,     2.072;, &nbsp;, &nbsp;)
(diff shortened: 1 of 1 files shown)


aG0aep6G:
This breaks existing links. That can be avoided by keeping anchors with the old names around.

> I was now annoyed a couple of times by the lack to be able to link directly to a deprecation page (or only with manually replacing spaces with %20, so here is a fix for it.

When I copy a URL from Firefox (address bar or right click on the link), it gives me one with `%20`s in it. When do you have to replace things manually?

wilzbach:
> This breaks existing links. That can be avoided by keeping anchors with the old names around.

Ok, fair enough.

with Chrom{ium,e}:

image

aG0aep6G:
> with Chrom{ium,e}:
> 
> image

Looks like that with Firefox, too. But if I actually copy the URL, I get one with `%20`s.

wilzbach:
> Looks like that with Firefox, too. But if I actually copy the URL, I get one with %20s.

Yep and for Chrome you don't get this automatically and it seems like I am not the only one who bumped into this:

> (Search for "scope for allocating classes on the stack".)

http://forum.dlang.org/post/fmxrxohnlzjwhkcompwn@forum.dlang.org

aG0aep6G:
> Yep and for Chrome you don't get this automatically

Ok. I think Chrome is being silly there, but there's little we can do about that.

We should probably scan everything for such anchors then. *sigh*

aG0aep6G:
@andralex, you just broke a bunch of URLs by merging this. It's probably not too bad in this case, but it's  something we usually try to avoid.

wilzbach:
> you just broke a bunch of URLs by merging this. It's probably not too bad in this case, but it's something we usually try to avoid.

Sorry lost track of this. I wanted to submit a non-breaking solution. There is a simple fix to just add another `<a name=""> with the title. Will submit this later tonight.

andralex:
urgh

andralex:
urgh https://github.com/dlang/dlang.org/pull/1457

andralex:
ok I misunderstood from the chatter that this is good to go... @wilzbach have at it

wilzbach:
> ok I misunderstood from the chatter that this is good to go... @wilzbach have at it

-> #1460 (contains a second named `<a>` tag for graceful deprecation of the deprecation links)

PR #1447, opened 2016-08-28
Add const to strcmp params

gvre:


AndrejMitrovic:
Should be `const(char)*`, not `const char *` in D.

andralex:
@AndrejMitrovic eh, in fact it doesn't matter - `const char *` is equivalent to `const(char*)` which makes the pointer const (irrelevant for the interface) as well as the pointed-to memory. I'll pull, please undo if I'm wrong.

AndrejMitrovic:
Yeah it's fine. It's `const char**` that's the real troublemaker.

PR #1452, opened 2016-08-29
Add tag & stats badges to the README

wilzbach:
- The tag badge will automatically show the latest git tag
- The issue stats badge is intended as a positive motivator for contributors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,9 @@
 ![D Logo](https://dlang.org/images/dlogo.svg) D Language Website
 ===============================================================
 
+[![GitHub tag](https://img.shields.io/github/tag/dlang/dlang.org.svg?maxAge=86400)](#)
+[![Issue Stats](https://img.shields.io/issuestats/p/github/dlang/dlang.org.svg?maxAge=2592000)](http://www.issuestats.com/github/dlang/dlang.org)
+
 This repository contains the source files for the [D Language
 Website](https://dlang.org), including the language specifications,
 documentation, and some articles.


wilzbach:
> Similar to the other repos.

Yep I thought that if I did it in one run the noise is created only once.
Thanks!

AndrejMitrovic:
Btw, we have so many files in root that the user has to scroll all the way down before the readme is visible.

wilzbach:
> Btw, we have so many files in root that the user has to scroll all the way down before the readme is visible.

Depends on your screen resolution, but if you want to say that all the information in the README might be easy to miss, I fully agree with you. I would love to group pages into subfolders.
@CyberShadow what's your opinion on this? With the Makefile we should even be able to keep the path, but on the other hand we can also use the htaccess to avoid broken links?

CyberShadow:
My opinion is that we really have better things to do :)

wilzbach:
>My opinion is that we really have better things to do :)

Hmm I disagree - a messy root folder leaves a very negative first impression.
At least for my part, figuring out how to build dlang.org was rather difficult in the beginning because I didn't scroll to the bottom and read the links.
So I would be happy to invest ten minutes to group these pages and cleanup the root folder.

> With the Makefile we should even be able to keep the path, but on the other hand we can also use the htaccess to avoid broken links?

Okay I realized that just modifying the Makefile doesn't work, because then the "Improve this page" wouldn't work, but moving a couple of pages and adding them to the htaccess should still work.
AFAIK `git` recognizes renames, so it wouldn't destroy the history.

For a start I could move all [articles](http://dlang.org/articles.html) to an `articles` folder?

CyberShadow:
> Hmm I disagree - a messy root folder leaves a very negative first impression.

Seriously? Do you really think that making the README more visible by moving some files around is more important than [any of these](https://issues.dlang.org/buglist.cgi?bug_severity=regression&bug_severity=blocker&bug_severity=critical&list_id=210317&query_format=advanced&resolution=---)?

Please, let's mind our priorities.

CyberShadow:
Essentially: http://forum.dlang.org/post/mbc83q$6dn$1@digitalmars.com

PR #1454, opened 2016-08-29
Mention the pragma inline in function.dd

BBasile:
The previous paragraph became misleading since inlining can be controlled and checked.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -872,15 +872,13 @@ void main()
 
 $(H4 $(LNAME2 inline-functions, Inline Functions))
 
-        $(P There is no inline keyword. The compiler makes the decision whether to
-        inline a function or not, analogously to the register keyword no
-        longer being relevant to a
-        compiler's decisions on enregistering variables.
-        (There is no register keyword either.))
-
-        $(P If a $(GLINK2 expression, FunctionLiteral) is immediately called,
-        its inlining would be enforced normally.)
-
+        $(P The compiler makes the decision whether to inline a function or not. 
+        This decision may be controlled by $(LINK2 pragma.html#inline, `pragma(inline)`),
+        assuming that the compiler implements it, which is not mandatory.)
+        
+        $(P Note that any $(GLINK2 expression, FunctionLiteral) should be inlined
+        when used in its declaration scope.
+        )
 
 $(H3 $(LNAME2 function-overloading, Function Overloading))
 


spec/function.dd
```diff
@@ -872,15 +872,10 @@ void main()
 
 $(H4 $(LNAME2 inline-functions, Inline Functions))
 
-        $(P There is no inline keyword. The compiler makes the decision whether to
-        inline a function or not, analogously to the register keyword no
-        longer being relevant to a
-        compiler's decisions on enregistering variables.
-        (There is no register keyword either.))
-
-        $(P If a $(GLINK2 expression, FunctionLiteral) is immediately called,
```
ZombineDev: Why remove this paragraph?

spec/function.dd
```diff
@@ -872,15 +872,10 @@ void main()
...
-        its inlining would be enforced normally.)
-
+        $(P There is no `inline` keyword, instead the compiler makes the decision whether to
```
ZombineDev: I would just remove "There is no `inline` keyword, instead" altogether, because it sounds confusing.

spec/function.dd
BBasile: imagine a `pragma(inline, false)` before the Function literal and then what's said in this paragraph is wrong.

BBasile: I see it was about thelambdas...this was a n error to remove it.

spec/function.dd
```diff
@@ -872,15 +872,12 @@ void main()
...
 $(H4 $(LNAME2 inline-functions, Inline Functions))
 
-        $(P There is no inline keyword. The compiler makes the decision whether to
-        inline a function or not, analogously to the register keyword no
-        longer being relevant to a
-        compiler's decisions on enregistering variables.
-        (There is no register keyword either.))
-
+        $(P The compiler makes the decision whether to inline a function or not. 
+        This decision may be controlled by the $(LINK2 pragma.html#inline, `pragma(inline)`),
```
andralex: drop the `the`

ZombineDev:
LGTM

spec/function.dd
```diff
@@ -872,15 +872,12 @@ void main()
...
-        (There is no register keyword either.))
-
+        $(P The compiler makes the decision whether to inline a function or not. 
+        This decision may be controlled by $(LINK2 pragma.html#inline, `pragma(inline)`),
+        assuming that the compiler implements it, which is not mandatory.)
+        
         $(P If a $(GLINK2 expression, FunctionLiteral) is immediately called,
-        its inlining would be enforced normally.)
-
+        its inlining would be enforced normally.)
```
schuetzm: What does "would be enforced normally" actually mean? I suppose, "users can expect the call to be inlined"? Please change it accordingly, if this is the intended meaning.

BBasile: @schuetzm, i agree, even if i was note the author. Rewritten with more accurate terms.

PR #1455, opened 2016-08-29
FAQ: D frontend is not C++ anymore

Cauterite:
>Why is the D front end written in C++ rather than D?

This shouldn't be frequently asked any more, since it's no longer true.

Also why is this entry named "dogfood" ?


JackStouffer:
👍

AndrejMitrovic:
> Also why is this entry named "dogfood" ?

Probably because someone thought of the expression _eating your own dogfood_. I guess they were hopeful that some day the front-end would be written in D. :)

PR #1456, opened 2016-08-29
fix issue 15442 - Eponymous template restrictions should be documented

BBasile:


spec/template.dd
```diff
@@ -868,21 +868,55 @@ $(H2 $(LNAME2 template_parameter_def_values, Template Parameter Default Values))
...
+
+        void main()
+        {
+            foo(1, 2, "test"); // foo!(int, int).foo(1, 2, "test")
+            foo(1, 2); // foo!(int, int).foo(1, 2)
+        }
+        ------
+
+        The eponymous members can't rely on a `static if` condition since the
+        parameters deduction relies on how the in members are used:
```
klickverbot: This is not correct as written. You could still use it as `foo!int(5);` just fine.

BBasile:
Ok, I've put the emphasize on the fact that this restriction only applies with IFTI. The explanation comes from https://issues.dlang.org/show_bug.cgi?id=1820 btw.

spec/template.dd
```diff
@@ -868,21 +868,57 @@ $(H2 $(LNAME2 template_parameter_def_values, Template Parameter Default Values))
 
 $(H2 $(LNAME2 implicit_template_properties, Eponymous Templates))
 
-    $(P If a template has exactly one member in it, and the name of that
-        member is the same as the template name, that member is assumed
-        to be referred to in a template instantiation:
+    $(P If a template contains members whose the name is the same as the
+        template identifier and that the type or the parameters type of these
```
andralex: s/and that the type/and if the type/

if I understand correctly

spec/template.dd
```diff
@@ -868,21 +868,57 @@ $(H2 $(LNAME2 template_parameter_def_values, Template Parameter Default Values))
...
+        members include at least all the template parameters then these members
+        are assumed to be referred to in a template instantiation:
 
         ------
-        template $(CODE_HIGHLIGHT Foo)(T)
+        template Foo(T)
         {
             T Foo; // declare variable Foo of type T
```
andralex: can you also s/Foo/foo/ please

BBasile:
Ok, the substitutions are made.

PR #1459, opened 2016-08-30
Remove comparisons with C++

WalterBright:
Also removed some obsolete points. The overview could still be a lot better, but this fixes some of the worst of it.


quickfur:
LGTM.

WalterBright:
Thanks! When can the website http://dlang.org/overview.html be updated?

wilzbach:
> When can the website http://dlang.org/overview.html be updated?

It's automatically deployed to dlang.org (sometimes with a bit of a delay). At least as of now it's live.

PR #1460, opened 2016-08-30
Fix relative links in deprecate page

wilzbach:
resubmission of #1445 which now contains  `<a name="$1">`  for graceful deprecation of whitespace links.

### DIFF
--- a/deprecate.dd
+++ b/deprecate.dd
@@ -9,40 +9,40 @@ $(SPEC_S Deprecated Features,
     )
 
     $(TABLE2 Deprecated Features,
-        $(THEAD Feature,                                                          Spec,  Dep,    Error,  Gone)
-        $(TROW $(DEPLINK Using the result of a comma expression),                 future, &nbsp;, &nbsp;, &nbsp;)
-        $(TROW $(DEPLINK delete),                                                 future, &nbsp;, &nbsp;, &nbsp;)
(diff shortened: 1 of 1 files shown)


deprecate.dd
```diff
@@ -713,7 +713,7 @@ $(H4 Rationale)
 
 
 
-$(H3 $(DEPNAME if (v; e)))
+$(H3 $(DEPNAME if (v; e, variable-delcaration-in-if-without-auto)))
```
aG0aep6G: Misplaced closing parenthesis here. `if (v; e` needs to be closed before `, variable-delcaration-in-if-without-auto`.

Also a typo: de**lc**aration. Needs to be fixed in the link, too.

deprecate.dd
```diff
@@ -814,8 +814,8 @@ $(H4 Rationale)
 )
 
 Macros:
-	DEPLINK=$(RELATIVE_LINK2 $0, $0)
+	DEPLINK=$(RELATIVE_LINK2 $1, $2)
 	DEPLINK2=$(LINK2 $1.html#$2, $2)
-	DEPNAME=$(LNAME2 $0, $0)
+	DEPNAME=<a name="$1">$(LNAME2 $2, $1)
```
aG0aep6G: Need to close the `<a>`.

Can we use the same order of arguments as with DEPLINK?

PR #1461, opened 2016-08-30
removed unused ddoc right paranthesis

wilzbach:
image


changelog/2.071.2_pre.dd
```diff
@@ -70,7 +70,6 @@ $(LI $(BUGZILLA 16225): [REG 2.068] Internal error cod1.c 1338 with -O)
 $(LI $(BUGZILLA 16316): [REG 2.071] $(LPAREN)Import deprecation$(RPAREN) fully qualified name of imports in mixin template not accessible)
 $(LI $(BUGZILLA 16348): [REG 2.070.2] ICE with package visibility)
 )
-)
```
aG0aep6G: I think this was the wrong one to remove. Three closing parentheses should be correct: P, BUGSTITLE, VERSION.

wilzbach: seems like ddoc doesn't care: http://dlang.org/changelog/2.071.2.html, but better fix the mess: #1463

PR #1462, opened 2016-08-31
more fixes for overview.dd

WalterBright:
A second pass through overview.dd

### DIFF
--- a/overview.dd
+++ b/overview.dd
@@ -10,8 +10,8 @@ $(DIVC page-contents quickindex,
         $(LI $(RELATIVE_LINK2 what_is_d, What is D?))
         $(LI $(RELATIVE_LINK2 why_d, Why D?)$(OL
             $(LI $(RELATIVE_LINK2 goals, Major Design Goals of D))
-            $(LI $(RELATIVE_LINK2 features_to_keep, Features To Keep From C++))
-            $(LI $(RELATIVE_LINK2 features_to_drop, Features To Drop From C++))
+            $(LI $(RELATIVE_LINK2 features_to_keep, Transitioning To D))
+            $(LI $(RELATIVE_LINK2 features_to_drop, Features To Leave Behind))
             $(LI $(RELATIVE_LINK2 for_d, Who D is For?))
             $(LI $(RELATIVE_LINK2 not_for_d, Who D is Not For?))
         ))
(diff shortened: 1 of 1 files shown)


WalterBright:
This could really use the attention of a better writer than I am.

overview.dd
```diff
@@ -419,12 +414,10 @@ $(SECTION3 $(LNAME2 productivity, Productivity),
...
     $(SECTION4 Declaration vs Definition,
 
-    $(P C++ usually requires that functions and classes be declared twice - the declaration
-    that goes in the .h header file, and the definition that goes in the .c source
-    file. This is an error prone and tedious process. Obviously, the programmer
-    should only need to write it once, and the compiler should then extract the
-    declaration information and make it available for symbolic importing. This is
-    exactly how D works.
+    $(P Functions and classes are defined once. There is no need
+    for declarations when they are forward referenced.
```
AndrejMitrovic: `no need for declarations` => `no need for forward declarations`

An example speaks more than words though, I'd add this below:

```
For example:

-----
class C : D { }  // D is already visible from here
class D { }  // does not need to be forward-declared
-----
```

overview.dd
```diff
@@ -419,12 +414,10 @@ $(SECTION3 $(LNAME2 productivity, Productivity),
...
+    A module can be imported, and all its public declarations
```
AndrejMitrovic: I think this is in the wrong section? It should be under the modules but it's listed under declaration vs definition.

AndrejMitrovic: And I'd word it as: "its public declarations become available to the importing module".

Technically we also have package symbols which might be available to the module depending on the package it belongs to.

overview.dd
```diff
@@ -439,36 +432,7 @@ class ABC
...
-
-class Bar
-{
-  public:
-    int bar() { return 3; }
-};
-)
-
-    $(P But the equivalent D code will work:
+    $(P All members are defined in the class or struct, not separately.
```
AndrejMitrovic: Perhaps "inside the class or struct" and "not externally" sounds better.

overview.dd
```diff
@@ -876,15 +840,10 @@ $(SECTION3 $(LNAME2 resource, Resource Management),
...
-    $(P D memory allocation is fully garbage collected. Empirical experience
-    suggests that a lot of the complicated features of C++ are necessary
-    in order to manage memory deallocation. With garbage collection, the
-    language gets much simpler.
-    )
-
-    $(P Garbage collection eliminates the tedious, error prone memory
-    allocation
-    tracking code necessary in C and C++. This not only means much
+    $(P D memory allocation is fully garbage collected.
```
AndrejMitrovic: This will raise red flags for people who think you can't allocate memory outside the garbage collector. Perhaps say "By default, D memory allocation.." but add a small sentence stating that manual memory management is still possible.

AndrejMitrovic:
Small nitpicks to fix and then LGTM.

andralex:
@AndrejMitrovic could you please implement yours as a subsequent pass? Let's empty @WalterBright 's plate a bit...

PR #1463, opened 2016-08-31
Cleanup parentheses in changelog 2.072.2_pre

wilzbach:


aG0aep6G:
Looks good to me.

Just noticed that the `P` around the `LI`s doesn't make sense. -> #1464

MartinNowak:
Stable would've been better.

wilzbach:
> Stable would've been better.

But the master branch gets automatically deployed to dlang.org ...

PR #1464, opened 2016-08-31
remove wrong `P` around `LI`s

aG0aep6G:


wilzbach:
Shouldn't in theory there be a `UL` around the `LI`s?
I guess we are lucky that HTML5 is so error-tolerant (this change doesn't change the visual output)

Btw I am really looking forward to the 2.072 branch-off, s.t. we can introduce more automation to this error-prone process.

aG0aep6G:
> Shouldn't in theory there be a UL around the LIs?

`BUGSTITLE` puts the `LI`s into an `OL` ([in changelog.ddoc](https://github.com/dlang/dlang.org/blob/8a7ffb28d2e96759efad8bd8dddfc8fbfc6e3e42/changelog/changelog.ddoc#L35)).

wilzbach:
> BUGSTITLE puts the LIs into an OL (in changelog.ddoc).

Oh sorry, LGTM then ;-)

AndrejMitrovic:
Wasn't this autogenerated with changed.d? So maybe the script needs to be fixed if using that macro is wrong.

aG0aep6G:
> Wasn't this autogenerated with changed.d? So maybe the script needs to be fixed if using that macro is wrong.

Indeed.-> https://github.com/dlang/tools/pull/199

MartinNowak:
Please stop manually editing those files and fix the tool instead.

MartinNowak:
Also there are plenty of existing changelogs w/ those <p> tags, so either fix all or none of them.

andralex:
tool got fixed

aG0aep6G:
> either fix all or none of them.

Fixing all of them: https://github.com/dlang/dlang.org/pull/1479

PR #1465, opened 2016-09-01
remove class {de,}allocators from the D spec

wilzbach:
> Class allocators are deprecated in D2.
> Class deallocators are deprecated in D2.

I couldn't find it on the [deprecation page](https://dlang.org/deprecate.html), so I assume it's a pretty old deprecation that went undocumented.

I don't know whether there was a ny rationale in keeping it in the spec. I guess initially to ease the migration?
If that's the case we could still move them to the deprecation page, but imho the spec should be up-to-date and not mention stuff from D1.

### DIFF
--- a/spec/class.dd
+++ b/spec/class.dd
@@ -66,8 +66,6 @@ $(UL
             $(LI $(GLINK SharedStaticDestructor)s)
             $(LI $(RELATIVE_LINK2 invariants, Class Invariants))
             $(LI $(DDSUBLINK spec/unittest, unittest, Unit Tests))
-            $(LI $(RELATIVE_LINK2 allocators, Class Allocators))
-            $(LI $(RELATIVE_LINK2 deallocators, Class Deallocators))
             $(LI $(RELATIVE_LINK2 alias-this, Alias This))
         )
         )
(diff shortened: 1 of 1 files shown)


ZombineDev:
No, they are still usable

AndrejMitrovic:
It should definitely be listed on the deprecations page.

They were being de-facto deprecated for many years but they still work in D2 The two small deprecation notes on this page were added back in 2012. Unfortunately we didn't really have a good deprecation & release process back then so this feature remained in the language.

I think this is the thread which initially started the talk about removing class (de)allocators: http://forum.dlang.org/post/hafpjn$1cu8$1@digitalmars.com

There's also this issue about warning / deprecation in the compiler: https://issues.dlang.org/show_bug.cgi?id=12081

Anyway, I think we need to properly deprecate them (via compiler warnings), and then **later** remove support for them at the same time as removing them from the spec as this pull does.

Do we have a deprecations lifecycle documented somewhere?

BBasile:
- libdparse explicitly doesn't handle them.
- we have the allocators now (make , copy the initializer, emplace, etc are available)

goodbye the class de/allocators

WalterBright:
Going to pass on this for now, pending coming up with a good RC object design.

PR #1466, opened 2016-09-01
Make opAssign return ref

Bolpat:
Assignment should return the assigned object and not be void. The example is misleading.

### DIFF
--- a/spec/operatoroverloading.dd
+++ b/spec/operatoroverloading.dd
@@ -532,10 +532,10 @@ $(H2 $(LEGACY_LNAME2 Assignment, assignment, Assignment Operator Overloading))
     struct S
     {
         // identiy assignment, allowed.
-        void $(CODE_HIGHLIGHT opAssign)(S rhs);
+        ref S $(CODE_HIGHLIGHT opAssign)(S rhs);
 
         // not identity assignment, also allowed.
-        void $(CODE_HIGHLIGHT opAssign)(int);
+        ref S $(CODE_HIGHLIGHT opAssign)(int);
     }
     S s;
     s = S();      // Rewritten to s.opAssign(S());


AndrejMitrovic:
That's not true at all. Where are you getting this from?

Bolpat:
> Where are you getting this from?

From my experience with *all* C-like languages I know. One can always do `x = y = z` (except for no-copy types) and `if ((a = f()) == b) { }` for built-in types and usually for user defined types. I've been taught to define assignment to work like this and never different.

Maybe I've said it wrong. The result of assignment is (a reference to) the lvalue after being assigned to (and not the value being assigned as one could interpret the term *assigned object*). It should not be `void`. At least in [Ali Çehreli's Book](http://ddili.org/ders/d.en/special_functions.html) it is the assigned lvalue, but not a reference to it which is an error from what I know.

When I read the example the first time, I concluded the D compiler made the `return this` and `ref` implicit.

> That's not true at all.

Give me a proper counterexample. Where is it useful *not* to return the lvalue after being assigned to?

I've taken a look through Phobos and found [Nullable's opAssign](https://dlang.org/library/std/typecons/nullable.op_assign.html) that returns `void`. I also found [BigInt's opAssign](https://dlang.org/library/std/bigint/big_int.op_assign.html) that does return `BigInt`. At least [BitFlags's opAssign](https://github.com/dlang/phobos/blob/dc56fc161572f9f70ee470a98fe84186c8190df9/std/typecons.d#L6748) returns `ref` sometimes.

Interestingly, any C++ course teaches to define assignment as `my_type& operator=(...)` but in D this seems not to be the case. I'd really like to understand the reason to it, so I've asked this in the [Forum](https://forum.dlang.org/post/ezeuvnvsfdpuaicgrmwr@forum.dlang.org).

AndrejMitrovic:
Returning `ref` can be useful for chaining assignments, sure. But it's not an actual requirement.

PR #1474, opened 2016-09-16
Clear up some array usage

schveiguy:
see https://forum.dlang.org/post/ivjnopgnimqedmsnxpeo@forum.dlang.org

### DIFF
--- a/spec/arrays.dd
+++ b/spec/arrays.dd
@@ -99,16 +99,10 @@ int* p;
 int[3] s;
 int[] a;
 
-int* q;
-int[3] t;
 int[] b;
 
-p = q;     // p points to the same thing q does.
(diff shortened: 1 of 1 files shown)


schveiguy:
Don't pull yet, just realized that the text is still incorrect

schveiguy:
OK, should be good now.

schveiguy:
Wait, `s = a` actually compiles.

Don't pull yet. I need to figure out what that line was actually supposed to be doing.

schveiguy:
So that line was there from the beginning. I think it was an error, or it may not have worked in the early days.

However, it now works, and does something completely different than what the section is talking about -- copying the elements.

I also removed the superfluous explanation that copying a pointer copies it.

PR #1481, opened 2016-09-19
use defined bootstrap compiler to build dmd

MartinNowak:
- avoids dependency on system dmd which can have any version (even newer
  than latest release), and might fail to build a correct compiler
- fixes auto-deployment of dlang.org which uses master dmd


CyberShadow:
LGTM :)

PR #1482, opened 2016-09-20
Added druntime's runtime gc switch to documentation

Jebbs:
Not sure if any other information should be added. It might be a good idea since a manual implementation was added, but I wasn't sure where to add that bit in the documentation.


Jebbs:
ping @MartinNowak

PR #1484, opened 2016-09-23
Update to latest DDOX.

s-ludwig:
- fixes an issue with macro parameter handling in conjunction with $+
- fixes a double-escape issue within the $(D) macro


s-ludwig:
Investigating the range error - would be an easy fix, but it's not obvious why std.getopt triggers it in the first place.

s-ludwig:
Need to put this on hold for a bit. Something in the macro processing code broke due to the first fix.

s-ludwig:
The `ndslice.slice` module is missing the "caption" parameter to the `BOOKTABLE` macro, so that the actual table contents are treated as the caption.

Otherwise looks to be alright now.

PR #1486, opened 2016-09-25
Specifies that `#line` with FileSpec doesn't create a new module

BBasile:


BBasile:
@CyberShadow , what's happening with DAutoTest ? I remember that this one was green last week.

PR #1489, opened 2016-10-01
Fix a link and typo in the changelog

9il:


MartinNowak:
Should've been on stable!!!

wilzbach:
> Should've been on stable!!!

Why? Afaik Master gets directly deployed because we want to avoid a stalled web page

MartinNowak:
> Why? Afaik Master gets directly deployed because we want to avoid a stalled web page

Because documentation incorporated into releases is build from stable. You can merge back stable into master as often as you want.

9il:
PR for stable: #1496

PR #1490, opened 2016-10-03
latex.ddoc: Remove breakatwhitespace

CyberShadow:
The option seems to be gone in the 2016 edition of the listings package.

This should fix the documentation CI.


CyberShadow:
Looks like it went from broken to broken:

Before:



After:

CyberShadow:
@andralex 

Fix for that comment: https://github.com/dlang/dlang.org/pull/1491

CyberShadow:
Thanks!

MartinNowak:
cherry picked to stable as 1fb1342a8115a666887d63bd0d85ce39a4d25a2e

PR #1491, opened 2016-10-03
Convert HTML comments to DDoc comments

CyberShadow:
The target output is not always HTML.

Fixes HTML comments visible in the PDF.


andralex:
ugh thx

andralex:
I merged in spite of the doc tests failing, could you please follow up with a fixup

CyberShadow:
Please look at #1490 foremost :) The doc tests are broken for everything #1490 should fix it.

PR #1493, opened 2016-10-06
Fix Issue 16592 - Building dlang.org does not work without a preexist…

Darredevil:
…ing dmd installation

https://issues.dlang.org/show_bug.cgi?id=16592


andralex:
QED (Quite Easily Done), thanks!

PR #1497, opened 2016-10-12
Minor tweaks to fix current Ddoc themes due to the new Ddoc theme

jacob-carlborg:
The new Ddoc theme [1] introduced three new macros: DDOC_MEMBER,
DDOC_MEMBER_HEADER and DDOC_HEADER_ANCHOR. This tweak configures these
new macros to have no impact on the existing ddoc themes.

[1] https://github.com/dlang/dmd/pull/6173


jacob-carlborg:
@andralex @CyberShadow fixed the existing dlang.org ddoc theme.

CyberShadow:
Awesome, thanks for the quick PR. Waiting on the doc autotester.

jacob-carlborg:
Since the new default macros are in a separate file, can this [1] section either be referencing the file on GitHub or somehow included in the `ddoc.dd` file? Then it would also always be up to date.

[1] http://dlang.org/spec/ddoc.html#predefined_macros

jacob-carlborg:
Thanks.

CyberShadow:
> Since the new default macros are in a separate file, can this [1] section either be referencing the file on GitHub or somehow included in the ddoc.dd file? Then it would also always be up to date.

I think it's doable with a Makefile rule.

andralex:
Nice, thanks very much @CyberShadow for watching things over.

jacob-carlborg:
Which approach is preferred?

andralex:
I'll let you and @CyberShadow consult on that. Works?

jacob-carlborg:
Works for me.

CyberShadow:
Well, I really should be working on other stuff right now. It seems like a low-priority thing, but if you're up for it, put up a PR, and if it passes the doc tester with a good diff ping me and I'll merge it.

jacob-carlborg:
@CyberShadow I can do the work myself, I just wanted an opinion on which approach to use.

CyberShadow:
Sorry, what are the options? Aside from the Makefile copying the file from the DMD repo to the dlang repo, then including it into the DDoc.

CyberShadow:
I guess you could also add a special DDoc macro that expands tho the default macro list (excluding itself of course), but I wouldn't be a good judge on how sensible that would be.

jacob-carlborg:
> Sorry, what are the options? Aside from the Makefile copying the file from the DMD repo to the dlang repo, then including it into the DDoc.

The other would be to link directly to the file on GitHub.

CyberShadow:
Ah, right. Well, that's an option, but it would have some issues:

- Documentation of older versions ought to remain working. It's not impossible that the link might change in the future.
- Some formats of documentation are for off-line use (PDF, CHM, HTML documentation included with DMD).
- The documentation should match the corresponding D release it was generated from, so at least the link should point to the specific tag/commit.

As things are, that file doesn't change very often, does it? Perhaps it's simplest to update the list manually for now.

jacob-carlborg:
> Ah, right. Well, that's an option, but it would have some issues

True.

> As things are, that file doesn't change very often, does it? Perhaps it's simplest to update the list manually for now.

It needs escaping, so it's not as easy as copy pasting the file.

PR #1502, opened 2016-10-23
[changelog] Refine comma expression examples

ntrel:
- Remove cruft, make clearer.
- Add link to comma bug examples.

Original PR: https://github.com/dlang/dmd/pull/5737

### DIFF
--- a/changelog/2.072.0.dd
+++ b/changelog/2.072.0.dd
@@ -120,30 +120,32 @@ $(HR)
 $(BUGSTITLE Language Changes,
     $(LI $(LNAME2 deprecated_commaexp, Using the result of a comma expression is deprecated.)
 
-    $(P Comma expressions expressions have proven to be a frequent source of confusion, and bug.
-        Using their result will now trigger a deprecation message.)
+    $(P Comma expressions have proven to be a frequent source of confusion, and bugs -
+    see $(HTTPS issues.dlang.org/show_bug.cgi?id=2659#c2, here) for examples.)
+
+    $(P Using the result of a comma expression will now trigger a deprecation message.)
 
     $(P Example:)
 
         ---
-        module comma;
-
         void main () {
-          size_t aggr;
-          MyContainerClass mc;
+          // OK, result of increment not directly used
+          for (int i, j; i != 3; i++, j++) {...}
+
           auto r1 = getRange!int, r2 = getRange!int;
-          // This is okay
-          for (; !r1.empty && r2.empty; r1.popFront, r2.popFront)
-            aggr += compute(r1.front, r2.front);
-          // This is not, as only the 3rd part (increment) allows commas
-          for (; !r1.empty, r2.empty; r1.popFront, r2.popFront)
-            aggr += compute(r1.front, r2.front);
-          // This is okay, the result is not used.
+          // OK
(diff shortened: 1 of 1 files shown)


changelog/2.072.0_pre.dd
```diff
@@ -124,30 +124,32 @@ $(HR)
 $(BUGSTITLE Language Changes,
     $(LI $(LNAME2 deprecated_commaexp, Using the result of a comma expression is deprecated.)
 
-    $(P Comma expressions expressions have proven to be a frequent source of confusion, and bug.
-        Using their result will now trigger a deprecation message.)
+    $(P Comma expressions have proven to be a frequent source of confusion, and bugs -
+    see $(LINK2 https://issues.dlang.org/show_bug.cgi?id=2659#c2, here) for examples.)
```
andralex: Use HTTPS instead of LINK2

aG0aep6G: > Use HTTPS instead of LINK2

Or don't. HTTPS doesn't buy us anything. This is getting old. I'm a little disappointed that you just keep pushing for it, @andralex, seemingly without reason. ([previously](https://github.com/dlang/phobos/pull/4411#issuecomment-226392170))

andralex: The simple reason is it's shorter. I've written a bunch of documentation and it helps. Please let's not argue on something this petty - just (preferably) do it or don't. I've approved the PR already so this is advisory. Thanks.

andralex: >The simple reason for LINK2 is that you can copy/paste from the browser. I think this wins.

I understand. It seems otherwise to me. Thanks.

ntrel:
Changed to use `HTTPS`, rebased. Sorry for the delay ;-)

changelog/2.072.0_pre.dd
CyberShadow: @andralex I think all of myself, @aG0aep6G, @wilzbach and anyone with an opinion agrees that `HTTPS` and such are harmful and should be replaced. An argument I haven't seen mentioned is that with `LINK2` etc. the link appears as it is in the source code in the same way as it will be in HTML and the browser URL. With `HTTP`/`HTTPS` it's no longer an URL, but some macro cacophony that we have created. This makes e.g. grepping for links harder. There is also the semantic issue that it moves the URL protocol, which is part of the URL (a self-contained reference to something) into a macro (which we use semantics for), thus it moves information across a boundary that should be left alone. The 6-7 characters saved in exchange make a poor contrary position, considering the length of typical URLs. As such, I recommend dropping this argument and converting the `HTTP`/`HTTPS` macros to `LINK2` or whatever whenever convenient.

CyberShadow: Also not sure what you meant by "The simple reason is it's shorter. I've written a bunch of documentation and it helps.", as URLs should never be typed, only copied. As such, removing the protocol and turning it into a macro actually creates more work than leaving the URL intact.

MartinNowak:
Mmpf, changes to the changelogs should go into stable branches so they are reflected in the documentation of point releases.

MartinNowak:
Mmpf, changes to the changelog should go into the stable branch so they are reflected in point releases.

PR #1503, opened 2016-10-23
changelog entries for unrestricted unions/scope value params

MartinNowak:
- see https://github.com/dlang/dmd/pull/5897
- see https://github.com/dlang/dmd/pull/5830


MartinNowak:
Ping, needed for 2.072.

MartinNowak:
Will have to merge this w/ a few more changelog edits.

PR #1504, opened 2016-10-25
add dlangupb-scholarship.dd

WalterBright:


andralex:
Thanks! I'll make an announcement shortly.

PR #1505, opened 2016-10-25
add dlangupb-scholarship to makefiles

WalterBright:

PR #1506, opened 2016-10-26
repair (hopefully) win32.mak build target

MartinNowak:
- fixup for #1505


MartinNowak:
Actually the win32.mak target should no longer generate any html files in our build script, they get copied over from linux. Maybe you're not building that html page on linux?

PR #1507, opened 2016-10-28
The foundation is tax-exempt

schveiguy:
We should proudly share that info!

### DIFF
--- a/foundation.dd
+++ b/foundation.dd
@@ -53,6 +53,9 @@ $(P If you want to help the D Language to advance further and have resources
     Your help is highly appreciated!
 )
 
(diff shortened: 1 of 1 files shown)


JackStouffer:
As this is a static page, it would make more sense to frame it as a statement of fact rather than an announcement IMO.

schveiguy:
It can be changed to something different. I wanted to put some sort of positive exclamation in there, and this is the only thing I could think of that sounded reasonable. Any other ideas?

CyberShadow:
@andralex

schveiguy:
Thanks!

PR #1510, opened 2016-11-01
Qualify C-style memory allocation functions as pure

nordlow:
Update according to https://github.com/dlang/druntime/pull/1683

I'm not sure if this is enough or if I should update the general formulation of purity aswell.

Made an existing statement more clear via a neither-nor formulation.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -173,7 +173,7 @@ int foo();
 
 $(H4 $(LNAME2 pure-functions, Pure Functions))
 
-        $(P Pure functions are functions which cannot access global or static, mutable
+        $(P Pure functions are functions which cannot access neither global nor static mutable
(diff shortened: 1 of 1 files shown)


spec/function.dd
```diff
@@ -173,7 +173,7 @@ int foo();
 
 $(H4 $(LNAME2 pure-functions, Pure Functions))
 
-        $(P Pure functions are functions which cannot access global or static, mutable
+        $(P Pure functions are functions which cannot access neither global nor static mutable
```
UplinkCoder: add , back in

nordlow: Do you mean I should undo this fix?

UplinkCoder: No neither global nor static {,} mutable state

PR #1511, opened 2016-11-05
name mangling: bring documentation closer to actual implementation

rainers:
This documents what I found in the current implementation. Not everything seems desirable, though...

### DIFF
--- a/spec/abi.dd
+++ b/spec/abi.dd
@@ -206,6 +206,7 @@ $(H3 $(LNAME2 name_mangling, Name Mangling))
     $(GNAME MangledName):
         $(B _D) $(GLINK QualifiedName) $(GLINK Type)
         $(B _D) $(GLINK QualifiedName) $(B M) $(GLINK Type)
+        $(B _D) $(GLINK QualifiedName) $(B Z)        $(GREEN // Internal)
 
     $(GNAME QualifiedName):
         $(GLINK SymbolName)
@@ -214,6 +215,7 @@ $(H3 $(LNAME2 name_mangling, Name Mangling))
     $(GNAME SymbolName):
         $(GLINK LName)
         $(GLINK TemplateInstanceName)
+        $(B 0)             $(GREEN // anonymous symbols)
     )
 
 	$(P The $(B M) means that the symbol is a function that requires
@@ -225,7 +227,11 @@ $(H3 $(LNAME2 name_mangling, Name Mangling))
 
     $(GRAMMAR
         $(GNAME TemplateInstanceName):
-            $(GLINK Number) $(B __T) $(GLINK LName) $(GLINK TemplateArgs) $(B Z)
+            $(GLINK Number) $(GLINK TemplateID) $(GLINK LName) $(GLINK TemplateArgs) $(B Z)
+
+        $(GNAME TemplateID):
+            $(B __T)
+            $(B __U)        $(GREEN // for symbols declared inside template constraint)
 
         $(GNAME TemplateArgs):
             $(GLINK TemplateArg)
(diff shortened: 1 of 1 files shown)


spec/abi.dd
```diff
@@ -206,12 +206,14 @@ $(H3 $(LNAME2 name_mangling, Name Mangling))
...
+        $(B _D) $(GLINK QualifiedName) $(B Z)        $(GREEN // Internal)
 
     $(GNAME QualifiedName):
         $(GLINK SymbolName)
         $(GLINK SymbolName) $(I QualifiedName)
 
     $(GNAME SymbolName):
+        $(B 0)             $(GREEN // anonymous symbols)
         $(GLINK LName)
         $(GLINK TemplateInstanceName)
```
MartinNowak: Let's move the less common anonymous symbol case to the bottom.

rainers: Amended.

BTW: I've only seen this with overload sets as template arguments, but it looks like a bug to me: the overloaded symbol has these anonymous parents, but something like the location of the lookup should be encoded instead.

wilzbach:
@rainers: how about linking to the bison spec file that you curated? I imagine that this could help other people as well  (at least those reading the spec about name mangling)

rainers:
@wilzbach A bison grammar matching the documentation is now here: https://gist.githubusercontent.com/rainers/6cdf73b48837defb9f88/raw/161db3fa7424fce90782b73fa1b12975074a9a22/dwebsite3.bison

Not sure where to place and link to it. I think it would have to be part of the dlang.org repository to have a chance of staying in sync.

BTW: It shows 2 ambiguities in the grammar:
- 'M' can be `scope` both on the function itself or the first function argument.
- `Y` can be the end of an function argument list or the beginning of the type of an Objective-C function argument.

wilzbach:
> Not sure where to place and link to it. I think it would have to be part of the dlang.org repository to have a chance of staying in sync.

How about creating a new `data` or `bison` folder?
What is your opinion @CyberShadow?

CyberShadow:
IMO? Put it in the same directory as the name mangling spec.

Also weird, can't comment via the web interface:
https://dump.thecybershadow.net/3b56e04862e0c61b7e3135ac95d3b129/13%3A14%3A06-upload.png

On 2016-12-19 08:30, Sebastian Wilzbach wrote:
>> Not sure where to place and link to it. I think it would have to be part of the dlang.org repository to have a chance of staying in sync.
>
> How about creating a new `data` or `bison` folder?
> What is your opinion @CyberShadow?
>

-- 
Best regards,
  Vladimir

spec/abi.dd
```diff
@@ -238,11 +244,11 @@ $(H3 $(LNAME2 name_mangling, Name Mangling))
         $(GNAME TemplateArgX):
             $(B T) $(GLINK Type)
             $(B V) $(GLINK Type) $(GLINK Value)
-            $(B S) $(GLINK LName)
+            $(B S) $(GLINK SymbolName)
+            $(B S) $(GLINK MangledName)
```
ibuclaw: Actually, this should be `QualifiedName` / `MangledName`.

Example:  https://github.com/dlang/phobos/blob/31e1b462d2311f9d8cecb33172b28eab587b255d/std/parallelism.d#L43-L77

Instantiates this symbol:
```
std.parallelism.Task!(run, real delegate(MapResult!(getTerm, Result), ulong, ulong) @system, MapResult!(getTerm, Result), ulong, ulong).Task.impl
```

Mangled as:
```
_D3std11parallelism301__T4TaskS213std11parallelism3runTDFS3std9algorithm9iteration91__T9MapResultS31_D8parallel4mainFZ7getTermMFiZeTS3std5range13__T4iotaTiTiZ4iotaFiiZ6ResultZ9MapResultmmZeTS3std9algorithm9iteration91__T9MapResultS31_D8parallel4mainFZ7getTermMFiZeTS3std5range13__T4iotaTiTiZ4iotaFiiZ6ResultZ9MapResultTmTmZ4Task4implFPvZv
```

Right at the beginning of the first template you have:
```
__T4TaskS213std11parallelism3run
           ^
```
Also, this cunningly forgets Bugzilla 3043 where symbol arguments are prefixed with their length.

The correct grammar should be:
```
         $(GNAME TemplateArgX):
             $(B T) $(GLINK Type)
             $(B V) $(GLINK Type) $(GLINK Value)
            $(B S) $(GLINK LName)
            $(B S) $(GLINK Number) $(GLINK QualifiedName)
            $(B S) $(GLINK Number) $(GLINK MangledName)
```

rainers: Yeah, noticed this too last night, see https://issues.dlang.org/show_bug.cgi?id=14591#c7

PR #1512, opened 2016-11-08
fix Issue 16147 - Provide shiny 404 error pages

CyberShadow:
https://issues.dlang.org/show_bug.cgi?id=16147

### DIFF
--- a/404.dd
+++ b/404.dd
@@ -1,0 +1,12 @@
+Ddoc
+
+$(D_S 404 Not Found,
+
+$(P
+Page not found!
+
+If you think there should be something here, please use the "Report a bug" link above to let us know.
+))
+
+Macros:
+    TITLE=404 Not Found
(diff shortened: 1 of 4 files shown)


404.dd
```diff
@@ -0,0 +1,12 @@
+Ddoc
+
+$(D_S 404 Not Found,
+
+$(P
+Page not found!
+
+If you think there should be something here, please use the "Report a bug" link above to let us know.
```
wilzbach: How about directly putting a report [link](https://issues.dlang.org/enter_bug.cgi?bug_file_loc=http%3A%2F%2Fdlang.org/&bug_severity=enhancement&component=dlang.org&op_sys=All&priority=P3&product=D&rep_platform=All&short_desc=%5BPage%20Not%20Found%5D&version=D2) here? One barrier less for a potential reporter ;-)

CyberShadow: DRY. If the bugzilla URL or link format changes, that's one less link to fix.

wilzbach: Fair point. Sorry for the nitpick.

PR #1516, opened 2016-11-20
put auto-highlights in <code> tags

aG0aep6G:
Makes them visually consistent with other code (monospaced font).
Carries more meaning that can be picked up by screen readers and such.


andralex:
@CyberShadow lgty?

CyberShadow:
Thanks!

PR #1517, opened 2016-11-21
[examples bug fix] template.dd

xiren7:
```D
int x;

template Foo(alias X)
{
    static int* p = &X;   
}

void main()
{
    alias bar = Foo!(x);  // <-- "Error: cannot take address of thread-local variable x at compile time"
}
```

Also change "test() => main()" to allow user can copy/paste/run.

### DIFF
--- a/spec/template.dd
+++ b/spec/template.dd
@@ -561,18 +561,18 @@ $(GNAME TemplateAliasParameterDefault):
         $(LI Global names
 
         ------
-        __gshared int x;
+        shared int x;
 
         template Foo(alias X)
         {
-            static int* p = &X;
+            auto p = &X;
         }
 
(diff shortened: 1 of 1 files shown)


ZombineDev:
Can you use `shared` instead of `__gshared`?

jmdavis:
> Can you use shared instead of __gshared?

Agreed. `__gshared` is only intended for interacting with C globals, and shouldn't ever be used in D-specific code. It arguably should be illegal to even declare a local variable which is `__gshared`. So, this should either be reworked to use `shared`, or it needs to be reworked in a way that neither `__gshared` or `shared` is required.

xiren7:
updated, use shared instead of __gshared.

PR #1524, opened 2016-12-16
Move changelog to a folder and use single file entries

wilzbach:
Single file allow to enforce that every PR comes with a changelog as one can't run into a merge conflicts.

For DMD the idea was to move the language changelog directly to dlang.org, s.t. the DMD repo only contains the compiler changes.

For more details, see dlang/tools#186

### DIFF
--- a/language-changelog/README.md
+++ b/language-changelog/README.md
@@ -1,0 +1,34 @@
+How to add a new changelog entry to the pending changelog?
+==========================================================
+
+This will get copied to dlang.org and cleared when master gets
+merged into stable prior to a new release.
+
+```
+My fancy new feature X
+
+A long description of the new feature.
+-------
+import std.range;
+import std.algorithm.comparison;
+
+assert([1, 2, 3, 4, 5].padLeft(0, 7).equal([0, 0, 1, 2, 3, 4, 5]));
+
+assert("Hello World!".padRight('!', 15).equal("Hello World!!!!"));
+-------
+```
+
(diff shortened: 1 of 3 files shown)


andralex:
@MartinNowak @CyberShadow what do you think?

language-changelog/README.md
```diff
@@ -0,0 +1,34 @@
...
+---------------
+
+If you have cloned the [tools](https://github.com/dlang/tools) and [dlang.org](https://github.com/dlang/dlang.org) repo),
+you can preview the changelog:
+
+```
+../tools/changed.d -o ../dlang.org/changelog/pending.dd && make -C ../dlang.org -f posix.mak html
+```
+
+and then open `../dlang.org/web/changelog/pending.html` with your favorite browser.
```
wilzbach: Yes of course and if I think about it we should probably only advertise to use a `Makefile` target, so that we can change such things in the future, how about e.g.:

```
make -C ../dlang.org -f posix.mak pending_changelog
```

CyberShadow: Yep, that would be nice.

wilzbach:
> Yes of course and if I think about it we should probably only advertise to use a Makefile target, so that we can change such things in the future, how about e.g.:

Okay set the `README.md` to the upcoming Makefile target and already created a "dummy" one in the Makefile.
(the Makefile target one depends on https://github.com/dlang/tools/pull/186)

PR #1528, opened 2016-12-20
Improve and tighten pure functions specification

andralex:
This is in keeping with the current implementation and tightens the spec to ensure no future shenanigans.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -226,22 +226,47 @@ pure int foo(int i,
     only parameters without mutable indirections, and (b) returns a result
     without mutable indirections, will have the same effect for all invocation
     with equivalent arguments, and is allowed to memoize the result of the
-    function. These are termed "strongly `pure`" functions in this document.
-    Note that a strongly pure function may still have behavior inconsistent with
-    memoization by e.g. using `cast`s or by changing behavior depending on the
-    address of its parameters. An implementation is currently not required to
-    enforce validity of memoization in all cases.)
+    function under the assumption that equivalent parameters always produce
+    equivalent results. Such functions are termed $(I strongly `pure`) functions
+    in this document. Note that a strongly pure function may still have behavior
+    inconsistent with memoization by e.g. using `cast`s or by changing behavior
+    depending on the address of its parameters. An implementation is currently
+    not required to enforce validity of memoization in all cases.)
 
-    $(P Any `pure` function that is not strongly pure cannot be memoized. The
-    compiler is required to honor all calls to the function, even if it appears
-    to do nothing. (Example: a pure function taking no arguments and returning
-    `int*` cannot be memoized. It also should not because it may be a typical
-    factory function returning a fresh pointer with each call.))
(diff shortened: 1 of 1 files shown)


spec/function.dd
```diff
@@ -224,6 +222,26 @@ pure int foo(int i,
...
+    $(P Any `pure` function that is not strongly pure cannot be memoized. The
+    compiler is required to honor all calls to the function, even if it appears
+    to do nothing. (Example: a pure function taking no arguments and returning
+    `int*` cannot be memoized. It also should not because it may be a typical
+    factory function returning a fresh pointer with each call.))
+
+    $(P `pure` functions returning `void` will be always called even if it is
+    strongly `pure`. The implementation must assume the function does something
+    outside the confines of the type system and is therefore not allowed to
+    elide the call, even if it appears to have no possible effect.)
```
ZombineDev: Wouldn't it be simpler to just say that `pure` functions returning `void` are not strongly `pure`?

ZombineDev: That's sort of implied by "... (b) returns a result without mutable indirections", if we agree that `void` functions don't return any result.

WalterBright:
Should cover what happens when an exception is thrown by the pure function. I suggest regarding the exception thrown as its 'return value' as far as uniqueness and memoization rules go.

andralex:
@WalterBright updated with no special requirements for exceptions so as to leave maximum freedom

andralex:
@WalterBright ping - guess we're now good

spec/function.dd
```diff
@@ -224,6 +228,51 @@ pure int foo(int i,
...
+
+    $(P Any `pure` function that is not strongly pure cannot be assumed to be
+    memoizable and enjoys no special treatment compared to a regular non-pure
+    function. (Example: a pure function taking no arguments and returning `int*`
+    should not be memoized. It also should not because it may be a typical
+    factory function returning a fresh pointer with each call.) Function calls
+    may still be elided, or results be memoized, by means of traditional
+    inlining and optimization techniques available for all functions.)
+
+    $(P A function that accepts only parameters without mutable indirections and
```
JohanEngelen: whoops, should be: "A _`pure`_ function" ?

andralex:
@JohanEngelen thx fixed

spec/function.dd
```diff
@@ -271,6 +271,9 @@ pure List* make(int a, int b)
     assumptions related to memoization and references do not carry to the thrown
     exception.)
 
+    $(P Destructors will always be called even if they appear to be strongly
```
JohanEngelen: "Destructors will always be called" is not want you want to say I think; instead, it should only say something about pure destructors.
Perhaps: "When a destructor is pure, it does not imply that it need not be called."


Maybe I am approaching this from too low-level viewpoint. When you write "destructors will always be called", do you mean it in the sense of debug-call-tracing-low-levelness, or in the high level sense of "the program will be executed as if destructors are always called" (thus allowing empty destructor calls to be eliminated)?

nordlow:
Can we go further with make free and GC add/remove Range pure now?

PR #1529, opened 2016-12-21
Improve runnable examples (v2)

wilzbach:
Two more small improvements:

-  Keep the correct editor/source code open if it's open on evalation
- use automatic height for CodeMirror boxes

The second commit (use automatic height for CodeMirror boxes) is important as it affects the main example as well. Hence, here are two examples of how its result:

### examples before:

image

(https://dlang.org/phobos-prerelease/std_algorithm_comparison.html#.among)

### examples after:

image

dlang.org frontpage before:
------------------------

image

dlang.org frontpage after:
---------------------

image

See also the [CodeMirror docs](https://codemirror.net/demo/resize.html).

### DIFF
--- a/js/run.js
+++ b/js/run.js
@@ -454,6 +454,8 @@ function setupTextarea(el, opts)
     var code = $(editor.getWrapperElement());
     code.css('display', 'none');
 
+    var plainSourceCode = parent.parent().children("div.d_code");
+
     var output = outputDiv.children("textarea.d_code_output");
     var outputTitle = outputDiv.children("span.d_code_title");
     if (opts.args) {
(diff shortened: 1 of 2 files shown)


js/run.js
```diff
@@ -469,20 +471,22 @@ function setupTextarea(el, opts)
         var orgStdin = stdin.val();
     }
 
-    var hideAllWindows = function()
+    var hideAllWindows = function(args)
     {
+        args = args || {};
```
andralex: whoa some good magic here

andralex:
Can't go wrong with `auto`!

PR #1531, opened 2016-12-23
Add documentation for "deprecate" cycle handling

schveiguy:
Add documentation for dlang/druntime#1720


schveiguy:
Unsure why there are so many diffs in the docs. @CyberShadow ?

CyberShadow:
Should be fixed on the next rebuild.

CyberShadow:
Fixed :)

PR #1532, opened 2016-12-24
Allow to run ddox unittest examples

wilzbach:
Pretty straight forward after https://github.com/rejectedsoftware/ddox/pull/138

FYI: @s-ludwig for some Ddoc magic reason one can wrap the code block with ddox :)

It uses the same script as for ddoc and thus also the same blacklist / configuration & is for now only enabled in the prerelease docs.

I did a minor UI tweak as well and moved the buttons to the left as this was suggested on the NG.

Ah and the reason for the rename to `optArguments` is that JS doesn't has any concept of scoping - it's time that D conquers the web!


wilzbach:
Preview e.g. [here](http://dtest.thecybershadow.net/artifact/website-dec5e59a4d0eb4d0f2a298e9a3b5fcaf263316e9-ed7ce13923ee095e617c17026098b62e/web/library-prerelease/std/algorithm/comparison/among.html)

wilzbach:
Hmm there's a bit of weirdness hapenning - the newest version of ddox seems to remove empty newlines in unittest blocks:

with ddox 0.15.15

image

with ddox 0.15.17

image

s-ludwig:
@wilzbach: It appears to be the macro processing added by the `DDOX_UNITTEST` macro. I'll not have time today anymore, but I will look into that to see if this is the right behavior. But apart from this issue, I've looked a bit at what this wrapping macro does and it's kind of ugly. Maybe we'll have to switch to the non-wrapping version anyway after all,

andralex:
So looking forward to this.

s-ludwig:
Okay, there is 0.15.18 now, which uses `DDOX_UNITTEST_HEADER` and `DDOX_UNITTEST_FOOTER`. Both also get the unittest function name as an argument in case that turns out to be useful some day. The marker method will be required to generate valid HTML (the `DDOX_UNITTEST` based one actually also generated an invalid nesting order).

wilzbach:
> Okay, there is 0.15.18 now

Wow - that was fast! Thanks a lot :)
I integrated the new tag and rebased the commits accordingly.
The code box doesn't show any difference anymore: :tada: 

image

wilzbach:
> So looking forward to this

Anything blocking this? :)

wilzbach:
>>So looking forward to this
> Anything blocking this? :)

Ping @andralex 


@s-ludwig - I found a small issue: When there's a real ddoc comment like for the [second example of among](https://github.com/dlang/phobos/blob/master/std/algorithm/comparison.d#L129), the `DDOX_UNITTEST_HEADER` macro is wrapped inside the block:

```
/**
Alternatively, $(D values) can be passed at compile-time, allowing for a more
efficient search, but one that only supports matching on equality:
*/
@safe unittest
```

```
<p>Alternatively, <code class="lang-d"><a href="../../../std/algorithm/comparison/among.html#values"><span class="pln">values</span></a></code> can be passed at compile-time, allowing for a more
efficient search, but one that only supports matching on equality:<span class="dlang_runnable"></span>
</p>
```

However, it's not a big problem and I added a simple workaround.

andralex:
I'll just take a leap of faith here!

wilzbach:
> I'll just take a leap of faith here!

Looks good so far:

http://dlang.org/library-prerelease/std/algorithm/searching/min_element.html

If there aren't any issues with this in next two weeks, we could enable it for the stable documentation with the 2.073 release.

PR #1533, opened 2016-12-26
update iasm instruction table for dlang/dmd#6364

BBasile:
Also adds `popcnt` that missed.

### DIFF
--- a/dstyle.dd
+++ b/dstyle.dd
@@ -252,10 +252,10 @@ $(P
     Phobos and other official D source code, there are additional requirements:
  )
 
-$(UL
-    $(LI Braces should be on their own line. There are a few exceptions to this
-         (such as when declaring lambda functions), but with any normal function
-         block or type definition, the braces should be on their own line.)
+$(H4 Brackets)
+    $(P Braces should be on their own line. There are a few exceptions to this
+        (such as when declaring lambda functions), but with any normal function
+        block or type definition, the braces should be on their own line.)
 
 -------------------------------
 void func(int param)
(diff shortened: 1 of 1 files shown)


spec/iasm.dd
```diff
@@ -1067,6 +1067,16 @@ $(H3 $(LNAME2 amd_opcode_support, AMD Opcodes Supported))
         )
     )
 
+$(H3 $(LNAME2 Bit Manipulation Instructions, Bit manipulation instructions supported))
```
Hackerpilot: Why do these instructions have their own section?

BBasile: three reasons:
- the big table is sorted alphabetically so it's unpleasant to insert something at the middle.
- these instructions are almost part of a special set called ABM1 and ABM2 (Advanced Bit Manipulation) which are, despite of not being specific to vectors, SSE2 only.
- I may propose more iasm ABM in a further PR.

BBasile: And because of the 3rd point This PR doesn't need to be merged right now !

wilzbach:
>> I may propose more iasm ABM in a further PR.
> And because of the 3rd point This PR doesn't need to be merged right now !

@BBasile could you please only close spec update PRs if you have replacements for them?
It's really easy to loose track of them otherwise.

(it seems that you have force-pushed this branch, so I can't reopen it)

BBasile:
Yes I didn't remember that I made the error to commit the change in master. I've forced the push of my fork to keep it up to date and closes against my will this PR. I'd say that here it doesn't matter as I might propose more later. The table with just one entry was ugly anyway.

PR #1537, opened 2016-12-27
Ddox: Show exact source code links for multiple declarations

wilzbach:
If there are multiple declarations, the source code link should go to the first declaration - not to the header of the file. 

Example e.g.: http://dlang.org/library/std/algorithm/comparison/among.html

### DIFF
--- a/dpl-docs/views/layout.dt
+++ b/dpl-docs/views/layout.dt
@@ -164,7 +164,7 @@ html(lang='en-US')
                 - project = "druntime", path_prefix = "src/";
               - else
                 - project = "phobos", path_prefix = "";
-              - if (info.docGroups.length == 1)
+              - if (info.docGroups.length >= 1)
                 - if (auto decl = cast(Declaration)info.docGroups[0].members[0]) line_suffix = "#L"~to!string(decl.line);
               - if (info.node.module_.isPackageModule)
                 - filename = replace(modname, ".", "/") ~ "/package.d";


wilzbach:
In case someone was wondering what this does:

### Before

http://dlang.org/library/std/algorithm/comparison/among.html
Source code link points to: https://github.com/dlang/phobos/blob/master/std/algorithm/comparison.d

### After

http://dtest.thecybershadow.net/artifact/website-99c70ce923b2fb7f7dfc8aec6a9d04c3a5f1406f-b7c033f85ecbb7d3947e28b0db7d1c5e/web/library-prerelease/std/algorithm/comparison/among.html
Source code link points to: https://github.com/dlang/phobos/blob/master/std/algorithm/comparison.d#L85

wilzbach:
I think this is a really trivial change (one char!) that sits here for more than a month. @CyberShadow could you please help and have a look?

wilzbach:
Hmm this had the disadvantage that it creates a huge diff for even small changes as all line numbers are shifted. Should we for Digger builds turn this off?

CyberShadow:
Link to example diff?

CyberShadow:
> Should we for Digger builds turn this off?

Digger builds the live website which would mean they would almost never be visible there. I guess you meant DAutoTest. We already do this for `NODATETIME`.

wilzbach:
> Link to example diff?

-> https://github.com/dlang/phobos/pull/5204
with; http://dtest.dlang.io/results/9936840768ecd85e4aef07f8f315b041d7839562/044ad84eca029973c3299cc0f5c92a665ade107a/

PR #1538, opened 2016-12-27
Add message about issue 16211 update

schveiguy:
It needs to be in the changelog for sure, but we had no bugzilla issue for it.


MartinNowak:
Rewrote this a proper changelog entry to not conflict with the auto-generated Bugzilla list, https://github.com/dlang/dlang.org/commit/e66a37204c869b439d5cc443e900434324709b97.
We plan to automate the whole changelog gen soon, https://trello.com/c/WIwLjrPE/243-create-changelog-builder-from-files.

schveiguy:
OK, that makes sense -- otherwise you need to hand edit the auto-generated list if anything gets added. Thanks.

PR #1544, opened 2017-01-02
Move LATEST parameter to a file

CyberShadow:
Previously, the posix.mak makefile would accept an optional LATEST
parameter, which controlled a number of things, including which DMD
and Phobos versions to use when building the stable documentation, and
which version to show on as latest on the front/download pages. If the
parameter was omitted, the makefile would determine it by looking at
the lexicographically last tag in the DMD directory matching the
`^v[0-9][0-9.]*$` regular expression.

The LATEST variable being a command-line parameter has caused a number
of issues:

1. When the tag is not specified on the command line, the makefile
performs some network activity. This is a small nuisance.

2. Since LATEST is an external variable whose computed value may
change at any moment, it makes performing reproducible builds of this
repository considerably more difficult. Scripts which require a
reliable result, such as those used for CI, must calculate the correct
value of LATEST for the point in time that they are trying to build.

3. Calculating the correct value of LATEST for any point in time has
shown to be more complicated than expected. This has especially
affected the ae.sys.d package, which is used by Digger, the
documentation CI, and automatic deployment of dlang.org. Trying to
find a correct algorithm and working around the current one's
deficiencies has eaten a lot of time from myself and Martin, who has
needed to bypass the tester/autodeployment and manually build and
upload the website for all of the past few releases.

As the "benefit" of not needing to maintain the latest
publicly-released version of DMD in the dlang.org repository (which I
argue is a bad decision by itself) is far outweighted by the problems
it creates, this patch moves the parameter to its own file,
LATEST. The file will need to be updated manually for each release;
however, as all releases involve repository changes anyway, this
shouldn't be much of an issue.


MartinNowak:
Can we please name the file VERSION like in dmd?

CyberShadow:
OK, renamed to `VERSION`.

PR #1545, opened 2017-01-06
set changelog VER=2.012.3 macro from file name

MartinNowak:


MartinNowak:
Auto-merge correctly didn't work here b/c approvals are required.

PR #1548, opened 2017-01-07
Fix  changelog/2.073.0_pre typo

tchaloupka:
minIndex -> maxIndex


tchaloupka:
Ah, didn't know that. Done :)

wilzbach:
CC @MartinNowak I wasn't sure whether to use `stable` or `master`, but as

- `master` is automatically deployed to dlang.org
- `stable` is only used for the documentation bundled with the DMD release
- It will probably get overridden with the next beta or final release

I went with `master`.

PR #1549, opened 2017-01-07
Add pending_changelog target

wilzbach:
@CyberShadow - do you need anything more than this target to display a preview of the upcoming changelog on your `DAutoTest`?

CC @MartinNowak

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -13,11 +13,14 @@
 ifeq (,${LATEST})
 LATEST:=$(shell cat VERSION)
 endif
+# Next major DMD release
+NEXT_VERSION:=$(shell bash -c 'version=$$(cat VERSION);a=($${version//./ });a[1]="10\#$${a[1]}";((a[1]++)); a[2]=0; echo $${a[0]}.0$${a[1]}.$${a[2]};' )
 
 # Externals
 DMD_DIR=../dmd
 PHOBOS_DIR=../phobos
 DRUNTIME_DIR=../druntime
+TOOLS_DIR=../tools
 DUB_DIR=../dub-${DUB_VER}
 DMD=$(DMD_DIR)/src/dmd
 DMD_REL=$(DMD_DIR)-${LATEST}/src/dmd
@@ -150,7 +153,8 @@ SPEC_ROOT=$(addprefix spec/, \
 	abi simd)
 SPEC_DD=$(addsuffix .dd,$(SPEC_ROOT))
 
-CHANGELOG_FILES=$(basename $(subst _pre.dd,.dd,$(wildcard changelog/*.dd)))
+CHANGELOG_FILES=$(basename $(subst _pre.dd,.dd,$(wildcard changelog/*.dd))) \
+				changelog/${NEXT_VERSION}
 
 # Website root filenames. They have extension .dd in the source
 # and .html in the generated HTML. Save for the expansion of
@@ -519,7 +523,17 @@ test:
 # Changelog generation
 ################################################################################
 
-pending_changelog:
-	@echo "This command will be available soon."
+changelog/${NEXT_VERSION}.dd: ${STABLE_DMD} ../tools ../installer
+	$(STABLE_RDMD) $(TOOLS_DIR)/changed.d "v${LATEST}..upstream/stable" -o changelog/${NEXT_VERSION}.dd \
+		--version "${NEXT_VERSION}"
+
+pending_changelog: changelog/${NEXT_VERSION}.dd html
+	@echo "Please open file:///$(shell pwd)/web/changelog/${NEXT_VERSION}.html in your browser"
+
+../tools:
+	git clone https://github.com/dlang/tools ../tools
+
+../installer:
+	git clone https://github.com/dlang/installer ../installer
 
 .DELETE_ON_ERROR: # GNU Make directive (delete output files on error)


posix.mak
```diff
@@ -519,7 +520,10 @@ test:
 # Changelog generation
 ################################################################################
 
-pending_changelog:
-	@echo "This command will be available soon."
+pending_changelog_impl:
+	$(RDMD) $(TOOLS_DIR)/changed.d -o changelog/pending.dd "v${LATEST}..upstream/stable"
```
wilzbach: FYI: if you only want to display the manual changes and hide the changes from resolved Bugzilla issues, we can just remove the `"v${LATEST}..upstream/stable"` part.
Querying Bugzilla takes about 10-15s.

posix.mak
```diff
@@ -519,7 +520,10 @@ test:
 # Changelog generation
 ################################################################################
 
-pending_changelog:
-	@echo "This command will be available soon."
+pending_changelog_impl: ${STABLE_DMD}
+	$(STABLE_RDMD) $(TOOLS_DIR)/changed.d -o changelog/pending.dd "v${LATEST}..upstream/stable"
```
wilzbach: FYI: if you only want to display the manual changes and hide the changes from resolved Bugzilla issues, we can just remove the `"v${LATEST}..upstream/stable"` part.
Querying Bugzilla takes about 10-15s.

wilzbach:
Ping @CyberShadow - is there anything else I can do from my side? 
You basically just need to merge this PR and add the Makefile target to your awesome DAutoTest

CyberShadow:
Any reason not to include it into the default target?

I.e., let's say I make DAutoTest invoke make to build this target too. Now, should this also be built by Digger when building and deploying the live website? If not, what's the argument for the discrepancy of what gets tested vs. what gets released? If yes, why not include it in the default target?

CyberShadow:
BTW, I've mentioned this before, but would be really cool if it would generate pages for the next versions using the current pending changelog, i.e. for 2.074.0 from `master` and and 2.073.2 from `stable`.

wilzbach:
> Any reason not to include it into the default target?

Okay I managed to hack this into the default target. The biggest problem was that the `changed.d` script requires the remote `upstream` to be set for all repos and it requires the repos `tools` and `installer`, but there's nothing that a Makefile can't solve ;-)
Btw I am impressed that it passes and renders correctly, because on my machine it was failing due to https://github.com/dlang/tools/pull/217.

> BTW, I've mentioned this before, but would be really cool if it would generate pages for the next versions using the current pending changelog, i.e. for 2.074.0 from master and for 2.073.2 from stable.

I am not sure whether I understand, but I changed the Makefile to write the generated changelog file to `changelog/$(cat ../dmd/VERSION).dd` which is `2.074.0` for master, but only `2.073.1` for `stable`.

wilzbach:
@CyberShadow: I am not sure how you want to handle the resulting changelog file, but I guess it's hard without special casing.

image

CyberShadow:
> The biggest problem was that the changed.d script requires the remote upstream to be set for all repos and it requires the repos tools and installer, but there's nothing that a Makefile can't solve ;-)

I think that should be solved in `changed`. It should not make assumptions like that.

Note that you can fetch from a URL directly without adding it as a remote. If adding a remote is required, `changed` should do that and namespace it properly (i.e. e.g. `changed_upstream`).

> I am not sure whether I understand, but I changed the Makefile to write the generated changelog file to changelog/$(cat ../dmd/VERSION).dd which is 2.074.0 for master, but only 2.073.1 for stable.

That might not be a good idea right now, since it doesn't look like updating the DMD `VERSION` is part of the release process - so it looks like unless it's updated properly, this might actually overwrite or conflict with a released version's changelog? I'd suggest using some simple arithmetics to calculate the next versions using the dlang repo's `VERSION` file.

Anyway, if you don't want to commit to all of that now, we can revert to the simpler version for the moment.

> @CyberShadow: I am not sure how you want to handle the resulting changelog file, but I guess it's hard without special casing.

That looks good to me, what's the problem?

wilzbach:
> I think that should be solved in changed. It should not make assumptions like that.
> Note that you can fetch from a URL directly without adding it as a remote. If adding a remote is required, changed should do that and namespace it properly (i.e. e.g. changed_upstream).

`changed.d` [parses the `git log` command](https://github.com/dlang/tools/blob/master/changed.d#L108). 

I am not sure how @MartinNowak depends on this script. I added this to the [tools PR](https://github.com/dlang/tools/pull/217/commits/f17490e69fab13717fe86debe490dbcec6026f31) as an example (it's just three lines).

Should I keep the Makefile initialization of `tools` and `installer` or do you prefer to do this in another way?

> That might not be a good idea right now, since it doesn't look like updating the DMD VERSION is part of the release process - so it looks like unless it's updated properly, this might actually overwrite or conflict with a released version's changelog? I'd suggest using some simple arithmetics to calculate the next versions using the dlang repo's VERSION file.

"Simple arithmetics" in a Makefile..
My workaround is to use an ugly `bash` expression.

> Anyway, if you don't want to commit to all of that now, we can revert to the simpler version for the moment.

Well, I am running a bit out of time for now....

> That looks good to me, what's the problem?

That as far as I understand it will always be shown as "385 additions" (?) in the status line

CyberShadow:
> Should I keep the Makefile initialization of tools and installer or do you prefer to do this in another way?

That looks good, but would be nice to drop the adding of the remotes. They will not be added if the repository already exists, and `changed` should add them by itself if it can't avoid it.

> My workaround is to use an ugly bash expression.

Yep, that's the ticket.

> That as far as I understand it will always be shown as "385 additions" in the status line?

No, only the first time the file is created.

andralex:
Cool, I see @wilzbach and @CyberShadow have this going. Thanks!

wilzbach:
> Cool, I see @wilzbach and @CyberShadow have this going. Thanks!

Well unfortunately this is a bit blocked on https://github.com/dlang/tools/pull/217 (cleaning up the way `changed.d` fetches the git log which it uses to parse the list of all fixed bugs from)

wilzbach:
@CyberShadow I think with https://github.com/dlang/tools/pull/217 merged we can finally enable this? :)

CyberShadow:
My wishlist of possible future improvements:

- Add a notice to the page to make it obvious it's a pre-release
- Generate a changelog for the `stable` branch as well
- Make the download link go to the nightlies
- Make it generate the release date according to our release schedule
   - Changed "Released" to "To be released"
- Link to it from the sidebar

First one is most important to avoid confusion. just in case someone stumbles upon that page somehow. I've been demanding enough for this series of changes, though.

wilzbach:
> First one is most important to avoid confusion. just in case someone stumbles upon that page somehow. 

Seems like this can't happen for the time being:
http://dlang.org/changelog/2.074.0.html

> Make the download link go to the nightlies
> Add a notice to the page to make it obvious it's a pre-release
> Changed "Released" to "To be released"

-> https://github.com/dlang/dlang.org/pull/1586

> Link to it from the sidebar

Hmm, then it would be easily visible to the public. Do we want this?
We could mark it as "to be released" though.
In any case, there's the `update_nav.sh` which could be invoked:

-> https://github.com/dlang/dlang.org/pull/1585

> Generate a changelog for the stable branch as well

Well I am not sure on the best path here as AFAIK currently @MartinNowak commits the changelog manually. Do you have sth. specific in mind?

CyberShadow:
> Hmm, then it would be easily visible to the public. Do we want this?

Yes, pertaining that the other issues are resolved.

> Well I am not sure on the best path here as AFAIK currently @MartinNowak commits the changelog manually. Do you have sth. specific in mind?

Sorry, but can we not use exactly the same method we use for the `master` branch but for the `stable` branch? And use the same method to generate the file name, except increment the minor version number instead of the major one.

MartinNowak:
Just name the changlog _pre.d will build a pre-release changlog.

posix.mak
```diff
@@ -519,7 +523,17 @@ test:
 # Changelog generation
 ################################################################################
 
-pending_changelog:
-	@echo "This command will be available soon."
+changelog/${NEXT_VERSION}.dd: ${STABLE_DMD} ../tools ../installer
```
MartinNowak: changelog/${NEXT_VERSION}_pre.dd

posix.mak
```diff
@@ -519,7 +523,17 @@ test:
...
+	$(STABLE_RDMD) $(TOOLS_DIR)/changed.d "v${LATEST}..upstream/stable" -o changelog/${NEXT_VERSION}.dd \
```
MartinNowak: Always using stable seems incorrect when the purpose of this target is trying to build the changelog for dmd/druntime/phobos PRs.

wilzbach: This was already changed, as we realized that showing the bug list from Bugzilla will lead to unrelated diffs on DAutoTest: https://github.com/dlang/dlang.org/pull/1588

MartinNowak:
This breaks the build script b/c it builds the pending changelog by default http://nightlies.dlang.org/dmd-master-2017-03-01/build.html.
Should we patch that away or make it configurable?
Does that actually make sense for nightlies (who reads offline nightly docs anyhow?), then we'd need to adapt the changed tool to not rely on `git -C path`.
See http://nightlies.dlang.org/dmd-master-2017-03-01/build.html.

wilzbach:
> Should we patch that away or make it configurable?

How about sth. like `BUILD_CHANGELOG`?

> Does that actually make sense for nightlies (who reads offline nightly docs anyhow?)

Nope doesn't make sense, you are absolutely correct ;-)

CyberShadow:
> Does that actually make sense for nightlies (who reads offline nightly docs anyhow?)

Does it matter?

Seems the problem is that the machine building the nightlies has an old version of git? Could that be solved instead?

`-C` seems like a fairly recent option, but it can be substituted with `--work-tree` and `--git-dir`.
https://github.com/CyberShadow/D-dot-git/blob/9ce30c21bd56c94494403886c1d9ff76735f68ab/repo.d#L26

MartinNowak:
Yes, it matters, we don't want to build 2.075 changelogs for 2.074 releases, so we need some option to skip that eventually.
For preview builds we'd rather want the changelog, though generation isn't yet fully automated (some small issues w/ changed.d and update_nav.sh) and the changed tools doesn't work w/ the shallow clones from the builds script.
So for now, I'd rather just disable this in the builds script.

CyberShadow:
> Yes, it matters, we don't want to build 2.075 changelogs for 2.074 releases, so we need some option to skip that eventually.

Ah, I thought it was already skipped for releases?

wilzbach:
> Ah, I thought it was already skipped for releases?

It's a pending PR: https://github.com/dlang/dlang.org/pull/1586

PR #1552, opened 2017-01-12
Tweak D for C, C++ programmer pages

ntrel:
Also expand factorial example.

### DIFF
--- a/cpptod.dd
+++ b/cpptod.dd
@@ -45,7 +45,7 @@ class Foo
 
 $(H4 The D Way)
 
-	Constructors are defined with the this keyword:
+	Constructors are defined with the `this` keyword:
 
 ------
 class Foo
@@ -76,7 +76,7 @@ class B : A
 
 $(H4 The D Way)
 
-	The base class constructor is called with the super syntax:
+	The base class constructor is called with the `super` syntax:
 
 ------
 class A { this() { ... } }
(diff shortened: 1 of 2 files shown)


ctod.dd
```diff
@@ -582,7 +582,7 @@ $(H3 <a name="tagspace">Struct tag name space</a>)
 
 $(H4 The C Way)
 
-       It's annoying to have to put the struct keyword every time a type is specified,
+       It's annoying to have to put the `struct` keyword every time a type is specified,
```
Hackerpilot: Also replace "put" with "use" or "write".

ctod.dd
```diff
@@ -295,7 +295,7 @@ for (i = 0; i &lt; sizeof(array) / sizeof(array[0]); i++)
 
 $(H4 The D Way)
 
-The length of an array is accessible through the property "length".
+The length of an array is accessible through the property `length`.
```
Hackerpilot: "`length` property"

cpptod.dd
```diff
@@ -800,7 +809,7 @@ template IsFunctionT(T)
 
 void test()
 {
-    typedef int fp(int);
+    alias int fp(int);
```
JackStouffer: new alias syntax please

ntrel: Not sure how, I tried `alias fp = int function(int)` but that didn't work.

ntrel:
@Hackerpilot done

cpptod.dd
Hackerpilot: I think I've run into this issue before. If it is what I think it is, it's not possible to translate this to the new alias syntax.

PR #1557, opened 2017-01-19
fix typo -mscrt -> -mscrtlib

rainers:


MartinNowak:
Thx, I'll just regenerate the changelog before the build anyhow.

PR #1558, opened 2017-01-20
Add missing changelog entry for new default Ddoc theme

jacob-carlborg:


jacob-carlborg:
@MartinNowak ping

MartinNowak:
The changelog is auto-generated, please add an entry to dmd/stable instead (looks like a compiler ddoc chage to me).
https://github.com/dlang/dmd/blob/stable/changelog/README.md

PR #1561, opened 2017-01-28
Use runnable examples for released docs

wilzbach:
With 2.073 released we could flip the switch and enable the runnable examples for the released doc pages.

Would be nice to get #1550 and #1551 as well.


wilzbach:
> With 2.073 released we could flip the switch and enable the runnable examples for the released doc pages.
> Would be nice to get #1550 and #1551 as well.

CC @andralex

PR #1567, opened 2017-02-09
update putao github url

zoujiaqing:


orgs-using-d.dd
```diff
@@ -118,9 +118,9 @@ $(DIVC orgs-using-d center,
     )
     $(DORG Putao, http://www.putao.com, putao.png,
         Science and technology for kids,
-        Large scale use of D for distributed systems and cloud storage.
+        Large scale use of Dlang for distributed systems and cloud storage.
```
klickverbot: We've been using "D" instead of "Dlang" consistently on this page - why deviate from that?

wilzbach:
(please feel free to submit another PR if you think that "Dlang" instead of "D" should be used in your organization description.

orgs-using-d.dd
zoujiaqing: We are difficult search D in article.
but dlang ok.

I love D used for language name, but dlang  is more practical.

PR #1568, opened 2017-02-11
posix.mak: Work around DMD bug manifesting on certain filesystems

CyberShadow:
If the files are passed to DMD in a different order, Phobos fails to
compile. For example, this happens when the Phobos source code is on a
tmpfs, but not on ext4 or btrfs. Work around this by sorting the
Phobos source file list before passing in to DMD.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -455,7 +455,7 @@ docs.json : ${DMD_REL} ${DRUNTIME_DIR}-${LATEST} \
 	find ${DRUNTIME_DIR}-${LATEST}/src -name '*.d' | \
 	  sed -e /unittest.d/d -e /gcstub/d > .release-files.txt
 	find ${PHOBOS_DIR}-${LATEST} -name '*.d' | \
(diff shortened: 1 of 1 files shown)


CyberShadow:
It's greeeen!

wilzbach:
I am going ahead and merge this, s.t. the approved PRs can pass smoothly. Thanks a lot @CyberShadow!

CyberShadow:
> I couldn't find a bug report for this. Did you already create one?

I've CC'd Walter (by email) but don't have a reasonable test case yet... will DustMite it after I finish some modifications to it... (yeah, quite the yak-shaving stack I'm in)

MartinNowak:
Thx for the fix.
Unfortunately order of semantic analysis sometimes unveils compiler bugs.

MartinNowak:
cherry-picked as 7f23e834c1a78e985b5f18a0ea02fcd9489b584c

PR #1570, opened 2017-02-14
Update FAQ: Shared does not cause memory barriers

jpf91:
There's been a question about this in D.learn recently:
http://forum.dlang.org/post/apoqhsoqksuwieklcnjj@forum.dlang.org

If I remember correctly the final decision from 2012 was that shared variables do not automatically cause any memory barriers, but we never documented this:
http://forum.dlang.org/post/k7pn19$bre$1@digitalmars.com
http://forum.dlang.org/post/mailman.1904.1352922666.5162.digitalmars-d@puremagic.com

ping @WalterBright @andralex

### DIFF
--- a/faq.dd
+++ b/faq.dd
@@ -118,8 +118,7 @@ $(ITEM shared_synchronized, What does shared have to do with synchronization?)
 
 $(ITEM shared_memory_barriers, What does shared have to do with memory barriers?)
 
-	$(P Reading/writing shared data emits memory barriers to ensure sequential consistency (not
-	implemented).
+	$(Currently the compiler does not insert memory barriers around shared variables.
 	)
 
 $(ITEM casting_to_shared, What are the semantics of casting FROM unshared TO shared?)


andralex:
I think we need to revisit this.

JackStouffer:
@andralex I think `shared` needs to be revisited ;)

d-random-contributor:
Hmm... was it **initially** meant so? Some people wanted it, but I don't think the compiler itself was meant to do it. Do you mean Bartosz proposal here?

jpf91:
> Hmm... was it initially meant so? Some people wanted it, but I don't think the compiler itself was meant to do it.

The FAQ entry this pull request changes is currently:

> Reading/writing shared data emits memory barriers to ensure sequential consistency (not implemented)

This is also mentioned in TDPL, chapter 13.12.1 Sequential consistency of shared data:
> [...] To effect that, shared accesses must be surrounded by special machine code instructions called memory barriers, [...]

The discussion linked in the pull request description is more recent than TDPL though:
http://forum.dlang.org/post/k7pn19$bre$1@digitalmars.com

andralex:
I think we made a mistake in the forum discussion. For now let's update this PR by making it state the current state of affairs: "Currently the compiler does not insert memory barriers around `shared` variables." Also please open a bugzilla issue to revisit this. Thanks.

d-random-contributor:
Somewhat broadly related issue is [14932](https://issues.dlang.org/show_bug.cgi?id=14932)

jpf91:
> For now let's update this PR by making it state the current state of affairs:

Done.

> Also please open a bugzilla issue to revisit this.

Do we need another issues or can we use 14932 and 16198 for this?

JackStouffer:
I think 14932 suffices

andralex:
OK we're good with bugzilla. Thanks.

PR #1571, opened 2017-02-15
Enable runnable examples on releases docs

wilzbach:
Turns out that I have added another protection which didn't even add the JS for the runnable examples. So this finally adds `js/run_examples.js` to the released docs.


wilzbach:
Yeah - it's finally public - e.g. https://dlang.org/phobos/std_algorithm_searching.html#.all

PR #1573, opened 2017-02-17
Point user to std.array.byPair if compatibility with Tuple is desired.

quickfur:
Based on [forum feedback](http://forum.dlang.org/post/osbkmrvyqqjjsxyzgcyd@forum.dlang.org).

### DIFF
--- a/spec/hash-map.dd
+++ b/spec/hash-map.dd
@@ -341,7 +341,11 @@ Properties for associative arrays are:
         ForeachStatement) which will iterate over key-value pairs of the
         associative array. The returned pairs are represented by an opaque type
         with $(D .key) and $(D .value) properties for accessing the key and
-	value of the pair, respectively.))
+        value of the pair, respectively.  Note that this is a low-level
+        interface to iterating over the associative array and is not compatible
+        with the $(LINK2 $(ROOT_DIR)phobos/std_typecons.html#.Tuple,`Tuple`)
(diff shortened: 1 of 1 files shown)


wilzbach:
Though I am wondering why no one has bothered to implement `opApply` for `.byKeyValue`

I could only find this discussion (https://github.com/dlang/druntime/pull/1070), but no mention of reasons against it.

quickfur:
@wilzbach Why would that be necessary?  `.byKeyValue` works fine inside a `foreach` loop.
````
int[string] aa;
aa["abc"] = 1;
aa["def"] = 2;
foreach (p; aa.byKeyValue) {
    writefln("%s: %s", p.key, p.value);
}
````

wilzbach:
> @wilzbach Why would that be necessary? .byKeyValue works fine inside a foreach loop.

I was referring to an `opApply(K, V)(K k, V v)` with two parameters because that would have covered most cases for which `byPair` is used ;-)

quickfur:
But why? This works just fine:
````
int[string] aa = ...;
foreach (key, value; aa) {
    writefln("%s: %s", key, value);
}
````

wilzbach:
> But why? This works just fine:

Oh I think I got totally confused, sorry.
-> I made an enhancement request to the specs to show an example of this to the reader:

https://github.com/dlang/dlang.org/pull/1577

PR #1575, opened 2017-02-18
Improve runnable: disable assert to writeln transformation + make output a pre element

wilzbach:
As of http://forum.dlang.org/post/enjjesmilghothyjlfid@forum.dlang.org

1) the Regex doesn't seem to handle all cases of `assert` -> removed
2) `textarea` shouldn't be used for a readonly output -> `pre`
3) For consistency I also moved the separated CSS up into its group

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1610,7 +1610,7 @@ div.changelog-nav
 
 /* Runnable-examples css */
 textarea.d_code {display: none;}
-textarea.d_code_output, textarea.d_code_stdin, textarea.d_code_args
+textarea.d_code_stdin, textarea.d_code_args
 {
     text-align: left;
     border: none;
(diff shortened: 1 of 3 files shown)


JackStouffer:
I'm getting "temporarily unavailable" errors when I try to test this.

wilzbach:
> I'm getting "temporarily unavailable" errors when I try to test this.

Yeah @nazriel doesn't allow CORS requests from the DAutoTest domain (dtest.dlang.io). I have written to him about this for a couple of times already since the last six months. Unfortunately as you have realized with no effect.

wilzbach:
For reference:
@JackStouffer If you open your console you should see something like:

> XMLHttpRequest cannot load https://dpaste.dzfl.pl/request/. No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://dtest.dlang.io' is therefore not allowed access.

wilzbach:
Ping @CyberShadow or @andralex for merging this fix, s.t. we can post an announcement to reddit ;-)

andralex:
Cool. Not a fan of the rewrite assert->writeln in the first place...

CyberShadow:
Thx. If possible, it would be nice to put all information describing a change in the commit message itself, e.g. what the assert rewrite was (I had no idea) and the justification for the HTML change.

PR #1576, opened 2017-02-19
Fix: Invisible input/textarea text for light-on-dark systems

Abscissa:


wilzbach:
Thanks a lot for submitting this change @Abscissa!
While this looks valid to me, my knowledge of dlang.org's CSS and quirks is also limited, so I assign this to our website tzar @CyberShadow 
(On my glance I couldn't find any side-effects of this change)

PR #1577, opened 2017-02-21
Hashmap: add example for iteration and tuple counting

wilzbach:
This is a modified version of my favorite DLang example for Associative Arrays because:

- shows iteration through arrays and associative arrays
- uses tuples (!)
- has real world use cases (e.g. DNA sequencing)
- it's very concise and "beautiful" (-> easy to understand)

Hence I think it could be worthwhile to list this as an example to the reader.

(Follow-up to https://github.com/dlang/dlang.org/pull/1573)

### DIFF
--- a/spec/hash-map.dd
+++ b/spec/hash-map.dd
@@ -304,7 +304,6 @@ $(H2 $(LNAME2 construction_and_ref_semantic, Construction and Reference Semantic
         assert(aa[2] == 2);      // now both refer to the same instance
         ------
 
-
 $(H2 $(LNAME2 properties, Properties))
 
 $(P Properties for associative arrays are:)
@@ -351,7 +350,7 @@ $(P Properties for associative arrays are:)
     )
 
 $(HR)
-$(H2 $(LNAME2 aa_example, Associative Array Example: word count))
+$(H3 $(LNAME2 aa_example, Associative Array Example: word count))
 
         $(P Let's consider the file is ASCII encoded with LF EOL.
         In general case we should use $(I dchar c) for iteration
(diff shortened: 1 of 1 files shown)


wilzbach:
> Doesn't this do UTF decoding? (Yes, it does, you even have `dchar` in there explicitly. Best not to encourage writing code that way?)

Yeah, I wasn't sure about the "best practice".
The [other example](https://dlang.org/spec/hash-map.html#aa_example) assumes the input to be ASCII only and states:

> In general case we should use dchar c for iteration

wilzbach:
> The reason why I think string (decoding) is not appropriate here is that DNA base sequences are not UTF text, but rather byte sequences from the alphabet 

Fair enough - I just wanted to keep the example as simple as possible, but you have made a good case. I updated the example to use `byCodeUnit` instead ;-)

CyberShadow:
TBH, this is nice, but if I were to see this code in the wild I'd just wonder why bother with tuples and ranges if you can use simple dull arrays and achieve the same thing with less code and better performance:
```
        import std.stdio : writefln;

        int[char[2]] aa;

        // The string `arr` has a limited alphabet: {A, C, G, T}
        // Thus, for better performance, iteration can be done _without_ decoding
        auto arr = "AGATAGA";

        // iterate over all pairs in the string and observe each pair
        // ('A', 'G'), ('G', 'A'), ('A', 'T'), ...
        foreach (i; 0..arr.length-1)
            aa[arr[i..$][0..2]]++;

        // iterate over all key/value pairs of the Associative Array
        foreach (key, value; aa)
            writefln("key: %s, value: %d", key, value);
```

So, not sure about this example's illustrative powers. D arrays are awesome too!

wilzbach:
> So, not sure about this example's illustrative powers. D arrays are awesome too!

This now uses this example & rebased to master.

CyberShadow:
I don't remember what the original version looked like any more, but LGTM!

PR #1580, opened 2017-02-22
Make modlist.d generic

wilzbach:
tl;dr: this is a first dependency to expose the built documentation DMD on dlang.org's menu and makes this creates usable independently from dlang.org.

* Accept arbitrary number of directories
* Replaced hard-coded tree dumping of Phobos and DRuntime with a --dump flag
* Detect the source folder of a directory automatically
* Use std.getopt as arg parser

### DIFF
--- a/modlist.d
+++ b/modlist.d
@@ -4,7 +4,7 @@
  * License: $(HTTP boost.org/LICENSE_1_0.txt, Boost License 1.0).
  * Authors: $(HTTP code.dawg.eu, Martin Nowak)
  */
-import std.algorithm, std.file, std.path, std.stdio, std.string, std.range;
+import std.algorithm, std.file, std.getopt, std.path, std.stdio, std.string, std.range;
 
 struct Tree
 {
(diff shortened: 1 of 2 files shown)


CyberShadow:
By itself this change seems to be killing library navigation, is that a bug or does it depend on another change?

wilzbach:
> By itself this change seems to be killing library navigation, is that a bug or does it depend on another change?

Sorry - forgot to rebase this as well.

> Documentation OK (no changes)

Looks good :)

modlist.d
```diff
@@ -115,15 +123,28 @@ int main(string[] args)
...
-    add(phobos, tree);
-    add(buildPath(druntime, "src"), tree);
+    foreach (dir; dirs)
+    {
+        // search for common root folders (fallback to the root directory)
+        ["source", "src", ""]
+            .map!(f => buildPath(dir, f))
+            .filter!exists
+            .front      // no UFCS for local symbols
+            .I!(name => add(name, tree));
```
CyberShadow: :+1:

PR #1581, opened 2017-02-22
Enable Ddoc DMD release + Ddox prerelease build of DMD docs + verbatim target for DMD

wilzbach:
I excluded the DDoc release build because it would be directly added to the menu (of the release build) and there's still a lot that isn't documented properly or shouldn't be exposed publicly.

However, as a quick preview here's a screenshot the Ddox menu (prerelease!):

image

And ddox's awesome auto-completion:

image

To enable dmd on the menu for the prerelease pages, https://github.com/dlang/dlang.org/pull/1580 is needed

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -511,12 +511,14 @@ docs.json : ${DMD_REL} ${DRUNTIME_STABLE_DIR} \
 	  --only-documented $(MOD_EXCLUDES_PRERELEASE)
 	rm .release-files.txt .release-dummy.html
 
+# DDox tries to generate the docs for all `.d` files. However for dmd this is tricky,
+# because the `{mach, elf, mscoff}` are platform dependent.
+# Thus the need to exclude these files (and the `objc.d` files).
 docs-prerelease.json : ${DMD} ${DMD_DIR} ${DRUNTIME_DIR} \
 		${PHOBOS_FILES_GENERATED} | dpl-docs
 	find ${DMD_DIR}/src -name '*.d' | \
-		sed -e /objc.d/d -e /mscoff/d -e '/dmd\/src\/id.d/d' \
-			-e /objc_glue.d/d ${DMD_EXCLUDE}  \
-		> .prerelease-files.txt
+		sed -e /objc.d/d -e /mscoff/d -e /objc_glue.d/d ${DMD_EXCLUDE}  \
+			> .prerelease-files.txt
 	find ${DRUNTIME_DIR}/src -name '*.d' | sed -e '/gcstub/d' \
 	  -e /unittest/d >> .prerelease-files.txt
 	find ${PHOBOS_DIR_GENERATED} -name '*.d' | sed -e /unittest.d/d \
--- a/posix.mak
+++ b/posix.mak
@@ -73,7 +73,8 @@ STABLE_RDMD=$(STABLE_DMD_ROOT)/dmd2/$(OS)/$(if $(filter $(OS),osx),bin,bin$(MODE
 MOD_EXCLUDES_PRERELEASE=$(addprefix --ex=, gc. rt. core.internal. core.stdc.config core.sys.	\
 	std.algorithm.internal std.c. std.concurrencybase std.internal. std.regex.internal.  \
 	std.windows.iunknown std.windows.registry etc.linux.memoryerror	\
-	std.experimental.ndslice.internal std.stdiobase)
+	std.experimental.ndslice.internal std.stdiobase \
+	tk. msvc_dmc msvc_lib)
 
 MOD_EXCLUDES_RELEASE=$(MOD_EXCLUDES_PRERELEASE)
 
@@ -91,6 +92,8 @@ REBASE = MYBRANCH=`git rev-parse --abbrev-ref HEAD` &&\
 CHANGE_SUFFIX = \
  for f in `find "$3" -iname '*.$1'`; do\
   mv $$f `dirname $$f`/`basename $$f .$1`.$2; done
+
+DMD_GEN=$(DMD_DIR)/generated/$(OS)/release/$(MODEL)
 
 # Disable all dynamic content that could potentially have an unrelated impact
 # on a diff
@@ -238,8 +241,8 @@ $(DOC_OUTPUT_DIR)/dmd-%.verbatim : %.ddoc dcompiler.dd verbatim.ddoc $(DMD)
 
 all : docs html
 
-docs : dmd-prerelease phobos-prerelease druntime-prerelease druntime-release \
-  phobos-release apidocs-release apidocs-prerelease
+docs : dmd-release dmd-prerelease phobos-prerelease druntime-prerelease \
+	   druntime-release phobos-release apidocs-release apidocs-prerelease
 
 html : $(ALL_FILES)
 
@@ -270,7 +273,7 @@ ${GENERATED}/modlist-${LATEST}.ddoc : modlist.d ${STABLE_DMD} $(DRUNTIME_STABLE_
 ${GENERATED}/modlist-prerelease.ddoc : modlist.d ${STABLE_DMD} $(DRUNTIME_DIR) $(PHOBOS_DIR)
 	mkdir -p $(dir $@)
 	$(STABLE_RDMD) modlist.d $(DRUNTIME_DIR) $(PHOBOS_DIR) $(MOD_EXCLUDES_PRERELEASE) \
-		$(addprefix --dump , object std etc core) >$@
+		$(addprefix --dump , object std etc core ddmd) >$@
 
 # Run "make -j rebase" for rebasing all dox in parallel!
 rebase: rebase-dlang rebase-dmd rebase-druntime rebase-phobos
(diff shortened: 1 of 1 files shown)


posix.mak
```diff
@@ -353,6 +353,11 @@ $(DMD) : ${DMD_DIR}
 $(DMD_REL) : ${DMD_DIR}-${LATEST}
 	${MAKE} --directory=${DMD_DIR}-${LATEST}/src -f posix.mak AUTO_BOOTSTRAP=1 -j 4
 
+dmd-release : $(STD_DDOC) $(DMD_DIR) $(DMD)
+	$(MAKE) AUTO_BOOTSTRAP=1 --directory=$(DMD_DIR) -f posix.mak html \
+		DOCDIR=${DOC_OUTPUT_DIR}/dmd \
+		DOCFMT="$(addprefix `pwd`/, $(STD_DDOC))"
```
wilzbach: The HTML for ddoc will get built from now on, but without a menu entry it's not visible to the user.

posix.mak
```diff
@@ -456,13 +467,18 @@ docs.json : ${DMD_REL} ${DRUNTIME_DIR}-${LATEST} \
...
+		sed -e /objc.d/d -e /mscoff/d \
+			-e /objc_glue.d/d $(DMD_EXCLUDE)  \
+		> .prerelease-files.txt
 	find ${DRUNTIME_DIR}/src -name '*.d' | sed -e '/gcstub/d' \
-	  -e /unittest/d > .prerelease-files.txt
+	  -e /unittest/d >> .prerelease-files.txt
 	find ${PHOBOS_DIR} -name '*.d' | sed -e /unittest.d/d \
 	  -e /windows/d >> .prerelease-files.txt
-	${DMD} -c -o- -version=CoreDdoc -version=StdDdoc -Df.prerelease-dummy.html \
+	${DMD} -J$(DMD_DIR)/res -J$(DMD_DIR)/src -c -o- -version=MARS -version=CoreDdoc \
```
wilzbach: `MARS` is needed due to [`backend/cc.d`](https://github.com/dlang/dmd/blob/master/src/backend/cc.d#L175). It's a [default flag for all D compilers](https://github.com/dlang/dmd/blob/master/src/posix.mak#L149).

CyberShadow:
BTW, http://dlang.org/dmd/ currently triggers that redirect (to dmd usage docs for your platform).

CyberShadow:
>  Enable Ddoc DMD release + Ddoc build of DMD docs #1581 

Did you mean "DDox" instead of one of those instances of "DDoc"?

CyberShadow:
This seems to add some unwanted packages / top-level modules to the navigation: 

- tk
- msvc_dmc
- msvc_lib

CyberShadow:
Please rebase, diffs are gone :)

wilzbach:
> Did you mean "DDox" instead of one of those instances of "DDoc"?

Yes - sorry :/ 

> This seems to add some unwanted packages / top-level modules to the navigation:

Removed. Thanks!
As mentioned I enabled the navigation only for the DDox build.
To enable it for DDoc we need #1580.

> Please rebase, diffs are gone :)

Done :)

wilzbach:
> To enable it for DDoc we need #1580.

Thanks a lot for pulling this so quickly.
I rebased this PR and:
- enabled ddoc menu on the prerelease page
- added DMD output to the verbatim target

posix.mak
```diff
@@ -55,7 +55,8 @@ STABLE_RDMD=$(STABLE_DMD_ROOT)/dmd2/$(OS)/$(if $(filter $(OS),osx),bin,bin$(MODE
 MOD_EXCLUDES_PRERELEASE=$(addprefix --ex=, gc. rt. core.internal. core.stdc.config core.sys.	\
 	std.c. std.algorithm.internal std.internal. std.regex.internal. 			\
 	std.windows.iunknown std.windows.registry etc.linux.memoryerror std.stream std.cstream	\
-	std.socketstream std.experimental.ndslice.internal)
+	std.socketstream std.experimental.ndslice.internal \
+	tk. msvc_dmc msvc_lib)
```
wilzbach: We should probably add more DMD files to this list (but this can be done after this PR)

posix.mak
```diff
@@ -281,8 +284,8 @@ ${GENERATED}/modlist-${LATEST}.ddoc : modlist.d ${STABLE_DMD} $(DRUNTIME_STABLE_
 
 ${GENERATED}/modlist-prerelease.ddoc : modlist.d ${STABLE_DMD} $(DRUNTIME_DIR) $(PHOBOS_DIR)
 	mkdir -p $(dir $@)
-	$(STABLE_RDMD) modlist.d $(DRUNTIME_DIR) $(PHOBOS_DIR) $(MOD_EXCLUDES_PRERELEASE) \
-		$(addprefix --dump , object std etc core) >$@
+	$(STABLE_RDMD) modlist.d $(DRUNTIME_DIR) $(PHOBOS_DIR) $(DMD_DIR) $(MOD_EXCLUDES_PRERELEASE) \
+		$(addprefix --dump , object std etc core ddmd) >$@
```
wilzbach: Btw is there any logic behind this order?
Imho `core`/ Druntime is more important the the `etc` C bindings ...

posix.mak
```diff
@@ -496,13 +523,18 @@ docs.json : ${DMD_REL} ${DRUNTIME_STABLE_DIR} \
 	  --only-documented $(MOD_EXCLUDES_PRERELEASE)
 	rm .release-files.txt .release-dummy.html
 
-docs-prerelease.json : ${DMD} ${DRUNTIME_DIR} \
+docs-prerelease.json : ${DMD} ${DMD_DIR} ${DRUNTIME_DIR} \
 		${PHOBOS_DIR_GENERATED} | dpl-docs
+	find ${DMD_DIR}/src -name '*.d' | \
+		sed -e /objc.d/d -e /mscoff/d -e '/dmd\/src\/id.d/d' \
```
wilzbach: The exclude of `src/id.d` was only necessary because I didn't completely clean the `dmd` cache. As I assume I won't be the only one, I added it to the "ignore list".

posix.mak
```diff
@@ -484,6 +504,13 @@ ${DOC_OUTPUT_DIR}/library-prerelease/.htaccess : dpl_prerelease_htaccess
 	@mkdir -p $(dir $@)
 	cp $< $@
 
+DMD_EXCLUDE =
+ifeq (osx,$(OS))
+	DMD_EXCLUDE += -e /scanelf/d -e /libelf/d
+else
+	DMD_EXCLUDE += -e /scanmach/d -e /libmach/d
+endif
```
wilzbach: DDox tries to generate the docs for all `.d` files. However for dmd this is tricky, because the `{mach, elf, mscoff}` are platform dependent. Thus the need to exclude these files (and similarly for the `objc.d` files).

wilzbach:
> BTW, http://dlang.org/dmd/ currently triggers that redirect (to dmd usage docs for your platform).

I realized that even Druntime is built under `phobos{-prerelease}` and the links are relative to the current folder only. So the easiest solution was to use `phobos{-prerelease}` as well.

posix.mak
CyberShadow: Probably not, if anything alphabetical makes more sense than the current one, and is consistent with the sorting of the subtrees.

CyberShadow: Current order is fine though.

CyberShadow:
@andralex So what's your opinion on having the `ddmd` package in the library reference's navigation tree? Is this what we want to do, and should we do this now, or wait until it's renamed to `dmd`?

wilzbach:
> @andralex So what's your opinion on having the ddmd package in the library reference's navigation tree? Is this what we want to do, and should we do this now, or wait until it's renamed to dmd?

I assumed that this will be an open question. Hence this PR only exposes the `ddmd` on the `prerelease` pages which aren't much known / visited anyways.
And FWIW without a menu, the (d)dmd documentation is quite useless see e.g.:
http://dlang.org/dmd-prerelease/mars.html

CC @RazvanN7 - this will be one bit that you can update soon ;-)

andralex:
>@andralex So what's your opinion on having the ddmd package in the library reference's navigation tree? Is this what we want to do, and should we do this now, or wait until it's renamed to dmd?

I'm not too worried about that, pages will be of narrow interest and it's unlikely anyone's life will depend on the links staying stable. Yes @RazvanN7 keep this on your list to change when we settle.

posix.mak
CyberShadow: I think these notes should be comments in the file itself.

CyberShadow:
The filenames under `/phobos-prerelease/` don't look right.

E.g.: `web/phobos-prerelease/intrange.html`

I'm guessing should be `ddmd_intrange.html`.

This causes the navigation to be broken as well.

wilzbach:
> The filenames under /phobos-prerelease/ don't look right.

Well it turns out that the HTML generation script at DMD is completely different to the one at Phobos and doesn't support this `mod_ule` notation for paths.
This means we need to modify the DMD script: 

https://github.com/dlang/dmd/pull/6581

wilzbach:
> This means we need to modify the DMD script:
> dlang/dmd#6581

Thanks a lot for pushing this through so quickly :)

-> rebased this, so it should be ready to ship :ship:

andralex:
I guess I'll...

wilzbach:
@andralex it's now live :party:

DDoc: https://dlang.org/phobos-prerelease/ddmd_mars.html
DDox: https://dlang.org/library-prerelease/ddmd/mars.html (enjoy the search feature on the top right!)

Next steps could be:
- add more ddoc headers for public declarations to `(d)dmd`
- exclude internal modules from the list ([the list of currently excluded modules is here](https://github.com/dlang/dlang.org/blob/master/posix.mak#L73))
- prepare dlang.org for the `ddmd` -> `dmd` rename (CC @RazvanN7)
...
- flip the switch and enable it for the stable  (aka released) pages

andralex:
Yowzers! cc @WalterBright

PR #1582, opened 2017-02-22
Add assert -> writeln transformation magic for runnable unittest examples

wilzbach:
Okay this is a more "sophisticated" approach towards rewriting the AssertExpressions in the runnable examples than the initial regex.

I think we all now the problem: "All test passed" is a quite boring output and not so much than being able to directly change a single character.
The objective of this tool is to be conservative as a broken example looks a lot worse than a few statements that could have potentially been rewritten.
For now:
 - only EqualExpressions are "transformed"
 - static asserts are ignored
 - only single-line asserts are rewritten
- this is only enabled for the prerelease pages

However
- that's still >12K automatic `writeln` transformations
- as `ddoc` is used to generate the documentation, it's ensured that no invalid transformations happened.

@CyberShadow I am not sure on the best way to integrate this into the existing Makefile is. At the moment the tool replaces occurrences directly in the Phobos directory which might be a bit inconvenient if enabled as default. Should we make a copy of the Phobos sources before?

This depends on these two fixes to libdparse (hence blocked):

- https://github.com/Hackerpilot/libdparse/pull/128
- https://github.com/Hackerpilot/libdparse/pull/130

### DIFF
--- a/assert_writeln_magic.d
+++ b/assert_writeln_magic.d
@@ -296,7 +296,7 @@ struct FileLines
         // within the docs we automatically inject std.stdio (hence we need to do the same here)
         // writeln needs to be @nogc, @safe, pure and nothrow (we just fake it)
         if (hasWrittenChanges)
-            outFile.writeln("void writeln(T)(T l) { }");
+            outFile.writeln("private void writeln(T)(T l) { }");
         outFile.flush;
 
         return s;
--- a/assert_writeln_magic.d
+++ b/assert_writeln_magic.d
@@ -275,6 +275,9 @@ struct FileLines
     // dumps all changes
     ~this()
     {
+        if (!hasWrittenChanges)
+            return;
+
         auto tmpFile = writeLinesToFile;
         tmpFile.copy(destFile);
         tmpFile.remove;
@@ -297,8 +300,8 @@ struct FileLines
         // writeln needs to be @nogc, @safe, pure and nothrow (we just fake it)
         if (hasWrittenChanges)
             outFile.writeln("// \nprivate void writeln(T)(T l) { }");
-        outFile.flush;
 
+        outFile.flush;
         return s;
     }
 
--- a/assert_writeln_magic.d
+++ b/assert_writeln_magic.d
@@ -1,7 +1,7 @@
 #!/usr/bin/env dub
 /++
 dub.sdl:
-dependency "libdparse" version="~>0.7.0-beta.6"
+dependency "libdparse" version="0.7.0-beta.7"
 name "assert_writeln_magic"
 +/
 /*
--- a/assert_writeln_magic.d
+++ b/assert_writeln_magic.d
@@ -108,7 +108,7 @@ class TestVisitor : ASTVisitor
     /// A single line
     override void visit(const DeclarationOrStatement expr)
     {
-    processLastAssert();
+        processLastAssert();
         expr.accept(this);
     }
 
@@ -118,7 +118,7 @@ class TestVisitor : ASTVisitor
         import std.format : format;
 
         if (fromAssert && !fromStaticAssert &&
-        lastEqualExpression !is null && lastAssert !is null)
+            lastEqualExpression !is null && lastAssert !is null)
         {
             auto e = lastEqualExpression;
             if (e.left !is null && e.right !is null)
@@ -154,7 +154,6 @@ private:
 
     void resetTestState()
     {
-
         fromAssert = false;
         fromStaticAssert = false;
         lastEqualExpression = null;
--- a/assert_writeln_magic.d
+++ b/assert_writeln_magic.d
@@ -1,0 +1,310 @@
+#!/usr/bin/env dub
+/++
+dub.sdl:
+dependency "libdparse" version="~>0.7.0-beta.6"
+name "assert_writeln_magic"
++/
+/*
+ * Tries to convert `assert`'s into user-friendly `writeln` calls.
+ * The objective of this tool is to be conservative as
+ * broken example look a lot worse than a few statements
+ * that could have potentially been rewritten.
+ *
+ *  - only EqualExpressions are "lowered"
+ *  - static asserts are ignored
+ *  - only single-line assers are rewritten
+ *
+ * Copyright (C) 2017 by D Language Foundation
+ *
+ * Author: Sebastian Wilzbach
+ *
+ * Distributed under the Boost Software License, Version 1.0.
+ *    (See accompanying file LICENSE_1_0.txt or copy at
+ *          http://www.boost.org/LICENSE_1_0.txt)
+*/
+// Written in the D programming language.
+
+import dparse.ast;
+import std.algorithm;
+import std.conv;
+import std.experimental.logger;
+import std.range;
+import std.stdio;
+import std.typecons;
+
+private string formatNode(T)(const T t)
+{
+    import dparse.formatter;
+    import std.array : appender;
+
+    auto writer = appender!string();
+    auto formatter = new Formatter!(typeof(writer))(writer);
+    formatter.format(t);
+    return writer.data;
+}
+
+class TestVisitor : ASTVisitor
+{
+    import dparse.lexer : tok, Token;
+
+    this(string fileName)
+    {
+        this.fileName = fileName;
+        fl = FileLines(fileName);
+    }
+
+    alias visit = ASTVisitor.visit;
+
+    override void visit(const Unittest test)
+    {
+        resetTestState();
+        inTest = true;
+        scope(exit) inTest = false;
+        test.accept(this);
+
+        processLastAssert();
+    }
+
+    override void visit(const EqualExpression expr)
+    {
+        enum eqToken = tok!"==";
+        if (inAssert && expr.operator == eqToken && expr.left !is null && expr.right !is null)
+            lastEqualExpression = expr;
+    }
+
+    override void visit(const AssertExpression expr)
+    {
+        if (inFunctionCall)
+            return;
+
+        lastAssert = expr;
+        inAssert = true;
+        expr.accept(this);
+        inAssert = false;
+        fromAssert = true;
+    }
+
+    // for now static asserts are ignored
+    override void visit(const StaticAssertStatement expr)
+    {
+        fromStaticAssert = true;
+        expr.accept(this);
+    }
+
+    /**
+    The following code (in std.concurrency) leads to false positives:
+
+        assertNotThrown!AssertError(assert(receiveOnly!int() == i));
+
+    Hence we simply ignore all asserts in function calls.
+    */
+    override void visit(const FunctionCallExpression expr)
+    {
+        inFunctionCall = true;
+        expr.accept(this);
+        inFunctionCall = false;
+    }
+
+    /// A single line
+    override void visit(const DeclarationOrStatement expr)
+    {
+    processLastAssert();
+        expr.accept(this);
+    }
+
+    void processLastAssert()
+    {
+        import std.uni : isWhite;
+        import std.format : format;
+
+        if (fromAssert && !fromStaticAssert &&
+        lastEqualExpression !is null && lastAssert !is null)
+        {
+            auto e = lastEqualExpression;
+            if (e.left !is null && e.right !is null)
+            {
+                // libdparse starts the line count with 1
+                auto lineNr = lastAssert.line - 1;
+
+                // only replace single-line expressions (for now)
+                if (fl[lineNr].endsWith(";"))
+                {
+                    auto wsLen = fl[lineNr].countUntil!(u => !u.isWhite);
+                    auto indent = fl[lineNr][0 .. wsLen];
+
+                    if (fl[lineNr][wsLen .. $].startsWith("assert", "static assert"))
+                    {
+                        auto left = lastEqualExpression.left.formatNode;
+                        auto right = lastEqualExpression.right.formatNode;
+
+                        if (left.length + right.length > 80)
+                            fl[lineNr] = format("%s// %s\n%swriteln(%s);", indent, right, indent, left);
+                        else
+                            fl[lineNr] = format("%swriteln(%s); // %s", indent, left, right);
+
+                        //writefln("line: %d, column: %d", lastAssert.line, lastAssert.column);
+                    }
+                }
+            }
+        }
+        resetTestState();
+    }
+
+private:
+
+    void resetTestState()
+    {
+
+        fromAssert = false;
+        fromStaticAssert = false;
+        lastEqualExpression = null;
+        lastAssert = null;
+    }
+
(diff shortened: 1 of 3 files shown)


CyberShadow:
This could use a bit more background (for myself included, sorry :)). What problem does this solve again?

From my limited understanding of this it really seems like we are fixing this in the wrong place.

For example, could this go into DPaste instead? Perhaps a CGI proxy between dlang.org and DPaste? I guess fixing this in the compiler/runtime is off the table.

> Should we make a copy of the Phobos sources before?

I don't remember offhand if Digger can cope with one repository's build script modifying another repository, but the test service shouldn't complain. Still, it would be confusing to users: what if you're testing a change, build the website, and suddenly your working directory is dirtied with hundreds of changes that you'd have to carefully clean up so you can commit just your intended changes?

wilzbach:
My fixes to libdparse have been merged (thanks a lot for being so reponsive @Hackerpilot), so now you can preview the effects of the script. 

- [`std.algorithm.iteration`](https://dlang.org/phobos/std_algorithm_iteration.html) with `assert` examples: 
- [`std.algorithm.iteration`](http://dtest.dlang.io/artifact/website-e2ad3d92df8634075b168d136f491ff290d735d7-fc1384e8b0cb8ba3361595308b2051e3/web/phobos-prerelease/std_algorithm_iteration.html) with `writeln` examples

> This could use a bit more background (for myself included, sorry :)). What problem does this solve again?

The idea is that `assert`'s while great for the compiler to test, aren't well-suited for examples as a reader often wants to copy/paste the example and/or play with it. Especially with the new runnable examples this becomes interesting because by default when he clicks on "Run" only "All tests passed" will be shown, which is suboptimal.

So this script tries to rewrite `EqualExpressions`of `AssertExpressions` in `Unittest` blocks into `writeln` calls. For example this snippet

```d
assert([2, 1, 4, 3].minElement == 1);
assert([5, 3, 7, 9].enumerate.minElement!"a.value" == tuple(1, 3));
```

would be rewritten as:

```d
writeln([2, 1, 4, 3].minElement); // 1
writeln([5, 3, 7, 9].enumerate.minElement!"a.value"); // tuple(1, 3))
```

Providing examples with output is a common style across almost all language and library documentations. A couple of examples:
- [C](http://www.cplusplus.com/reference/cstdio/ferror/)
- [C++](http://en.cppreference.com/w/cpp/container/list) (this one also has a "run" feature)
- [Ruby](http://ruby-doc.org/core-2.4.0/Hash.html#method-c-5B-5D)
- [Python](https://docs.python.org/3/library/itertools.html#module-itertools)
- [NumPy](https://docs.scipy.org/doc/numpy/reference/generated/numpy.dot.html)
- [GoLang](https://golang.org/pkg/container/heap/)
...

Even all examples on dlang.org (and the examples included in the Ddoc comment) use `writeln`

CC @MartinNowak and @andralex as they also had the same idea.

> For example, could this go into DPaste instead? Perhaps a CGI proxy between dlang.org and DPaste? I guess fixing this in the compiler/runtime is off the table.

Well it's really intended to be shown on `dlang.org`.

> I don't remember offhand if Digger can cope with one repository's build script modifying another repository, but the test service shouldn't complain. Still, it would be confusing to users:

Yeah I am quite aware of this downside, but doing this transformation in JavaScript with the examples rendered as syntax-highlighted HTML didn't turn it to be a great idea.

CyberShadow:
I see, thanks for the explanation.

What do you think about integrating it into DDox?

CyberShadow:
Something seems to have gone terribly wrong here:
http://dtest.dlang.io/artifact/website-e2ad3d92df8634075b168d136f491ff290d735d7-fc1384e8b0cb8ba3361595308b2051e3/web/library-prerelease/std/algorithm/setops/writeln.html

wilzbach:
> What do you think about integrating it into DDox?

[It seems](https://github.com/dlang/dlang.org/pull/1526) that the switch to DDox won't happen in the foreseeable near future, so this wouldn't allow us to use this "feature" for the default documentation and thus for the majority of all readers/users.
In any case, DDox already uses a libdparse, so it wouldn't be difficult to integrate this into Ddox if we decide to go this road.

> Something seems to have gone terribly wrong here:
http://dtest.dlang.io/artifact/website-e2ad3d92df8634075b168d136f491ff290d735d7-fc1384e8b0cb8ba3361595308b2051e3/web/library-prerelease/std/algorithm/setops/writeln.html

The script automatically adds a `writeln` stub to every module, so that Ddoc compiles successfully.
I am pretty sure it's due to this, but this symbol is set to `private`.

CC @s-ludwig

wilzbach:
> Something seems to have gone terribly wrong here:

I found & fixed the problem: [`setops` ends with a `/++` comment`](https://github.com/dlang/phobos/blob/master/std/algorithm/setops.d#L1272), hence the `writeln` symbol is assumed to be public, even though it's set to private.

CyberShadow:
Alright, well, I think this is fine provided that we can avoid modifying Phobos source code in-place with the caveat I described. It's just not great that this puts in an additional step between the source code and the final output, which makes the entire thing more complicated, but if that can't be avoided then so be it. So, if it's not too difficult, would be good if the changes were done to another private copy, either by copying the source files and modifying them in-place, or having the tool read from one location and write to another. Otherwise we should probably exclude this from the default target (maybe we should introduce a new target that specifies that the files are built by DAutoTest, so we'd only have to change DAutoTest once).

CyberShadow:
> It's just not great that this puts in an additional step between the source code and the final output,

A PR which does that as well: https://github.com/dlang/dlang.org/pull/1214

Though, I think the justification for that one is much weaker (which is also why it's been stalled forever).

wilzbach:
> So, if it's not too difficult, would be good if the changes were done to another private copy,

Sure! I used
-  `rsync --update` s.t. only the exact changes will be propagated to ddoc.
- `wildcard`, s.t. changes to source files will be propagated

Also as `rsync` can't be assumed to be existent, I added a fallback with `cp` that will copy everything.

CyberShadow:
Yeah, I guess this works, though probably the way I'd do it is a makefile wildcard rule to copy the files that need to be copied. This is probably fine too.

wilzbach:
> Yeah, I guess this works, though probably the way I'd do it is a makefile wildcard rule to copy the files that need to be copied. This is probably fine too.

You know, I don't really care. If you want to, I am happy to change ;-)

When I introduced the runnable examples for the prerelease pages, they sat there for half a year without any feedback/notices. Hence I also added this writeln transformation process to the stable pages, because if we decide to go with it, this "testing phase" would be useless. The preview of the changes on DAutoTest is awesome and should be enough to make an informed decision.

CyberShadow:
> You know, I don't really care. If you want to, I am happy to change ;-)

No, that's completely fine.

posix.mak
```diff
@@ -508,6 +518,34 @@ d.tag : chmgen.d $(STABLE_DMD) $(ALL_FILES) phobos-release druntime-release
...
+# Assert -> writeln magic
+# -----------------------
+#
+# - This transforms assert(a = b) to writeln(a); // b
+# - It creates a copy of Phobos to apply the transformations
+################################################################################
+
+# --update allows to copy only the newer files and thus only propagate these
+#  changes
+HAS_RSYNC := $(shell command -v rsync22 2> /dev/null)
```
CyberShadow: Why `rsync22`? Typo? The commands below run `rsync`.

CyberShadow: Is `command` like `which` here?

assert_writeln_magic.d
```diff
@@ -0,0 +1,313 @@
...
+    this(string repoDir, string path)
+    {
+        writeln("opening: ", path);
+
+        this.path = path;
+        destFile = buildPath(repoDir, path);
+        lines = File(destFile).byLineCopy.array;
+
+        // it's a good practise to use a common tmp folder -> easier to look at or clean
+        tmpDir = buildPath(tempDir, "file_tester", path.stripExtension.replace("/", "_"));
```
CyberShadow: `.generated` ?

assert_writeln_magic.d
```diff
@@ -0,0 +1,313 @@
...
+            }
+        }
+        resetTestState();
+    }
+
+private:
+
+    void resetTestState()
+    {
+
```
CyberShadow: And here

wilzbach:
> I'd say that the `.generated` directory would be a better fit, to avoid more littering

I didn't do this initially because of build errors due to the hard-coded paths, but I forgot that one can simply overwrite variables in `make` :)

> Why `rsync22`? Typo? The commands below run `rsync`.

Yeah :/
Somehow I was trying whether `HAS_RSYNC` works and somehow this ended up in the git diff. Sorry

> Is `command` like `which` here?

Yes
-> http://stackoverflow.com/questions/592620/check-if-a-program-exists-from-a-bash-script

CyberShadow:
> -> http://stackoverflow.com/questions/592620/check-if-a-program-exists-from-a-bash-script

TIL, thanks.

assert_writeln_magic.d
wilzbach: `tmpDir` is only used when writing the file line by line and doing the swap. It will be immediately be removed afterwards. Isn't this the dedicated purpose of a `tempDir()` or is there another reason to use `.generated`?

CyberShadow:
I have no further feedback so the only thing remaining is to squash any comments you think ought to be squashed before merging.

assert_writeln_magic.d
CyberShadow: OK. Currently on DAutoTest `.generated` is on a RAM disk but `/tmp` is not, but I should set the `TMP` variable to fix that.

wilzbach:
> the only thing remaining is to squash any comments you think ought to be squashed before merging.

Currently `~master` of libdparse is fetched as there hasn't been a new release with the bug fixes included (though they have already been merged :)).
It's quite unlikely that @Hackerpilot will break libdparse, but it's still a bad practice to run a CI without a fixed version, so I think we should either wait a bit for a new version (or work around this limitation in DUB by e.g. embedding libdparse as a git submodule.)

CyberShadow:
Does dub allow specifying dependencies by commit hash yet? I think Sönke said he'd do it at some point.

wilzbach:
> Does dub allow specifying dependencies by commit hash yet?

Yes it's on the roadmap, but not implemented (as Sönke is quite busy with Vibe.d and other projects).
There's a DEP for it though: https://github.com/dlang/dub/wiki/DEP6

Hackerpilot:
@wilzbach https://github.com/Hackerpilot/libdparse/releases/tag/v0.7.0-beta.7

assert_writeln_magic.d
wilzbach: (I wouldn't mind changing this)

wilzbach:
> @wilzbach https://github.com/Hackerpilot/libdparse/releases/tag/v0.7.0-beta.7

Wow. Thanks a lot!

wilzbach:
> CyberShadow/DAutoTest — Build failed

Arrgh it seems that there was an error with the DUB version. Fixed & it's green again :)

CyberShadow:
Nothing else is blocking this, right?

I'll be looking forward to removing the intermediate source file generation whenever DDox becomes the main documentation method.

BTW, I went ahead and hacked DPaste to accept dtest.dlang.io for cross-side requests, so that previews of runnable examples work.

wilzbach:
> Nothing else is blocking this, right?

Not that I would know of.

> I'll be looking forward to removing the intermediate source file generation whenever DDox becomes the main documentation method.

Ditto!
It's really a hassle to fight against two documentation engines :/

> BTW, I went ahead and hacked DPaste to accept dtest.dlang.io for cross-side requests, so that previews of runnable examples work.

Sweet! Thanks a lot! I haven't been able to reach Damian lately, but my wishlist would include:
- support `dmd-nightly` (for the pre-release pages, e.g. `checkedint` doesn't work currently)
- add a (simple) cache (most users press on "run" without editing)
- check whether there's a speed difference in using `dmd -run` instead of `rdmd`

CyberShadow:
> Sweet! Thanks a lot! I haven't been able to reach Damian lately, but my wishlist would include:

If you want to have a go at it I'll gladly give you access :)

CyberShadow:
Uh oh, the dlang.org autodeploy build failed:

```
Linking...
Running ./assert_writeln_magic -i .generated/phobos-prerelease
std.file.FileException@std/file.d(2229): /tmp/file_tester/unittest: Permission denied
```

CyberShadow:
Right, because the same directory is being shared by multiple users.

I'm going to work around this by setting `TMP` to a new directory under the working directory, which should also improve performance and encapsulation, but going forward I think we should avoid using globally unique file names especially when we need to write to them.

wilzbach:
> Uh oh, the dlang.org autodeploy build failed:
> ...
> but going forward I think we should avoid using globally unique file names especially when we need to write to them.

I am happy to make a follow-up PR to change from `/tmp` to `.generated` -  just let me know.

assert_writeln_magic.d
```diff
@@ -0,0 +1,312 @@
...
+    foreach (file; files)
+        if (!ignoredFiles.any!(x => file.name.canFind(x)))
+            file.name.parseFile;
+}
+
+/**
+A simple line-based in-memory representation of a file.
+ - will automatically write all changes when the objct is destructed
+ - will use a temporary file to do safe, whole file swaps
```
CyberShadow: BTW if you wanted safe, atomic file replacements I think writing to `<filename>.tmp` and renaming over it would be closer to that goal, since POSIX renames are atomic.

CyberShadow:
https://github.com/CyberShadow/ae/commit/c919ef28f218b7f200e7c05e4fd6c24371dbb9b1

Seems to be fixed:

https://dlang.org/library-prerelease/std/algorithm/comparison/among.html

assert_writeln_magic.d
wilzbach: My motivation to save `<filename>.tmp` in a `tmp` folder was that the process could be stopped/interrupted during the write of the temporary file.

CyberShadow: Yes, but the same applies to your `copy` call.

andralex: Yah, renames are atomic only within the same filesystem which is why it's important to put the xyz.tmp file in the same place as xyz (/tmp might be on a different filesystem etc).

BTW the pattern @CyberShadow has saved my behind many times, and is considered par for the course across Facebook eng.

MartinNowak:
This broke the builds script b/c your invocation is relying on some system-wide dmd. Please change it to use `--compiler=$(STABLE_DMD)`.

wilzbach:
> This broke the builds script b/c your invocation is relying on some system-wide dmd. Please change it to use --compiler=$(STABLE_DMD).

-> https://github.com/dlang/dlang.org/pull/1592

PR #1584, opened 2017-02-26
documents the limitation of attributes on overridden functions

BBasile:
Just to prevent any report tagged "reject-valid" after this D front-end change: https://github.com/dlang/dmd/pull/6570

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -921,6 +921,31 @@ void main()
 }
 ------
 
+    $(P It's not allowed to mark an overridden method with the attributes
+    $(LINK2 attribute.html#disable, $(D @disable)) or
+    $(LINK2 attribute.html#deprecated, $(D deprecated)).
+    To stop the compilation or to output the deprecation message, the compiler
+    must be able to determine the target of the call, which can't be guaranteed
+    when it is virtual.
+    )
+
+------
+class B
+{
+    void foo() {}
+}
+
+class D : B
+{
+    @disable override void foo() {}
+}
+
+void main()
+{
+    B b = new D;
+    b.foo(); // would compiles and then the most derived would be called even if disabled.
+}
+------
 
 $(H4 $(LNAME2 inline-functions, Inline Functions))
 


spec/function.dd
```diff
@@ -921,6 +921,31 @@ void main()
 }
 ------
 
+    $(P It's not allowed to mark an overridden method with the attributes
+    $(LINK2 attribute.html#disable, $(D @disable)) or
```
wilzbach: please use a macro for such links, e.g `RELATIVE_LINK2` would work ;-)

adamdruppe:
On Mon, Feb 27, 2017 at 09:17:56PM -0800, Sebastian Wilzbach wrote:
> +    $(P It's not allowed to mark an overridden method with the attributes
> +    $(LINK2 attribute.html#disable, $(D @disable)) or
> 
> please use a macro for such links, e.g `RELATIVE_LINK2` would work ;-)

I'm not really following this (just saw an email), but why? Linking with the ddoc macros is complicated enough as it is, so why use another for a standard url feature like relative linking? I don't think I have ever seen RELATIVE_LINK2 before (though again, I haven't been following closely) and don't really see why you'd need it, even for other bases like library-prerelease.

I can understand having a macro for linking to the spec which prepends /whatever-base/spec/ to it, but I'd call that SPEC_LINK not RELATIVE_LINK2...

wilzbach:
> I'm not really following this (just saw an email), but why? Linking with the ddoc macros is complicated enough as it is, so why use another for a standard url feature like relative linking? I don't think I have ever seen RELATIVE_LINK2 before (though again, I haven't been following closely) and don't really see why you'd need it, even for other bases like library-prerelease.

I have just been trying to review this and compared the link style with the prevalent one in the document.
My assumption for the need was that from the spec a PDF, mobi and Kindle ebook is generated?

adamdruppe:
On Mon, Feb 27, 2017 at 09:28:10PM -0800, Sebastian Wilzbach wrote:
> I have just been trying to review this and compared the link style with the prevalent one in the document.

huh, yeah, it is used elsewhere so I guess we can just be consistent with that here.

> My assumption for the need was that from the spec a PDF, mobi and Kindle ebook is generated?

As far as I can tell, the definition is pretty consistently just an alias for LINK2. It is reasonable for a contributor to be confused by it...

wilzbach:
> huh, yeah, it is used elsewhere so I guess we can just be consistent with that here.

@BBasile would you mind updating your PR, so that we can merge it and move forward?

adamdruppe:
@wilzbach RELATIVE_LINK2 cannot possibly replace that LINK2... the definition in html.ddoc puts a `#` on the href meaning it can't cross pages.

BBasile:
@wilzbach, not done because not needed. LINK2 is good here.

BBasile:
@wilzbach, do we agree on this PR finally ? That's already too much discussion for such a detail.

PR #1585, opened 2017-02-27
Add generated changelog to the changelog menu

wilzbach:
(setting this to blocked until we find a nice way to mark it as "to be released")


wilzbach:
Superseded by https://github.com/dlang/dlang.org/pull/1769

PR #1586, opened 2017-02-27
Use nightly changelog theme

wilzbach:
It seems like we are stuck with the `VERSION` macro.
The best solution I found is to overwrite temporarily it with `NIGHTLY_VERSION`, however this depends on https://github.com/dlang/tools/pull/221

To make it "stable" (aka without the nightly information), the easiest way is to set `BUILD_NIGHTLY=0`

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -121,7 +121,8 @@ ifeq (1,$(DIFFABLE))
 	NODATETIME := nodatetime.ddoc
 	DPL_DOCS_PATH_RUN_FLAGS := --no-exact-source-links
 else
-	CHANGELOG_VERSION := "v${LATEST}..upstream/stable"
+	CHANGELOG_VERSION_MASTER := "v${LATEST}..upstream/master"
+	CHANGELOG_VERSION_STABLE := "v${LATEST}..upstream/stable"
 endif
 
 # Documents
@@ -171,8 +172,8 @@ SPEC_ROOT=$(addprefix spec/, \
 	abi simd)
 SPEC_DD=$(addsuffix .dd,$(SPEC_ROOT))
 
-CHANGELOG_FILES=$(basename $(subst _pre.dd,.dd,$(wildcard changelog/*.dd))) \
-				changelog/${NEXT_VERSION}
+CHANGELOG_FILES=changelog/${NEXT_VERSION}_pre \
+				$(basename $(subst _pre.dd,.dd,$(wildcard changelog/*.dd))) \
 
 # Website root filenames. They have extension .dd in the source
 # and .html in the generated HTML. Save for the expansion of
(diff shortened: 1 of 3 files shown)


CyberShadow:
Awesome!

posix.mak
```diff
@@ -42,6 +42,9 @@ STABLE_DMD_CONF=$(STABLE_DMD).conf
 STABLE_RDMD=$(STABLE_DMD_ROOT)/dmd2/$(OS)/$(if $(filter $(OS),osx),bin,bin$(MODEL))/rdmd \
 	--compiler=$(STABLE_DMD) -conf=$(STABLE_DMD_CONF)
 
+# whether to use a special theme for nightly changelog builds
+BUILD_NIGHTLY=1
```
CyberShadow: Why is this option needed?

wilzbach: The idea was that Martin can call the Makefile with `BUILD_NIGHTLY=0` and then it will generate the to be released page without the nightly texts. In other words:
- current date instead of "to be released"
- `VERSION` instead of `NIGHTLY_VERSION`
- no `(upcoming)` in the title
- ...

wilzbach: See also https://github.com/dlang/tools/pull/222 for an example on more "customization" based on `BUILD_NIGHTLY`

wilzbach:
>> OK, LGTM and leaving this for @MartinNowak to review as it affects the release process.
> Just name the changlog _pre.d will build a pre-release changlog.

-> renamed to `changelog/${NEXT_VERSION}_pre.dd` for the default generation (nightly).

There's also a `changelog/${NEXT_VERSION}.dd` target, which can be used by to generate the "release version" of the changelog (without "to be released", nightlies download link, ...).

As we now have two different files I added the `_pre` files to the `.gitignore`, s.t. a default build of `dlang.org` doesn't generate any "new files".

> Always using stable seems incorrect when the purpose of this target is trying to build the changelog for dmd/druntime/phobos PRs.

I have changed it, s.t. for the `_pre` changelog files `upstream/master` is used instead.
Due to the new `DIFFABLE` Makefile variable this doesn't matter on DAutoTest anyways.

CyberShadow:
@MartinNowak If you're happy with these could you please merge them? Thanks!

MartinNowak:
THX

posix.mak
```diff
@@ -588,12 +589,16 @@ test:
 # Changelog generation
 ################################################################################
 
+changelog/${NEXT_VERSION}_pre.dd: ${STABLE_DMD} ../tools ../installer
+	$(STABLE_RDMD) $(TOOLS_DIR)/changed.d $(CHANGELOG_VERSION_MASTER) -o $@ \
+	--version "${NEXT_VERSION} (upcoming)" --date "To be released" --nightly
+
 changelog/${NEXT_VERSION}.dd: ${STABLE_DMD} ../tools ../installer
```
MartinNowak: Do we still need that rule for anything?

PR #1587, opened 2017-02-27
Changelog: build next stable (aka minor) version

wilzbach:
> Sorry, but can we not use exactly the same method we use for the `master` branch but for the `stable` branch? And use the same method to generate the file name, except increment the minor version number instead of the major one.

Sure, but I am not sure whether this is useful because at the moment the manual text changelog files are only intended for major releases, so the page will only included a list of bugs from Bugzilla, which might introduce "semi-random" changes in the DAutoTest output.

OTOH it's nice to have a preview for this as well and we already have this Bugzilla bug problem for the generated major version.

Hence leaving this up to @MartinNowak

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -14,7 +14,8 @@ ifeq (,${LATEST})
 LATEST:=$(shell cat VERSION)
 endif
 # Next major DMD release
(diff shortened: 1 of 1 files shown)


wilzbach:
It seems like this isn't important to @MartinNowak, so I am closing this ;-)

PR #1588, opened 2017-02-27
Better diff mode (DIFFABLE=1)

wilzbach:
Follow-up to https://github.com/dlang/tools/pull/222 and https://github.com/dlang/dlang.org/pull/1537/

>> Btw as this could also lead to unrelated diffs, should we disable the Bugzilla query for bugs for the changelog builder in "nightly mode"?

> Hmm, yeah, changes that shouldn't result in documentation changes should not have documentation changes indicated in the status line, because then inadvertent documentation changes will get through. I guess with this and https://github.com/dlang/dlang.org/pull/1537 we should expand `NODATETIME` to something more generic.

### DIFF
--- a/dpl-docs/views/layout.dt
+++ b/dpl-docs/views/layout.dt
@@ -6,6 +6,7 @@ html(lang='en-US')
     | http://digitalmars.com
   - import std.process : environment;
   - string version_id = environment["GIT_TARGET"];
+  - bool noExactSourceCodeLinks = environment["NO_EXACT_SOURCE_CODE_LINKS"] == "1";
   - bool haveVersion = version_id.startsWith("v");
   - string version_name = haveVersion ? version_id[1 .. $] : "Prerelease";
   - string root_dir = info.linkTo(null) ~ (req is null ? "../" : "");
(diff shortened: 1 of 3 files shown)


dpl-docs/views/layout.dt
```diff
@@ -6,6 +6,7 @@ html(lang='en-US')
     | http://digitalmars.com
   - import std.process : environment;
   - string version_id = environment["GIT_TARGET"];
+  - bool noExactSourceCodeLinks = environment["NO_EXACT_SOURCE_CODE_LINKS"] == "1";
```
wilzbach: Apparently that's the "workaround" to inject parameters into the layout file - it's also the prevalent style here (see above).

CyberShadow:
```
[2017-02-27 09:13:16.384055] Running: 'make' '-f' 'posix.mak' 'DIFFABLE=1' 'all' 'kindle' 'pdf' 'verbatim'
```

Seems to be working from my side :)

CyberShadow:
Relevant change: https://github.com/CyberShadow/ae/commit/a3134576583d1919995a85b8282cf955b1cfd7de

CyberShadow:
Why the broken "Previous version.html" link? Will that be resolved by https://github.com/dlang/tools/pull/222?

wilzbach:
> Why the broken "Previous version.html" link? Will that be resolved by dlang/tools#222?

Good catch!
It's due to the `changed.d` trying to detect the previous version based on the git revision pattern, which isn't passed anymore. Luckily there's the `--prev-version` flag (fixed).

PR #1593, opened 2017-03-01
Added padding to links in Phobos book table

JackStouffer:
Just a small improvement. I think the links can use a little breathing room.

CC @wilzbach

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -997,6 +997,11 @@ table td:not(:last-child), table th:not(:last-child)
     padding-right: 1em;
 }
 
+table.book tbody a
+{
+    padding-right: .6em;
+}
+
 hr
 {
     margin: 2em 0;


JackStouffer:
Before

screen shot 2017-03-01 at 10 22 49 am

After

screen shot 2017-03-01 at 10 22 00 am

wilzbach:
Hmm also note that this will modify existing pages like, e.g.

image

After:

image

Given that this page has its issues - it looks a bit weirder afterwards. However on the regex page it looks definitely better, so I am undecided.
@CyberShadow what's your opinion?
How about enabling it only for BOOKTABLES with the `quickindex` class?

A bit OT
-----------

What do you think about using "Match, Build, Replace, Split, Objects" as category names instead?

CyberShadow:
> @CyberShadow what's your opinion?

I guess it is a minor improvement.

Would be nice to avoid `px` units, especially since this is something that is intended to be relative to the font size.

JackStouffer:
> Would be nice to avoid px units, especially since this is something that is intended to be relative to the font size.

I'll change it

> Hmm also note that this will modify existing pages like, e.g.

I will note that that page looks much better on the /phobos/ version

screen shot 2017-03-01 at 10 54 13 am

JackStouffer:
The odd spaces come from these `&nbsp;`s that are inserted everywhere for some reason

wilzbach:
>  The odd spaces come from these `&nbsp;`s that are inserted everywhere for some reason. If you delete them from the HTML, the page looks normal

Alrighty - yet another weirdness of Ddoc. I will have a look at it. Maybe I can find the reason ...

wilzbach:
> The odd spaces come from these &nbsp;s that are inserted everywhere for some reason. 

It's comes from the `SUBREF` macro here:
https://github.com/dlang/phobos/blob/master/std/algorithm/package.d#L175

It seems like a hack to get a better text justification...

JackStouffer:
Styling stuff should live in CSS.

Delete the non-breaking space and see what happens.

PR #1595, opened 2017-03-02
Footer: show always the current year and the D Language Foundation

wilzbach:
Why is this important?
------------------------------

An updated copyright that is older than X years shows that a file hasn't been maintained properly and the trust of the user sink immediately if he reads that the file was touched for the last time ten years ago.
For most Phobos modules the copyright is __never__ updated as this is regarded as unnecessary churn and/or because different copyright year formats are used.

For an impression, a couple of randomly selected modules:

- https://dlang.org/phobos/std_encoding.html
- https://dlang.org/phobos/std_signals.html
- https://dlang.org/phobos/std_zlib.html

Why the D Language Foundation?
----------------------------------------

- The footer is inserted by the _owner_ of the website and documentation which is the D Language Foundation. The author header at the beginning of each module, which is a lot more visible than the tiny foooter still shows all authors of a file (that the "Authors" there don't reflect the actual list of Authors is a different topic for another discussion)
- This is _free_ advertisement for D and its foundation
- Boost only requires us to include the full Boost header in all source files, which btw we mostly don't do:

> The copyright notices in the Software and __this entire statement__, including
the above license grant, this restriction and the following disclaimer,
must be included in all copies of the Software, in whole or in part, and
all derivative works of the Software, unless such copies or derivative
works are solely in the form of machine-executable object code generated by
a source language processor.

http://www.boost.org/LICENSE_1_0.txt

Examples from other Languages
---------------------------------------------

- https://docs.python.org/3/library/itertools.html
- http://www.scala-lang.org/api/2.12.1/scala/collection/index.html
- http://php.net/manual/en/language.exceptions.php
- https://docs.perl6.org/language/nativecall
- https://nodejs.org/api/http.html (no Copyright notice!)
- https://doc.rust-lang.org/std/hash (no Copyright notice!)
- http://en.cppreference.com/w/cpp/header/bitset (only About)


dlang.org.ddoc
```diff
@@ -54,9 +54,9 @@ DDOC=
 <!DOCTYPE html>
 <html lang="en-US">
 <!--
-    Copyright (c) 1999-$(YEAR) by Digital Mars
-    All Rights Reserved Written by Walter Bright
-    http://digitalmars.com
+    Copyright (c) 1999-$(YEAR) by the D Language Foundation
+    All Rights Reserved.
+    http://dlang.org/foundation.html
```
andralex: Use https

dpl-docs/views/layout.dt
```diff
@@ -1,9 +1,11 @@
...
 html(lang='en-US')
+  - import std.datetime : Clock;
+  - auto year = Clock.currTime.year;
   //
-    | Copyright (c) 1999-2016 by Digital Mars
-    | All Rights Reserved Written by Walter Bright
-    | http://digitalmars.com
+    | Copyright (c) 1999-$#{year} by the D Language Foundation
+    | All Rights Reserved.
+    | http://dlang.org/foundation.html
```
andralex: ditto

dpl-docs/views/layout.dt
```diff
@@ -212,9 +214,8 @@ html(lang='en-US')
         |     
         #quickindex.quickindex
 
-    #copyright
-      block copyright
-      | | Page generated by <a href="https://github.com/rejectedsoftware/ddox">ddox</a>.
+        #copyright.smallprint
+          | Copyright &copy; 1999-2017 by the <a href="#{root_dir}/foundation.html">D Language Foundation</a> | Page generated by <a href="https://github.com/rejectedsoftware/ddox">ddox</a>.
```
andralex: year is hardcoded here, can you use `$#{year}`?

ebook.ddoc
```diff
@@ -3,9 +3,9 @@ DDOC=
 <html lang="en-US">
 
 <!--
-	Copyright (c) 1999-$(YEAR) by Digital Mars
-	All Rights Reserved Written by Walter Bright
-	http://www.digitalmars.com
+    Copyright (c) 1999-$(YEAR) by the D Language Foundation
+    All Rights Reserved.
+    http://dlang.org/foundation.html
```
andralex: https

dlang.org.ddoc
CyberShadow: I'm guessing @WalterBright is OK with what essentially indicates a copyright assignment and transfer of ownership to the D foundation :)

wilzbach:
> I'm guessing @WalterBright is OK with what essentially indicates a copyright assignment and transfer of ownership to the D foundation :)

Sorry this time I did a rebase, so the comment from @CyberShadow got lost.
@andralex or @WalterBright could you quickly confirm this?

andralex:
This should be fine, if @WalterBright wants to amend things we can always change it back.

wilzbach:
> This should be fine, if @WalterBright wants to amend things we can always change it back.

Awesome! As I fixed the nits, I am adding "auto-merge" now :)

dlang.org.ddoc
```diff
@@ -40,7 +40,7 @@ COMMON_SCRIPTS_DLANG =
 _=
 
 CONSOLE=$(TC pre, console notranslate, $0)
-COPYRIGHT=Copyright &copy; 1999-$(YEAR) by the $(LINK2 $(ROOT)/foundation.html, D Language Foundation)
+COPYRIGHT_FOUNDATION=Copyright &copy; 1999-$(YEAR) by the $(LINK2 $(ROOT)/foundation.html, D Language Foundation)
```
wilzbach: Meh apparently it's better to use `ROOT_DIR` without the leading `/`

wilzbach:
FYI @CyberShadow it seems that for `dlang.org` the bot still can't merge "auto-merge" PRs.
It's a bit hard to debug without much [logging output](https://github.com/dlang-bots/dlang-bot/issues/59).

However, maybe it simply doesn't receive a hook event?

CyberShadow:
This repository has 2 webhooks, while Phobos has 8. So... maybe? :)

One of those on Phobos but not here is https://dlang-bot.herokuapp.com/github_hook

@MartinNowak

wilzbach:
> One of those on Phobos but not here is https://dlang-bot.herokuapp.com/github_hook

Yep that's the one :)

PR #1596, opened 2017-03-02
include osmodel.mak from tools repo

wilzbach:
I realized that it's quite hard to get an outdated file, if there's only one and as we already require the `dmd` repo, we can directly use the `osmodel.mak` file from there. It rarely changes.

https://github.com/dlang/druntime/pull/1780
https://github.com/dlang/phobos/pull/5228
https://github.com/dlang/tools/pull/227

edit: rebased to use `tools` instead


braddr:
We don't already require the dmd repo, we require a dmd executable, very different things.  Please don't add cross package dependencies like this.  The same applies to the other changes that are adding git clone and similar cross package dependencies.. just don't.

CyberShadow:
@braddr What does your autotester have to do with the dlang.org repo though? The documentation tester does in fact require the DMD repo because we build documentation for the DMD source code.

wilzbach:
> Please don't add cross package dependencies like this. The same applies to the other changes that are adding git clone and similar cross package dependencies.. just don't.

Have you had a recent look at the `posix.mak` here?
The "cross-package" have existed since ages ...

https://github.com/dlang/dlang.org/blob/master/posix.mak#L350

To build `dlang.org` one needs to clone `dmd`, `druntime`, `phobos` and `tools`, so I honestly don't see the issue.

FYI: As suggested on the other PRs, I changed the "one source of truth" to `tools`.

braddr:
Apologies, the straw that pushed me into objecting to the cross package dependency changes turned out to be the one that was already so deep in that hole that another hardly matters.  The objection still holds for the others and if you need me to, I'll repeat myself across each of them.

wilzbach:
> The objection still holds for the others and if you need me to, I'll repeat myself across each of them.

Discussions are at:

https://github.com/dlang/tools/pull/227
https://github.com/dlang/dmd/pull/6584

Maybe we move this discussion to the internal ML, s.t. you don't need to repeat yourself ;-)

wilzbach:
Superseded by https://github.com/dlang/dlang.org/pull/1791

PR #1598, opened 2017-03-03
Updated makefile to use dmd/generated/os/release/model/dmd

RazvanN7:
I saw there is also the variable $(DMD_STABLE_DIR) which is used for $(DMD-REL) that uses the hierarchy dmd/src/dmd. I didn't modify that one to point to generated/.../dmd since I assume that the changes in [1] are not stable yet.

[1] https://github.com/dlang/dmd/pull/6562

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -27,7 +27,7 @@ INSTALLER_DIR=../installer
 DUB_DIR=../dub-${DUB_VER}
 
 # External binaries
-DMD=$(DMD_DIR)/src/dmd
+DMD=$(DMD_DIR)/generated/$(OS)/release/$(MODEL)/dmd
 DUB=${DUB_DIR}/bin/dub
 
 # External directories


wilzbach:
Two things I found while scanning `posix.mak`:

1) I think you should add `dmd` to the list of dumped modules for the prerelease docs
(this isn't a blocker, you can also do it after your ddmd -> dmd move)

-> https://github.com/dlang/dlang.org/blob/master/posix.mak#L277

```
$(addprefix --dump , object std etc core ddmd) >$@
```

2) Would you be so kind to remove the redundant `DMD_GEN` variable and replace its occurrences with `DMD`?

```
L96:
DMD_GEN=$(DMD_DIR)/generated/$(OS)/release/$(MODEL)
```

Thanks!

wilzbach:
> MartinNowak added the auto-merge label 4 minutes ago

@MartinNowak the dlang-bot doesn't receive hooks from `dlang.org`.
See also: https://github.com/dlang/dlang.org/pull/1595#issuecomment-283729705

PR #1600, opened 2017-03-05
Make link padding only apply to more specific cases

adamdruppe:
The old rule was too general and cause silly looking spaces as Andrei
described here:

http://forum.dlang.org/post/ohjurtkzhvufqsahmimj@forum.dlang.org

This makes it only happen where I believe it was intended to happen:
on the hand-written index replacement, such as here:
https://dlang.org/phobos-prerelease/std_algorithm.html

where we have a long list of links all right next to each other under
the quickindex div.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -997,7 +997,28 @@ table td:not(:last-child), table th:not(:last-child)
     padding-right: 1em;
 }
 
-table.book tbody a
+/*
+  These next two rules are for pages like:
+  http://dlang.org/phobos-prerelease/std_regex.html
+
+  The idea is if we have a list of links, we want space
+  to the right of each one. But detecting a list of links
+  vs just a single link requires some thought: the css
+  adjacent sibling rule does well detecting all but the first
+  one in a list, so we'll add space to the right of them (that's
+  the second rule here), but that leaves an inconsistent space
+  after the first item.
+
+  That's where the following rule comes in: on the item directly
+  following the first link, add some left padding to simulate the
+  first one having the same space as the others.
+*/
+table.book tbody a:first-child + a
+{
+    padding-left: .6em;
+}
+
+table.book tbody a + a
 {
     padding-right: .6em;
 }


wilzbach:
> The old rule was too general

(and was introduced three days ago):

https://github.com/dlang/dlang.org/pull/1593

CC @JackStouffer

wilzbach:
> though I didn't actually test it... is there an online preview for that?

Hmm while your CSS rule makes a lot of sense, it doesn't make std.algorithm look better :/

image

http://dtest.dlang.io/artifact/website-261e2794158ca0d40ba40d6e8230dd9447135a50-49809e8d61a51aa8f109fddf642ae1a7/web/phobos-prerelease/std_algorithm.html

JackStouffer:
@wilzbach Was just about to draw attention to that. Here's the before pic for reference

screen shot 2017-03-04 at 11 30 28 pm

adamdruppe:
Hmm, I didn't think about the line wrapping... so let's go back to padding-right... but I still want to do it only on a list. The quickindex isn't consistently present in the right places either :S 

Well, let's try trusting the adjacent selector and special casing the second element, to act like the first has the space too (standard css afaik doesn't permit us to do `a:has(+a)` (my dom.d does though :P dom.d rox)... let's try it.

ZombineDev:
@adamdruppe there's a small CSS bug when there's room for only a single link in the rightmost column - in that case the next link goes on the next line, but gets padded to the right. Look at largestPartialIntersection:
screenshot_2017-03-05-10-49-07

screenshot_2017-03-05-10-51-58

adamdruppe:
Hmm, maybe margin instead of padding can help with that. But really what we might be forced to do is go through the source files and add a class name to target for these lists on the container.

But, I do think that's relatively minor so at least we're a step ahead from where we were last night.

BTW, I don't really wanna spend a lot of time on this since IMO these tables are just a hack papering over a ddoc deficiency. The other two generators can do this kind of thing automatically anyway...

ZombineDev:
> But, I do think that's relatively minor so at least we're a step ahead from where we were last night.

Agreed. My comment just just FYI.

wilzbach:
Meh it seems that if the links in std.functional would be proper links, it would also be affected:
https://github.com/dlang/phobos/pull/5245

PR #1601, opened 2017-03-05
Fix dmd source links for DDoc and DDox

wilzbach:
For Ddoc we have the problem that `ddoc` is run in the `src` directory, which means the `SRCFILENAME` starts with `ddmd/...`
For `druntime` this worked before as `ddoc` is run outside of the `src` directory.
My simple solution is to add a `PROJECT_SOURCE_DIR` variable to the `project.ddoc` file at `dmd`.

@RazvanN7 as you would need to update these files anyways, I was nice and already set them to `dmd` (nobody uses the online documentation of dmd at the moment anyways).

### DIFF
--- a/dlang.org.ddoc
+++ b/dlang.org.ddoc
@@ -136,7 +136,7 @@ DDOCKEYVAL2=$(DIVC keyval $1, $(SPANC key key$1, $2:) $(DIVC val val$1, $(TAIL $
 DDSUBLINK=$(LINK2 $(ROOT_DIR)$1.html#$2, $3)
 _=
 
-DMDSRC=$(HTTPS github.com/D-Programming-Language/dmd/blob/master/src/$0, $0)
+DMDSRC=$(HTTPS github.com/dlang/dmd/blob/master/src/dmd/$0, $0)
 DOT_PREFIXED=.$1$(DOT_PREFIXED $+)
 DOT_PREFIXED_SKIP=$(DOT_PREFIXED $+)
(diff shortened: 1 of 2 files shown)


wilzbach:
> My simple solution is to add a PROJECT_SOURCE_DIR variable to the project.ddoc file at dmd.

-> https://github.com/dlang/dmd/pull/6591

rainers:
According to the preview, this doesn't work correctly (anymore?), e.g. ddmd.access links to https://github.com/dlang/dmd/blob/master/src/dmd/access.d. Maybe `PROJECT_SOURCE_DIR` needs to be updated again?

Also, a lot of the backend source file links look very broken, e.g. http://dtest.dlang.io/artifact/website-4edd0524db57eff02e7186756f620f099ce55137-c3942c8fa8aa4fd3efeed3fda27b4068/web/phobos-prerelease/ddmd_backend_cc.html

wilzbach:
> According to the preview, this doesn't work correctly (anymore?)

I think I have submitted the PR before dlang/dmd#6591 was merged, so the links never worked on DAutoTest.
I just rebased, s.t. DAutoTest can test it again.

wilzbach:
>> I think I have submitted the PR before dlang/dmd#6591 was merged, so the links never worked on DAutoTest.
>I just rebased, s.t. DAutoTest can test it again.

Yep it worked well. DAutoTest nows shows only the fixed links.
I am going to merge this now as it's a very simple, approved PR that fixes broken source links, DAutoTest likes it and I intend to write a small blog post on the NG about docs for DMD as a library.

PR #1602, opened 2017-03-06
Removed redundant DMD_GEN variable and replaced it's occurences with DMD

RazvanN7:
The makefile had a redundant variable : $DMD_GEN which stored the same path as $DMD. I erased it and replaced it's occurences with $DMD. 

CC @wilzbach : regarding the other matter, I think it would be best to add dmd to the list of dumped modules after the switch: ddmd -> dmd

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -525,7 +525,7 @@ docs-prerelease.json : ${DMD} ${DMD_DIR} ${DRUNTIME_DIR} \
 	  -e /unittest/d >> .prerelease-files.txt
 	find ${PHOBOS_DIR_GENERATED} -name '*.d' | sed -e /unittest.d/d \
 	  -e /windows/d >> .prerelease-files.txt
-	${DMD} -J$(DMD_DIR)/res -J$(DMD) -c -o- -version=MARS -version=CoreDdoc \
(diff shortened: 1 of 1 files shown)


wilzbach:
> , I think it would be best to add dmd to list of dumped module after the switch ddmd -> dmd

Sure. Good luck!

RazvanN7:
@wilzbach @CyberShadow any ideas why the auto-tester is failing?

posix.mak
```diff
@@ -527,7 +525,7 @@ docs-prerelease.json : ${DMD} ${DMD_DIR} ${DRUNTIME_DIR} \
 	  -e /unittest/d >> .prerelease-files.txt
 	find ${PHOBOS_DIR_GENERATED} -name '*.d' | sed -e /unittest.d/d \
 	  -e /windows/d >> .prerelease-files.txt
-	${DMD} -J$(DMD_DIR)/res -J$(DMD_GEN) -c -o- -version=MARS -version=CoreDdoc \
+	${DMD} -J$(DMD_DIR)/res -J$(DMD) -c -o- -version=MARS -version=CoreDdoc \
```
wilzbach: Use sth. like `$(dirname $(DMD))`  here, then it should pass

RazvanN7:
@wilzbach thanks!

PR #1603, opened 2017-03-06
Make links in 404 pages absolute

CyberShadow:
Fixes display of 404 pages for any non-root URL (e.g. http://dlang.org/phobos/nonexisting.html).

There's more than one way to skin this cat, such as to use full URLs (set `ROOT` to `//dlang.org`) or convert `404.html` to a PHP file, but this is probably the simplest.

For some reason putting the `ROOT_DIR` / `ROOT` macros in the `404.dd` file itself didn't work.

### DIFF
--- a/errorpage.ddoc
+++ b/errorpage.ddoc
@@ -1,0 +1,2 @@
+ROOT_DIR=/
+ROOT=
--- a/posix.mak
+++ b/posix.mak
@@ -189,6 +189,9 @@ $(DOC_OUTPUT_DIR)/changelog/%.html : changelog/%_pre.dd $(CHANGELOG_PRE_DDOC) $(
 $(DOC_OUTPUT_DIR)/spec/%.html : spec/%.dd $(SPEC_DDOC) $(DMD)
 	$(DMD) -c -o- -Df$@ $(SPEC_DDOC) $<
 
+$(DOC_OUTPUT_DIR)/404.html : 404.dd $(DDOC) $(DMD)
+	$(DMD) -conf= -c -o- -Df$@ $(DDOC) errorpage.ddoc $<
+
 $(DOC_OUTPUT_DIR)/%.html : %.dd $(DDOC) $(DMD)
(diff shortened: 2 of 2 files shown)


CyberShadow:
Note that you won't be able to preview it properly on the autotester, because it doesn't have the resource files in the URL root.

(I guess I could sort of fix that by putting information in the subdomain, but then goodbye HTTPS...)

andralex:
Thanks!!

andralex:
Wondering how long the auto-deploy will take :). Is there a place where I can monitor the process?

wilzbach:
> Wondering how long the auto-deploy will take :).

About five to ten minutes (=the same time DAutoTest takes plus a little for the rsync) 

> Is there a place where I can monitor the process?

AFAIK is  this done in the shadows..

CyberShadow:
> Is there a place where I can monitor the process?

*You* can `ssh dlang@k3.1azy.net tail -F update.log upload.log`. It's not publicly visible right now.

CyberShadow:
> Wondering how long the auto-deploy will take :).

Did you mean to merge this BTW? Because it's not merged yet :)

andralex:
What the...

andralex:
thanks!

MartinNowak:
> (I guess I could sort of fix that by putting information in the subdomain, but then goodbye HTTPS...)

Let's encrypt makes it fairly simple to automatically obtains certs, though they don't have wildcard certs.

MartinNowak:
BTW, chmgen can't yet deal with `<a href="">` URLs.

CyberShadow:
> Let's encrypt makes it fairly simple to automatically obtains certs, though they don't have wildcard certs.

That doesn't scale for that purpose, I'd need to request one cert per build and that would be way over the limit.

> BTW, chmgen can't yet deal with `<a href="">` URLs.

What's the problem?

MartinNowak:
I'll just make a PR, let's discuss it there (https://github.com/dlang/dlang.org/pull/1617).

PR #1605, opened 2017-03-08
Add style rule for Phobos about field declarations

JackStouffer:


JackStouffer:
Ping @wilzbach @andralex

andralex:
noice

PR #1606, opened 2017-03-09
Update safed.dd

insideoutclub:
Fixing Typos.

### DIFF
--- a/safed.dd
+++ b/safed.dd
@@ -13,7 +13,7 @@ $(D_S SafeD&mdash;The Safe Subset of D,
 	)
 
 	$(P
-	The universal reason I've heard from the turncoats was $(DOUBLEQUOTE productivity.) The consensus seems to be that programmers are more productive using Java, C#, Ruby, or Python then they are using C++.
+	The universal reason I've heard from the turncoats was $(DOUBLEQUOTE productivity.) The consensus seems to be that programmers are more productive using Java, C#, Ruby, or Python than they are using C++.
 	)
 
 	$(P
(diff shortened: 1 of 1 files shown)


safed.dd
```diff
@@ -189,14 +189,14 @@ writeln("Hello Safe World!");
...
+	The function $(CODE writeln) is the equivalent of the C $(CODE printf) (more precisely, it's the representative of a family of output functions including $(CODE write) and its formatting versions, $(CODE writef) and $(CODE writefln)). Just like $(CODE printf),  $(CODE writeln) accepts a variable number of arguments of arbitrary types. But here the similarity ends. As long as you pass SafeD-arguments to $(CODE writeln), you are guaranteed not to encounter any undefined behavior. Here, $(CODE writeln) is called with a single argument of the type $(CODE string). In contrast to C, a D $(CODE string) is not a pointer. It is an array of $(CODE immutable char), and arrays are built into the safe subset of D.
```
wilzbach: There is a double white space after `printf`, but it has been there before and is ignored in HTML anyways

safed.dd
```diff
@@ -160,7 +160,7 @@ std:cout &lt;&lt; "Hello World!" &lt;&lt; std::endl;	)
 	)
 
 	$(P
-	Whereas pointers were important in C, C++ embraced them as the main vehicle for the Standard Library. STL algorithms use iterators, objects that are either pointers themselves or imitate the behavior (and the pitfalls) of pointers. Just like with pointers, a programmer's error in using iterators leads to undefined behavior (see the $(CODE swap_range) example).
+	Whereas pointers were important in C, C++ embraced them as the main vehicle for the Standard Library. STL algorithms use iterators, objects that are either pointers themselves or imitate the behavior (and the pitfalls) of pointers. Just like with pointers, a programmer's error in using iterators leads to undefined behavior (see the $(CODE swap_ranges) example).
```
wilzbach: Ehm maybe a dumb question, but to which example does the text refer? There is none one the page..

insideoutclub: The example he's referring to is in paragraph 6 that reads:

"One notorious example is the std::swap_ranges algorithm, which takes three iterators."

wilzbach: Ah right - thanks a lot. As a thought for the future, a link like http://www.cplusplus.com/reference/algorithm/swap_ranges might be nice as well ;-)

PR #1607, opened 2017-03-09

MartinNowak:


CyberShadow:
Um, this seems to have broken building master... In the email you mentioned it's not deterministic, but it was working before.

CyberShadow:
Yep, worked on second try...

PR #1609, opened 2017-03-17
Do not use deprecated syntax in spec.

Bolpat:
Done corrective action as proposed in the [List of Deprecated Features](https://dlang.org/deprecate.html#Using%20the%20result%20of%20a%20comma%20expression).


quickfur:
Are you sure this is what the compiler's lowerings are?  We should not simply replace the syntax if it doesn't reflect what the compiler is actually doing.

Bolpat:
@klickverbot Thanks. I knew it doesn't compile, but thought it did some time longer ago. I find it weird to use non compiling code-looking text. Code on the page that is not obviously something else than D should compile.

PR #1613, opened 2017-03-23
Avoid auto-hyphenation of code in the dpldocs documentation.

s-ludwig:
Fixes rejectedsoftware/ddox#126

### DIFF
--- a/css/ddox.css
+++ b/css/ddox.css
@@ -14,6 +14,10 @@ h1 code a {
 	border-bottom: 1px dashed #CCC;
 }
 
+.hyphenate code {
+	hyphens: manual;
+}
+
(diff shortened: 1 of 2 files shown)


JackStouffer:
Can you please apply this fix to all links in the summary tables? On the main ddox page I still see this

screen shot 2017-03-24 at 12 39 43 am

Applying your CSS rule to the left column made it look way better

s-ludwig:
Overview page links are now also `<code>` and contain manual breaks before the dots.

s-ludwig:
But honestly, the overview page should be replaced by something else altogether. This plain module list is far too long to be useful.

PR #1614, opened 2017-03-24
merge stable into master

MartinNowak:
- to prevent 2.074.0.html -> 2.074.0.dd -> ../tools/changed -o
  2.074.0.dd chaing when 2.074.0_pre.dd is already present


MartinNowak:
Mmh, doesn't integrate to well w/ the changelog preview, but should be fine as master.

PR #1615, opened 2017-03-28
fix a few issues/failures with doc release builds

MartinNowak:


MartinNowak:
Thx, another one #1616.
@CyberShadow. Not sure, but it seems Digger (or the tester) isn't using parallelism for builds. You might want to change that or _this_ PR will cause a slow-down.

CyberShadow:
The bulk of the build time is building ddox pages and the PDF. Both of these are not paralellizable, so I'm not overly concerned.

PR #1617, opened 2017-03-29
fix empty href for 404.html dlang logo

MartinNowak:
- caused chmgen to fail
- no nice idea how to get the real root from arbitrary
  path/to/nonexistent

### DIFF
--- a/errorpage.ddoc
+++ b/errorpage.ddoc
@@ -1,2 +1,3 @@
 ROOT_DIR=/
-ROOT=
+_=ROOT is only used for top left dlang.org logo and cannot be empty `<a href="$(ROOT)">`, so use / even though it breaks when served under /subpath/.
+ROOT=/


MartinNowak:
The d.tag target fails at `Processing 404.html` with https://github.com/dlang/dlang.org/blob/3039c73461a0c1ecd50bd51f3089c36f55b41f67/chmgen.d#L477.

MartinNowak:
We could use JS and try to go `..` step by step, until finding some specific file that's only present on root.

CyberShadow:
Weird, `chmgen` works fine here

MartinNowak:
`rdmd chmgen --root=web --only-tags`

MartinNowak:
How does your 404.html look like?

MartinNowak:
```
Processing 404.html
Error while processing file: web/404.html
object.Exception@chmgen.d(477): Empty URL
----------------
??:? pure @safe void std.exception.bailOut!(Exception).bailOut(immutable(char)[], ulong, const(char[])) [0x54633a]
??:? pure @safe ulong std.exception.enforce!(Exception, ulong).enforce(ulong, lazy const(char)[], immutable(char)[], ulong) [0x5462b5]
??:? immutable(char)[] chmgen.absoluteUrl(immutable(char)[], immutable(char)[]) [0x5451c2]
??:? _Dmain [0x543418]
??:? _D2rt6dmain211_d_run_mainUiPPaPUAAaZiZ6runAllMFZ9__lambda1MFZv [0x590eeb]
??:? void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).tryExec(scope void delegate()) [0x590e13]
??:? void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).runAll() [0x590e90]
??:? void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).tryExec(scope void delegate()) [0x590e13]
??:? _d_run_main [0x590d7f]
??:? main [0x58c49d]
??:? __libc_start_main [0xf6ee7400]
```

MartinNowak:
Anyhow the link is broken atm. so fixing makes sense, on http://dlang.org/path/nonexistent, the logo points to the nonexistent page itself.

CyberShadow:
> so fixing makes sense, on http://dlang.org/path/nonexistent, the logo points to the nonexistent page itself.

Agreed

> How does your 404.html look like?

errorpage.ddoc is not used in win32.mak, so it has the problem fixed by #1603. 

> `rdmd chmgen --root=web --only-tags`

Oh I see, I thought the problem was with building the CHM so I ran it on WIndows.

MartinNowak:
> Oh I see, I thought the problem was with building the CHM so I ran it on WIndows.

Ah sorry, should've been more specific.

CyberShadow:
(was waiting on CI mostly)

MartinNowak:
Thx, hope that was the last issue, release script changes upcoming.

PR #1621, opened 2017-04-08
fix issue 17303 - type error in the href url under the link Systems P…

aG0aep6G:
…rogramming


aG0aep6G:
Closing in favor of #1623 which does the exact same thing.

PR #1623, opened 2017-04-08
 Fix Issue 17303 - type error in the href url under the link Systems Programming

alouanchi:


aG0aep6G:
This fixes the same issue as my #1621. I should probably have expressed more explicitly in Bugzilla that I'm submitting a fix.

Ideally, you'd change the commit message to include the word "fix". The bot picks it up then, which automates a couple of things. I don't know if you can do that with GitHub's web interface. You may have to make a local clone for this. If you have experience with git, it should be relatively straight-forward (pull, amend, force push). If you don't know git, you'd probably have to learn some basics first. If this is too much hassle, I can proceed with #1621 instead. But if you want you're very welcome to go through with this.

On the actual changes: Changing the target ID is a minor breaking change. If anyone has a link/bookmark to `#system-programming`, you're breaking that link. I don't know how important it is to keep this particular URL alive, but generally we try not to break them.

alouanchi:
Thanks for your comments,
I will give a try.

alouanchi:
I have updated my commit to follow @aG0aep6G advice.

@aG0aep6G I agree changing the target ID may break saved links and as our fix is the same is up to you to merge the one you want.

@aG0aep6G thanks a lot for you help.

aG0aep6G:
> @aG0aep6G I agree changing the target ID may break saved links and as our fix is the same is up to you to merge the one you want.

I don't have merge rights. We'll have to wait for a team member to take a look. I've closed my PR, so it's all yours now.

> @aG0aep6G thanks a lot for you help.

You're welcome.

CyberShadow:
I'm fine with the ID breakage in this case, as this is a relatively new page and is not part of the language spec.

aG0aep6G:
> I'm fine with the ID breakage in this case, as this is a relatively new page and is not part of the language spec.

Note that @alouanchi has removed the ID change. No breakage here.

CyberShadow:
Thanks for the clarification.

PR #1624, opened 2017-04-08
Remove gratuitous "empty" clause.

rainers:
FuncAttrs is already optional, so FuncAttr does not need to contain an empty clause.


ibuclaw:
@CyberShadow - is the automerge label the defacto way to mark PRs now?

CyberShadow:
Yes.

PR #1626, opened 2017-04-08
Allow function parameter types in QualifiedName

rainers:
Nested symbols encode the function arguments of the outer function, similar to template arguments.

### DIFF
--- a/spec/abi.dd
+++ b/spec/abi.dd
@@ -211,6 +211,7 @@ $(H3 $(LNAME2 name_mangling, Name Mangling))
     $(GNAME QualifiedName):
         $(GLINK SymbolName)
         $(GLINK SymbolName) $(I QualifiedName)
+        $(GLINK SymbolName) $(GLINK TypeFunctionNoReturn) $(I QualifiedName)
 
     $(GNAME SymbolName):
         $(GLINK LName)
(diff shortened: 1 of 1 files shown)


ibuclaw:
When testing this against your bison files, it adds 6 shift/reduce conflicts for `dwebsite3.bison`, and 5 shift/reduce conflicts for `d_noconflicts.bison`. ;-)

rainers:
> When testing this against your bison files, it adds 6 shift/reduce conflicts for dwebsite3.bison, and 5 shift/reduce conflicts for d_noconflicts.bison. ;-)

Yes, I had to add a token that tells the parser that there is no function return type after TypeFunctionNoReturn (see token QFT in https://gist.githubusercontent.com/rainers/6cdf73b48837defb9f88/raw/d6dc446362f4e41dedfa8e24f11b5a43f85285b4/dwebsite4.bison). That's a deficiency of the LALR parser, but not an ambiguity in the grammar. I think an adhoc parser will be able to handle this without too much trouble.

I also wondered why bison doesn't detect https://issues.dlang.org/show_bug.cgi?id=14591, maybe it's a similar case?

ibuclaw:
> I also wondered why bison doesn't detect https://issues.dlang.org/show_bug.cgi?id=14591, maybe it's a similar case?

I could be swayed into believing that the problem is with my hand rolled parser.  I'll have another look at debugging it.  Perhaps the problem would go away if I refactor the symbol name handling into two functions - I think it's currently the case that `QualifiedName` and `SymbolName` are handled by the same function. (see dland_parse_symbol for comments about possible ambiguity between pascal and template value parameter https://github.com/ibuclaw/gdb/blob/dlang/libiberty/d-demangle.c).

ibuclaw:
I may have to check the change history of the mangle abi too, as I based that program on the current documentation of the time. I can already see some new parts that need adding. ☺️

ibuclaw:
By the way, I noticed this in your comments:
```
FuncAttrScope:
    'M'                // 'M' indistinguishable from first parameter having scope modifier, use Nm instead?
  ;
```
Looks like no one updated the documentation after this change:
https://github.com/dlang/dmd/commit/7b1b10f384253e32fca64fa9ce6972906e63186c

rainers:
Yeah, my plea wasn't heard (https://github.com/dlang/dmd/pull/6496#issuecomment-277186471). Here's a PR to change that: https://github.com/dlang/dlang.org/pull/1630

rainers:
Oops, just noticed that "Improve this page" created this PR as a banch of the dlang.org repository (I didn't even know I could merge PRs in this repo). I'll close and delete this one and reopen another from my fork.

PR #1627, opened 2017-04-09
clarify that MixinTemplateName can refer to a TemplateMixinDeclaration

aG0aep6G:
Before, the spec didn't actually allow `mixin template foo() {} mixin foo!();`.


aG0aep6G:
This has been sitting here for three months. Can someone have a look?

ZombineDev:
CC @andralex

andralex:
thx!

PR #1631, opened 2017-04-11
fix Issue 17293 - 'Using C++ Classes From D' example no longer works

WalterBright:


wilzbach:
BTW I just had the idea to show linker commands, s.t. an unexperienced reader has an easier time:

https://github.com/dlang/dlang.org/pull/1679

> but what changed here? Are extern (C++) interface-s no longer supported?

Hmm at least with recent compilers it seemed to have never worked.
Maybe at some point the interface mangling was changed as its an abstract class on the C++ side as well.

qchikara:
2.071 has the important change of ABI, as the Issue 15644.
It must be strongly highlighted.

PR #1632, opened 2017-04-13
abi: Template symbol parameters are either Qualified or Mangled.

ibuclaw:
An example of a QualifiedName found in a TemplateArg can be seen when dump the symbols of this compiled program: https://github.com/dlang/phobos/blob/31e1b462d2311f9d8cecb33172b28eab587b255d/std/parallelism.d#L43-L77

Also symbol parameters are prefixed with their symbol length, as per bugzilla 3043.

Implementation and Notes: https://github.com/dlang/dmd/blob/15b3ee1603b7dcc0317ba63c54f77ee9718f90e4/src/ddmd/dtemplate.d#L8323-L8331

### DIFF
--- a/spec/abi.dd
+++ b/spec/abi.dd
@@ -244,8 +244,8 @@ $(H3 $(LNAME2 name_mangling, Name Mangling))
         $(GNAME TemplateArgX):
             $(B T) $(GLINK Type)
             $(B V) $(GLINK Type) $(GLINK Value)
-            $(B S) $(GLINK SymbolName)
-            $(B S) $(GLINK MangledName)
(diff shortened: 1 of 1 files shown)


ibuclaw:
FYI @rainers

rainers:
Thanks.

If we want to document non-D symbols as template parameters, we should extend it by
```
    $(B S) $(GLINK Number) $(GLINK ExternallyMangledName)
```
or even use this instead of the MangledName rule. The symbol compression in https://github.com/dlang/dmd/pull/5855 prefers to reencode the name and type, so this would rather remove the MangledName rule instead.

PR #1634, opened 2017-04-13
add link to ndslice in arrays.dd

9il:
arrays.dd contains rectangular-arrays section.
Motivation: http://forum.dlang.org/post/fsaohvwffyxgtlbholhv@forum.dlang.org

### DIFF
--- a/spec/arrays.dd
+++ b/spec/arrays.dd
@@ -379,6 +379,8 @@ double[3][3] matrix;
 double matrix[3,3];
 ---------
 
+        $(P Dynamic rectangular multidimensional arrays (ndslices) are implemented as generic library solution in $(LINK2 https://github.com/libmir/mir-algorithm, mir-algorithm).)
+
 $(H2 $(LNAME2 array-length, Array Length))
 
 


JackStouffer:
I don't know about this. Should we really be promoting a third party library in the spec itself?

CyberShadow:
> I don't know about this. Should we really be promoting a third party library in the spec itself?

No, we shouldn't.

9il:
The paragraph should cover somehow that dynamics tensors are exists. Can this be replaced with: See also https://wiki.dlang.org/Dense_multidimensional_arrays ? Please reopen at least for others to see and discuss it.

CyberShadow:
> Please reopen at least for others to see and discuss it.

OK. But I still don't think the spec is a place to list third-party libraries and such.

The spec is supposed to be, well, the specification: It's meant as a technical reference for implementers and speaks about the language itself. It makes very few references even to the standard library, much less third-party libraries.

We used to have articles hosted on dlang.org. I think the best place to improve ndslice visibility is an article on wiki.dlang.org, which can be linked from the site navigation (Documentation -> Articles).

wilzbach:
Hmm ndslice has been a part of Phobos and as such isn't a usual third-party library as people might still be using ndslice from `std.experimental`, also there are precedents of linking to other projects from the spec. Just by quickly browsing I found this:

- Link to Deimos: https://dlang.org/spec/interfaceToC.html
- std.experimental.allocator: https://dlang.org/spec/cpp_interface.html

So imho we should at least link to the wiki and even linking to Mir would be helpful as it's the de-facto library  for multi-dimensional data in which most readers might be interesting. However, as the later is controversial, I am just voting for the wiki link ;-)

9il:
Updated with the wiki link

wilzbach:
@CyberShadow @JackStouffer are you okay with linking to the Wiki? (this happens fairly often in the specs).

CyberShadow:
> Link to Deimos: https://dlang.org/spec/interfaceToC.html

Deimos is not third-party, it is managed by the same people managing github.com/dlang.

> std.experimental.allocator: https://dlang.org/spec/cpp_interface.html

That's in the standard library.

> and even linking to Mir would be helpful

Here is why I am concerned:

1. Once we have one link to a third-party library in the spec, this sets a precedent and everyone with their own pet library will have a justification of adding a link to it in the spec.

2. Because the two are different projects managed by different people, there is no guarantee that Mir will be around as long as the D programming language is. I am reminded of http://www.digitalmars.com/d/dlinks.html, which may have made sense 10-15 years ago when the community was small, but has been a link graveyard for a very long time now.

3. As I said before, the spec is a technical document describing the language itself. Recommendations for third-party libraries are out of place there.

Therefore I'm still uneasy about this change as it is. Please move the link to http://dlang.org/articles.html and leave http://dlang.org/spec/arrays.html unchanged.

CyberShadow:
Actually it looks mostly fine now. Only nit is that the wiki article should be presenting built-in solutions first and third-party libraries afterwards. Link text could be a bit better. Could use a second opinion, @JackStouffer ?

9il:
Native D has not built-in multidimensional arrays. Only arrays of arrays.

wilzbach:
> Actually it looks mostly fine now. Only nit is that the wiki article should be presenting built-in solutions first and third-party libraries afterwards. Link text could be a bit better. Could use a second opinion, @JackStouffer ?

Ping @JackStouffer or maybe @jmdavis ?

dkorpel (approved):

dkorpel:
> Only nit is that the wiki article should be presenting built-in solutions first and third-party libraries afterwards.

Looks like that's the case now

PR #1635, opened 2017-04-13
Allow function parameter types in QualifiedName

rainers:
Nested symbols encode the function arguments of the outer function, similar to template arguments.

Continuation of https://github.com/dlang/dlang.org/pull/1626

### DIFF
--- a/spec/abi.dd
+++ b/spec/abi.dd
@@ -211,6 +211,7 @@ $(H3 $(LNAME2 name_mangling, Name Mangling))
     $(GNAME QualifiedName):
         $(GLINK SymbolName)
         $(GLINK SymbolName) $(I QualifiedName)
+        $(GLINK SymbolName) $(GLINK TypeFunctionNoReturn) $(I QualifiedName)
 
     $(GNAME SymbolName):
         $(GLINK LName)
@@ -454,7 +455,10 @@ $(H3 $(LNAME2 type_mangling, Type Mangling))
         $(B Nh) $(GLINK Type)
 
     $(GNAME TypeFunction):
-        $(GLINK CallConvention) $(GLINK FuncAttrs)$(OPT) $(GLINK Parameters)$(OPT) $(GLINK ParamClose) $(GLINK Type)
+        $(GLINK TypeFunctionNoReturn) $(GLINK Type)
+
+    $(GNAME TypeFunctionNoReturn):
+        $(GLINK CallConvention) $(GLINK FuncAttrs)$(OPT) $(GLINK Parameters)$(OPT) $(GLINK ParamClose)
 
     $(GNAME CallConvention):
         $(B F)       $(GREEN // D)
--- a/spec/abi.dd
+++ b/spec/abi.dd
@@ -212,7 +212,7 @@ $(H3 $(LNAME2 name_mangling, Name Mangling))
         $(GLINK SymbolName)
         $(GLINK SymbolName) $(I QualifiedName)
         $(GLINK SymbolName) $(GLINK TypeFunctionNoReturn) $(I QualifiedName)
-        $(GLINK SymbolName) $(B M) $(GLINK TypeFunctionNoReturn) $(I QualifiedName)
+        $(GLINK SymbolName) $(B M) $(GLINK TypeModifiers)$(OPT) $(GLINK TypeFunctionNoReturn) $(I QualifiedName)
 
     $(GNAME SymbolName):
         $(GLINK LName)
--- a/spec/abi.dd
+++ b/spec/abi.dd
@@ -212,6 +212,7 @@ $(H3 $(LNAME2 name_mangling, Name Mangling))
         $(GLINK SymbolName)
         $(GLINK SymbolName) $(I QualifiedName)
         $(GLINK SymbolName) $(GLINK TypeFunctionNoReturn) $(I QualifiedName)
+        $(GLINK SymbolName) $(B M) $(GLINK TypeFunctionNoReturn) $(I QualifiedName)
 
     $(GNAME SymbolName):
         $(GLINK LName)


ibuclaw:
OK, I've found one discrepancy, in relation to this.  Parent function decls that need a 'this' pointer *always* have `M` put out.

e.g:
```
module demangle;

void test()
{
    void foo()
    {
        void bar() { }
        pragma(msg, bar.mangleof);
    }
    pragma(msg, foo.mangleof);
}
```

spec/abi.dd
```diff
@@ -211,6 +211,7 @@ $(H3 $(LNAME2 name_mangling, Name Mangling))
     $(GNAME QualifiedName):
         $(GLINK SymbolName)
         $(GLINK SymbolName) $(I QualifiedName)
+        $(GLINK SymbolName) $(GLINK TypeFunctionNoReturn) $(I QualifiedName)
```
ibuclaw: This also needs:
```
$(GLINK SymbolName) M $(GLINK TypeFunctionNoReturn) $(I QualifiedName)
```
I think that is the right place to put it.

rainers: Yes, I've added it.

rainers:
I've also updated the grammer to not need the special token QFT: https://gist.githubusercontent.com/rainers/6cdf73b48837defb9f88/raw/b91ff6b3ae3ce0c7039f1a95c9f6a4e24b7d9849/dwebsite4.bison

Instead I disallowed function types in a number of places where the compiler rejects them, too (e.g. as function return type, as arrays, etc). This now reveals the conflict for 'V' and another one on 'M' (context pointer in qualified name or 'scope' for parameter).

ibuclaw:
> Instead I disallowed function types in a number of places where the compiler rejects them, too (e.g. as function return type, as arrays, etc). 

It's starting to look more and more like my demangler in binutils. ;-)

ibuclaw:
> another one on 'M' (context pointer in qualified name or 'scope' for parameter).

I'm doubtful that could happen. Maybe whilst decoding a parameter list?  Not sure, as the context pointer M relies on the next mangle being a continuation of QualifiedName, whereas parameter M is just a prefix.

rainers:
> I'm doubtful that could happen.

```
void global() {}
struct outer { static struct Strukt {} }

void fun()
{
	auto nested() { struct T {} return T(); }
	static void fn(typeof(nested()) f) {}
	static void fg(outer.Strukt s, scope typeof(&global) f) {}
	pragma(msg, fn.mangleof);
	pragma(msg, fg.mangleof);
}
```
yields
```
_D4test3funFZ2fnFNaNbNiNfS4test3funFZ6nestedMFZ1TZv
_D4test3funFZ2fgFNaNbNiNfS4test5outer6StruktMPFZvZv
```
The 'M' after 'nested' has a differrent meaning than the one after 'Strukt'. I can work around this by look ahead to the next character(s): if it's a CallConvention, it cannot be a Parameter and vice versa.

ibuclaw:
OK, let's have a look (by the way, do you mind if I throw some of these examples in the iberty testsuite?)

rainers:
> OK, let's have a look (by the way, do you mind if I throw some of these examples in the iberty testsuite?)

No problem. BTW: core.demangle is unable to handle these...

ibuclaw:
> No problem. BTW: core.demangle is unable to handle these...

There's not much that core.demangle *can* handle I'm afraid...

https://issues.dlang.org/show_bug.cgi?id=14576

As far as I know, the one I wrote for binutils can correctly handle 99.99% of all symbols thrown at it - or at least there were no failures when I dump all Phobos symbols at it.  There are far too many hundreds of thousands of them to check that what it demangles is actually representative of the actual function, but that is what the testsuite is supposed to catch on the small scale... :-)

ibuclaw:
I don't see the problem on my side.
```
$ ./cxxfilt -s dlang

_D8demangle3funFZ2fnFS8demangle3funFZ6nestedMFZ1TZv
demangle.fun().fn(demangle.fun().nested().T)

_D8demangle3funFZ2fgFS8demangle5outer6StruktMPFZvZv
demangle.fun().fg(demangle.outer.Strukt, scope void() function)
```

rainers:
I guess we could have avoided a lot of to and fro if I had read your implementation in libiberty carefully. Would it make sense to tanslate it to D and replace core.demangle? At the additional benefit of not needing exceptions.

ibuclaw:
> Would it make sense to tanslate it to D and replace core.demangle? At the additional benefit of not needing exceptions.

Yes, I would allow this.  There's probably quite a number of similarities between the two anyway.

rainers:
> Yes, I would allow this. There's probably quite a number of similarities between the two anyway.

Is this (https://github.com/ibuclaw/gdb/blob/dlang/libiberty/d-demangle.c) the latest version? It seems it doesn't output any variable types or function return types... Seems deliberate as there is a comment "Consume the decl type of symbol" in dlang_parse_mangle.

ibuclaw:
FYI, I've sent new patches out, this is the current implementation I'm proposing to be merged in:

https://github.com/ibuclaw/gdb/blob/c218fb20c9cab850322a110ebf6f0f82b3a28344/libiberty/d-demangle.c

Some implementation details:
- Mini-string library you can just replace with D strings.
- You may want to use `checkedint` when decoding integer values.
- This doesn't pretend to support UTF-8, that's OK for function names with UTF8, as the length is correct.  But maybe you want pretty UTF-8 strings in demangled template parameters (`fn!("пример")`).
- You may want to use `strtold` for decoding real values.  The implementation just forwards the hexadecimal float with no attempt to make it pretty.

I think that's all quirks to look out for.

rainers:
Thanks, that's the version I tried against the druntime unittests (after running it through Visual D's C++ to D converter + a few manual edits). It throws away the type of a symbol, though: https://github.com/ibuclaw/gdb/blob/c218fb20c9cab850322a110ebf6f0f82b3a28344/libiberty/d-demangle.c#L1390

ibuclaw:
Ah, you beat me to it. :-)

> Is this (https://github.com/ibuclaw/gdb/blob/dlang/libiberty/d-demangle.c) the latest version? It seems it doesn't output any variable types or function return types... Seems deliberate as there is a comment "Consume the decl type of symbol" in dlang_parse_mangle.

That is correct, I intend to support this in future, along with `--no-params` which skips parameters.

ibuclaw:
> That is correct, I intend to support this in future, along with --no-params which skips parameters.

Turned out not to be *so* difficult on the surface (https://github.com/ibuclaw/gdb/commit/8210b2d64ec88946fce3753c28fb68ff981b6005), but I don't really have the time to go deeper and update the testsuite. ;-)

rainers:
> That is correct, I intend to support this in future, along with --no-params which skips parameters.

Good. I used to have a patch for Visual D to omit the (return) type in the function display for the profile browser.

ibuclaw:
> Thanks, that's the version I tried against the druntime unittests (after running it through Visual D's C++ to D converter + a few manual edits).

Feel free to incorporate (and improve) my own testsuite into druntime, some of which are copied from druntime itself.  https://github.com/ibuclaw/gdb/blob/dlang/libiberty/testsuite/d-demangle-expected

Sorry about the format, but should be a simple mechanical change to convert into a list of asserts. ;-)

rainers:
I noticed some of the symbols in the druntime unittest are not demangled. Should I file bug reports?

BTW: d-demangle.c is currently LGPL licensed. Would changing/dual-licensing this to Boost be an option?

ibuclaw:
> I noticed some of the symbols in the druntime unittest are not demangled. Should I file bug reports?

I had a look, one was caused by this change (https://github.com/ibuclaw/gdb/commit/f650fd95f2d5be135ed014ce7dade4feabee3282) (spec change: https://github.com/dlang/dlang.org/pull/1511/files#diff-7bd92f948c3c1d8d0d16a465bb464b99L243)

In this case, it looks like the unittest case is wrong.
```
UnitTest: _D8demangle29__T2fnVa97Va9Va0Vu257Vw65537Z2fnFZv
Compiler: _D8demangle34__T2fnVai97Vai9Vai0Vui257Vwi65537Z2fnFZv  <-- this is what we accept
```
It looks like the compiler emit in the old style up until 2.065, where it was dropped in the version that followed.  I'll revert it then.  :-(

ibuclaw:
> BTW: d-demangle.c is currently LGPL licensed. Would changing/dual-licensing this to Boost be an option?

Yes, you can re-license as boost.

ibuclaw:
I am about to make the commit to gcc the relevant demangler changes.

I ran it past a quick test demangling all symbols in the phobos unittester - this is the result:
```
(dlang) $ ./ddemangle-old < ~/src/gcc/build-test/phobos_symvers.txt | grep "^_D" | wc -l
140
(dlang) $ ./ddemangle-new < ~/src/gcc/build-test/phobos_symvers.txt | grep "^_D" | wc -l
523
```

Hmm, this doesn't seem promising, looks like there are nearly 400 regressions, lets have a look at a few of them.
```
(dlang) $ ./ddemangle-new < ~/src/gcc/build-test/phobos_symvers.txt | grep "^_D" | head > failsyms.txt
(dlang) $ cat failsyms.txt 
_D2rt8typeinfo10ti_cdouble10TypeInfo_r11initializerMxFNaNbNeZ1ryr
_D2rt8typeinfo11ti_delegate10TypeInfo_D11initializerMxFNaNbNeZ1dyDFiZv
_D2rt8typeinfo7ti_char10TypeInfo_a11initializerMxFNaNbNeZ1cya
_D2rt8typeinfo7ti_real10TypeInfo_e11initializerMxFNaNbNeZ1rye
_D2rt8typeinfo8ti_creal10TypeInfo_c11initializerMxFNaNbNeZ1ryc
_D2rt8typeinfo8ti_dchar10TypeInfo_w11initializerMxFNaNbNeZ1cyw
_D2rt8typeinfo8ti_float10TypeInfo_f11initializerMxFNaNbNeZ1ryf
_D2rt8typeinfo8ti_wchar10TypeInfo_u11initializerMxFNaNbNeZ1cyu
_D2rt8typeinfo9ti_cfloat10TypeInfo_q11initializerMxFNaNbNeZ1ryq
_D2rt8typeinfo9ti_double10TypeInfo_d11initializerMxFNaNbNeZ1ryd
```
```
(dlang) $ ./ddemangle-old < failsyms.txt 
rt.typeinfo.ti_cdouble.TypeInfo_r.initializer() const.r
rt.typeinfo.ti_delegate.TypeInfo_D.initializer() const.d
rt.typeinfo.ti_char.TypeInfo_a.initializer() const.c
rt.typeinfo.ti_real.TypeInfo_e.initializer() const.r
rt.typeinfo.ti_creal.TypeInfo_c.initializer() const.r
rt.typeinfo.ti_dchar.TypeInfo_w.initializer() const.c
rt.typeinfo.ti_float.TypeInfo_f.initializer() const.r
rt.typeinfo.ti_wchar.TypeInfo_u.initializer() const.c
rt.typeinfo.ti_cfloat.TypeInfo_q.initializer() const.r
rt.typeinfo.ti_double.TypeInfo_d.initializer() const.r
```
All others have pretty much the same thing, either demangling a `const.foo` or `shared.foo` symbol.  So looks like there's no regression here, the behaviour is now "reject" instead of "obviously wrong".

Here's the reduced test:
```
_D4testMxFZ1rye  =>  test() const.r
_D4testMyFZ1rye  =>  test() immutable.r
_D4testMOFZ1rye  =>  test() shared.r
```

spec/abi.dd
```diff
@@ -211,6 +211,8 @@ $(H3 $(LNAME2 name_mangling, Name Mangling))
...
+        $(GLINK SymbolName) $(B M) $(GLINK TypeFunctionNoReturn) $(I QualifiedName)
```
ibuclaw: > Here's the reduced test:
> _D4testMxFZ1rye  =>  test() const.r

The fix would be:
```
$(GLINK SymbolName) $(B M) $(GLINK TypeModifiers)$(OPT) $(GLINK TypeFunctionNoReturn) $(I QualifiedName)
```
:disappointed:

ibuclaw: Only need it for this rule.  As you can't have TypeModifiers without a 'this'.

rainers:
> The fix would be:
> 
> $(GLINK SymbolName) $(B M) $(GLINK TypeModifiers)$(OPT) $(GLINK TypeFunctionNoReturn)  $(I QualifiedName)

You're right. Updated.

ibuclaw:
And I applied the fix in libiberty, and I now get:
```
(dlang) $ ./ddemangle-new < ~/src/gcc/build-test/phobos_symvers.txt | grep "^_D" | wc -l
138
```
https://gist.github.com/ibuclaw/d15fb1b30027dc1a314f9cca23eda442

Sweet, should probably add one of those newly supported symbols to the testsuite, just to make such that we don't regress.

ibuclaw:
Committed: https://gcc.gnu.org/viewcvs/gcc?view=revision&revision=247450

PR #1636, opened 2017-04-17
Remove special casing for `objc.d` from the Posix makefile

jacob-carlborg:
This file is now shared for all platforms after the Objective-C integration refactoring: https://github.com/dlang/dmd/pull/6679.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -518,11 +518,11 @@ docs.json : ${DMD_REL} ${DRUNTIME_STABLE_DIR} \
 
 # DDox tries to generate the docs for all `.d` files. However for dmd this is tricky,
 # because the `{mach, elf, mscoff}` are platform dependent.
-# Thus the need to exclude these files (and the `objc.d` files).
+# Thus the need to exclude these files (and the `objc_glue.d` file).
 docs-prerelease.json : ${DMD} ${DMD_DIR} ${DRUNTIME_DIR} \
 		${PHOBOS_FILES_GENERATED} | dpl-docs
 	find ${DMD_DIR}/src -name '*.d' | \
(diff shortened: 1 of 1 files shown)


jacob-carlborg:
@ibuclaw @WalterBright This is required for the DAutoTest to pass for https://github.com/dlang/dmd/pull/6679.

I guess this will fail until https://github.com/dlang/dmd/pull/6679 is merged.

ibuclaw:
Looks fine. FYI @CyberShadow.

jacob-carlborg:
> I guess the same will need to be done for objc_glue too if you plan on changing that too?

Yes, one step at the time 😃. I'll wait until I have the PR for the compiler ready.

CyberShadow:
LGTM, but I'm not going to merge this because it looks like by itself it would break master. I understand that merging this simultaneously with https://github.com/dlang/dmd/pull/6679 would fix things again, so whoever merges https://github.com/dlang/dmd/pull/6679 should merge this immediately after.

Even better would be to do this in three steps to avoid breaking master at all. E.g. in the dlang.org makefile, detect whether `objc.d` should be ignored (i.e. whether https://github.com/dlang/dmd/pull/6679 is merged). That would completely avoid the risk of breaking master should merging both PRs somehow still result in tests failing.

jacob-carlborg:
@CyberShadow auto merging has been enabled for the other PR https://github.com/dlang/dmd/pull/6679, can this be merged now? Looks like Iain doesn't have permission to merge this.

jacob-carlborg:
@CyberShadow https://github.com/dlang/dmd/pull/6679 has been merged. This can now be merged as well.

jacob-carlborg:
Thanks.

CyberShadow:
Thanks for handling this @schuetzm, just got back home :)

PR #1423, opened 2016-07-10
Issue 12558 - Document implicit catch statement deprecation

Geod24:
See dlang/dmd#5183

PR #1637, opened 2017-04-18
Implicit catch statements are deprecated.

veelo:
Bring spec/statement.dd in line with spec/grammar.dd. See https://github.com/dlang/dlang.org/pull/1423 and https://github.com/dlang/dmd/pull/5183.

### DIFF
--- a/spec/statement.dd
+++ b/spec/statement.dd
@@ -1485,12 +1485,8 @@ $(GNAME TryStatement):
     $(D try) $(PSSCOPE) $(GLINK FinallyStatement)
 
 $(GNAME Catches):
-    $(GLINK LastCatch)
     $(GLINK Catch)
     $(GLINK Catch) $(I Catches)
-
-$(GNAME LastCatch):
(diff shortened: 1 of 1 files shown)


wilzbach:
Ping @ other reviewers. What's our policy on spec updates during the deprecation period?
There's also another PR with a similar question the queue: https://github.com/dlang/dlang.org/pull/1639

veelo:
Whether the spec should document deprecated features or not is not really the main concern for this PR. [`spec/grammar.d` is already updated](https://github.com/dlang/dlang.org/pull/1423/files#diff-8dfb9f9aa450e9831e4ef5c45a97af8b), so now there is a discrepancy been `spec/grammar.dd` and `spec/statement.dd`. The correct thing to do is either revert the change to `spec/grammar.'d or pull this request. I prefer the latter :-)

quickfur:
I think we should go ahead with this.  But just in case, let's ping @andralex just to be sure.

PR #1639, opened 2017-04-26
remove deprecated StringLiterals (implicit concat) from expressions

BBasile:
As observed [here](https://forum.dlang.org/post/odoqjn$17eh$1@digitalmars.com) the _StringLiterals_ rule is now deprecated.

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -1125,7 +1125,7 @@ $(GNAME PrimaryExpression):
     $(GLINK2 lex, IntegerLiteral)
     $(GLINK2 lex, FloatLiteral)
     $(GLINK CharacterLiteral)
-    $(GLINK StringLiterals)
+    $(GLINK StringLiteral)
     $(GLINK ArrayLiteral)
     $(GLINK AssocArrayLiteral)
     $(GLINK FunctionLiteral)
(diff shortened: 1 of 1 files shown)


rainers:
I'm not sure we should erase deprecated stuff from the language description as it can be confusing to understand why code is accepted by the compiler. There is no common pattern about this in the specification, though, e.g. class allocators are still there (https://github.com/dlang/dlang.org/blob/master/spec/class.dd#L855), but typedef has only some indirect traces (https://github.com/dlang/dlang.org/blob/master/spec/abi.dd#L391). 

IMO adding a comment to the grammar rule might be better.

BBasile:
I don't think there's a precedent case where a deprecation is indicated directly in the grammar so I keep this PR opened for the day the deprecation will end...if ever.

wilzbach:
There's another, similar PR in the queue that also updates the spec of an deprecated, but not removed feature: 

https://github.com/dlang/dlang.org/pull/1637

spec/expression.dd
```diff
@@ -1233,9 +1233,8 @@ $(H4 $(LEGACY_LNAME2 CharacterLiteral, character-literal, Character Literals))
 $(H4 String Literals)
 
 $(GRAMMAR
-$(GNAME StringLiterals):
+$(GNAME StringLiteral):
     $(GLINK2 lex, StringLiteral)
-    $(I StringLiterals) $(GLINK2 lex, StringLiteral)
```
CyberShadow: This isn't right, you are defining `StringLiteral` as being defined as itself. Instead it should link to the `StringLiteral` definition in `lex.dd` in the first hunk.

PR #1647, opened 2017-05-07
Test tour.dlang.org as alternative backend

wilzbach:
Proof of concept PR (it's ugly!) to switch over to the DLang Tour as code executor for the executable blocks on dlang.org

However, unfortunately DLang Tour doesn't support `STDIN` nor `ARGS`, so it would only work for the runnable code block in the documentation.

### DIFF
--- a/js/run.js
+++ b/js/run.js
@@ -66,53 +66,96 @@ function safeVar(data, path)
     return res;
 }
 
-function parseOutput(data, o, oTitle)
+var backends = {
+  dpaste: {
+    url: "https://dpaste.dzfl.pl/request/",
+    contentType: "application/x-www-form-urlencoded; charset=UTF-8",
+    requestTransform: function(data) {
+      return data;
+    },
+    parseOutput: function(data, opts) {
+      var r = {};
+      if (data.compilation === "undefined")
+        return null;
+      r.cout = safeVar(data, "compilation.stdout");
+      r.stdout = safeVar(data, "runtime.stdout");
+      r.stderr = safeVar(data, "runtime.stderr");
+      r.ctime = parseInt(safeVar(data, "compilation.time"));
+      r.rtime = parseInt(safeVar(data, "runtime.time"));
+      r.cstatus = parseInt(safeVar(data, "compilation.status"));
+      r.rstatus = parseInt(safeVar(data, "runtime.status"));
+      r.cerr = safeVar(data, "compilation.err");
+      r.rerr = safeVar(data, "runtime.err");
+      r.defaultOutput = data.output || opts.defaultOutput;
+      return r;
+    }
+  },
+  tour: {
+    url: "https://tour.dlang.org/api/v1/run",
+    contentType: "application/json; charset=utf-8",
+    requestTransform: function(data) {
+        return JSON.stringify({
+            source: data.code
+        });
+    },
+    parseOutput: function(data, opts) {
+      var r = {};
+      if (data.success === "undefined") {
+        return null;
+      }
+      r.cout = data.success === false ? data.output : "";
+      r.stdout = data.success === true ? data.output : "";
+      r.stderr = "";
+      r.ctime = "";
+      r.rtime = "";
+      r.cstatus = data.errors.length === 0 ? 0 : 1;
+      r.rstatus = data.success === true ? 0 : 1;
+      r.cerr = "";
+      r.rerr = "";
+      r.defaultOutput = data.output || opts.defaultOutput;
+      return r;
+    }
+  }
+};
+
+function parseOutput(res, o, oTitle)
 {
-    if (typeof data.compilation == "undefined")
+    if (!res)
     {
         o.text("Temporarily unavailable");
         return;
     }
 
     var output = "";
-    var cout = safeVar(data, "compilation.stdout");
-    var stdout = safeVar(data, "runtime.stdout");
-    var stderr = safeVar(data, "runtime.stderr");
-    var ctime = parseInt(safeVar(data, "compilation.time"));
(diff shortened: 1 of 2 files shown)


JackStouffer:
Why?

wilzbach:
> Why?

'Cause DPaste is down a lot of time (like now) and we have no control over it (i.e. can't fix it), which means all executable blocks on dlang.org don't work...

The DTour already exposes an API to executable runnable examples, so this was a proof-of-concept PR to investigate whether we could switch to the DTour as API backend.

CyberShadow:
> and we have no control over it (i.e. can't fix it),

This is no longer true - DPaste's frontend and backend is hosted on my server.

The only thing that's not under our control is the domian. I suggest we fix that by making in redirect to paste.dlang.io instead of the other way around.

wilzbach:
> This is no longer true - DPaste's frontend and backend is hosted on my server.

Oh it seems that I forgot to post the motivation here. There are many problems with DPaste and I don't think it's easy to fix all of them:

- segfaults (people care/complain about this, e.g. [this post on the NG](https://forum.dlang.org/post/jhcyjnidqzxinlpyzuyn@forum.dlang.org) or [the bugzilla issue](https://issues.dlang.org/show_bug.cgi?id=17560#c2))
- no permission to write files
- no caching
- frequent downtimes
- slow
- no dmd-nightly (to be fair, currently the dlang-tour doesn't support this either, but it's just one pull request away for me)

A good "real life" example can be found on https://github.com/dlang/phobos/pull/5517#issuecomment-311534310 as well.
I don't think _anyone_ cares which backend is used in the background as long as they can run code _smoothly_ on dlang.org.
This PR just changes the backend provider for all Phobos examples.

> The only thing that's not under our control is the domian. I suggest we fix that by making in redirect to paste.dlang.io instead of the other way around.

It's not like I haven't tried to improve DPaste in the past, even the rebranding: https://github.com/nazriel/dpaste-frontend/issues/4

@CyberShadow you still have access to dlang.io via NameCheap right? Feel free to set up a A / CNAME record for `paste.dlang.io` (I guess that will be nicer than a redirect?)

#perf matters
-------------------

DPaste:

image

DTour:

image

Note that it didn't cache the request as this is currently not activated (-> https://github.com/dlang-tour/core/pull/545).

CyberShadow:
> I don't think anyone cares which backend is used in the background as long as they can run code smoothly on dlang.org.

Agreed.

> This PR just changes the backend provider for all Phobos examples.

I'm fine with going forward with this PR if you consider things to be ready here and on the backend's side.


> @CyberShadow you still have access to dlang.io via NameCheap right? Feel free to set up a A / CNAME record for paste.dlang.io (I guess that will be nicer than a redirect?)

If there is no interest in maintaining DPaste, then a better solution would be a migration plan to upgrade the tour backend with DPaste's critical features (saving pastes and supporting its set of compilers), migrating its database of pastes, and replacing / decommissioning it.

CyberShadow:
CC @nazriel

wilzbach:
> I'm fine with going forward with this PR if you consider things to be ready here and on the backend's side.

Yep, I do. As shown, even without the caching PR, the tour backend is still a lot faster :)
Also it has a [nice & clean API](https://github.com/dlang-tour/core/blob/master/source/exec/iexecprovider.d) and allows plugging in new backends (e.g. Linux LXC containers, ...) easily.

> If there is no interest in maintaining DPaste, then a better solution would be a migration plan to upgrade the tour backend with DPaste's critical features (saving pastes and supporting its set of compilers), migrating its database of pastes, and replacing / decommissioning it.

Could you please fill all required tasks at the DTour repo?
https://github.com/dlang-tour/core/issues

We might also want to take inspirations from the playgrounds of other languages, e.g.

https://play.rust-lang.org/
https://try.kotlinlang.org/#/Examples/Hello,%20world!/Simplest%20version/Simplest%20version.kt
https://play.golang.org/

CyberShadow:
I think this breaks the front page examples but I'm not sure how. I can only see that they don't work on dtest.dlang.io, even though the master branch works.

wilzbach:
> I think this breaks the front page examples but I'm not sure how.

DPaste sends formencoded requests, while I set the default to json. However, this means that the request is "preflighted" (see: https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Preflighted_requests), which DPaste doesn't support:

> XMLHttpRequest cannot load https://dpaste.dzfl.pl/request/. Response for preflight is invalid

Another good resource: https://www.soasta.com/blog/options-web-performance-with-single-page-applications.
I will look into accepting `text/plain;charset=UTF-8` payloads as well to avoid the double latency due to the `OPTIONS` preflight request.

wilzbach:
> I will look into accepting text/plain;charset=UTF-8 payloads as well to avoid the double latency due to the OPTIONS preflight request.

Okay, it's in the queue

-> https://github.com/dlang-tour/core/pull/507

> I think this breaks the front page examples 

Should be fixed now - let's better wait and have a look at Dtest though ;-)

CyberShadow:
> Could you please fill all required tasks at the DTour repo?

OK, I filed the ones that came up so far.

> Should be fixed now - let's better wait and have a look at Dtest though ;-)

Seems to work!

CyberShadow:
BTW, I noticed something a little weird - the "Reset" button is visible by default, but it goes away after you click it once.

wilzbach:
> OK, I filed the ones that came up so far.

Thanks. I will work on them over the next weeks - you will be notified ;-)

> BTW, I noticed something a little weird - the "Reset" button is visible by default, but it goes away after you click it once.

-> https://github.com/dlang/dlang.org/pull/1779

PR #1648, opened 2017-05-09
add D-Man images

WalterBright:
Add D-Man images collected from:

https://github.com/wilzbach/d-mans


wilzbach:
Thanks a lot for your interest in this collection @WalterBright!
However, I think these 200 images are quite large and thus add a substantial size to the dlang.org repository, which will lead to longer CI for _all_ `dlang` repositories with DAutoTest enabled.
At least that's the reason why I put them in a separate repository, which of course could be transferred to the `dlang` organization for better visibility ;-)

CC @CyberShadow

CyberShadow:
DAutoTest doesn't reclone repos like Travis, so it shouldn't affect it, but it will affect clone times of anyone wanting to contribute as well as possible future CI services we might want to use.

Laeeth:
why not just have lower-quality/smaller versions of these?

andralex:
@wilzbach @CyberShadow so what's the action item here?

wilzbach:
> @wilzbach @CyberShadow so what's the action item here?

I am in favor of _not_ merging this adds quite a bit of memory that will _always_ need to be downloaded and it doesn't add any value. Imho the images have a better space in a separate repository as [they are currently](https://github.com/wilzbach/d-mans).
If you want to increase the exposure, I am happy to transfer this repo to the `dlang` organization and/or we can try to find other ways to make this easier accessible (wiki entry, blog post, link on dlang.org, ...).

WalterBright:
For reference, it's 35Mb

Is it really that big a deal these days?

andralex:
FWIW the current site source is 14MB including all images, to which .git/ adds 16 MB. The built site we push has 8.3 MB. So it seems the bump in size is significant.

CyberShadow:
> Is it really that big a deal these days?

Yes; it adds seconds to clone time, and as these files will not be useful to practically everyone who intends to clone the repository, it will become an unnecessary source of frustration.

> The built site we push has 8.3 MB.

Excluding DDox, I guess (which is like half a gigabyte).

wilzbach:
>  Imho the images have a better space in a separate repository as [they are currently](https://github.com/wilzbach/d-mans).
If you want to increase the exposure, I am happy to transfer this repo to the `dlang` organization and/or we can try to find other ways to make this easier accessible (wiki entry, blog post, link on dlang.org, ...).

FWIW there's also a crude & simplistic listing page to view all D-mans: https://wilzbach.github.io/d-mans/meiz

> @wilzbach @CyberShadow so what's the action item here?

Could we please come to a consensus here? @WalterBright is there another option that would work for you? We can also move the repo to [`dlang-community`](https://github.com/dlang-community) if `dlang` is too official.

CyberShadow:
@wilzbach I suggest you take the initiative and put them anywhere else so that we can close this PR as solved.

wilzbach:
> Well I see / remember you already have them in their own repo, so...

Yup and I now opened a discussion to move them to dlang-community, s.t. the visibility is increased:

https://github.com/dlang-community/discussions/issues/19

>  so...

I think we all agree here that a separate repository for the d-mans is the best solution, so I am taking "the initiative here" and closing this PR.

andralex:
@wilzbach how about putting the images on dlang.rocks?

wilzbach:
FYI: the repo is now part of dlang-community https://dlang-community.github.io/d-mans/meiz/

May the D-mans rest happily ever after :)

> @wilzbach how about putting the images on dlang.rocks?

I have a better idea (for now): http://dlang.rocks/

(source: https://github.com/wilzbach/dancing-d-man)

CyberShadow:
> I have a better idea (for now): http://dlang.rocks/

Getting a 400 Bad Request on that one right now, FYI.

PR #1590, opened 2017-02-27
Use Makefile's pattern rules for the intermediate assert/writeln translation

wilzbach:
- only regeneration when real changes happen
- removes the ugly `rsync` workaround

PR #1650, opened 2017-05-10
Ensure that the phobos-stable directory exists

wilzbach:
A simple fix to solve the problem of a non-existent stable Phobos directory as introduced in https://github.com/dlang/dlang.org/pull/1590.

The current repository layout assumes that the latest, stable phobos
(e.g. ../phobos-2.074.0) exists. For now, this is automatically ensured
before the Makefile targets are evaluated.

A similar trick is used at the [DMD Makefile](https://github.com/dlang/dmd/blob/master/src/posix.mak#L48) and we would use `git clone` anyways.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -55,6 +55,8 @@ PHOBOS_STABLE_DIR_GENERATED=$(GENERATED)/phobos-release
 #   Makefile dependencies and rules
 PHOBOS_FILES := $(shell find $(PHOBOS_DIR) -name '*.d' -o -name '*.mak' -o -name '*.ddoc')
 PHOBOS_FILES_GENERATED := $(subst $(PHOBOS_DIR), $(PHOBOS_DIR_GENERATED), $(PHOBOS_FILES))
+$(shell [ ! -d $(PHOBOS_DIR) ] && git clone --depth=1 ${GIT_HOME}/phobos $(PHOBOS_DIR))
+$(shell [ ! -d $(PHOBOS_STABLE_DIR) ] && git clone -b v${LATEST} --depth=1 ${GIT_HOME}/phobos $(PHOBOS_STABLE_DIR))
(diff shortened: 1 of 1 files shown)


CyberShadow:
Still getting:
```
find: ‘../phobos’: No such file or directory
```

wilzbach:
> find: ‘../phobos’: No such file or directory

Hmm, of course we could check for this as well and clone it as well as in the most likely case it will be present, but I am starting to dislike this simple solution ...

CyberShadow:
> Hmm, of course we could check for this as well and clone it as well as in the most likely case it will be present, but I am starting to dislike this simple solution ...

I think it's important that we find a working solution, even if it means reverting #1590.

MartinNowak:
Look at https://github.com/dlang/dlang.org/pull/1214/files#diff-228b518f4545d060f900578830ade5da for a rough guide on how to do this correctly.

Make downstream rules depend on some `assert_writeln/.done` file, make that depend on the cloned repo, within the rule us find to only rewrite updated sources.
So when the repo isn't yet cloned it will get cloned. Since `/.done` is a phony target it will always run.
I know no good way to first clone and then still add all wildcard (found) files into make's dependencies in a single invocation. Doing it as an unconditional script (similar to the `shell` call now) will likely cause troubles later on, e.g. to just build phobos-release without cloning.

As always, maybe there is a better solution I'm not aware of, but the above is feasible and fixes the problem at hand.

Overall I hope we finally have a green light for dpl-docs and can dump all of that messy stuff soon.

MartinNowak:
In case you guys want mentions @wilzbach, @CyberShadow

CyberShadow:
@wilzbach ping

CyberShadow:
I'm going to merge this for now as it fixes the problem at least partially, but @wilzbach please don't forget to return to this.

CyberShadow:
@wilzbach https://issues.dlang.org/show_bug.cgi?id=17649 looks related to this, perhaps you should have a look.

PR #1653, opened 2017-05-13
posix.mak: Fix default target

CyberShadow:
The default target of a Makefile is usually the first defined
non-pattern target. The intended default target of dlang.org's
posix.mak is the 'all' target, however this was broken in
44dd86f9a22bffbc057e76a2cf861037f8f34760 by introducing a non-pattern
target among the pattern targets (in order to specialize building
error pages as necessary to fix relative links to resources).

Move the non-pattern rules block (which includes the intended default
target, 'all') above the pattern rules block to fix the intended
default target.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -186,58 +186,6 @@ $(PREMADE) $(STYLES) $(IMAGES) $(JAVASCRIPT))
 ALL_FILES = $(ALL_FILES_BUT_SITEMAP) $(DOC_OUTPUT_DIR)/sitemap.html
 
 ################################################################################
-# Pattern rulez
-################################################################################
-
-# NOTE: Depending on the version of make, order matters here. Therefore, put
-# sub-directories before their parents.
-
-$(DOC_OUTPUT_DIR)/changelog/%.html : changelog/%_pre.dd $(CHANGELOG_PRE_DDOC) $(DMD)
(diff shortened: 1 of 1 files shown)


posix.mak
```diff
@@ -299,6 +247,58 @@ rsync-only :
...
+$(DOC_OUTPUT_DIR)/changelog/%.html : changelog/%_pre.dd $(CHANGELOG_PRE_DDOC) $(DMD)
+	$(DMD) -conf= -c -o- -Df$@ $(CHANGELOG_PRE_DDOC) $<
+
+$(DOC_OUTPUT_DIR)/changelog/%.html : changelog/%.dd $(CHANGELOG_DDOC) $(DMD)
+	$(DMD) -conf= -c -o- -Df$@ $(CHANGELOG_DDOC) $<
+
+$(DOC_OUTPUT_DIR)/spec/%.html : spec/%.dd $(SPEC_DDOC) $(DMD)
+	$(DMD) -c -o- -Df$@ $(SPEC_DDOC) $<
+
+$(DOC_OUTPUT_DIR)/404.html : 404.dd $(DDOC) $(DMD)
```
andralex: I'll note that this is not a pattern rule although it's under the "pattern rules" section.

PR #1654, opened 2017-05-13
dpl-docs: Disable SSL

CyberShadow:
As SSL is not necessary to for anything dpl-docs is currently doing,
disable it. This fixes building dlang.org on systems with OpenSSL
1.1.0 or newer, such as Arch Linux, which introduced breaking API
changes that currently cause linking errors for Vibe.d apps not built
with version=VibeNoSSL.


MartinNowak:
Wouldn't it make sense to remove OpenSSL from dub.selections.json then?
https://github.com/dlang/dlang.org/blob/fbe65938d8ffb8e8d9f48897ed66a8270512e014/dpl-docs/dub.selections.json#L12

CyberShadow:
Done

PR #1655, opened 2017-05-13
posix.mak: Move .{pre,}release-files.txt to .generated/

CyberShadow:
Avoid cluttering the root directory and .gitignore.


wilzbach:
I had a similar idea, but went a step further and moved the json files to generated as well: https://github.com/dlang/dlang.org/pull/1678

PR #1656, opened 2017-05-17
Add OpenGraph meta tags to the front page

CyberShadow:


CyberShadow:
@wilzbach If we're not going to do anything with Travis, can we disable it? It's making everything red.

CyberShadow:
I went ahead and disabled it.

PR #1657, opened 2017-05-22
Fix target 'all' in makefile

andralex:
The introduction of target 404.html makes it the first target, which makes `make` without a target build just that instead of everything.


andralex:
cc @CyberShadow - can we have this tested automatically?

CyberShadow:
Duplicate of https://github.com/dlang/dlang.org/pull/1653 :)

PR #1659, opened 2017-05-23
Issue 17419 - add __traits(getLinkage, s) to the the linkage of symbol s

WalterBright:


spec/traits.dd
```diff
@@ -493,6 +494,33 @@ pragma(msg, __traits(getFunctionAttributes, (int x) @trusted { return x * 2; }))
...
+	$(LI $(D "Objective-C"))
+	$(LI $(D "System"))
+	)
+
+---
+extern (C) int fooc();
+alias aliasc = fooc;
+
+static assert(__traits(getLinkage, fooc) == "C");
+static assert(__traits(getLinkage, aliasc) == "C");
```
JohanEngelen: I'd add tests for a few of the other linkages too (so we are sure it doesnt always return "C" ;). And also for implicit extern(D), for a pointer to function, and perhaps for a lambda and class method.

ZombineDev: That's already in dmd's testsuite, btw: https://github.com/dlang/dmd/pull/6823/files#diff-4fa5742db248994843ba04a77a6d70f1L17

JohanEngelen: Ah, sorry, I didn't see that this is in a spec file. Disregard!

PR #1662, opened 2017-05-25
Update orgs-using-d.dd

Laeeth:
I haven't checked results as no time.  Please check it builds and looks okay.


orgs-using-d.dd
```diff
@@ -62,6 +62,7 @@ $(DIVC orgs-using-d center,
     $(DORG eBay, https://ebay.com, ebay.jpg,
             One of the world’s largest marketplaces,
             Large scale data mining tools.
+            $(HTTP http://dlang.org/blog/2017/05/24/faster-command-line-tools-in-d/, Command line tools in D)
```
CyberShadow: When using `HTTP` macro, omit `http://`

PR #1665, opened 2017-05-28
Set bottom subnav to a single pix

BBasile:
It's needed in the  forum, sometimes the subnav doesn't snap to the page bottom, eg here https://forum.dlang.org/post/yjpgswjigrenryplyvji@forum.dlang.org


BBasile:
ping @CyberShadow you adviced me to propose the change here rather than on DFeed: https://github.com/CyberShadow/DFeed/pull/81

PR #1667, opened 2017-06-05
DIP1003: Remove body as a Keyword

WalterBright:
https://github.com/dlang/DIPs/blob/master/DIPs/DIP1003.md

### DIFF
--- a/spec/contracts.dd
+++ b/spec/contracts.dd
@@ -70,15 +70,15 @@ out (result)
 {
     ...contract postconditions...
 }
-body
+do
 {
-    ...code...
+    ...function body...
 }
 ------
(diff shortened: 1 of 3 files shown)


wilzbach:
Don't we have to update:
- [the list of keywords](http://dlang.org/spec/lex.html)
- [functions spec](http://dlang.org/spec/function.html#BodyStatement)
- [the pragma mangling example](http://dlang.org/spec/pragma.html#mangle) as "body" is now available

as well?

Also for convenience, here's the link to the respective DMD PR: https://github.com/dlang/dmd/pull/6855

wilzbach:
Ah this has already been done in another PR (https://github.com/dlang/dlang.org/pull/1829) in the meantime -> closing.

PR #1668, opened 2017-06-05
Fix 17469 - View source code for object.d is broken

wilzbach:
Due to [17131](https://issues.dlang.org/show_bug.cgi?id=17131) a DUB dependency upgrade was necessary. I discovered that the assert-rewrite pipeline doesn't like empty files like the ones generated from the `publictests` target. A proper solution is to make the assert_writeln pipeline more selective, but the first commit should is a simple solution for now.


wilzbach:
Oh I didn't realize that I already submitted a fix for the "ddmd" ones: https://github.com/dlang/dlang.org/pull/1601

wilzbach:
(rebased to master)

PR #1671, opened 2017-06-05
Add DMD to the release pages of Ddoc and Ddox

wilzbach:
This will add the "DMD as a library" documentation to the release pages of Ddoc and Ddox.
The DMD pages have been part of the prerelease pages [since four months](https://github.com/dlang/dlang.org/pull/1581):

Ddoc: https://dlang.org/phobos-prerelease/ddmd_mars.html
Ddox: https://dlang.org/library-prerelease/ddmd/mars.html

As we got _no_ feedback I feel like the only way people will notice (and thus maybe start to improve it), is exposing the documentation on the release docs.

CC @andralex @RazvanN7 

(changing the documentation generation from `ddmd` to `dmd` is rather trivial and already needs to be done for the prerelease pages anyways.)

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -40,7 +40,7 @@ TMP?=/tmp
 
 # Last released versions
 DMD_STABLE_DIR=${DMD_DIR}-${LATEST}
-DMD_REL=$(DMD_STABLE_DIR)/src/dmd
+DMD_REL=$(DMD_STABLE_DIR)/generated/$(OS)/release/$(MODEL)/dmd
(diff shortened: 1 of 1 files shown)


andralex:
ah cool thx!!

wilzbach:
> ah cool thx!!

Wow that was quick! Thanks!
As many pages still need some work, I opened a post on the NG:
http://forum.dlang.org/post/rkhbkdskleboxyenxqqa@forum.dlang.org

PR #1672, opened 2017-06-06
Deprecation: set .sort and .reverse props are 'Error' starting with 2.075

wilzbach:
CC @UplinkCoder @MartinNowak


CyberShadow:
Kind of a nitpick, but I think the "error" column is for explicit errors given when the compiler explicitly detects the use of the deprecated feature. AIUI these properties were removed outright, so it would be more correct to leave the "error" column blank, like what was done with `clear` below in the table.

UplinkCoder:
well together with the removal comes an error when they are used.

2017-06-07 15:30 GMT+02:00 Vladimir Panteleev <notifications@github.com>:

CyberShadow:
> well together with the removal comes an error when they are used.

If you apply that logic, then that makes the `error` column completely useless, because if a feature is completely removed from the compiler code, of course you're going to get some error or another when trying to use it.

wilzbach:
> well together with the removal comes an error when they are used.

Hehe, that was my logic as well. 

> then that makes the error column completely useless, because if a feature is completely removed from the compiler code, of course you're going to get some error or another when trying to use it.

That's a fair point:
https://github.com/dlang/dlang.org/pull/1680

PR #1673, opened 2017-06-06
Deprecation: Issue 6227 - Comparison of different enums

wilzbach:
From https://github.com/dlang/dmd/pull/6821

CC @UplinkCoder

### DIFF
--- a/deprecate.dd
+++ b/deprecate.dd
@@ -10,6 +10,7 @@ $(SPEC_S Deprecated Features,
 
     $(TABLE2 Deprecated Features,
         $(THEAD Feature,                                                          Spec,  Dep,    Error,  Gone)
+        $(TROW $(DEPLINK Implicit comparison of different enums),                 2.075,  2.075, &nbsp;, &nbsp;)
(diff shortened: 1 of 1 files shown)


UplinkCoder:
looks good.

wilzbach:
The deprecation is already part of DMD, so anything blocking this?

wilzbach:
CC @CyberShadow - can we get this shipped before the 2.075 release which includes this deprecation? :)

deprecate.dd
```diff
@@ -57,6 +58,41 @@ $(SPEC_S Deprecated Features,
         $(DD The feature is completely gone)
     )
 
+$(H3 $(DEPNAME Implicit comparison of different enums))
+    $(P Currently, comparison of different enumerated type is allowed:
```
CyberShadow: "Currently" is not going to be current for much longer

wilzbach:
> "Currently" is not going to be current for much longer

Hehe, I copied this from the entry below -> I changed the tense for both entries.

PR #1674, opened 2017-06-07
Add labels to spec/lex.dd

wilzbach:
Labels are nice and necessary if you want to direct a reader to a specific place of the D spec.

### DIFF
--- a/spec/lex.dd
+++ b/spec/lex.dd
@@ -370,7 +370,7 @@ $(GNAME TokenString):
 	$(P In all string literal forms, an $(GLINK EndOfLine) is regarded as a single
 	$(D \n) character.)
 
-$(H4 Wysiwyg Strings)
+$(H4 $(LNAME2 wysiwyg, Wysiwyg Strings))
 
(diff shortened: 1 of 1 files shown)


wilzbach:
CC @CyberShadow - anything blocking this? It just adds labels to the headings ...

spec/lex.dd
```diff
@@ -551,7 +551,7 @@ q{ __TIME__ }       // " __TIME__ "
     // __EOF__ is not a token, it's end of file
 ---
 
-$(H4 Escape Sequences)
+$(H4 $(LNAME2 escape-sequences, Escape Sequences))
```
CyberShadow: Why dash here and underscore everywhere else?

wilzbach:
> Why dash here and underscore everywhere else?

It seems there's no standard and both dash and underscores are used in various spec pages, but underscores are used in `spec/lex.dd`, so it makes sense to be at least consistent on a file basis.
-> changed the dash label to an underscore label

PR #1675, opened 2017-06-07
Add labels to spec/statement.dd

wilzbach:


wilzbach:
CC @CyberShadow - anything blocking this? It just adds labels to the headings ...

PR #1676, opened 2017-06-07
Add labels to spec

wilzbach:

This contains the remaining bits of my pass through the D spec:

```
ls | parallel -j1 "echo {}; cat {} | grep -E 'H(2|3|4|5)'" | grep -v LNAME2
```

See also:
- https://github.com/dlang/dlang.org/pull/1677
- https://github.com/dlang/dlang.org/pull/1675
- https://github.com/dlang/dlang.org/pull/1674


spec/property.dd
```diff
@@ -65,7 +65,7 @@ $(THEAD Property, Description)
 $(TROW $(RELATIVE_LINK2 classinfo, $(D .classinfo)), Information about the dynamic type of the class)
 )
 
-$(SECTION3 $(LNAME2 init, .init) Property,
+$(SECTION3 $(LNAME2 init, .init Property),
```
wilzbach: This unifies this page (same headings contain "Property" in the link, some didn't).

spec/ddoc.dd
```diff
@@ -1043,7 +1043,7 @@ $(P
     cross-site scripting).
 )
 
-$(H2 Links to D documentation generators)
+$(H2 $(LNAME2 links_to_d_documentation_generators, Links to D documentation generators))
 
 $(P
     A list of current D documentation generators which use Ddoc
```
ZombineDev: Can you give "Comparing D Immutable and Const with C++ Const" the same treatment?

spec/cpp_interface.dd
```diff
@@ -315,7 +315,7 @@ instead disabled on the D side; an alternative would be to reimplement the
 constructor in D. See the $(DDSUBLINK spec/cpp_interface, lifetime-management,
 section below on lifetime management) for more information.)
 
-$(H3 Using D Classes From C++)
+$(H3 $(LNAME2 using_d_classes_from_cpp, Using D Classes From C++))
 
     $(P Given D code like:)
```
ZombineDev: Can you give "Comparing D Immutable and Const with C++ Const" the same treatment?

ZombineDev: Can you give "Comparing D Immutable and Const with C++ Const" the same treatment?

wilzbach:
> modulo one missed label.

Nice catch. I didn't think about checking for `SECTION2`.

PR #1677, opened 2017-06-07
Add labels to spec/expressions.dd

wilzbach:

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -49,7 +49,7 @@ $(GNAME CommaExpression):
         operand, and the result is the result of the right operand.
     )
 
-$(H3 Assign Expressions)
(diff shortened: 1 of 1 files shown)


spec/expression.dd
```diff
@@ -381,7 +381,7 @@ $(GNAME RelExpression):
         longer array.
     )
 
-$(H4 Integer comparisons)
+$(H4 LNAME2 integer_comparisons, Integer comparisons))
```
CyberShadow: `$(` is missing here, breaks PDF build

wilzbach:
> `$(` is missing here, breaks PDF build

I shouldn't make so many PRs at night ... Thanks a lot! Fixed!

PR #1679, opened 2017-06-07
spec/cpp: Add compile instructions

wilzbach:
Probably quite helpful to people who don't know that much about linkers or just want to copy/paste.
I wasn't sure whether we should recommend the dynamic or static library.

BTW the final example seems to be broken:

```
foo.o:foo.d:function _Dmain: error: undefined reference to 'void increment<int>(Foo<int>&)'
foo.o:foo.d:function _Dmain: error: undefined reference to 'void increment<char>(Foo<char>&)'
foo.o:foo.d:function _Dmain: error: undefined reference to 'void printThreeNext<char>(Foo<char>)'
collect2: error: ld returned 1 exit status
```

### DIFF
--- a/spec/cpp_interface.dd
+++ b/spec/cpp_interface.dd
@@ -88,6 +88,8 @@ void main()
     and then running it yields:)
 
 $(CONSOLE
+> g++ -c foo.cpp
+> dmd bar.d foo.o -L-lstdc++ $(ANDAND) ./bar
 i = 1
 j = 2
 k = 3
@@ -151,6 +153,8 @@ void bar()
     $(P Compiling, linking, and running produces the output:)
 
 $(CONSOLE
+> dmd -c foo.d
+> g++ bar.cpp foo.o -lphobos2 -pthread -o bar $(ANDAND) ./bar
 i = 6
 j = 7
 k = 8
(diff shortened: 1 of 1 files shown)


wilzbach:
Rebased to latest -> anything blocking this from being merged?
It just adds helpful information to the user (=how to run & compile the examples)

wilzbach:
> LGTM except we should either remove or change the failing example until we can fix it.

@MetaLang (or others) I absolutely agree, but that might be trickier than simply merging this PR which is _independent_ to this problem. After all, the errors probably have been there for years.
In any case I opened [a bug for this https://issues.dlang.org/show_bug.cgi?id=18043], s.t. it doesn't get forgotten once this is merged.

CC @ZombineDev maybe?

spec/cpp_interface.dd
```diff
@@ -361,6 +368,14 @@ int callE(E *e)
 }
 )
 
+$(CONSOLE
+> dmd -c base.d
+> g++ klass.cpp class.o -lphobos2 -pthread -o klass $(ANDAND) ./klass
```
JinShil: shouldn't that be `base.o`?

wilzbach: Yes. Fixed.

PR #1681, opened 2017-06-07
Allow to build the master branch instead of stable

wilzbach:
I think we should get rid of the error-prone manual patching of the dlang.org repo (e.g. https://github.com/dlang/installer/pull/230, [full list](https://github.com/dlang/installer/pulls?utf8=%E2%9C%93&q=patch)).

@MartinNowak This is a first step and should reduce the size of your patch drastically.
Does it help?

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -39,10 +39,18 @@ REMOTE_DIR=d-programming@digitalmars.com:data
 TMP?=/tmp
 
 # Last released versions
+# By default the master branch is built under {library,phobos}-prerelease
+# with USE_MASTER_AS_STABLE=1 you can build the master branch instead of a stable release
+ifeq (,${USE_MASTER_AS_STABLE})
 DMD_STABLE_DIR=${DMD_DIR}-${LATEST}
-DMD_REL=$(DMD_STABLE_DIR)/generated/$(OS)/release/$(MODEL)/dmd
 DRUNTIME_STABLE_DIR=${DRUNTIME_DIR}-${LATEST}
 PHOBOS_STABLE_DIR=${PHOBOS_DIR}-${LATEST}
+else
+DMD_STABLE_DIR=${DMD_DIR}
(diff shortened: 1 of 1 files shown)


CyberShadow:
This looks like it's replacing one hack with another. Not really my area of expertise so no comment on this one.

wilzbach:
> This looks like it's replacing one hack with another. Not really my area of expertise so no comment on this one.

Yes, but the idea is that this hack doesn't break due to changes in the patch set.
(I asked for review as you are the dlang.org tzar and have veto rights against such hacks).

CyberShadow:
> (I asked for review as you are the dlang.org tzar and have veto rights against such hacks).

Deferring all such rights to @MartinNowak for this PR.

posix.mak
```diff
@@ -39,10 +39,18 @@ REMOTE_DIR=d-programming@digitalmars.com:data
 TMP?=/tmp
 
 # Last released versions
+# By default the master branch is built under {library,phobos}-prerelease
+# with USE_MASTER_AS_STABLE=1 you can build the master branch instead of a stable release
```
CyberShadow: Actually one comment - I think all such hacks should be accompanied with a verbose explanation of why it's needed, and implied or explicit terms for when the hack can be removed.

MartinNowak:
It's not a hack, the release docs are a different target than the normal doc builds, b/c they are built before the git tags are published and do not include prerelease docs.
Looks good, but might not be sufficient, since the Makefile will still try to rebuild dmd/druntime/phobos and whatnot, instead of just using the passed DMD compiler.

MartinNowak:
Let's finish the ddox transition before investing more into this stuff, patching really doesn't hurt much atm.

CyberShadow:
> they are built before the git tags are published

Why?

Also, why do they need to be published? Why not simply created locally?

wilzbach:
> patching really doesn't hurt much atm.

Well, it's your time ;-)

PR #1683, opened 2017-06-08
Add new __traits to the D grammar spec

wilzbach:
It would be quite cool if we could __ensure__ that the D spec never gets outdated ...

@WalterBright it seems like you forgot to update the main grammar spec.


andralex:
Ah, thanks so very much. Yes, the spec is in dire need of "unittesting". I'm not sure how additions can be unittested though - if something's not there, you don't know it isn't. Changes are much easier to test for.

CyberShadow:
How about extracting the BNF blocks and checking if the code in DMD's test suite conforms to them?

I know two efforts of having a complete machine-readable declarative grammar of D:

https://rawgit.com/Hackerpilot/DGrammar/master/grammar.html
http://www.dsource.org/projects/visuald/browser/grammar

wilzbach:
> Crazy idea: at least for highly structured parts of the compiler - such as pragmas or __traits - we should be able to generate the documentation from the compiler ddoc comments in the code implementing things.

I gave this a shot: https://github.com/dlang/dmd/issues/6877 & #1688

> How about extracting the BNF blocks and checking if the code in DMD's test suite conforms to them?

How would you do this? Randomly generating valid blocks?

CyberShadow:
> How would you do this? Randomly generating valid blocks?

Feed the grammar into a parser generator, then use the resulting parser to parse the test suite.

PR #1682, opened 2017-06-07
Revert some PRs, fix master

CyberShadow:
Let's see if this works

PR #1684, opened 2017-06-08
Rename DMD_REL to DMD_STABLE v2

wilzbach:
Resubmission due to #1682


MartinNowak:
Those should be named {DMD,DRUNTIME,PHOBOS}_LATEST, as they are a specific version, but not intended to be used as stable tool compiler.
In particular confusing since we have a `STABLE_DMD_VER` in that same makefile.

PR #1685, opened 2017-06-08
Move generated docs{,-prerelease}.json to $GENERATED v2

wilzbach:
Due to https://github.com/dlang/dlang.org/pull/1682

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,8 +32,6 @@ css/cssmenu.css
 *.min.*
 
 deleteme.*
-docs-prerelease.json
-docs.json
 .dub/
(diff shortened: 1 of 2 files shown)


wilzbach:
At least we know know the error - it's due to the `rm` at the end which I left in there, in case someone uses an old dlang.org build ...

```
Writing filtered docs...
rm .generated/.release-files.txt .generated/.release-dummy.html
rm: cannot remove '.generated/.release-dummy.html': No such file or directory
posix.mak:513: recipe for target '.generated/docs.json' failed
make: *** [.generated/docs.json] Error 1
```

PR #1687, opened 2017-06-09
Unify DDoc Makefile build variables

wilzbach:
Ideally this should cleanup our Ddoc variables mess. It's noteworthy that Druntime uses completely different ones, so the idea is to change the build files at Druntime afterwards.

If I did it right, _no_ changes should be detected by DAutoTest.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -10,6 +10,7 @@
 #
 
 include osmodel.mak
+PWD=$(shell pwd)
 
 # Latest released version
 ifeq (,${LATEST})
@@ -19,19 +20,19 @@ endif
 NEXT_VERSION:=$(shell bash -c 'version=$$(cat VERSION);a=($${version//./ });a[1]="10\#$${a[1]}";((a[1]++)); a[2]=0; echo $${a[0]}.0$${a[1]}.$${a[2]};' )
 
 # DLang directories
-DMD_DIR=../dmd
-PHOBOS_DIR=../phobos
-DRUNTIME_DIR=../druntime
-TOOLS_DIR=../tools
-INSTALLER_DIR=../installer
-DUB_DIR=../dub-${DUB_VER}
+DMD_DIR=$(PWD)/../dmd
+PHOBOS_DIR=$(PWD)/../phobos
+DRUNTIME_DIR=$(PWD)/../druntime
+TOOLS_DIR=$(PWD)/../tools
+INSTALLER_DIR=$(PWD)/../installer
+DUB_DIR=$(PWD)/../dub-${DUB_VER}
 
 # External binaries
 DMD=$(DMD_DIR)/generated/$(OS)/release/$(MODEL)/dmd
 DUB=${DUB_DIR}/bin/dub
 
 # External directories
-DOC_OUTPUT_DIR:=$(shell pwd)/web
+DOC_OUTPUT_DIR:=$(PWD)/web
 GIT_HOME=https://github.com/dlang
 DPL_DOCS_PATH=dpl-docs
 DPL_DOCS=$(DPL_DOCS_PATH)/dpl-docs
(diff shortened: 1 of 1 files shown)


posix.mak
```diff
@@ -10,6 +10,7 @@
 #
 
 include osmodel.mak
+PWD=$(realpath .)
```
wilzbach: The idea here to get rid of all the repeated `realpath` clutter

posix.mak
```diff
@@ -107,6 +108,31 @@ else
 endif
 
 ################################################################################
+# Ddoc build variables
+################################################################################
+DDOC_VARS_STABLE=\
```
wilzbach: If we agree on [renaming _STABLE to _LATEST](https://github.com/dlang/dlang.org/pull/1689), this should be changed as well.
However without #1689 it the naming is `stable`...

posix.mak
```diff
@@ -375,48 +401,39 @@ $(DMD_STABLE) : ${DMD_STABLE_DIR}
...
 ################################################################################
 # druntime, latest released build and current build
+# TODO: remove DOCDIR and DOCFMT once they have been removed at Druntime
 ################################################################################
 
 druntime-prerelease : ${DRUNTIME_DIR} $(DMD) $(STD_DDOC_PRE)
-	${MAKE} --directory=${DRUNTIME_DIR} -f posix.mak target doc \
+	${MAKE} --directory=${DRUNTIME_DIR} -f posix.mak target doc $(DDOC_VARS_HTML) \
 		DOCDIR=${DOC_OUTPUT_DIR}/phobos-prerelease \
 		DOCFMT="$(addprefix `pwd`/, $(STD_DDOC_PRE))"
```
wilzbach: https://github.com/dlang/druntime/pull/1834 changes `DOCDIR` and `DOCFMT` to their more common names. After the next release (2.074.2), these lines can be removed for all druntime doc targets.

posix.mak
CyberShadow: `realpath` resolves symlinks and unnecessarily exposes implementation details in log files or even built files. We should get rid of it and just use `$(pwd)` instead.

wilzbach:
> @`realpath` resolves symlinks and unnecessarily exposes implementation details in log files or even built files. We should get rid of it and just use `$(pwd)` instead.

You mean `$(shell pwd)`, right?
Done.

CyberShadow:
Thanks!

>     Cloning into '/dev/shm/dtest/work/repo/dlang.org/../phobos-2.074.1'...

Looks like there's still a `readlink` somewhere...

wilzbach:
> The term `DDOC_VARS_STABLE` doesn't relate to `DDOC_VARS_HTML` and `DDOC_VARS_VERBATIM`.
Following the existing naming schemas, it prolly should be named `DDOC_VARS_HTML_STABLE`,

https://github.com/dlang/dlang.org/pull/1714

MartinNowak:
You also added a subtle breakage, now this makefiles requires any `DMD` or such variables to be absolute, while previously could deal well with relative paths.

posix.mak
```diff
@@ -355,14 +381,14 @@ dlangspec.verbatim.txt : $(DMD) verbatim.ddoc dlangspec-consolidated.d
...
 ################################################################################
 
-../%-${LATEST} :
-	git clone -b v${LATEST} --depth=1 ${GIT_HOME}/$* $@
+$(PWD)/%-${LATEST} :
+	git clone -b v${LATEST} --depth=1 ${GIT_HOME}/$(notdir $*) $@
 
-../%-${DUB_VER} :
-	git clone --depth=1 -b v${DUB_VER} ${GIT_HOME}/$* $@
+$(PWD)/%-${DUB_VER} :
```
MartinNowak: Furthermore this broke the clone targets, they are located at `$(PWD)/../%-${LATEST}`.

PR #1689, opened 2017-06-09
Rename {DMD,DRUNTIME,PHOBOS,VERSION}_STABLE -> _LATEST

wilzbach:
CC @MartinNowak @CyberShadow (follow-up to #1684)

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -528,11 +528,13 @@ $G/docs.json : ${DMD_LATEST} ${DMD_LATEST_DIR} \
 	find ${DMD_LATEST_DIR}/src -name '*.d' | \
 		sed -e /mscoff/d -e /objc_glue.d/d ${DMD_EXCLUDE}  \
 			> $G/.release-files.txt
+	find ${DRUNTIME_LATEST_DIR}/src -name '*.d' | \
 	  sed -e /unittest.d/d -e /gcstub/d >> $G/.release-files.txt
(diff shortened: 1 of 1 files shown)


CyberShadow:
Why? Are all these renames achieving something?

wilzbach:
Cleaning up the Makefile or in [Martin's words](https://github.com/dlang/dlang.org/pull/1684#issuecomment-307365531):

> Those should be named {DMD,DRUNTIME,PHOBOS}_LATEST, as they are a specific version, but not intended to be used as stable tool compiler.
In particular confusing since we have a STABLE_DMD_VER in that same makefile.

CyberShadow:
OK, deferring to Martin then

MartinNowak:
> Why? Are all these renames achieving something?

Yes, we want to introduce a new target named `release` for the installer builds, and use the correct term `latest` for the current one.
Also the term `stable` has 2 meanings in the current Makefile.

wilzbach:
Rebased to the latest, greatest `posix.mak`. I am assuming that Martin needs to update his diff, so it probably make sense to wait for him here.

wilzbach:
Rebased and this should be good to go now.
IIRC the nightly build server doesn't build the dlang.org anymore, so there's no risk at breaking it again.

CyberShadow:
Build is failing.

MartinNowak:
> Why? Are all these renames achieving something?

Clarifying terminology, the makefile is quite a mess atm.

PR #1691, opened 2017-06-10
Add documentation for uClibc and musl C libraries.

ibuclaw:
These are supported by existing compilers (i.e: `gdc -mmusl` or `configure --target=i386-linux-musl`), but there's currently no way to express that in D code.

https://github.com/dlang/dmd/pull/6885


ibuclaw:
Two positive reviews -> auto-merge

PR #1692, opened 2017-06-13
More expressive variable name in opApply example.

chadjoan:
Replace the variable name "result" with "abort".

### DIFF
--- a/spec/statement.dd
+++ b/spec/statement.dd
@@ -628,15 +628,15 @@ $(H4 $(LNAME2 foreach_over_struct_and_classes, Foreach over Structs and Classes 
 
             int opApply(scope int delegate(ref uint) dg)
             {
-                int result = 0;
+                int abort = 0;
 
                 for (int i = 0; i < array.length; i++)
                 {
-                    result = dg(array[i]);
-                    if (result)
+                    abort = dg(array[i]);
+                    if (abort)
                         break;
                 }
-                return result;
+                return abort;
             }
         }
         --------------


wilzbach:
Hmm its a return code and in Phobos actually `res` is used:

https://github.com/dlang/phobos/blob/master/std/range/package.d#L4608
https://github.com/dlang/phobos/blob/master/std/algorithm/iteration.d#L3005
...

chadjoan:
This might end up being subjective bikeshedding and whatnot, so feel free to close the pull if you're not feeling it.

I felt this would be more descriptive than anything like "result", "res", "return", "rc", and so on.  When I see something like "result", I wonder: what kind of result is it? is it a true/false with special casing (given that it's an int) or possibly more enumeration-like? maybe it has directionality like opCmp? if there are special cases or a list of possibilities, then what are they? perhaps the delegate returns an iteration count?  Reading the surrounding text will tell us how to use it, but that isn't as user-friendly as just making the usage obvious upon glance, as in the spirit of "show, don't tell".  The reading doesn't explain the rationale either (why an int and not a bool?  should D implementors be aware of some hidden utility in passing values besides 0 or 1?).

Something like "abort" has the advantage of specificity.  It suggests a boolean nature: iteration either aborts or it doesn't.  It suggests that halting has occured.  It even suggests that the halting occured _before the thing finished normally_.

I am also somewhat convincing/reminding myself that the "int" really should be "bool", because this would add to the example being more immediately descriptive and self-documenting.  I'm not quite sure if I'm up to doing or pushing that change though; it might have... compatibility implications.

CyberShadow:
> I am also somewhat convincing/reminding myself that the "int" really should be "bool",

I believe the result can have other values than 0 and 1.

Perhaps you may find "status" a better name.

chadjoan:
"status" would be slightly more specific than "result", but it opens another can of worms: so there's some list of statuses somewhere that no one put into an enum, and the values are never mentioned in the language spec.  So the author of an opApply method is receiving this extra information from the delegate they call, but they don't know how to use that extra information; the docs just tell them to break out of the loop and return when it's nonzero.  I'd also expect to see something like this:
```
status = dg(array[i]);
if ( status == ForeachStatus.Aborted )
	break;
```

dkorpel:
Considering the lack of consensus, the inactivity, and the fact that the example has been clarified in https://github.com/dlang/dlang.org/pull/3192, I'm closing this for now.

PR #1651, opened 2017-05-10
[WIP] Add Travis CI to test building dlang.org from scratch

wilzbach:
This will need some work.

PR #1693, opened 2017-06-15
Fix Issue 17501 - Runnable unittest problem with AST rewrite

wilzbach:
@CyberShadow are you okay with enabling Travis here, s.t. we can easily run `test` on a CI?

It works on my fork:
https://travis-ci.org/wilzbach/dlang.org

image

(I also wanted to revive the PR in which we build dlang.org from scratch on Travis, but that would be another step).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+dist: trusty
+sudo: false
+
+language: d
+os:
+ - linux
+d:
+ - dmd
+ - dmd-beta
+ - dmd-nightly
+ - ldc
+ - ldc-beta
+
+script:
+  - make -f posix.mak test
--- a/assert_writeln_magic.d
+++ b/assert_writeln_magic.d
@@ -43,14 +43,13 @@ private string formatNode(T)(const T t)
     return writer.data;
 }
 
-class TestVisitor : ASTVisitor
+class TestVisitor(Out) : ASTVisitor
 {
     import dparse.lexer : tok, Token;
 
-    this(string fileName, string destFile)
+    this(Out fl)
     {
-        this.fileName = fileName;
-        fl = FileLines(fileName, destFile);
+        this.fl = fl;
     }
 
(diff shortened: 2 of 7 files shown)


wilzbach:
CC @ZombineDev  - thanks a lot for reporting!

CyberShadow:
> @CyberShadow are you okay with enabling Travis here, s.t. we can easily run test on a CI?

Yeah but it could use DAutoTest just as well.

wilzbach:
> Yeah but it could use DAutoTest just as well.

I know, but it's a free CI & we get a nice UX separation between failed tests and failed documentation build.
Also [I wanted to use Travis to build dlang.org from scratch to ensure consistency](https://github.com/dlang/dlang.org/pull/1651) in the near future.

CyberShadow:
> I know, but it's a free CI & we get a nice UX separation between failed tests and failed documentation build.

Is this going to eat into Travis time for the other repositories, though?

wilzbach:
> Is this going to eat into Travis time for the other repositories, though?

Yes, but it currently takes less than one minute to run, so imho marginal. 
If I think about it, it doesn't make sense to test it with different versions as we download and use `STABLE_DMD` always anyhow (will make a short-follow, sorry).

.travis.yml
```diff
@@ -0,0 +1,15 @@
...
+
+language: d
+os:
+ - linux
+d:
+ - dmd
+ - dmd-beta
+ - dmd-nightly
+ - ldc
+ - ldc-beta
```
CyberShadow: I think this is overkill. Just test with the DMD built by the makefile.

wilzbach:
>> If I think about it, it doesn't make sense to test it with different versions as we download and use STABLE_DMD always anyhow (will make a short-follow, sorry).
> I think this is overkill. Just test with the DMD built by the makefile.

-> https://github.com/dlang/dlang.org/pull/1695

PR #1695, opened 2017-06-15
Use only one instance for the Travis build

wilzbach:
Follow-up to #1693

There's no point in testing with many compilers as we download and use `STABLE_DMD` anyhow.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,6 @@ sudo: false
 language: d
 os:
  - linux
-d:
- - dmd
- - dmd-beta
- - dmd-nightly
- - ldc
- - ldc-beta
 
 script:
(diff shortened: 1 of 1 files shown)


CyberShadow:
> but it currently takes less than one minute to run, 

It's been half an hour and it looks like it hasn't even started.

wilzbach:
> It's been half an hour and it looks like it hasn't even started.

Maybe some kind of weird deadlock due to enabling it today? It looks like the builds after it have started:

image

I tried rekicking it and it seems like it worked:

https://travis-ci.org/dlang/dlang.org/builds/243100080

CyberShadow:
Made Travis required, so @dlang-bot minds it for auto-merge.

PR #1694, opened 2017-06-15
Fix 17480 - Update file contents of the release archives & link to them on downloads.html

wilzbach:
- The issue is about the fact that we never mention that DUB is part of the release archives
- I tried to update the file contents archive and link to them (img + text)
- See below for the actual file listings from the archives

CC @MartinNowak @CyberShadow

File archives
----------------

- 2.074.1
- for better readability I removed the hits from other directories (the default configs of every OS are part of nearly all archives)

```
> als dmd.2.074.1.windows.7z | grep bin/
2017-05-31 17:29:06 ....A         1011               dmd2/windows/bin/README.TXT
2017-05-31 17:41:22 ....A      3772330               dmd2/windows/bin/d.chm
2017-05-31 17:28:32 ....A         4630               dmd2/windows/bin/sc.ini
2017-05-31 17:29:06 ....A       215274               dmd2/windows/bin/windbg.hlp
2017-05-31 17:41:50 ....A       540188      4129070  dmd2/windows/bin/ddemangle.exe
2017-05-31 17:29:08 ....A        85776               dmd2/windows/bin/dm.dll
2017-05-31 17:42:18 ....A      2300956               dmd2/windows/bin/dman.exe
2017-05-31 17:37:44 ....A      3165724               dmd2/windows/bin/dmd.exe
2017-05-31 17:43:06 ....A      8267908               dmd2/windows/bin/dub.exe
2017-05-31 17:42:06 ....A       891420               dmd2/windows/bin/dustmite.exe
2017-05-31 17:29:06 ....A       150288               dmd2/windows/bin/eecxxx86.dll
2017-05-31 17:29:08 ....A        80656               dmd2/windows/bin/emx86.dll
2017-05-31 17:29:08 ....A        65564               dmd2/windows/bin/lib.exe
2017-05-31 17:29:06 ....A       451072               dmd2/windows/bin/libcurl.dll
2017-05-31 17:29:08 ....A       223260               dmd2/windows/bin/link.exe
2017-05-31 17:29:08 ....A        49692               dmd2/windows/bin/make.exe
2017-05-31 17:29:06 ....A       275216               dmd2/windows/bin/mspdb41.dll
2017-05-31 17:41:36 ....A       796700               dmd2/windows/bin/rdmd.exe
2017-05-31 17:29:06 ....A        32768               dmd2/windows/bin/replace.exe
2017-05-31 17:29:08 ....A        57616               dmd2/windows/bin/shcv.dll
[...]

PR #1697, opened 2017-06-15
Link to release description on downloads.dd

wilzbach:
Cherry-picked from #1694 as this was controversial:

> Not sure about making the platform name a link to the documentation. I'd think that being on the 
download page, clicking it would take me to all Windows downloads.

We could also add a link "Docs", but I though that this is the leaner way to do it.

### DIFF
--- a/download.dd
+++ b/download.dd
@@ -267,6 +267,16 @@ Macros:
         </div>
         )
     )
+    C_A=$1:$2;
+    EXTRA_HEADERS=$(T style,
+        .download_paragraph > span {
+            visibility: hidden;
+            $(C_A font-size, 125%)
+        }
+        .download_paragraph:hover > span {
+            visibility: visible;
+        }
+    )
     DOWNLOAD_OTHER =
     $(DIV,
         $(DIVC download_image, $1)
--- a/download.dd
+++ b/download.dd
@@ -64,17 +64,29 @@ $(DIVC download_channels,
 
     $(BR)$(BR)
 
-    $(DOWNLOAD $(WINDOWS), Windows, $(SBTN $(WINEXE), exe) $(SBTN $(ARCH windows, 7z), 7z))
+    $(DOWNLOAD Windows, $(WINDOWS), windows,
+        $(SBTN $(WINEXE), exe) $(SBTN $(ARCH windows, 7z), 7z)
+    )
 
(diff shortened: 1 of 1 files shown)


CyberShadow:
Hmm, how about something like show an 🛈 on hover next to the platform name that links to the help page or something like that?

wilzbach:
> Hmm, how about something like show an 🛈 on hover next to the platform name that links to the help page or something like that?

Before:

image

After:

image

CyberShadow:
```css
.download_paragraph > span {
    visibility: hidden;
    font-size: 125%;
}
.download_paragraph:hover > span {
    visibility: visible;
}
```
?

wilzbach:
Well I like the increased font-size, but I personally don't like user interfaces on which I don't immediately see all buttons, but I don't feel strongly about this & would like to move forward here
-> I added your CSS definitions -> should be good to go?

CyberShadow:
Not everything on the page is equally important, so bombarding visitors with visual clutter won't result in the best experience. CSS hover tricks are probably not the best way to avoid that, but I think it's still better than nothing.

MartinNowak:
I've first taken this for a security warning or sth. Looks definitely more like an alert than an info. Let's better change the font color to blue or black.

download.dd
```diff
@@ -64,17 +64,29 @@ $(DIVC download_channels,
...
+        $(SBTN $(ARCH osx, tar.xz), tar.xz)
+    )
 
-    $(DOWNLOAD $(UBUNTU) &nbsp; $(DEBIAN), Ubuntu/Debian, $(SBTN $(DEB32), i386) $(SBTN $(DEB64), x86_64) $(SBTN $(ARCH linux, tar.xz), tar.xz))
+    $(DOWNLOAD Ubuntu/Debian, $(UBUNTU) &nbsp; $(DEBIAN), linux,
+        $(SBTN $(DEB32), i386) $(SBTN $(DEB64), x86_64) $(SBTN $(ARCH linux, tar.xz), tar.xz)
+    )
 
-    $(DOWNLOAD $(FEDORA) &nbsp; $(CENTOS), Fedora/CentOS, $(SBTN $(RPM32), i386) $(SBTN $(RPM64), x86_64) $(SBTN $(ARCH linux, tar.xz), tar.xz))
+    $(DOWNLOAD Fedore/CentOS, $(FEDORA) &nbsp; $(CENTOS), linux,
```
CyberShadow: It's "Fedora", not "Fedore".

CyberShadow:
-> https://github.com/dlang/dlang.org/pull/2059

PR #1699, opened 2017-06-15
Fix Issue 17265 - WithStatement: Find better Example for what with really does

wilzbach:
AFAICT this simple change should work.

CC @UplinkCoder

### DIFF
--- a/spec/statement.dd
+++ b/spec/statement.dd
@@ -1368,7 +1368,7 @@ with (expression)
 }
 --------------
 
-        $(P Note that $(EXPRESSION) only gets evaluated once.
+        $(P Note that $(EXPRESSION) only gets evaluated once and is not copied.
         The with statement does not change what $(D this) or
         $(D super) refer to.
         )


CyberShadow:
I'm not sure that this is correct, as it would imply that `with` is not usable with rvalues, which is clearly not true.

UplinkCoder:
@CyberShadow the r values get assigned to a temporary variable.
The deref of the addr of the temporary variable is used.

UplinkCoder:
@wilzbach looks correct ... I am not sure if we want to expose that implementation detail though.

wilzbach:
> @wilzbach looks correct ... I am not sure if we want to expose that implementation detail though.

It's already exposed - the bug report was about the fact the current "pseudo-lowering' isn't correct for structs.

CyberShadow:
> @CyberShadow the r values get assigned to a temporary variable.
> The deref of the addr of the temporary variable is used.

That's not what the pseudocode here shows.

spec/statement.dd
```diff
@@ -1372,14 +1372,14 @@ with (expression)
...
-    tmp = expression;
+    Object* tmp;
+    tmp = &expression;
     ...
     tmp.ident;
 }
 --------------
 
-        $(P Note that $(EXPRESSION) only gets evaluated once.
+        $(P Note that $(EXPRESSION) only gets evaluated once and not copied by-value.
```
andralex: We should improve this. The address taking example is not edifying because it doesn't work with rvalues. Furthermore, there's no obvious lowering. I think it's enough to leave the example as is and simply mention in the text the following:

`$(P Note that $(EXPRESSION) only gets evaluated once and is not copied.`

andralex: ping @wilzbach

wilzbach:
> I think it's enough to leave the example as is and simply mention in the text the following:

Okay, rebased.

> That's not what the pseudocode here shows.

Hmm, so what's the consensus on this? Better ideas?

CyberShadow:
> Okay, rebased.

I think you still need to revert the example change, if you're going by what Andrei said.

wilzbach:
> I think you still need to revert the example change, if you're going by what Andrei said.

Oh sorry, done.

PR #1701, opened 2017-06-15
Fix Issue 17224 - Foreach documentation still refers to TypeTuples, rather than AliasSequences

wilzbach:


wilzbach:
Phobos PRs: https://github.com/dlang/phobos/pull/5484 and  https://github.com/dlang/phobos/pull/5485

PR #1702, opened 2017-06-15
Fix Issue 17159 - Behavior of unions at compile time is not documented

wilzbach:
```d
void main(string[] args)
{
    union U {
        int a;
        double b;
    }
    enum u = U(1);
    enum c = u.b;
}
```

```
> rdmd foo.d
foo.d(8): Error: reinterpretation through overlapped field b is not allowed in CTFE
```

CC @UplinkCoder

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -2104,6 +2104,7 @@ $(H3 $(LNAME2 interpretation, Compile Time Function Execution (CTFE)))
     $(LI Non-portable casts (eg, from $(D int[]) to $(D float[])), including
         casts which depend on endianness, are not permitted.
         Casts between signed and unsigned types are permitted)
+    $(LI Reinterpretation of overlapped fields in a Union.)
     )
 
     $(P Pointers are permitted in CTFE, provided they are used safely:)


don-clugston-sociomantic:
Note that C99 annex J states that it is unspecified behaviour to access a union by any element other than the one that was last written.
CTFE disallows almost all behaviour that is unspecified in C.

And the word 'of' appears in your sentence twice.

wilzbach:
> Note that C99 annex J states that it is unspecified behaviour to access a union by any element other than the one that was last written.
CTFE disallows almost all behaviour that is unspecified in C.

Thanks a lot for the explanation, Do you want to reword the addition or note this in the specification?

> And the word 'of' appears in your sentence twice.

Oh apparently I fixed too many bugs tonight. Thanks.

UplinkCoder:
This looks fine.

spec/function.dd
```diff
@@ -2105,6 +2105,7 @@ $(H3 $(LNAME2 interpretation, Compile Time Function Execution (CTFE)))
         casts which depend on endianness, are not permitted.
         Casts between signed and unsigned types are permitted)
     )
+    $(LI Reinterpretation of overlapped fields in a Union)
```
CyberShadow: This needs to be one line higher (the `)` above it closes the `$(OL`).

wilzbach:
> This needs to be one line higher (the `)` above it closes the `$(OL`).

Fixed.

PR #1703, opened 2017-06-15
Fix Issue 17121 - DDoc documentation is out of date

wilzbach:
Not really sure on the best course of action here. Just dumping the entire file with all its stylesheets feels wrong.

> The new predefined macros are in a separate file [1]. Is it possible to embed that in the rest of the Ddoc documentation?

@jacob-carlborg I guess you would like to be able to preprocess DDoc, e.g. https://github.com/dlang/dlang.org/pull/1688?

CC @acehreli

### DIFF
--- a/spec/ddoc.dd
+++ b/spec/ddoc.dd
@@ -682,6 +682,8 @@ $(P
     minimal definitions needed by Ddoc to format and highlight
     the presentation.
     The definitions are for simple HTML.
+    The $(LINK2 https://github.com/dlang/dmd/blob/master/res/default_ddoc_theme.ddoc,
+    curent defintion) is a bit enhanced and includes more styling.
 )
 
 $(DDOCCODE
--- a/spec/ddoc.dd
+++ b/spec/ddoc.dd
@@ -683,7 +683,7 @@ $(P
     the presentation.
     The definitions are for simple HTML.
     The $(LINK2 https://github.com/dlang/dmd/blob/master/res/default_ddoc_theme.ddoc,
-    curent defintion) is a bit enhanced and includes more styling.
+    current definition) is a bit enhanced and provides more options for styling.
 )
 
 $(DDOCCODE


jacob-carlborg:
> jacob-carlborg I guess you would like to be able to preprocess DDoc, e.g. #1688?

I'm not sure what you mean with preprocess Ddoc. #1688 suggests to use Ddoc comments but there are no Ddoc comments. If the Ddoc document are but in comments as is, the compiler would evaluate the macros. I'm thinking of if there's some way to escape the whole Ddoc default theme document to be able to render that on the site.

wilzbach:
> . #1688 suggests to use Ddoc comments but there are no Ddoc comments. 

#1688 allows to evaluate D code and inject arbitrary variables into you the Ddoc template, i.e. it makes proper templates out of the Ddoc files, e.g. on the D site you could do sth. like this:

```d
context["DDOC_COMPILER_TEMPLATE"] = "default_ddoc_theme".readText.rescapeDdoc;
```

and then in the Ddoc file:

```
...
{{ DDOC_COMPILER_TEMPLATE }}
...
```

> If the Ddoc document are but in comments as is, the compiler would evaluate the macros. 

You can escape it manually as well, but I thought the main point of this issue was to automatically keep the specification in sync?

jacob-carlborg:
Aha, cool. As long as it works 😃.

jacob-carlborg:
> IMO, the addition of the new DDoc theme was not executed correctly - the default theme should've been left as it was, and the new theme should've been shipped as a separate file that was parsed by default by the compiler, similar to object.d.

How is that different from now?

CyberShadow:
> How is that different from now?

I see that the DDoc theme is built into the DMD executable with an `import` expression; it would have been better to include it directly with the binary and make DMD load it at runtime, so that it's more transparent to the users. Also, I don't know of any way to get the old theme with a command-line switch.

andralex:
Close this as superseded by https://github.com/dlang/dlang.org/pull/1805?

jacob-carlborg:
> Close this as superseded by #1805?

Fine by me.

PR #1708, opened 2017-06-15
Switch to CircleCi

wilzbach:
> Travis stuck again.

OK, apparently we can't use Travis for any other repo except for DMD.
Then let's switch to CircleCi. It's very trivial here as all setup is done within the Makefile :)

Worked immediately on my fork: https://circleci.com/gh/wilzbach/dlang.org/2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,0 @@
-dist: trusty
-sudo: false
-
-language: d
-os:
- - linux
-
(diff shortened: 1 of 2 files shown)


wilzbach:
From #1713

> This is now live, so I'm going to disable Travis

Alrighty, closing this one then as well.

PR #1709, opened 2017-06-15
Allow runnable examples anywhere

wilzbach:
With this addition the D specification could be made runnable as it's then a simple question of going through it once and adding `$(RUNNABLE_EXAMPLE )` around everything all code blocks that are runnable.
As an example I have "runnified" spec/module.html

Of course, unlike for Phobos, we don't verify the run status yet.


wilzbach:
Update: As explained in https://github.com/dlang/dlang.org/pull/1711#issuecomment-308732215, there's absolutely no need for the MD5 hashing - saving STDIN next to the example is a lot friendlier to the user,
faster, easier to implement (and a [quick search through the git log](https://github.com/dlang/dlang.org/commits/master/js/run-main-website.js) showed that people have already forgotten to update the MD5 hash ... even with peer review) .

wilzbach:
For the lazy - a link to the modified module spec: http://dtest.dlang.io/artifact/website-a4153fc60f57d597ed0ef253643679077b7374af-6e5e8ea6c097cbc8c88dba65ceeb3f89/web/spec/module.html

CC @andralex - you might like this ;-)

ZombineDev:
> As explained in #1711 (comment), there's absolutely no need for the MD5 hashing - saving STDIN next to the example is a lot friendlier to the user,
faster, easier to implement (and a quick search through the git log showed that people have already forgotten to update the MD5 hash ... even with peer review) .

Indeed, the implementation looks much better know.

wilzbach:
And it's online: http://dlang.org/spec/module.html

It looks that most of the spec examples actually don't compile, because they explicitly trigger errors. Any good idea how we could mitigate this?
Maybe simply commenting these bits?

andralex:
Oh, I see. Well I'd say in the cases where the error is part of the very point of the code, leave it in. For example:

```d
import io = std.stdio;
void main()
{
    io.writeln("hello!");        // ok, calls std.stdio.writeln
    std.stdio.writeln("hello!"); // error, std is undefined
    writeln("hello!");           // error, writeln is undefined
}
```

I suppose part of the impetus to edit and run that example is to mess with it and see what fails and what works, starting under the assumption it's not buildable as is.

PR #1710, opened 2017-06-15
Show latest article of the DBlog on the frontpage

wilzbach:
I just had this crazy idea and funnily it works quite well - even with Ddoc :)

This definition is only used when `DIFFABLE` is not set, so it won't be possible to preview this on the DAutoTest, but here's how it will look in action:

image.

For a local build without internet, `curl` will fail silently and it will just create an empty file.
On subsequent runs, this target won't be executed. (No `.PHONY`).

This approach is also nicer than the previous JavaScript injection, because search engines see the new content & they like sites that are updated often.

It's fairly unlikely that the DBlog theme will change soon and if - as said before - it will just silently fail ...

CC @mdparker @JackStouffer @CyberShadow


index.dd
```diff
@@ -94,10 +94,11 @@ $(DIVC boxes,
                 http://youtube.com/playlist?list=PL3jwVPmk_PRxo23yyoc0Ip_cP3-rCm7eB,
                 this YouTube playlist).
             )
-            $(P Stay updated with
-                $(LINK2 http://arsdnet.net/this-week-in-d,
-                    <cite>This Week in D</cite>) -
-                $(SPANC twid)
```
ZombineDev: Let's keep This Week in D. It's pretty cool, despite not having the "official" stamp on it. Just move it below the official blog. May be something like:

> Stay updated with the [official D Blog](https://dlang.org/blog/) and [This Week in D](http://arsdnet.net/this-week-in-d):
> * [D Blog - Compile-Time Sort in D](http://dlang.org/blog/2017/06/05/compile-time-sort-in-d/) by Michael Parker from June 5, 2017.
> * [This Week in D - May 28](http://arsdnet.net/this-week-in-d/2017-may-28.html): Blog article on optimization and Andrei Israel trip, Beta D 2.074, tip on mixin w/ anon func.

wilzbach:
> Let's keep This Week in D. It's pretty cool, despite not having the "official" stamp on it. Just move it below the official blog. May be something like:

I considered this for a moment, but then dropped it because of space problems:

image

Any ideas?

adamdruppe:
I didn't get a chance to update twid again since I was out of town again.... I wouldn't mind if you simplified or removed it to save space.

ZombineDev:
If space is an issue we can try with this:

> Stay updated with:
> * The official D Blog: [Compile-Time Sort in D](http://dlang.org/blog/2017/06/05/compile-time-sort-in-d/) from June 5 2017, and
> * This Week in D:  [May 28](http://arsdnet.net/this-week-in-d/2017-may-28.html).

Or with an even shorter version:

> Stay updated with the [official D Blog - Compile-Time Sort in D](http://dlang.org/blog/2017/06/05/compile-time-sort-in-d/) from June 5 2017 and [This Week in D - May 28](http://arsdnet.net/this-week-in-d/2017-may-28.html).

wilzbach:
CC @aG0aep6G - as you rescued the Dlang.org template - what's your opinion on this?

CC @popen2 - at DConf you mentioned that you were planning on redesigning dlang.org - maybe you have already an idea in mind?

aG0aep6G:
> CC @aG0aep6G - as you rescued the Dlang.org template - what's your opinion on this?

I really dislike how much the home page currently depends on JavaScript, so this is interesting. It feels weird to generate the "latest" stuff when building the site, but if you do that regularly enough, sure.

Does it work, though? The doc tester shows "from : by . ". Also, https://blog.dlang.org doesn't work. Needs "http://".

Regarding space problems, you could cut the DConf text down. One paragraph instead of two; shorter "thank you" or none at all; "through this YouTube playlist" -> "on YouTube" - stuff like that.

Regarding TWID, it doesn't seem very alive. So I'd go with the blog if only one can get in.

CyberShadow:
I think the DConf bit needs to be moved out of the news section and somewhere else, a banner or a sidebar widget. Then there should be enough room for the blog and TWiD in the News section (no need to listify them either, just two paragraphs).

wilzbach:
> Does it work, though? The doc tester shows "from : by . ".

Yes as mentioned before, I explicitly disable this (as dynamic content) on the DAutoTest to avoid random changes (`DIFFABLE=1` is set on DAutotTest).
However on the deployment, `DIFFABLE` isn't set. For example, the [pending changelog](https://dlang.org/changelog/2.075.0_pre.html) is built similarly.
If the `curl` trick looks too weird, I can always plugin a simple DScript instead...

> Also, https://blog.dlang.org doesn't work. Needs "http://".

Oh thanks!

> I think the DConf bit needs to be moved out of the news section and somewhere else, a banner or a sidebar widget. Then there should be enough room for the blog and TWiD in the News section (no need to listify them either, just two paragraphs).

Okay,
- I also parse TWID at "Compile-Time" now, so there's one ugly layout reflow less on the main page.
- I removed the "Thank you DConf 2017" notice 

image

Now adding the DConf link to "Learn" yields (and slightly adjusting the News intro):

image

@CyberShadow the only place I could find to put the playlist link was the "Learn" section. Better ideas?

CyberShadow:
> Better ideas?

I was thinking maybe take http://dconf.org/2017/images/dconf_logo_2017.svg, crop it, and put it in the sidebar above or under the Tweeter widget, what do you think? Plenty of space there.

wilzbach:
> I was thinking maybe take http://dconf.org/2017/images/dconf_logo_2017.svg, crop it, and put it (with some text) in the sidebar above or under the Tweeter widget, what do you think? Plenty of space there.

Nice idea - like this? Maybe a bit more spacing?

image

CyberShadow:
Yep! Could use a bit of text under the logo (essentially what we have in the news section right now).

wilzbach:
> Yep! Could use a bit of text under the logo (essentially what we have in the news section right now).

default

image

12px

image

sitemap.dd
```diff
@@ -0,0 +1,203 @@
+Ddoc
```
CyberShadow: Did you add this accidentally?

posix.mak
```diff
@@ -378,6 +383,22 @@ dlangspec.verbatim.txt : $(DMD) verbatim.ddoc dlangspec-consolidated.d
 	$(DMD) -conf= -Df$@ verbatim.ddoc dlangspec-consolidated.d
 
 ################################################################################
+# Fetch the latest article from the official D blog
+################################################################################
+
+$G/dblog_latest.ddoc:
+	@echo "Receiving the latest DBlog article. Disable with DIIFABLE=1"
```
CyberShadow: DIFFABLE, not DIIFABLE

posix.mak
```diff
@@ -378,6 +383,22 @@ dlangspec.verbatim.txt : $(DMD) verbatim.ddoc dlangspec-consolidated.d
...
+	curl -s --retry 3 --retry-delay 5 http://blog.dlang.org | grep -m1 'entry-title' | \
+		sed -E 's/^.*<a href="(.+)" rel="bookmark">([^<]+)<\/a>.*<time class="updated" datetime="[^"]+">([^<]*)<\/time>.*Author *<\/span><a [^>]+>([^<]+)<\/a>.*/DBLOG_LATEST_TITLE=\2|DBLOG_LATEST_LINK=\1|DBLOG_LATEST_DATE=\3|DBLOG_LATEST_AUTHOR=\4/' | \
+		tr '|' '\n' > $@
+
+$G/twid_latest.ddoc:
+	@echo "Receiving the latest TWID article. Disable with DIIFABLE=1"
+	curl -s --retry 3 --retry-delay 5 http://arsdnet.net/this-week-in-d/twid-latest.js | grep setTwid | \
+		sed -E 's/.*setTwid\("([^"]+)", "([^"]+)", "([^"]+)"\).*/TWID_LATEST_LINK=\1|TWID_LATEST_DATE=\2|TWID_LATEST_TITLE=\3/' | \
```
CyberShadow: Is this going to handle special characters nicely?

images/dconf_logo_2017.svg
```diff
@@ -0,0 +1,213 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
CyberShadow: Hotlinking it would've been fine, but this is fine too

sitemap.dd
wilzbach: Yes, I really need some sleep :P
In any case, we should add this to the gitignore or even better: simply put it into `.generated`
https://github.com/dlang/dlang.org/pull/1725

index.dd
```diff
@@ -128,7 +142,10 @@ $(DIVC boxes,
             	check out $(LINK2 https://wiki.dlang.org/Books, books) about D,
             	among others Ali Çehreli's
                 $(LINK2 http://ddili.org/ders/d.en/index.html,
-                <cite>Programming in D</cite>).
+                <cite>Programming in D</cite>) or
+                $(LINK2 http://youtube.com/playlist?list=PL3jwVPmk_PRxo23yyoc0Ip_cP3-rCm7eB, watch)
+                talks from
+                $(LINK2 http://dconf.org/2017/index.html, DConf17)
```
wilzbach: Should I still keep this if we add the DConf banner?

CyberShadow: Yeah, not sure how good the talks are effective for actually learning D.

wilzbach:
> Is this going to handle special characters nicely?

- I don't see why not? [Example with Regex101](https://regex101.com/r/cGz0vE/1/)
- [RFC 3966](https://tools.ietf.org/html/rfc3986#section-2) defines `|` as an invalid symbol and would need to be escaped

CyberShadow:
If I understand that regex correctly, if a TWiD issue has a " in it, it's not going to show up at all.

CyberShadow:
I don't understand how RFC 3966 plays in this. The TWiD data is a JavaScript file, it's not URL-encoded.

wilzbach:
> If I understand that regex correctly, if a TWiD issue has a " in it, it's not going to show up at all.
> don't understand how RFC 3966 plays in this. The TWiD data is a JavaScript file, it's not URL-encoded.

Yes & Yes, sorry.
Updated the RegEx: https://regex101.com/r/cGz0vE/2 (at least for TWID I hope that this is only a temporary solution anyways as for @adamdruppe it's probably the same effort to write his latest article information in this extracted Ddoc format than JS).

images/dconf_logo_2017.svg
wilzbach: I prefer copying - just in case the layout of dconf.org changes ...

wilzbach:
Alrighty, I have fixed all open "issues" and this should be good to go. Anything else blocking this?
(I also squashed the DConf banner changes into one commit).

CyberShadow:
@adamdruppe Any news on twid_latest.dd?

posix.mak
```diff
@@ -378,6 +383,22 @@ dlangspec.verbatim.txt : $(DMD) verbatim.ddoc dlangspec-consolidated.d
...
+$G/dblog_latest.ddoc:
+	@echo "Receiving the latest DBlog article. Disable with DIFFABLE=1"
+	curl -s --retry 3 --retry-delay 5 http://blog.dlang.org | grep -m1 'entry-title' | \
+		sed -E 's/^.*<a href="(.+)" rel="bookmark">([^<]+)<\/a>.*<time class="updated" datetime="[^"]+">([^<]*)<\/time>.*Author *<\/span><a [^>]+>([^<]+)<\/a>.*/DBLOG_LATEST_TITLE=\2|DBLOG_LATEST_LINK=\1|DBLOG_LATEST_DATE=\3|DBLOG_LATEST_AUTHOR=\4/' | \
+		tr '|' '\n' > $@
+
+$G/twid_latest.ddoc:
+	@echo "Receiving the latest TWID article. Disable with DIFFABLE=1"
+	curl -s --retry 3 --retry-delay 5 http://arsdnet.net/this-week-in-d/twid-latest.js | grep setTwid | \
+		sed -E 's/.*setTwid\("([^"]+)", "([^"]+)", "(.+)"\);$$/TWID_LATEST_LINK=\1|TWID_LATEST_DATE=\2|TWID_LATEST_TITLE=\3/' | \
```
CyberShadow: Quotes (and other JS escapes) in TWID titles are still going to show up wrong.

adamdruppe:
http://arsdnet.net/this-week-in-d/twid_latest.dd

wilzbach:
Thanks a lot @adamdruppe! Integrated & rebased. Anything else missing? :)

wilzbach:
For reference, this is how it currently looks:

image

image

CyberShadow:
Looks great!

wilzbach:
And fetching the latest resources on the deploy for dlang.org worked - it's live now :)

Thanks a lot @adamdruppe @mdparker @ZombineDev @aG0aep6G @CyberShadow !

@adamdruppe dlang.org gets built and deployed everytime a PR gets merged to dlang/{dmd,druntime,phobos,dlang.org} - according to your stats that's about 30 times a week or ~ three times a day. So, in the worst case, there might be a lag of a few hours until new articles are propagated on dlang.org.

CyberShadow:
If you want, I can add a trigger that forces a rebuild

adamdruppe:
On Tue, Jun 27, 2017 at 11:58:35AM -0700, Sebastian Wilzbach wrote:
> @adamdruppe dlang.org gets built and deployed everytime a PR gets merged to dlang/{dmd,druntime,phobos,dlang.org}

There's a lag with me updating it too, some days I don't get around
to it till late monday, so not a big deal.

PR #1711, opened 2017-06-15
Make average line length example more concise and move it to the top

ZombineDev:
The primary motivation is that we need to have a simple example for first time visitors.

For comparison: [before](http://dtest.dlang.io/artifact/website-36ff9d12adcb438f7dfae521c67249fc1b96d4e7-6138877b18af2c2ca98cf2e1c871bc14/web/index.html) and [after](http://dtest.dlang.io/artifact/website-bdfbe11d2fd5710bff433c6b988cde641b28af9f-8c7e2386e90ebff5b8e64e2165294823/web/index.html).

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -30,6 +30,27 @@ $(DIVC intro, $(DIV, $(DIV,
         )
         $(DIVID your-code-here-default,
 $(RUNNABLE_EXAMPLE
+$(RUNNABLE_EXAMPLE_STDIN
+The D programming language
+Modern convenience.
+Modeling power.
+Native efficiency.)
+----
+// Compute average line length for stdin
+void main()
+{
+    import std.range, std.stdio;
+
+    auto sum = 0.0;
+    auto count = stdin.byLine
+        .tee!(l => sum += l.length).walkLength;
+
+    writeln("Average line length: ",
+        count ? sum / count : 0);
+}
+----
+)
+$(EXTRA_EXAMPLE
 $(RUNNABLE_EXAMPLE_STDIN 2.4 plus 2.4 equals 5 for sufficiently large values of 2.)
 ----
 // Round floating point numbers


index.dd
```diff
@@ -227,19 +227,13 @@ $(P Built-in linear and associative arrays, slices, and ranges make daily
...
-    foreach (line; stdin.byLine())
-    {
-        ++lines;
-        sumLength += line.length;
-    }
-    writeln("Average line length: ",
-        lines ? cast(double) sumLength / lines : 0.0);
+    import std.range, std.stdio;
+    auto sum = 0.0;
+    auto count = stdin.byLine().tee!(l => sum += l.length).walkLength;
```
wilzbach: Nit: maybe removing the parentheses here to be more in the D-style?

ZombineDev:
> I really like this example. How about adding it to the example code roulette as well?

Is there a way to share the source code between the different examples? Or perhaps we should swap this one with the rounding one>

wilzbach:
Btw I just realized that we probably need to update the MD5 hash:

https://github.com/dlang/dlang.org/blob/master/js/run-main-website.js

Imho this is a really annoying approach and we get rid of this approach entirely and store the STDIN in the DOM as well. This manual hashing approach wouldn't scale anyways if we go with https://github.com/dlang/dlang.org/pull/1709

> Is there a way to share the source code between the different examples?

From `js/dlang.js`:

```js
        // [your code here] rotation for index.html
        var $examples = $('.your-code-here-extra > pre');
        if ($examples.length) {
            var n = Math.floor(Math.random() * ($examples.length+1));
            if (n)
                $('#your-code-here-default > pre').replaceWith($examples[n-1]);
        }
```

So in theory just adding another $(EXTRA_EXAMPLE should work...

>  Or perhaps we should swap this one with the rounding one?

I would leave it in the roulette for now as we only have two examples at the moment.
Maybe we can steal a couple from the DTour (or other)?

wilzbach:
> Btw I just realized that we probably need to update the MD5 hash:
> Imho this is a really annoying approach and we get rid of this approach entirely and store the STDIN in the DOM as well. This manual hashing approach wouldn't scale anyways if we go with #1709

And a lot of chunk is gone: https://github.com/dlang/dlang.org/pull/1709/commits/44cbc937104d2c1529bc020f5856b8ab47366453 (#1709)

ZombineDev:
So what's left for me to do? Copy & paste the example to the top?

wilzbach:
> So what's left for me to do? Copy & paste the example to the top?

Yep, but better do this as a separate PR, so that this one isn't blocked
(I don't know how long it will take for #1709 to get in ...)

wilzbach:
> So what's left for me to do? Copy & paste the example to the top?

Once you have rebased your PR, you should be able to do this :)

ZombineDev:
@wilzbach done. @CyberShadow @aG0aep6G what do you think?

index.dd
```diff
@@ -243,20 +243,17 @@ Modern convenience.
...
-    foreach (line; stdin.byLine())
-    {
-        ++lines;
-        sumLength += line.length;
-    }
+    import std.range, std.stdio;
+
+    auto sum = 0.0;
+    auto count = stdin.byLine
+        .tee!(l => sum += l.length).walkLength;
```
CyberShadow: Even shorter would be:
```
size_t count = 0;
double sum = stdin.byLine.tee!(l => count++).sum;
```
but I guess that demonstrates fewer range features.

ZombineDev: I don't think that would compile, unless you add a `map!(l => l.length)` between the `tee` and the `sum`.

wilzbach:
> I would leave it in the roulette for now as we only have two examples at the moment.
> Maybe we can "steal" a couple from the DTour (or other sources)?

https://github.com/dlang/dlang.org/pull/1755

ZombineDev:
Ping @wilzbach  @CyberShadow is this good to go?

index.dd
CyberShadow: Oops, you're right :)

ZombineDev:
Thanks @CyberShadow!

PR #1712, opened 2017-06-15
move std.{outbuffer,signals,socket,xml} to undeaD

wilzbach:
This hides the respective modules from the documentation sidenav.
See https://github.com/dlang/phobos/pull/5488


adamdruppe:
I think std.socket is somewhat heavily used and there isn't really that much wrong with it... even just hiding the docs I think will make people wtf a bit; this kind of basic functionality is expected at a glance.

but the others are pretty clear, outbuffer and signals as far as I know are extremely rarely used anyway, and everyone says xml sucks.

CyberShadow:
Closing for now as discussed in https://github.com/dlang/phobos/pull/5488.

PR #1713, opened 2017-06-15
Search for undefined macros

wilzbach:
Follow-up to https://github.com/dlang/dlang.org/pull/1694
I had to exclude the stable branches as the fixes are only on master yet (and if we ensure that master has no undefined macros, then stable should safe as well).

@CyberShadow what's the best way to run this on all repos? As post hook to DAutoTest?

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -615,10 +615,11 @@ $(PHOBOS_STABLE_FILES_GENERATED): $(PHOBOS_STABLE_DIR_GENERATED)/%: $(PHOBOS_STA
 # Style tests
 ################################################################################
 
-test: $(ASSERT_WRITELN_BIN)_test
+test: $(ASSERT_WRITELN_BIN)_test all
 	@echo "Searching for trailing whitespace"
-	@echo "Check for trailing whitespace"
-	grep -n '[[:blank:]]$$' $$(find . -type f -name "*.dd") ; test $$? -eq 1
+	@grep -n '[[:blank:]]$$' $$(find . -type f -name "*.dd") ; test $$? -eq 1
+	@echo "Searching for undefined macros"
+	@grep -n "UNDEFINED MACRO" $$(find $(DOC_OUTPUT_DIR) -type f -name "*.html" -not -path "$(DOC_OUTPUT_DIR)/phobos/*") ; test $$? -eq 1
 	@echo "Executing assert_writeln_magic tests"
 	$<
 


CyberShadow:
> what's the best way to run this on all repos?

Add support for running tests for the dlang.org repo to ae.sys.d.manager, and making DAutoTest do it. I'll do this.

We can disable Travis after that too.

posix.mak
```diff
@@ -617,8 +617,9 @@ $(PHOBOS_STABLE_FILES_GENERATED): $(PHOBOS_STABLE_DIR_GENERATED)/%: $(PHOBOS_STA
 
 test: $(ASSERT_WRITELN_BIN)_test
```
CyberShadow: Now that we scan the output, this needs a dependency on `web` target.

andralex:
Awesome work!

CyberShadow:
> Add support for running tests for the dlang.org repo to ae.sys.d.manager, and making DAutoTest do it.

This is now live, so I'm going to disable Travis. This PR works as it is, but formally it's still missing the `web` dependency.

posix.mak
```diff
@@ -617,8 +617,9 @@ $(PHOBOS_STABLE_FILES_GENERATED): $(PHOBOS_STABLE_DIR_GENERATED)/%: $(PHOBOS_STA
...
 	@echo "Searching for trailing whitespace"
-	@echo "Check for trailing whitespace"
-	grep -n '[[:blank:]]$$' $$(find . -type f -name "*.dd") ; test $$? -eq 1
+	@grep -n '[[:blank:]]$$' $$(find . -type f -name "*.dd") ; test $$? -eq 1
+	@echo "Searching for undefined macros"
+	@[ -d web ] && grep -n "UNDEFINED MACRO" $$(find web -type f -name "*.html" -not -path "web/phobos/*") ; test $$? -eq 1
```
CyberShadow: Use `DOC_OUTPUT_DIR` instead of hard-coding `web`?

posix.mak
CyberShadow: ping

CyberShadow: Sorry, the target is called `all`, not `web`

wilzbach:
> ping

Sorry - I was on a hackathon over the weekend and unfortunately time is rather precious at these events.

CyberShadow:
GitHub forced me to write something in order to submit a review :)

wilzbach:
> This is now live, so I'm going to disable Travis. This PR works as it is, but formally it's still missing the web dependency

Okay just to be sure, let's have a more closer look at DAutoTest over the next day..

PR #1715, opened 2017-06-16
Frontpage: D foundation -> D Language Foundation

wilzbach:


wilzbach:
@andralex - how about giving merge/review rights for dlang.org to @ZombineDev, so that his reviews count? :)
We are currently really low-staffed on reviewers at dlang.org ...

andralex:
@wilzbach great idea stay tuned

PR #1718, opened 2017-06-16
[TRIVIAL] fix indefinite article "an" before uint/ubyte/ushort etc

BBasile:


aG0aep6G:
The commit message doesn't make sense. It says: "fix indefinite article "an" before uint/ubyte/ushort etc", but you're fixing it for "enum", not for "uint" etc.

PR #1722, opened 2017-06-18
Fix Issue 17510 - How could 509 be a byte value?

CyberShadow:


spec/lex.dd
```diff
@@ -574,8 +574,8 @@ $(H4 Escape Sequences)
         specified as two hexadecimal digits.$(BR)For example: $(D \xFF)
         represents the character with the value 255.)
     $(TROW $(D \)$(I n)$(BR)$(D \)$(I nn)$(BR)$(D \)$(I nnn), Byte value in
-        octal.$(BR)For example: $(D \775) represents the character with the
-        value 509.)
+        octal.$(BR)For example: $(D \101) represents the character with the
+        value 65 ($(D 'A')).)
```
ZombineDev: Maybe you should also mention that the maximum value is `\377`, just for clarity.

wilzbach:
> Maybe you should also mention that the maximum value is \377, just for clarity
(https://github.com/dlang/dmd/blob/v2.074.1/src/ddmd/lexer.d#L1233). The hex version already implies that, because obviously the largest representable number made of two hex digits is 255 - i.e. there's no room for error.

@CyberShadow I took the liberty to add this to your PR, s.t. we can move forward. Feel free to edit or drop the commit if you don't like it.

CyberShadow:
LGTM :)

PR #1727, opened 2017-06-21
Ignore $(RUNNABLE_EXAMPLE) for eBook & LaTeX build

wilzbach:
When I added the `$(RUNNABLE_EXAMPLE)` macro to `dlang.org.ddoc` I falsely assumed that this macro wouldn't be part of the eBook / LaTeX build.

Btw when I had the `dlangspec` open, I took the opportunity to skim through it and check for obvious layout errors.


.gitignore
```diff
@@ -25,6 +25,7 @@ dlangspec.log
 dlangspec.out
 dlangspec.tex
 dlangspec.mobi
+dlangspec.toc
```
CyberShadow: What creates this file? It's not mentioned in the Makefile.

PR #1753, opened 2017-06-21
[run-spec] Allow examples of the D spec to be runnable: spec_version

wilzbach:

PR #1754, opened 2017-06-21
Allow examples of the D spec to be runnable

wilzbach:
A combined PR instead of [the wave](https://github.com/dlang/dlang.org/pulls?utf8=%E2%9C%93&q=run-spec).

For convenience copy/pasted comment from https://github.com/dlang/dlang.org/pull/1753:

In general I was a quite liberate with adding `$(RUNNABLE_EXAMPLE)` as even playing with the compile error should be helpful to a reader.
In some cases
- I changed `test` -> `main` to allow execution
- As it's available as book, I try to add `import std.stdio` where it was missing (though it doesn't matter for dlang.org as `std.stdio` is always imported)
- Sometimes e.g. for unittests I added `$(RUNNABLE_EXAMPLE_ARGS ...)` - that's "new" way to clarify  arguments (there's also $(RUNNABLE_EXAMPLE_STDIN)`
- This isn't intended to be complete run, but I hope a good start with ~90% of the ones that are interesting to be runnable

### DIFF
--- a/spec/arrays.dd
+++ b/spec/arrays.dd
@@ -884,10 +884,12 @@ printf("the string is '%.*s'\n", str.length, str.ptr);
         $(P The best way is to use std.stdio.writefln, which can handle
         D strings:)
 
+$(RUNNABLE_EXAMPLE
 ---------
 import std.stdio;
 writefln("the string is '%s'", str);
 ---------
+)
 
 $(H3 $(LNAME2 void_arrays, Void Arrays))
 
@@ -906,6 +908,7 @@ $(H3 $(LNAME2 void_arrays, Void Arrays))
     and it is an error to convert to an array type whose element size does not
     evenly divide the length of the void array.)
 
+$(RUNNABLE_EXAMPLE
 ---------
 void main()
 {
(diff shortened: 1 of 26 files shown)


CyberShadow:
>  [run-spec] Allow examples of the D spec to be runnable: posix 

This looks like an accidental revert of an earlier PR.

CyberShadow:
> [run-spec] Allow examples of the D spec to be runnable: spec_arrays

Last I checked, _ was not a directory separator :)

spec/arrays.dd
```diff
@@ -884,10 +884,12 @@ printf("the string is '%.*s'\n", str.length, str.ptr);
         $(P The best way is to use std.stdio.writefln, which can handle
         D strings:)
 
+$(RUNNABLE_EXAMPLE
 ---------
 import std.stdio;
 writefln("the string is '%s'", str);
 ---------
+)
```
CyberShadow: This one doesn't work. (For one, a statement outside a function, for two, referencing a variable declared in another block.)

I thought you used a script to check which example works and enable it only for those? Or did you choose examples manually?

wilzbach:
Closing in favor of https://github.com/dlang/dlang.org/pull/1998

PR #1755, opened 2017-06-21
Add example from the DLang Tour to the example roulette

wilzbach:
Full credit for this example goes to @JackStouffer https://github.com/dlang-tour/english/pull/137 - I merely reformatted it and added `@nogc` to prove the point ;-)

While adding the example, I realized that the existing roulette mechanism was broken.

CC @CyberShadow @ZombineDev

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -161,6 +161,26 @@ void main()
 }
 ----
 )
+$(EXTRA_EXAMPLE_RUNNABLE Sort in-place across multiple arrays,
+----
+void main()
+{
+    import std.stdio : writefln;
+    import std.algorithm.sorting : sort;
+    import std.range : chain;
+
+    int[] arr1 = [4, 9, 7];
+    int[] arr2 = [5, 2, 1, 10];
+    int[] arr3 = [6, 8, 3];
+    // @nogc functions are guaranteed by the compiler
+    // to be without any GC allocation
+    () @nogc {
+        sort(chain(arr1, arr2, arr3));
+    }();
+    writefln("%s\n%s\n%s\n", arr1, arr2, arr3);
+}
+----
+)
 
 ) $(COMMENT your-code-here)
 ))) $(COMMENT intro, div, div)


index.dd
```diff
@@ -53,6 +53,25 @@ void main()
 )
 )
 $(EXTRA_EXAMPLE
+----
+void main()
```
CyberShadow: Needs a short comment at the top describing what the program is doing.

CyberShadow:
Would be nice to avoid the example switching after the page loads.

Could be done by forcing the JavaScript to print a random example as the page loads (using `document.write`), and having a fallback in a `<noscript>` tag.

wilzbach:
> Needs a short comment at the top describing what the program is doing.

Ok. Added.

> Would be nice to avoid the example switching after the page loads.
> Could be done by forcing the JavaScript to print a random example as the page loads (using document.write), and having a fallback in a <noscript> tag.

Yeah I don't like the current solution either. Will make a PR with something smarter ;-)

CyberShadow:
This example would be a good deal more exciting if all of main was `@nogc`... as it is, the lambda literal is going to be incomprehensible to most people not familiar with the language.

Perhaps it's better to just wrap the literal and only leave the comment.

@andralex What do you think? I think you would like to review all new non-trivial front page changes anyway.

ZombineDev:
I think it would be nice to have a drop down, like on the [golang page](https://golang.org/). I remember back some time ago when I wanted to show one particular example to a friend that I had to reload the page a couple of times to get to it.

andralex:
The example is fine. Would be great to have a little button somewhere "Another" that loads another random example. Also examples should have nice titles such as "Sort", "Find min and max", etc.

index.dd
andralex: Comment should be before main - perhaps even a ddoc comment.

index.dd
```diff
@@ -53,6 +53,28 @@ void main()
...
+    import std.stdio : writefln;
+    import std.algorithm.sorting : sort;
+    import std.range : chain;
+
+    int[] arr1 = [4, 9, 7];
+    int[] arr2 = [5, 2, 1, 10];
+    int[] arr3 = [6, 8, 3];
+    // @nogc functions are guaranteed by the compiler
+    // to be without any GC allocation
+    () @nogc {
```
andralex: nogc would indeed be best in main

CyberShadow:
> Would be great to have a little button somewhere "Another" that loads another random example.

https://github.com/dlang/dlang.org/pull/1777

index.dd
wilzbach: Yep, of course, but this depends on `writefln` (or similar) to be `@nogc` ...
And the [`assumeNoGC`](https://p0nce.github.io/d-idioms/#Bypassing-@nogc) hack seemed to ugly for a frontpage example.

wilzbach:
> Would be great to have a little button somewhere "Another" that loads another random example.

As #1777 has been merged & deployed, I rebased this PR.

> Also examples should have nice titles such as "Sort", "Find min and max", etc.

#1777 implies this ;-)

wilzbach:
Ping @CyberShadow - I think we are all set here :)

wilzbach:
> Ping @CyberShadow - I think we are all set here :)

Rebased -> should be good to go (it has even been approved by Andrei).

CyberShadow:
Looks like you need to rebase.

wilzbach:
> Looks like you need to rebase.

That was weird - rebase worked without interaction on my machine ...

PR #1756, opened 2017-06-21
Add hashmap + D arrays example

wilzbach:
From @CyberShadow's [comment](https://github.com/dlang/dlang.org/pull/1577#issuecomment-282944600):

> So, not sure about this example's illustrative powers. D arrays are awesome too!

I really liked the example and imho it makes a good fit for the frontpage.

(this temporarily includes the first commit from https://github.com/dlang/dlang.org/pull/1755 to fix the roulette picker, s.t. it can be previewed)

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -181,6 +181,25 @@ void main()
 }
 ----
 )
+$(EXTRA_EXAMPLE_RUNNABLE Count frequencies of all 2-tuples,
+----
+void main()
+{
+    import std.stdio : writefln;
+    int[char[2]] aa;
+    auto arr = "ABBBA";
+
+    // Iterate over all pairs in the string and observe each pair
+    // ('A', 'B'), ('B', 'B'), ('B', 'A'), ...
+    // String slicing doesn't allocate a copy
+    foreach (i; 0 .. arr.length - 1)
+        aa[arr[i .. $][0 .. 2]]++;
+
+    foreach (key, value; aa)
+        writefln("key: %s, value: %d", key, value);
+}
+----
+)
 
 ) $(COMMENT your-code-here)
 ))) $(COMMENT intro, div, div)


index.dd
```diff
@@ -53,6 +53,28 @@ void main()
 )
 )
 $(EXTRA_EXAMPLE
+----
+void main()
```
CyberShadow: Needs a comment at the top explaining what the program does.
"Counts frequencies of adjacent pairs of DNA bases" (though I'm fairly sure I got the terminology wrong).

index.dd
```diff
@@ -53,6 +53,28 @@ void main()
...
+{
+    import std.stdio : writefln;
+    int[char[2]] aa;
+
+    // The string `arr` has a limited alphabet: {A, C, G, T}
+    // Thus, iteration can be done _without_ decoding
```
CyberShadow: I don't think we should bring up auto-decoding in a front page example.

index.dd
```diff
@@ -53,6 +53,28 @@ void main()
...
+    auto arr = "AGATAGA";
+
+    // iterate over all pairs in the string and observe each pair
+    // ('A', 'G'), ('G', 'A'), ('A', 'T'), ...
+    foreach (i; 0 .. arr.length - 1)
+        aa[arr[i .. $][0 .. 2]]++;
```
CyberShadow: Might be worth mentioning/implying that slicing a string does not allocate a copy

index.dd
```diff
@@ -53,6 +53,28 @@ void main()
...
+
+    // iterate over all key/value pairs of the Associative Array
```
CyberShadow: This comment seems redundant.

wilzbach:
@ others: what do you feel about this one? Is showing cool slicing with nice iteration a good example?

image

wilzbach:
> @ others: what do you feel about this one? Is showing cool slicing with nice iteration a good example?

There was a long discussion on the NG one months ago with the gist that the current examples are too advanced and that Python.org shows simple things like [list comprehensions on their front page](https://www.python.org/).

index.dd
```diff
@@ -142,6 +142,25 @@ void main()
...
+{
+    import std.stdio : writefln;
+    int[char[2]] aa;
+    auto arr = "ABBBA";
+
+    // Iterate over all pairs in the string and observe each pair
+    // ('A', 'B'), ('B', 'B'), ('B', 'A'), ...
+    // String slicing doesn't allocate a copy
+    foreach (i; 0 .. arr.length - 1)
+        aa[arr[i .. $][0 .. 2]]++;
```
ZombineDev: The above looks cool, but can be simplified to:
```D
aa[arr[i .. i + 2]]++;
```

wilzbach: I think in this case it's necessary:

```
onlineapp.d(11): Error: cannot implicitly convert expression `arr[i..i + 2LU]` of type `string` to `char[2]`
```

PR #1757, opened 2017-06-21
Fix JavaScript example roulette

wilzbach:
As discussed [here](https://github.com/dlang/dlang.org/pull/1755#issuecomment-310029805), to avoid page reflows, we should pick an example ASAP and render its HTML.

Unfortunately this required quite a bit of work as the original code heavily used jQuery for which we don't want to wait until it's loaded. I am not really happy with it, but I guess that's the necessary evil if we want to avoid the ugly page redraws (and don't have a server to do the selection).

> I think it would be nice to have a drop down, like on the golang page. I remember back some time ago when I wanted to show one particular example to a friend that I had to reload the page a couple of times to get to it.

That was fairly trivial to add and requiring a title for an example and reduces the manual checking from @CyberShadow ;-)

The example selector still needs styling, hence I am marking this as "Needs work" for now.

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -28,8 +28,7 @@ $(DIVC intro, $(DIV, $(DIV,
                 $(P Upon approval it will be showcased here on a random schedule.)
             )
         )
-        $(DIVID your-code-here-default,
-$(RUNNABLE_EXAMPLE
+$(EXTRA_EXAMPLE
 $(RUNNABLE_EXAMPLE_STDIN
 The D programming language
 Modern convenience.
@@ -98,8 +97,14 @@ void main()
 ----
 )
 
+) $(COMMENT your-code-here)
+))) $(COMMENT intro, div, div)
+
+$(COMMENT
+Picking the frontpage example after the page has loaded leads to reflows
+and re-rendering. Hence, we pick the example as soon as possible
 )
-)))
+<script src="$(ROOT_DIR)js/example_roulette.js"></script>
 
 $(DIVC more,
 
--- a/js/example_roulette.js
+++ b/js/example_roulette.js
@@ -1,0 +1,116 @@
+/**
+Picking the frontpage example after the page has loaded leads to reflows
+and re-rendering. Hence, we pick the example as soon as possible
+
+Copyright: 1999-2017 by D Language Foundation
+License:   http://boost.org/LICENSE_1_0.txt, Boost License 1.0
+*/
+
+/**
+Prepares the DOM layout of a runnable code example.
+Sibling nodes for program arguments (.runnable-examples-args)
+and stdin (.runnable-examples-stdin) are detected.
+The parent node has usually the class runnable-examples.
+pre = element containing the code example
+*/
+function buildRunnableExampleLayout(pre) {
+    var parent = pre.parentNode;
+
+    // helper method to construct a button in vanilla JavaScript
+    function createButton(className, value) {
+        var btn = document.createElement("input");
+        btn.type = "button";
+        btn.className = className;
+        btn.value = value;
+        return btn;
+    }
+
+    // helper method to construct an overlay in vanilla JavaScript
+    // typically overlays are output, args, stdin and the editor instance
+    function createOverlay(className, title, type, value) {
+        var el = document.createElement("div");
+        el.className = className;
+        var titleEl = document.createElement("span");
+        titleEl.className = "d_code_title";
+        titleEl.innerText = title;
+        el.appendChild(titleEl);
+        el.appendChild(document.createElement("br"));
+        var inner = document.createElement(type);
+        inner.className = className;
+        inner.value = value;
+        el.appendChild(inner);
+        return el;
+    }
+
+    // check whether args has been defined (RUNNABLE_EXAMPLE_ARGS macro in Ddoc)
+    var args = findFirst(parent.children, function(e){ return e.className === "runnable-examples-args";});
+    if (typeof args !== "undefined") { args = args.innerText; }
+
+    // check whether stdin has been defined (RUNNABLE_EXAMPLE_STDIN macro in Ddoc)
+    var stdin = findFirst(parent.children, function(e){ return e.className === "runnable-examples-stdin";});
+    if (typeof stdin !== "undefined") { stdin = stdin.innerText; }
+
+    // wraps the <pre> code box in a <div> element
+    // this rewrapping is needed for the editor framework
+    var dCode = document.createElement("div");
+    dCode.className = "d_code";
+    dCode.append(pre);
+
+    // create the CodeMirror editor overlay
+    var el = document.createElement("div");
+    el.style.display = "block";
+    el.className = "d_run_code";
+    // code box (CodeMirror will be loaded into this element)
+    var dCodeText = document.createElement("textarea");
+    dCodeText.className = "d_code";
+    dCodeText.style.display = "none";
+    el.appendChild(dCodeText);
+
+    // create optional stdin overlay
+    if (stdin) { el.appendChild(createOverlay("d_code_stdin", "Input", "textarea", stdin)); }
+
+    // create optional args overlay
+    if (args) { el.appendChild(createOverlay("d_code_stdin", "Command line arguments", "textarea", args)); }
+
+    // create main output overlay
+    el.appendChild(createOverlay("d_code_output", "Application output", "pre", "Running..."));
+
+    // user interaction buttons
+    el.appendChild(createButton("editButton", "Edit"));
+    if (stdin) { el.appendChild(createButton("inputButton", "Input")); }
+    if (args) { el.appendChild(createButton("argsButton", "Args")); }
+    el.appendChild(createButton("runButton", "Run"));
+    el.appendChild(createButton("resetButton", "Reset"));
+
+    // replace the <pre> element with its enhanced wrapper div
+    var root = document.createElement("div");
+    root.appendChild(dCode);
+    root.appendChild(el);
+    parent.appendChild(root);
+    return root;
+}
+/**
+Searches any object for the first hit.
+Params:
+    el = object to filter
+    fn = function to be applied as filter
+Returns: first hit, `undefined` otherwise
+*/
+function findFirst(el, fn) {
+    return Array.prototype.filter.call(el, fn)[0];
+}
+
+(function() {
+// randomly pick one example and bootstrap the runnable editor
+var examples = document.getElementsByClassName('your-code-here-extra');
+var rouletteIndex = Math.floor(Math.random() * examples.length);
+var rouletteChild = examples[rouletteIndex];
+
+// step 1: select the .runnable-example <div>
+var rouletteChildNode = findFirst(rouletteChild.children, function(e) { return e.className.indexOf("runnable-example") != -1; });
+// step 2: select the first <pre> element (of the .runnable-example <div>)
+var el = findFirst(rouletteChildNode.children, function(e) { return e.nodeName == "PRE";});
+// step 3: Construct runnable example layout within the <pre> element
+buildRunnableExampleLayout(el);
+rouletteChild.style.display = "block";
+})();
--- a/js/dlang.js
+++ b/js/dlang.js
@@ -58,14 +58,6 @@
             $('#search-query input').focus();
         });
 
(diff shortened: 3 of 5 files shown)


index.dd
```diff
@@ -100,6 +96,148 @@ void main()
 
 )
 )))
+$(SCRIPT
+// Picking the frontpage example after the page has loaded leads to reflows
+// and rerendering. Hence, we pick the example as soon as possible
+/**
+Bootstrap a parent node containing a <pre> element into an editor instance.
```
CyberShadow: What does "bootstrap" mean here?

index.dd
```diff
@@ -100,6 +96,148 @@ void main()
...
+    // helper method to construct a button in vanilla JavaScript
+    function createButton(className, value) {
+        var btn = document.createElement("input");
+        btn.type = "button";
+        btn.className = className;
+        btn.value = value;
+        return btn;
+    }
+    // helper method to construct a textarea in vanilla JavaScript
+    function createTextArea(className, value, text) {
```
CyberShadow: This helper function is used only once, inline?

index.dd
```diff
@@ -100,6 +96,148 @@ void main()
...
+        var el = document.createElement("textarea");
+        el.className = className;
+        el.innerText = text;
+        return el;
+    }
+    // helper method to construct an editor overlay in vanilla JavaScript
+    // typically overlays are output,args,stdin and the editor instance
```
CyberShadow: What does "overlay" mean here?

index.dd
```diff
@@ -100,6 +96,148 @@ void main()
...
+        titleEl.innerText = title;
+        el.appendChild(titleEl);
+        el.appendChild(document.createElement("br"));
+        var inner = document.createElement(type);
+        inner.className = className;
+        inner.value = value;
+        el.appendChild(inner);
+        return el;
+    }
+    // JavaScript filter for humans
```
CyberShadow: Pretty sure this function is not used for filtering humans. :P

CyberShadow:
Good stuff. I hadn't thought of actually putting the code inline, but this approach does have merits. I wonder though if it's possible to reduce code duplication (if there is any) and the number of things that are in index.dd, e.g. by putting the JavaScript in a separate file and including it as a DDoc macro (e.g. js/index-examples.js.ddoc).

It's not immediately obvious from the code (and I haven't looked at it in depth) whether this is what it already does, but doing just the bare minimum to show the example would be the goal for the inline JavaScript code. Showing the buttons and getting syntax highlighting for the editor working can be a task for the rest of the JavaScript code that's set up. The inline code's goal is to just do the bare minimum for the layout (ideally even avoid a re-layout later) and hand things over to the main code once it's loaded.

wilzbach:
>  I wonder though if it's possible to reduce code duplication (if there is any)

The first commit is a more or less literal translation of the code in `js/run.js`.
Sadly the only comment in there is from me (when I added an improvement to it).
Hence, some of the comments I added for better understanding might have a confusing terminology.

> and the number of things that are in index.dd, e.g. by putting the JavaScript in a separate file and including it as a DDoc macro (e.g. js/index-examples.js.ddoc).

I thought about this as well, but I wasn't sure whether it's worth the extra step.
OTOH this would allow us to minify/uglify the inline JavaScript.

> It's not immediately obvious from the code (and I haven't looked at it in depth) whether this is what it already does, but doing just the bare minimum to show the example would be the goal for the inline JavaScript code.

Yes, that was my goal.

> Showing the buttons and getting syntax highlighting for the editor working can be a task for the rest of the JavaScript code that's set up.

You need the buttons - they will trigger a reflow as well ;-)
The code just creates an empty `textarea` into which the existing `setupTextarea` injects the CodeMirror editor.

> The inline code's goal is to just do the bare minimum for the layout (ideally even avoid a re-layout later) and hand things over to the main code once it's loaded.

Yes it does exactly this - I basically moved the entire layout setup function inline.

index.dd
wilzbach: Everything that "lays over" the code box, e.g.

image
image
image

As mentioned, there weren't any comments, so I have chosen my own terminology. Better ideas?

wilzbach:
> The example selector still needs styling, hence I am marking this as "Needs work" for now.

Done. I think this is styling should get us started:

image

>  e.g. by putting the JavaScript in a separate file and including it as a DDoc macro (e.g. js/index-examples.js.ddoc).

Done. I also tried to clean the script a bit further. For example, only the `pre` elementNode (i.e. the one with the code) is passed in and touched now.

CyberShadow:
Looks nice :) @andralex What do you think?

CyberShadow:
> You need the buttons - they will trigger a reflow as well ;-)

Why? Just reserve space for them.

I don't understand why the code needs to do anything beyond displaying one of the divs that are hidden by default.

index.dd
```diff
@@ -454,7 +455,11 @@ Macros:
     TAG2=<$1 $2>$3</$1>
     D=<span class="d_inlinecode">$0</span>
     EXAMPLE=$(TAG2 a, id="a$1-control" class="example-control", )$(TAG2 div, id="a$1" class="example-box", $(RUNNABLE_EXAMPLE $2))
-    EXTRA_EXAMPLE=<div class="your-code-here-extra" style="display:none">$(RUNNABLE_EXAMPLE $0)</div>
+    EXTRA_EXAMPLE=<div class="your-code-here-extra" style="display:none">
```
ZombineDev: EXAMPLE_ROULETTE_ENTRY sounds better to me.

index-examples.js.ddoc
```diff
@@ -0,0 +1,139 @@
+_=
+$(COMMENT
+Picking the frontpage example after the page has loaded leads to reflows
+and re-rendering. Hence, we pick the example as soon as possible
+Be careful, Ddoc forbids certain patterns, e.g.:
+- using < or >
+- direct assignments (foo = bar) are interpreted as new Ddoc variables
+)
```
ZombineDev: What prevents adding this as a plain js file under e.g. `js/example_roulette.js`? Can ddoc string-import-mixin other files?

ZombineDev:
Overall I really like the improvements in this PR, especially the added `<select>` box!
As an improvement,  I would like to propose creating a `homepage-examples` folder so we can add plain `*.d` files inside it. This should make it easier for people not well versed with ddoc to contribute and it should be easier to maintain. golang also does it like that: https://golang.org/doc/play/, however the list of examples is hard-coded and they have the first example in-line (I guess in order to improve the page load time / prevent reflows), see https://github.com/golang/go/blob/master/doc/root.html#L35.

ZombineDev:
Sorry, for the close - wrong button :(

wilzbach:
> > Is loading an external file via <script src="${DLANG_SITE}/js/example_roulette.js" type="text/javascript"></script> not an option?

I thought about this for a while and while this approach has the disadvantage of potentially being a bit slower, it allows to use a proper JavaScript file (syntax-highlighting) and avoids the constant fights against Ddoc when writing inline JS.
Moreover, browsers do peek through the entire DOM and look for source tags while downloading the found ones in parallel.

The simulated page load waterfall for "Fast 3G" shows this nicely:

image

Of course the simulation, isn't perfect, but it does show that a lot of other stuff needs to be download before the page can be initially drown. DOMContentLoaded (the blue line) is at 4.13s

> Or alternatively, mix the js script in, after ddoc processing, as part of the build.

This would add another step to the build process and currently @CyberShadow is opposed to this (see e.g. https://github.com/dlang/dlang.org/pull/1688#issuecomment-307409033).

wilzbach:
> As an improvement, I would like to propose creating a homepage-examples folder so we can add plain *.d files inside it. This should make it easier for people not well versed with ddoc to contribute and it should be easier to maintain.

Excellent idea! But let's do this in another PR. This one is already big enough.

js/example_roulette.js
```diff
@@ -0,0 +1,139 @@
...
+    function createOverlay(className, title, type, value) {
+        var el = document.createElement("div");
+        el.className = className;
+        var titleEl = document.createElement("span");
+        titleEl.className = "d_code_title";
+        titleEl.innerText = title;
+        el.appendChild(titleEl);
+        el.appendChild(document.createElement("br"));
+        var inner = document.createElement(type);
+        inner.className = className;
```
ZombineDev: So, the returned element and the inner "type" element have same class. Is that on purpose and why?

wilzbach: Yes, because I tried to make a 1:1 copy of the jQuery version, see e.g. https://github.com/dlang/dlang.org/blob/master/js/run.js#L138

```js
stdin = '<div class="d_code_stdin"><span class="d_code_title">Standard input</span><br>'
        + '<textarea class="d_code_stdin">'+stdin+'</textarea></div>';
```

Imho "translation" and refactoring shouldn't be mixed (even though the temptation is very strong).

CyberShadow:
OK, so, as I understand, this PR does three things right now:

1. Fix the JS roulette
2. Select an example during page load instead of after, to avoid a flash of the default noscript example
3. Add a dropdown to select an example

IMO these should be separate PRs, or at least commits. Also I still have a strong suspicion that the second one could be much simpler as [mentioned above](https://github.com/dlang/dlang.org/pull/1757#issuecomment-310267525).

ZombineDev:
@wilzbach something in this PR broke somehow the layout, see:

Before:
before

After:
after

(That's with Chrome 58 on Windows 10.)

Last time I tried the DAutoTest preview (I think it was yesterday) I didn't notice any layout problems.

wilzbach:
> Fix the JS roulette

-> https://github.com/dlang/dlang.org/pull/1767

> Select an example during page load instead of after, to avoid a flash of the default noscript example
> Add a dropdown to select an example

Hehe. I initially had separate commits, but then squashed because it easier due to the several steps (inline Ddoc -> Ddoc as separate file -> separate Js).

I for now simply dropped the commit.
I will do this as follow-up

> @wilzbach something in this PR somehow broke the layout, see:

Yeah, I realized this myself. Already fixed. Sorry.

js/example_roulette.js
```diff
@@ -0,0 +1,116 @@
...
+var examples = document.getElementsByClassName('your-code-here-extra');
+var rouletteIndex = Math.floor(Math.random() * examples.length);
+var rouletteChild = examples[rouletteIndex];
+
+// step 1: select the .runnable-example <div>
+var rouletteChildNode = findFirst(rouletteChild.children, function(e) { return e.className.indexOf("runnable-example") != -1; });
+// step 2: select the first <pre> element (of the .runnable-example <div>)
+var el = findFirst(rouletteChildNode.children, function(e) { return e.nodeName == "PRE";});
+// step 3: Construct runnable example layout within the <pre> element
+buildRunnableExampleLayout(el);
```
wilzbach: @CyberShadow 
> Also I still have a strong suspicion that the second one could be much simpler as mentioned above.

Yes, these lines without `buildRunnableExampleLayout` would - in theory - be enough. However, it still "jumps" due to the buttons added and just
- adding the buttons isn't a good solution imho, because (1) it creates code entanglement (at another random, JS place the script needs to be aware of this) and (2) it creates code redundancy and makes it a lot harder if someone wants to change the overall layout
- adding sth. `el.style.marginBottom = "20px"` is a similar hack (and still flashes/"jumps")

wilzbach:
Rebased.

@CyberShadow let's come to a conclusion here: are you okay with this now or do you want me to come up with an ugly hack that "fakes" the bottoms with sth. like `marginBottom`? [As explained](https://github.com/dlang/dlang.org/pull/1757#discussion_r123895338) I would prefer to have one source of truth for the layout generation.

CyberShadow:
Less code is better. There is way too much code with a bus factor of 1 on dlang.org already. As far as I can see, index.dd just needs like 5 lines of JavaScript, at most. If you don't feel like doing that, I can put it on my list.

wilzbach:
> Less code is better. There is way too much code with a bus factor of 1 on dlang.org already. As far as I can see, index.dd just needs like 5 lines of JavaScript, at most. If you don't feel like doing that, I can put it on my list.

Opened another PR with this approach

https://github.com/dlang/dlang.org/pull/1775

As mentioned the redraw flashes are still partially visible ...

wilzbach:
Closing this with regrets as it wasn't trivial to refactor the jQuery code, but in the end the smaller diffs won.

PR #1758, opened 2017-06-21
Add example; Compile-Time Sort in D

wilzbach:
Shamelessly adapted from http://dlang.org/blog/2017/06/05/compile-time-sort-in-d/

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -94,6 +94,23 @@ void main()
 }
 ----
 )
+$(EXTRA_EXAMPLE Sort an Array at Compile-Time,
+----
+void main()
+{
+    import std.algorithm, std.conv, std.stdio;
+
+    "Starting program".writeln;
+
+    // Sort a constant declaration at Compile-Time
+    enum a = [ 3, 1, 2, 4, 0 ];
+    static immutable b = sort(a);
+
+    // Print the result _during_ compilation
+    pragma(msg, text("Finished compilation: ", b));
+}
+----
+)
 
 ) $(COMMENT your-code-here)
 ))) $(COMMENT intro, div, div)


index.dd
```diff
@@ -76,6 +76,21 @@ void main()
 }
 ----
 )
+$(EXTRA_EXAMPLE
+----
+// Compile-Time Sort in D
+void main() {
```
CyberShadow: Brace style?

wilzbach: That's copied from the blog article ;-)
(But yeah will fix that soon.)

index.dd
```diff
@@ -76,6 +76,22 @@ void main()
...
 ----
 )
+$(EXTRA_EXAMPLE
+----
+// Compile-Time Sort in D
+void main()
+{
+    import std.algorithm, std.stdio;
+    writeln("Starting program.");
+    // Constant declaration
```
quickfur: What about adding some blank lines to make the code "breathe" better? That, or move the comments to EOL comments instead of occupying a single line each?

wilzbach: I think we can't really go with EOL comments because they might not be directly visible on mobile devices,  but sure a new line doesn't hurt :)

CyberShadow:
This looks ready, shall we merge it?

index.dd
```diff
@@ -94,6 +94,23 @@ void main()
 }
 ----
 )
+$(EXTRA_EXAMPLE Compile-Time Sort in D,
+----
+void main()
+{
+    import std.algorithm, std.stdio;
+
+    "Starting program".writeln;
```
CyberShadow: `writeln("Starting program");` I guess as per Andrei's yesterday comment

index.dd
```diff
@@ -94,6 +94,23 @@ void main()
 }
 ----
 )
+$(EXTRA_EXAMPLE Compile-Time Sort in D,
```
CyberShadow: " in D" seems a little redundant here :)

wilzbach:
> `writeln("Starting program");` I guess as per Andrei's yesterday comment

Do you mean this one: https://github.com/dlang/dlang.org/pull/1761#issuecomment-312312742?
AFAICT he just said that people aren't used to new, "fancy syntax" yet (in regard to coding styles and practices), but isn't "fancy syntax" exactly what we should advertise on dlang.org?

CyberShadow:
Now that you mentioned it, I'm not sure.

Anyway, I don't think it really matters.

wilzbach:
> Anyway, I don't think it really matters.

Same here, so we are good to go here? :)

CyberShadow:
Yep, was waiting for CI.

PR #1759, opened 2017-06-22
[Idea] Add subtyping example to frontpage

wilzbach:
Shamelessly adapted from https://tour.dlang.org/tour/en/gems/subtyping

Here the Allman brace style really costs a lot of vertical space ...

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -240,6 +240,29 @@ void main()
 }
 ----
 )
+$(EXTRA_EXAMPLE_RUNNABLE Subtyping with alias this,
+----
+struct Point
+{
+    private double[2] p;
+    // Forward all undefined symbols to p
+    alias p this;
+    double dot(Point rhs)
+    {
+        return p[0] * rhs.p[0] + p[1] * rhs.p[1];
+    }
+}
+void main()
+{
+    import std.stdio : writeln;
+    // Point behaves like a `double[2]` ...
+    Point p1, p2; p1 = [2, 1], p2 = [1, 1];
+    assert(p1[$ - 1] == 1);
+    // ... but with extended functionality
+    writeln("p1 dot p2 = ", p1.dot(p2));
+}
+----
+)
 
 ) $(COMMENT your-code-here)
 ))) $(COMMENT intro, div, div)


index.dd
```diff
@@ -76,6 +76,32 @@ void main()
...
+    {
+        return p[0] * rhs.p[0] + p[1] * rhs.p[1];
+    }
+}
+
+void main()
+{
+    import std.stdio : writeln;
+    Point p1, p2;
+    // we're basically talking with double[]
```
quickfur: Correction: `double[2]`.

index.dd
```diff
@@ -76,6 +76,32 @@ void main()
 }
 ----
 )
+$(EXTRA_EXAMPLE
+----
+// subtyping with alias this
+struct Point
+{
+    private double[2] p;
+    // p is used by the compiler to lookup unknown symbols
```
quickfur: Unknown symbols and operators? Just to clarify, since a newcomer may not know how looking up unknown symbols allows assigning arrays to `Point`.

wilzbach:
> Unknown symbols and operators? Just to clarify, since a newcomer may not know how looking up unknown symbols allows assigning arrays to Point.

@quickfur space comes at a premium on the front page. This clarification would require yet another line and I think we are at the outer limit here:

image

Maybe someone has an idea how to simply this one or are we good to go here?

index.dd
```diff
@@ -240,6 +240,30 @@ void main()
 }
 ----
 )
+$(EXTRA_EXAMPLE_RUNNABLE Subtyping with alias this,
+----
+struct Point
+{
+    private double[2] p;
+    // If Point.<function> isn't defined, use p.<function>
```
JinShil: I think the `Point.<function>` looks like some special syntax, and users students might be wondering what such syntax means.  I suggest, instead, saying "Forward all undefined symbols to `p`"

wilzbach:
Okay rebased to master. So should we go with this?
CC @JackStouffer @MetaLang @ZombineDev

index.dd
```diff
@@ -240,6 +240,30 @@ void main()
...
+    {
+        return p[0] * rhs.p[0] + p[1] * rhs.p[1];
+    }
+}
+void main()
+{
+    import std.stdio : writeln;
+    Point p1, p2;
+    // Point behaves like a double[2]
+    p1 = [2, 1], p2 = [1, 1];
```
JinShil: To save vertical space you could try...

```D
// Point behaves like a double[2]
Point p1, p2; p1 = [2,1]; p2 = [1,1];
```

index.dd
```diff
@@ -240,6 +240,30 @@ void main()
...
+    assert(p1[$ - 1] == 1);
+    // but with extended functionality
```
JinShil: I would prefer ellipses to show continuity:

```D
// `Point` behaves like `double[2]`...
p1 = [2, 1], p2 = [1, 1];
assert(p1[$ - 1] == 1);
// ...but with extended functionality
```

And you can tell I prefer tick marks to disambiguate code from English even in the comments.

index.dd
wilzbach: I think that's a bit too dense, but I will try it.

index.dd
wilzbach: Hehe I had something like this in the beginning, but then changed it to this, because I thought it's clearer to people -> reverting :)

wilzbach:
So if we could use K&R brace style, it would look really good:

image

Otherwise:

image

For reference, this examples was taken from the DTour, then improved and we have already "backported" the improvements: https://tour.dlang.org/tour/en/gems/subtyping

PR #1761, opened 2017-06-22
Add example about std.process

wilzbach:
image

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -111,6 +111,18 @@ void main()
 }
 ----
 )
+$(EXTRA_EXAMPLE Invoke external programs,
+----
+void main()
+{
+    import std.exception, std.stdio, std.process;
+
+    auto result = ["whoami"].execute;
+    enforce(result.status == 0);
+    result.output.write;
+}
+----
+)
 
 ) $(COMMENT your-code-here)
 ))) $(COMMENT intro, div, div)


quickfur:
This example actually makes it look harder than it should be.  My favorite `std.process` example is `execute`, which automatically forks, execs a new process, redirect its output into a buffer, and returns said buffer. So you can just write:
````
auto result = execute(["echo", "hello", "world"]);
assert(result.output == "hello world\n");
````

**Comment by quickfur** [comment]
Of course, you can also check return codes and what-not, but this shows the basic functionality.

**Comment by ZombineDev** [comment]
+1 @quickfur. At most something like this:
```D
auto result = execute(["echo", "hello", "world"]);
enforce(result.output == "hello world\n");
enforce(result.status == 0);
```
We can add a more advanced example after we get an example selection box (#1777).

index.dd
```diff
@@ -96,6 +96,21 @@ void main()
...
+    import std.stdio, std.process;
+
+    auto result = ["echo", "hello", "world"].execute;
+    result.output.write; // hello world
+    writefln("All good?: %d, result.status); // 0
+}
+----
+)
+
+)
```
wilzbach: Not sure yet why this is needed. Will debug later / tomorrow.

index.dd
```diff
@@ -96,6 +96,21 @@ void main()
...
+$(EXTRA_EXAMPLE
+----
+// std.process simplifies invoking external programs
+void main()
+{
+    import std.stdio, std.process;
+
+    auto result = ["echo", "hello", "world"].execute;
+    result.output.write; // hello world
+    writefln("All good?: %d, result.status); // 0
```
CyberShadow: Unclosed string literal (`%d,` should be `%d",`)

Subjective but I'd put an enforce here instead of writeln :)

wilzbach:
> Subjective but I'd put an enforce here instead of writeln :)
> (and move the status check before reading output)

Sounds good to me. It just wanted to avoid an empty result page - that's a bit boring for a reader.

Btw what do you think about using D to generate the arguments?

```d
void main()
{
    import std.algorithm, std.conv, std.exception, std.range, std.stdio, std.process;

    auto result = "echo".only.chain(5.iota.map!(to!string)).array.execute;
    result.output.write; // 0 1 2 3 4
    enforce(result.status == 0);
}
```

Unfortunately it's not that elegant, so it's probably very confusing for newcomers?

wilzbach:
For convenience:

image

CyberShadow:
I think it's fine as it is. Maybe it would be worth it if someone thinks of a still concise but more interesting example than echo.

BTW, where are we stylistically regarding optional parens on the last function of a UFCS chain? I know we're officially not doing optional parens within the chain. With no parens the line doesn't look like it has a function call in it, not sure if that would be confusing or not.

wilzbach:
> Maybe it would be worth it if someone thinks of a still concise but more interesting example than echo.

How about `whoami` - we could leave the solution out and thus force people to play with the snippet to find out.
[Spoiler](https://tour.dlang.org/editor?source=void%20main()%0A%7B%0A%20%20%20%20import%20std.algorithm%2C%20std.conv%2C%20std.exception%2C%20std.range%2C%20std.stdio%2C%20std.process%3B%0A%0A%20%20%20%20auto%20result%20%3D%20%5B%22whoami%22%5D.execute%3B%0A%20%20%20%20result.output.write%3B%0A%7D)

> BTW, where are we stylistically regarding optional parens on the last function of a UFCS chain? I know we're officially not doing optional parens within the chain. With no parens the line doesn't look like it has a function call in it, not sure if that would be confusing or not.

Not sure. I have seen Andrei commenting on a couple of PRs in favor of avoid parenthesis for function calls with zero arguments (e.g. `foo()` -> `foo`), so I guess that's the semi-official style guideline. Should we make it official?

CyberShadow:
> Not sure. I have seen Andrei commenting on a couple of PRs in favor of avoid parenthesis for function calls with zero arguments (e.g. foo() -> foo), so I guess that's the semi-official style guideline. Should we make it official?

I think only @andralex can answer that :)

CyberShadow:
> I have seen Andrei commenting on a couple of PRs in favor of avoid parenthesis for function calls with zero arguments (e.g. foo() -> foo),

Were the function calls sub-expressions, or entire statements?

CyberShadow:
> How about whoami - we could leave the solution out and thus force people to play with the snippet to find out.

I think `whoami` is a great idea. It's even available on Windows, a rarity among shell commands!

andralex:
Placing parens after a function name is a sacred cow. It will take time to convince people to eat burgers.

wilzbach:
> I think `whoami` is a great idea

Rebased (due to the other changes).

> It's even available on Windows, a rarity among shell commands!

Wow.

index.dd
```diff
@@ -111,6 +111,18 @@ void main()
...
 )
+$(EXTRA_EXAMPLE Invoke external programs,
+----
+void main()
+{
+    import std.exception, std.stdio, std.process;
+
+    auto result = ["whoami"].execute;
+    enforce(result.status == 0);
+    result.output.write; // so, who are you?
```
CyberShadow: I think the comment is a bit over the top :) Someone non-technical might get confused or insulted (it happens).

wilzbach:
> result.output.write; // so, who are you?
> I think the comment is a bit over the top :) Someone non-technical might get confused or insulted (it happens).

Couldn't think of a good, neutral way, so I simply removed the comment.

wilzbach:
Ping @CyberShadow - I think we are all set here :)

PR #1762, opened 2017-06-22
Add example about code generation

wilzbach:
With Allman brace style this one got quite long, any ideas on trimming it down?

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -240,6 +240,32 @@ void main()
 }
 ----
 )
+$(EXTRA_EXAMPLE_RUNNABLE Generate a config parser at compile-time,
+----
+import std.algorithm, std.conv, std.range, std.stdio, std.traits;
+struct Config {
+    int a, b; string c;
+}
+void main() {
+    Config conf;
+    // use the generated parser for each entry
+    "b=1,a=2,c=hello".splitter(",").each!(e => conf.parse(e));
+    conf.writeln;
+}
+void parse(Conf)(ref Conf c, string entry) {
+    auto r = entry.splitter("=");
+    Switch: switch (r.front) {
+        foreach (idx, _; c.tupleof) {
+            case Conf.tupleof[idx].stringof:
+                c.tupleof[idx] = r.dropOne.front
+                    .to!(typeof(Conf.tupleof[idx]));
+                break Switch;
+        }
+        default: assert (0, "Unknown member name.");
+    }
+}
+----
+)
 
 ) $(COMMENT your-code-here)
 ))) $(COMMENT intro, div, div)


ZombineDev:
There's not need for mixins here. I think the following looks better:
```D
// Use compile-time introspection to generate
// a configuration parser.
import std.algorithm, std.conv, std.range, std.stdio,
       std.typecons : staticIota; // XXX: staticIota is undocumented and should be in std.meta

void main()
{
    Config conf;

    "b=1,a=2,c=hello".splitter(",").each!(e => conf.read(e));
    
    conf.writeln;
}

struct Config
{
    int a, b;
    string c;
}

void read(Conf)(ref Conf c, string entry)
{
    auto r = entry.splitter("=");
    
    switch (r.front)
    {
        default:
            assert (0, "Unknown member name.");
        foreach (idx; staticIota!(0, Conf.tupleof.length))
        {
            case Conf.tupleof[idx].stringof:
                c.tupleof[idx] = r.dropOne.front.to!(typeof(Conf.tupleof[idx]));
                break;
        }
    }   
}

```

CyberShadow:
> `                break;`

You need to break with a label there :) Otherwise it'll break the `foreach`, not `switch`. Common gotcha.

ZombineDev:
TBH, I wasn't sure, so I first wrote it with `break LSwitch`, but then I tested it without a didn't find a difference and used a plain `break;`. Still I'm not sure about the sematics of breaking out of unrolled loop - what if the break was guarded by a runtime check (that always passes) then it won't be possible to stop the loop unrolling and I would expect the runtime check + the `break;` to be present in all switch cases.
Similarly, I think you can't do an early return inside an unrolled loop, because you would get "statement unreachable" warnings.

aG0aep6G:
> Still I'm not sure about the sematics of breaking out of unrolled loop

You can see the effect when you put the `default` case behind the `foreach`:

```d
void main()
{
    import std.meta: AliasSeq;
    import std.stdio: writeln;
    sw: switch (1)
    {
        foreach (x; AliasSeq!(1, 2, 3))
        {
            case x: writeln(x); break;
        }
        default: writeln("default");
    }
}
```

Prints:
```
1
default
```

The `break` jumps behind the `foreach`, and the `default` case gets executed. `break sw;` would jump behind the `switch`, skipping over `default`.

> what if the break was guarded by a runtime check (that always passes) then it won't be possible to stop the loop unrolling and I would expect the runtime check + the break; to be present in all switch cases.

As far as I understand, you can't stop the unrolling. `break` is always a run-time jump, even in a static foreach.

> Similarly, I think you can't do an early return inside an unrolled loop, because you would get "statement unreachable" warnings.

Would make sense to me, but dmd accepts this without warning:

```d
int main()
{
    import std.meta: AliasSeq;
    foreach (x; AliasSeq!(1, 2, 3)) return x;
}
```

wilzbach:
Okay I finally got around rebasing this PR.
Thanks @ZombineDev for the excellent suggestion!
I modified it a bit to save space, but the example still uses up a lot:

image

Does anyone have a good idea to save space? I feel like this example has three lines too many.

wilzbach:
> I feel like this example has three lines too many.

I tried switching from Allman brace style to K&R.

index.dd
```diff
@@ -240,6 +240,34 @@ void main()
...
+struct Config {
+    int a, b;
+    string c;
+}
+void main() {
+    Config conf;
+    // use the generated parser for each entry
+    "b=1,a=2,c=hello"
+        .splitter(",")
+        .each!(e => conf.read(e));
```
JinShil: `"b=1,a=2,c=hello".splitter(",").each!(e => conf.read(e));`

That would get rid of 2 lines. The syntax coloring will help to make it readable.

index.dd
```diff
@@ -240,6 +240,34 @@ void main()
 }
 ----
 )
+$(EXTRA_EXAMPLE_RUNNABLE Generate a config parser at Compile-Time,
```
JinShil: I don't the "Compile-Time" should be in caps here.  I recommend "compile-time".

index.dd
```diff
@@ -240,6 +240,34 @@ void main()
...
+----
+import std.algorithm, std.conv, std.range, std.stdio, std.traits;
+struct Config {
+    int a, b;
+    string c;
+}
+void main() {
+    Config conf;
+    // use the generated parser for each entry
```
JinShil: Using "generated parser" but there is nothing called "parser" in the example.  Perhaps rename the `read` function below to `parser`.

wilzbach: Fair point - I renamed it to `parse` as then `"b=1,a=2,c=hello".splitter(",").each!(e => conf.parse(e));` reads nicer

wilzbach:
@JinShil thanks a lot for your ideas! It's still a bit. Maybe one or two lines less and we it's good?

image

JinShil:
That newline really stinks.  What about?

```D
c.tupleof[idx] = r.dropOne.front
    .to!(typeof(Conf.tupleof[idx]));
```

JinShil:
Is it still too tall?  If you really need to shrink it down you might try...

```D
struct Config { int a, b; string c; }
```

wilzbach:
> Is it still too tall? If you really need to shrink it down you might try...

Yeah gave it a try. I could try to remove the comment too?

image

JinShil:
Look OK to me.  How many lines are you aiming for?

wilzbach:
> Look OK to me. How many lines are you aiming for?

So that people don't shout. WTF? What monster is this?
People already complained that the [example at the Dlang-Tour](https://tour.dlang.org/) is too complex, so I'm not so sure whether we should move forward with this. Maybe it's better to add a "D by example" section to tour.dlang.org? This would also allow more room for comments and explanations.

JinShil:
I don't see this getting any simpler.  I thought this was for the front-page example, not the D tour.  The D tour's examples should be just enough complexity to illustrate the concept under discussion and nothing more, but the front-page examples are intended to showcase D's capabilities, are they not?

I wouldn't recommend this being in the D tour, but as a front-page example showcase, I think it's OK.

wilzbach:
> I don't see this getting any simpler. I thought this was for the front-page example, not the D tour. The D tour's examples should be just enough complexity to illustrate the concept under discussion and nothing more, but the front-page examples are intended to showcase D's capabilities, are they not?

Just, but it should be obvious what they are doing, e.g. "Initialize an array in parallel" is a very good example. It's easy to understand and you even get a visual proof when you hit "Run" (well, apart from the deprecation message which appears atm).

> I wouldn't recommend this being in the D tour, but as a front-page example showcase, I think it's OK.

Oh I think you misunderstood me. I was talking about adding it at the end of the tour as an addendum for people interesting in powerful, one-page examples.

Have a look at: https://github.com/dlang-tour/english/pull/217

I believe that's the better way to go than shocking users on the main page.
CC @andralex @CyberShadow @ZombineDev @MetaLang @JackStouffer for opinions.

CyberShadow:
Oof ouch owie my inbox.

Perhaps we can exclude some examples from the "random roulette", so that the page doesn't appear too broken by default, but allow adding larger examples that the user could still select manually from the dropdown?

wilzbach:
> I was talking about adding it at the end of the tour as an addendum for people interesting in powerful, one-page examples.

https://tour.dlang.org/tour/en/byexample/code-generation-parser

> Perhaps we can exclude some examples from the "random roulette", so that the page doesn't appear too broken by default, but allow adding larger examples that the user could still select manually from the dropdown?

I like the idea, but won't the list then get too big? IIRC was the magic number eight.

wilzbach:
I will close this now as I feel that the example has found a good place at the end of the DTour. If someone wants to include it on dlang.org, it will be easy to find.

PR #1763, opened 2017-06-22
[Idea] Add example about running a minimal web server

wilzbach:
Marking this "needs work" as I would need to remove `RUNABLE_EXAMPLE` from the `EXTRA_EXAMPLE` macro.

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -30,7 +30,7 @@ $(DIVC intro, $(DIV, $(DIV,
                 $(P Upon approval it will be showcased here on a random schedule.)
             )
         )
-$(EXTRA_EXAMPLE Compute average line length for stdin,
+$(EXTRA_EXAMPLE_RUNNABLE Compute average line length for stdin,
 $(RUNNABLE_EXAMPLE_STDIN
 The D programming language
 Modern convenience.
@@ -50,7 +50,7 @@ void main()
 }
 ----
 )
-$(EXTRA_EXAMPLE Round floating point numbers,
+$(EXTRA_EXAMPLE_RUNNABLE Round floating point numbers,
 $(RUNNABLE_EXAMPLE_STDIN 2.4 plus 2.4 equals 5 for sufficiently large values of 2.)
 ----
 import std.algorithm, std.conv, std.functional,
@@ -71,7 +71,7 @@ void main()
 }
 ----
 )
-$(EXTRA_EXAMPLE Sort lines,
+$(EXTRA_EXAMPLE_RUNNABLE Sort lines,
 $(RUNNABLE_EXAMPLE_STDIN
 Mercury
 Venus
@@ -120,6 +120,25 @@ void main()
     auto result = ["whoami"].execute;
     enforce(result.status == 0);
     result.output.write;
+}
+----
+)
+$(EXTRA_EXAMPLE Start a minimal web server,
+----
+#!/usr/bin/env dub
+/+ dub.sdl:
+name "hello_vibed"
+dependency "vibe-d" version="~>0.8.0"
++/
+void main()
+{
+    import vibe.d;
+    auto settings = new HTTPServerSettings;
+    settings.port = 8080;
+    listenHTTP(settings, (req, res) {
+        res.writeBody("Hello, World: " ~ req.path);
+    });
+    runApplication();
 }
 ----
 )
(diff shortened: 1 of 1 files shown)


MartinNowak:
Even vibe.d's front page finally switched to a normal main function ;). http://vibed.org/
```d
#!/usr/bin/env dub
/+ dub.sdl:
 name "hello_vibed"
 dependency "vibe-d" version="~>0.8.0-rc.2"
 +/
import vibe.vibe;

void main()
{
    auto settings = new HTTPServerSettings;
    settings.port = 8080;

    listenHTTP(settings, (req, res) { res.writeBody("Hello, World: " ~ req.path); });
    runApplication();
}
```

wilzbach:
@CyberShadow - what's your opinion on this? It wouldn't be hard to hide the run buttons and maybe show additional explanation and links if we want to go this road.
I am not really decided here,
+ Vibe.d is the defacto flagship library
+ We want to show D code for different areas
+ It shows how awesome DUB's single file descriptions are

- it's not runnable
- it's a third-party library and not "official" D

CyberShadow:
I think it's still worth adding.

When I asked an IRL colleague whether he'd be interested in using D, the first reason he said was that D "doesn't even have a web server in its standard library!". Starting up a web server is one of the more common front page examples of other languages, so I think we shouldn't stay behind here.

Lack of dub support on the example runner is just a lucky coincidence, since it's not meaningful to start a web server there either :)

wilzbach:
Rebased the example to the option chooser example.

- `runApplication` isn't `@safe`. With my initial `VibeDefaultMain`,  `shared static this() @safe` and would work. What option do you prefer?
- Should we use `#!/usr/bin/env rdmd` for the other examples?

wilzbach:
Oh and for convenience:

image

CyberShadow:
> What option do you prefer?

`void main()` IMO

> Should we use #!/usr/bin/env rdmd for the other examples?

I think there's no need for that - vertical space is at a premium anyhow

index.dd
```diff
@@ -94,6 +94,25 @@ void main()
 }
 ----
 )
+$(EXTRA_EXAMPLE Start a minimal web server,
+----
+#!/usr/bin/env dub
+/+ dub.sdl:
+name "hello_vibed"
+dependency "vibe-d" version="~>0.8.0"
```
wilzbach: Vibe.d 0.8.0 will be released very [soon](http://forum.dlang.org/post/oj198q$1j9q$1@digitalmars.com), so I already accounted for the approval / review lag.

index.dd
```diff
@@ -94,6 +94,25 @@ void main()
 }
 ----
 )
+$(EXTRA_EXAMPLE Start a minimal web server,
+----
+#!/usr/bin/env dub
+/+ dub.sdl:
+name "hello_vibed"
```
wilzbach: FYI: With newer DUB instances this line won't be needed anymore (-> https://github.com/dlang/dub/pull/1081), but for now we have to wait a couple of months until https://github.com/dlang/dub/pull/1081 has gone through a couple of releases.

CyberShadow:
@wilzbach Is this good to go? I see it still has the "needs work" label.

CyberShadow:
Oh, or is it blocked by the Vibe.d release?

wilzbach:
> Oh, or is it blocked by the Vibe.d release?

Yes, it's in three days. I think we can wait until then. I will ping this PR then ;-)

wilzbach:
> Yes, it's in three days. I think we can wait until then. I will ping this PR then ;-)

Finally Vibe.d 0.8 was relased :tada:

http://forum.dlang.org/post/mmfrsonxrfxwltkfvliw@forum.dlang.org

CyberShadow:
Darn it, doesn't work on my machine because of the OpenSSL issue.

Do you think we should go ahead with this despite this problem, or wait until Vibe's OpenSSL bindings are updated?

wilzbach:
> Darn it, doesn't work on my machine because of the OpenSSL issue.

Hehe that's a well-known problem, see e.g. http://forum.dlang.org/post/bpjzqbdhwzplqrvsjfca@forum.dlang.org

In short, the workarounds are:
- `versions "VibeNoSSL"`
-  [setting libs-posix to openssl1.0](https://github.com/rejectedsoftware/vibe.d/issues/1748#issuecomment-297647832)
-  Use the botan library (pure D) with:
```
dependency "vibe-d:tls" version="~>0.8.0"
subConfiguration "vibe-d:tls" "botan"
```
This was a last minute change to 0.8.0: https://github.com/rejectedsoftware/vibe.d/pull/1823

The [proper fix](https://github.com/rejectedsoftware/vibe.d/pull/1759) has still problems due to DUB's inability to pass the linked library version.

> Do you think we should go ahead with this despite this problem, or wait until Vibe's OpenSSL bindings are updated?

- Sönke merged a similar PR to the README last week: https://github.com/rejectedsoftware/vibe.d/pull/1803
- A similar example is on [vibed.org](http://vibed.org)

-> Yeah - I think Arch users are a minority and they are used to breakage ;-)

PR #1764, opened 2017-06-22
Fix link to FuncDeclarator

wilzbach:
@CyberShadow any smart idea on easily verifying that all links are valid?
The ugly way would be to parse the generated HTML, create name/id lists for each file and then check all links ...

### DIFF
--- a/spec/declaration.dd
+++ b/spec/declaration.dd
@@ -325,7 +325,7 @@ $(H3 $(LNAME2 alias, Alias Declarations))
 $(GRAMMAR
 $(GNAME AliasDeclaration):
     $(D alias) $(GLINK StorageClasses)$(OPT) $(GLINK BasicType) $(GLINK Declarators) $(D ;)
-    $(D alias) $(GLINK StorageClasses)$(OPT) $(GLINK BasicType) $(GLINK FuncDeclarator) $(D ;)
+    $(D alias) $(GLINK StorageClasses)$(OPT) $(GLINK BasicType) $(GLINK2 function, FuncDeclarator) $(D ;)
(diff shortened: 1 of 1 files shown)


CyberShadow:
> Thanks for your pull request, @wilzbach! We are looking forward to reviewing it, and you should be hearing from a maintainer soon.

Oh, what is this?

CyberShadow:
> The ugly way would be to parse the generated HTML, create name/id lists for each file and then check all links ...

That's the way.

chmgen already does it BTW.

wilzbach:
> Oh, what is this?

-> https://github.com/dlang-bots/dlang-bot/pull/44

The idea is to always "reserve" the first comment and update it with helpful information.
We are still at the beginning and it's not super helpful (yet), but this is deployed:
- hello world message
- [warning about a regression PR not targeted at stable](https://github.com/dlang-bots/dlang-bot/pull/57)

Anyways, there's [more to come](https://github.com/dlang-bots/dlang-bot/issues/8).

If you want to change the text -> https://github.com/dlang-bots/dlang-bot/blob/master/source/dlangbot/github.d#L57

ZombineDev:
(Closing & reopening in order to restart the CI.)

ZombineDev:
@CyberShadow @wilzbach can you guys add a "Ddoc bug fix" label? The intent is to tag pull requests that fix ddoc errors (such as this one) but don't otherwise deserve a Bugzilla entry.

wilzbach:
@ZombineDev You can add them yourself - click on labels on the pull requests overview list...

ZombineDev:
@wilzbach Thanks, not sure how I missed it.

PR #1765, opened 2017-06-24
Explain D as a general-purpose programming language

wilzbach:
From http://forum.dlang.org/post/oij5t5$14tq$1@digitalmars.com:

While D isn't a "scripting language" as stated in the thread, it can be used literally anywhere and for anything nowadays and thus imho "general-purpose" is a better description than "systems"

FWIW C++ is a "general-purpose" programming language as well. From [Wikipedia](https://en.wikipedia.org/wiki/C%2B%2B):

> C++ is a general-purpose programming language. It has imperative, object-oriented and generic programming features, while also providing facilities for low-level memory manipulation.

CC @aG0aep6G

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -6,9 +6,10 @@ $(SECTION1 ,
 
 $(DIVC intro, $(DIV, $(DIV,
     $(DIV,
-        $(TC p, pitch, $(B D) is a systems programming language with C-like
-        syntax and static typing. It combines efficiency, control and modeling
-        power with safety and programmer productivity.)
+        $(TC p, pitch, $(B D) is a general-purpose programming language with
+        static typing, systems-level access, and C-like syntax.
+        It combines efficiency, control and modeling power with safety
+        and programmer productivity.)
 
         $(DIVC download,
             $(DIVC btn-group-vertical,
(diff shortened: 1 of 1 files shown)


CyberShadow:
Well, this definitely needs W&A's approval :) @WalterBright @andralex

ZombineDev:
What do you think about adding below something along the lines of:
"D scales from one-off scripts, through high-level application programming all the way down to low-level systems programming tasks like implementing operating system kernels." ?

index.dd
```diff
@@ -6,7 +6,7 @@ $(SECTION1 ,
 
 $(DIVC intro, $(DIV, $(DIV,
     $(DIV,
-        $(TC p, pitch, $(B D) is a systems programming language with C-like
+        $(TC p, pitch, $(B D) is a general-purpose programming language with C-like
         syntax and static typing. It combines efficiency, control and modeling
```
andralex: Let's do this:

D is a general-purpose language with static typing, systems-level access, and C-like syntax.

wilzbach:
> Let's do this:
> D is a general-purpose language with static typing, systems-level access, and C-like syntax.

Done.

aG0aep6G:
Followup to avoid excessive hyphenation with the new text: #1783.

PR #1766, opened 2017-06-25
Add release dates to changelog version list

CyberShadow:
Currently, there is no at-a-glance timeline of D releases, and there is no simple way to find the latest version at a certain point in time. This can be useful when examining older bug reports or commits.

Done using:

```
( IFS=$'\n' ; sed -i changelog.ddoc $(grep '(VERSION' *.dd | sed -E 's/^(.*)\.dd:\$\(VERSION (.*), ==.*/-e\ns#CHANGELOG_VERSION \1)#CHANGELOG_VERSION \1,\2)#/') )
```


css/style.css
```diff
@@ -454,6 +454,11 @@ in the narrow layouts without JS. */
     text-decoration: underline;
 }
 
+.subnav .hide-from-nav
+{
+    display: none;
+}
+
```
CyberShadow: I couldn't see a way to render the dates only on the changelog listing page and not in the sidebar, hence this CSS hack.

I wish we could do `$($1, ...)`, then we could pass DDoc macro names as parameters.

wilzbach:
Good idea, but I think you need to update the [generation script](https://github.com/dlang/dlang.org/blob/master/changelog/update_nav.sh) as well.

A bit unrelated, but is there a good way for a user to find this overview listing? Maybe we could link to it from the changelog sidenav?

CyberShadow:
> A bit unrelated, but is there a good way for a user to find this overview listing?

Wow, I didn't realize that. I typed `dlang.org/changelog` in my browser's address bar every time.

> Maybe we could link to it from the changelog sidenav?

I faintly recall this being proposed earlier, when the changelog was split up from being all on a single page, but at that time there was nothing interesting to link to - that page duplicated the sidebar's content..

css/style.css
CyberShadow: https://issues.dlang.org/show_bug.cgi?id=17549

CyberShadow:
> Good idea, but I think you need to update the generation script as well.

Done.

changelog/changelog.ddoc
```diff
@@ -63,96 +63,96 @@ DRUNTIMEPR = $(PULL_REQUEST druntime,$1)
...
-    $(CHANGELOG_VERSION 2.008)
-    $(CHANGELOG_VERSION 2.007)
-    $(CHANGELOG_VERSION 2.006)
-    $(CHANGELOG_VERSION 2.005)
-    $(CHANGELOG_VERSION 2.004)
-    $(CHANGELOG_VERSION 2.003)
-    $(CHANGELOG_VERSION 2.002)
-    $(CHANGELOG_VERSION 2.001)
-    $(CHANGELOG_VERSION 2.000)
+    $(CHANGELOG_VERSION 2.074.1,May 30, 2017)
```
ZombineDev: Space after comma, or this doesn't bode well with ddoc macros?

CyberShadow: Shouldn't matter. Fixed.

(Could've sworn I wrote this comment already...)

CyberShadow:
> A bit unrelated, but is there a good way for a user to find this overview listing? Maybe we could link to it from the changelog sidenav?

-> #1768

css/style.css
wilzbach: > I wish we could do 

Or simply use a proper template engine (i.e. one that supports conditions) ...

css/style.css
```diff
@@ -454,6 +454,11 @@ in the narrow layouts without JS. */
     text-decoration: underline;
 }
 
+.subnav .hide-from-nav
```
ZombineDev: There was no need to add a new css rule when `.subnav .subnav-duplicate` does the same a few lines above.
-> #1770

MartinNowak:
Thanks for updating the update_nav.sh script as well :).

PR #1767, opened 2017-06-25
Temporarily fix the example roulette

wilzbach:
From https://github.com/dlang/dlang.org/pull/1757#issuecomment-310889979

> OK, so, as I understand, this PR does three things right now:
> Fix the JS roulette
> IMO these should be separate PRs, or at least commits

Well #1757 removes this bit completely (hence no separate commit), but as the discussion seems to take a bit longer, there's no reason why we couldn't fix the roulette in the meantime.

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -50,6 +50,7 @@ void main()
 }
 ----
 )
+)
 $(EXTRA_EXAMPLE
 $(RUNNABLE_EXAMPLE_STDIN 2.4 plus 2.4 equals 5 for sufficiently large values of 2.)
 ----
@@ -71,7 +72,6 @@ void main()
         .each!writeln;
 }
 ----
-)
 )
 $(EXTRA_EXAMPLE
 $(RUNNABLE_EXAMPLE_STDIN
--- a/js/dlang.js
+++ b/js/dlang.js
@@ -59,11 +59,11 @@
         });
 
         // [your code here] rotation for index.html
-        var $examples = $('.your-code-here-extra > pre');
+        var $examples = $('.your-code-here-extra .d_code');
         if ($examples.length) {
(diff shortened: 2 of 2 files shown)


CyberShadow:
> Thanks for your pull request, @wilzbach! We are looking forward to reviewing it, and you should be hearing from a maintainer soon.

Shouldn't this have been the brief version? You're a committer here.

wilzbach:
> Shouldn't this have been the brief version? You have push access here.

The [GH API](https://developer.github.com/v3/orgs/members/#public-members-list) lists only 30 items per requests (e.g. https://api.github.com/orgs/dlang/public_members), and as it mostly affects only me (https://api.github.com/orgs/dlang/public_members?page=2) I was too lazy/busy too fix it for now.

CyberShadow:
How about checking if https://api.github.com/repos/dlang/dlang.org/commits?author=wilzbach is non-empty? Then contributors would also start getting the shorter message once their first PR is accepted.

CyberShadow:
> The GH API lists only 30 items per requests

?per_page=100 but either way that solution doesn't scale.

index.dd
```diff
@@ -50,6 +50,7 @@ void main()
 }
 ----
 )
+)
```
CyberShadow: HTML diff says:
```diff
- 
+) 
```
so this doesn't seem right.

wilzbach:
> How about checking if https://api.github.com/repos/dlang/dlang.org/commits?author=wilzbach is non-empty? Then contributors would also start getting the shorter message once their first PR is accepted.

-> let's continue this discussion over at the bot: https://github.com/dlang-bots/dlang-bot/issues/94

PR #1768, opened 2017-06-25
changelog: Link to all versions page in the sidebar unconditionally

CyberShadow:
For #1766.

### DIFF
--- a/changelog/changelog.ddoc
+++ b/changelog/changelog.ddoc
@@ -8,8 +8,7 @@ CHANGELOG_NAV_FIRST=$(DIVC changelog-nav, next version: $(A $1.html, $1))
 CHANGELOG_NAV_LAST=$(DIVC changelog-nav, previous version: $(A $1.html, $1))
 
 SUBNAV=$(SUBNAV_TEMPLATE
-    $(SUBNAV_HEAD Change Log, $(ROOT_DIR)changelog/index.html, List of all versions,
-        subnav-duplicate)
+    $(SUBNAV_HEAD Change Log, $(ROOT_DIR)changelog/index.html, List of all versions, )
     $(UL $(CHANGELOG_VERSIONS))
 )
 


changelog/changelog.ddoc
```diff
@@ -8,8 +8,7 @@ CHANGELOG_NAV_FIRST=$(DIVC changelog-nav, next version: $(A $1.html, $1))
 CHANGELOG_NAV_LAST=$(DIVC changelog-nav, previous version: $(A $1.html, $1))
 
 SUBNAV=$(SUBNAV_TEMPLATE
-    $(SUBNAV_HEAD Change Log, $(ROOT_DIR)changelog/index.html, List of all versions,
-        subnav-duplicate)
+    $(SUBNAV_HEAD Change Log, $(ROOT_DIR)changelog/index.html, List of all versions)
```
ZombineDev: Even though it visually looks correct, I don't think this is the right fix.
`SUBNAV_HEAD` is defined in `dlang.org.ddoc` as:
```
SUBNAV_HEAD=
    $(DIVC head,
        $(H2 $1)
        $(TC p, $4,
            $(LINK2 $2, $3))
```
And `TC` is defined in `html.ddoc` as:
```
TC=<$1 class="$2">$(TAIL $+)</$1>
```

I.e. it takes `SUBNAV_HEAD` 4 arguments, where the fourth is the `class` attribute of the `<p>` tag.
Inspecting the generated html reveals that
`$(SUBNAV_HEAD Change Log, $(ROOT_DIR)changelog/index.html, List of all versions)`  gets expanded to:
```html
<div class="head">
    <h2>Change Log</h2>
    <p class="Change Log, ../changelog/index.html, List of all versions">
        <a href="../changelog/index.html">List of all versions</a>
    </p>
</div>
```

While before
```
    $(SUBNAV_HEAD Change Log, $(ROOT_DIR)changelog/index.html, List of all versions,
        subnav-duplicate)
```

expanded to:

```html
<div class="head">
    <h2>Change Log</h2>
    <p class="        subnav-duplicate">            
        <a href="../changelog/index.html">List of all versions</a>
    </p>
</div>
```

Where `subnav-duplicate` is defined in `css/style.css` as:
```
/* Hide links that just duplicate the subnav contents. They're only displayed
in the narrow layouts without JS. */
.subnav .subnav-duplicate
{
    display: none;
}
```
( Added in commit https://github.com/dlang/dlang.org/commit/736a8bbf3d39cf44da672a62bedc819dc35121f1 )

Hence why the link in the sidebar was not visible.

@aG0aep6G why did you hide link in #1190 (at least it seems intentional to me)?

ZombineDev: The correct fix is add a forth empty parameter to the `SUBNAV_HEAD` macro invocation, though I'm not well-enough versed with ddoc to know if simply:
```
$(SUBNAV_HEAD Change Log, $(ROOT_DIR)changelog/index.html, List of all versions, )
```
would work.

CyberShadow: Nice catch!

ZombineDev: Alternatively, I think we can change the definition of `SUBNAV_HEAD` from:
```
SUBNAV_HEAD=
    $(DIVC head,
        $(H2 $1)
        $(TC p, $4,
            $(LINK2 $2, $3))
```
to:
```
SUBNAV_HEAD=
    $(DIVC head,
        $(H2 $1)
        $(P
            $(LINK2 $2, $3))
```

though I haven't tested this yet.

CyberShadow:
> @aG0aep6G why did you hide link in #1190 (at least it seems intentional to me)?

I think I can answer that.

The links in the side navigation are accessible either by simply clicking them directly, as they are always visible when the browser viewport is wide enough, or by expanding the secondary navigation dropdown on narrow views. However, when the viewport is narrow and JavaScript is not available, there is no way to access them. Therefore, some index pages were made only accessible for this use case: a list of links that you can get to through a link in the navigation that's only visible when the viewport is narrow and JS is off.

wilzbach:
Or in more visual words, that's what clients without JavaScript + narrow viewport see:

image

The spec navigation allows this as well:

image

ZombineDev:
> wilzbach dismissed ZombineDev’s review 6 hours ago

Thanks, I was busy, otherwise I would have merged this myself.

PR #1770, opened 2017-06-25
Remove duplicate css rule

ZombineDev:
There's was an existing one defined as:

```css
.subnav .subnav-duplicate
{
    display: none;
}
```

### DIFF
--- a/changelog/changelog.ddoc
+++ b/changelog/changelog.ddoc
@@ -63,7 +63,7 @@ DRUNTIMEPR = $(PULL_REQUEST druntime,$1)
 PHOBOSPR = $(PULL_REQUEST phobos,$1)
 _=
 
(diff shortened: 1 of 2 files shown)


CyberShadow:
Consider that in the future, someone is going to look at the definition of `CHANGELOG_VERSION`, notice the `subnav-duplicate` style, then look at its definition in the CSS file. There, the comment says:

> Hide links that just duplicate the subnav contents. They're only displayed
> in the narrow layouts without JS.

Not only is the date part (wrapped in the `<span>` tag) not a link, but its (in)visibility is completely unrelated to narrow layouts.

Although you could update the comment, it's better to just have separate rules for their separate semantic meanings. Because they do different things, their implementation may change independently of each other in the future.

ZombineDev:
Sorry, it seems that I misunderstood your [comment](https://github.com/dlang/dlang.org/pull/1768#issuecomment-310895341), because I though that the intentions were the same.

PR #1771, opened 2017-06-25
fix support of relative variable overrides

MartinNowak:


wilzbach:
> make[1]: Leaving directory '/dev/shm/dtest/work/repo/dmd/src'
> make: *** No rule to make target '../druntime-2.074.1', needed by '.generated/modlist-2.074.1.ddoc'.  Stop.

Hmm the `git clone` targets are absolute (due to `$(PWD)`) as well.
How about replacing the first git clone rule with sth. like this?

```
${DMD_DIR}-${LATEST} ${DRUNTIME_DIR}-${LATEST} ${PHOBOS_DIR}-${LATEST} ${TOOLS_DIR}-${LATEST} ${INSTALLER_DIR}-${LATEST}:
	git clone --depth=1 -b v${LATEST} ${GIT_HOME}/$(word 1, $(subst -, ,$(notdir $(@F)))) $@
```

posix.mak
```diff
@@ -113,16 +113,16 @@ endif
 DDOC_VARS_STABLE=\
 	  DOC_OUTPUT_DIR="${DOC_OUTPUT_DIR}/phobos" \
 	  STDDOC="$(addprefix $(PWD)/, $(STD_DDOC))" \
-	  DMD="$(DMD_STABLE)" \
-	  DRUNTIME_PATH="${DRUNTIME_DIR}" \
+	  DMD="$(realpath $(DMD_STABLE))" \
```
CyberShadow: Please no `realpath`, it resolves symlinks, and the final paths (which may be temporary and expose implementation details) end up all over the place, sometimes even in outputs.

MartinNowak: It's what we've used during the previous years, but I could use make's [abspath](https://www.gnu.org/software/make/manual/make.html#index-file-name_002c-abspath-of) if that helps.

MartinNowak:
> Hmm the git clone targets are absolute (due to $(PWD)) as well.
> How about replacing the first git clone rule with sth. like this?

This is unrelated to *this* change and was introduced here https://github.com/dlang/dlang.org/pull/1687#discussion_r123899158.

posix.mak
CyberShadow: Yes please.

wilzbach:
> This is unrelated to this change and was introduced here #1687 (comment).

Maybe, but it would fix the broken build on DAutoTest.

MartinNowak:
See https://github.com/dlang/dlang.org/pull/1772 @wilzbach.
And again, I'd rather we spent time to finish the last few missing bits for ddox than throwing even more at the ever more complex ddoc build.

MartinNowak:
Also fixes the clone targets now.

wilzbach:
> MartinNowak  wants to merge 2 commits into dlang:stable

Btw as `master` gets auto-deployed on dlang.org, we could also declare `stable` as default branch for dlang.org if that makes your life easier?

MartinNowak:
> Btw as master gets auto-deployed on dlang.org, we could also declare stable as default branch for dlang.org if that makes your life easier?

It's fine for the moment and not much effort.

PR #1772, opened 2017-06-26
fix clone targets

MartinNowak:
- broken by 4129e45951 which changed target folders from
  ../druntime-2.074.1 to $PWD/druntime-2.074.1
- no need to make those targets absolute since the makefile will
  always run in the dlang.org root folder


wilzbach:
> I might submit a short follow-up with #1771 (comment)

It seems that you directly need this without #1771

MartinNowak:
I see, needs to be incorporated with #1771 due to $PWD in the DMD_DIR et.al. vars.

PR #1775, opened 2017-06-27
Fix redraw flashes at the JavaScript example roulette

wilzbach:
"Meh" solution instead of https://github.com/dlang/dlang.org/pull/1757

CC @CyberShadow

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -28,8 +28,7 @@ $(DIVC intro, $(DIV, $(DIV,
                 $(P Upon approval it will be showcased here on a random schedule.)
             )
         )
-        $(DIVID your-code-here-default,
-$(RUNNABLE_EXAMPLE
+$(EXTRA_EXAMPLE
 $(RUNNABLE_EXAMPLE_STDIN
 The D programming language
 Modern convenience.
@@ -49,7 +48,6 @@ void main()
         count ? sum / count : 0);
 }
 ----
-)
 )
 $(EXTRA_EXAMPLE
 $(RUNNABLE_EXAMPLE_STDIN 2.4 plus 2.4 equals 5 for sufficiently large values of 2.)
@@ -98,8 +96,22 @@ void main()
 ----
 )
 
+) $(COMMENT your-code-here)
+))) $(COMMENT intro, div, div)
+
+$(COMMENT
+Picking the frontpage example after the page has loaded leads to reflows
+and re-rendering. Hence, we pick the example as soon as possible
 )
-)))
+$(SCRIPT
+    (function() {
+    // randomly pick one example and bootstrap the runnable editor
+    var examples = document.getElementsByClassName('your-code-here-extra');
+    var rouletteIndex = Math.floor(Math.random() * examples.length);
+    var rouletteChild = examples[rouletteIndex];
+    rouletteChild.style.display = "block";
+    })();
+)
 
 $(DIVC more,
 
--- a/js/dlang.js
+++ b/js/dlang.js
@@ -58,14 +58,6 @@
             $('#search-query input').focus();
         });
 
-        // [your code here] rotation for index.html
-        var $examples = $('.your-code-here-extra .d_code');
-        if ($examples.length) {
-            var n = Math.floor(Math.random() * ($examples.length+1));
-            if (n)
-                $('#your-code-here-default .d_code').replaceWith($examples[n-1]);
-        }
-
         // Insert the show/hide button if the contents section exists
         $('.page-contents-header').append('<span><a href="javascript:void(0);">[hide]</a></span>');
 


index.dd
```diff
@@ -98,8 +96,14 @@ void main()
...
 
+) $(COMMENT your-code-here)
+))) $(COMMENT intro, div, div)
+
+$(COMMENT
+Picking the frontpage example after the page has loaded leads to reflows
+and re-rendering. Hence, we pick the example as soon as possible
 )
-)))
+<script src="$(ROOT_DIR)js/example_roulette.js"></script>
```
CyberShadow: Inline the script please?

js/example_roulette.js
```diff
@@ -0,0 +1,30 @@
...
+// randomly pick one example and bootstrap the runnable editor
+var examples = document.getElementsByClassName('your-code-here-extra');
+var rouletteIndex = Math.floor(Math.random() * examples.length);
+var rouletteChild = examples[rouletteIndex];
+
+// step 1: select the .runnable-example <div>
+var rouletteChildNode = findFirst(rouletteChild.children, function(e) { return e.className.indexOf("runnable-example") != -1; });
+// step 2: select the first <pre> element (of the .runnable-example <div>)
+var el = findFirst(rouletteChildNode.children, function(e) { return e.nodeName == "PRE";});
+el.style.marginBottom = "20px";
```
CyberShadow: BTW, this is not necessary. You can use the `have-javascript` class and do it with a CSS rule.

index.dd
wilzbach: I still would like to add the example selection as well (so the file will grow) and doing Js in Ddoc is really a pain.
Didn't you believe me, when [I showed that using a separate file doesn't matter performance-wise](https://github.com/dlang/dlang.org/pull/1757#issuecomment-310866543)?

js/example_roulette.js
wilzbach: I need to reset this _after_ the entire body has been parsed (and the JS for the building the examples layout gets triggered), not on start:

```html
<body id='Home' class='doc'>
<script type="text/javascript">document.body.className += ' have-javascript'</script>
```

CyberShadow: Err, why?

index.dd
CyberShadow: > I still would like to add the example selection as well (so the file will grow)

I don't understand how adding the dropdown should affect this part. The dropdown can only work when JavaScript is enabled, so it would need to be added at the end, no?

CyberShadow: > Didn't you believe me, when I showed that using a separate file doesn't matter performance-wise?

1. It doesn't matter today
2. It doesn't matter in that browser
3. Not an excuse to make a sub-ideal situation worse. What that shows is that more things need to be loaded lazily.

wilzbach: > The dropdown can only work when JavaScript is enabled, so it would need to be added at the end, no?

No it should be done here as well - same motivation:  preventing redraws

js/example_roulette.js
wilzbach: Isn't this how a browser behaves?

- Page gets fetched
- browser fetches all scripts in the header
- it starts evaluating the body -> adds the class
- it hits the examples and executes this js (with the marginBottom hack)
- it fetches all the scripts and stuff before and at the page bottom
- it reaches the end of the page and reverts the marginBottom hack

js/run.js
```diff
@@ -127,6 +127,8 @@ $(document).ready(function()
     {
         var root = $(this);
         var el = root.children("pre");
+        // hack from the front page to avoid redraw flashes due to the buttons showing up
+        el[0].style.marginBottom = "0px";
```
CyberShadow: Instead of removing the margin, can this be left in? As mentioned in the previous PR, my suggestion was to simply create a container (whose size is not going to change) for the buttons to be placed in. I think that would avoid a relayout.

index.dd
CyberShadow: I think we need to worry about relayouts first; redraws shouldn't be much of a problem as long as they aren't too distracting. Avoiding relayouts should be easy by reserving space for the interactive stuff that will be filled in later by JS. 

As for redraws (I assume you mean pop-in when new UI elements are displayed), we have too much stuff redrawing anyway with all the widgets etc. Not sure if it's worth complicating the implementation much just to be able to have the drop-down there from the start, though even so, I think it would be like 1 more line to select the right option, right? And the buttons are going to be added in later either way, right?

js/example_roulette.js
CyberShadow: Oh, I understand what you mean now. Just because you need to reset it doesn't mean you have to add it via JS too - JS styles override CSS ones because they're going to be set on the element itself. Regardless I agree that undoing/resetting it is suboptimal, see my comment below.

wilzbach:
> As for redraws (I assume you mean pop-in when new UI elements are displayed), we have too much stuff redrawing anyway with all the widgets etc. Not sure if it's worth complicating the implementation much just to be able to have the drop-down there from the start, though even so, I think it would be like 1 more line to select the right option, right? And the buttons are going to be added in later either way, right?
> ...
> Instead of removing the margin, can this be left in? As mentioned in the previous PR, my suggestion was to simply create a container (whose size is not going to change) for the buttons to be placed in. I think that would avoid a relayout.

Okay, I have removed the `marginBottom` hack and the code now contains only what's necessary to immediately select an example and shouldn't be controversial anymore.
Regarding the buttons: I won't be going for the container hack as imho the proper solution is to do the entire layout at once and not splitting it into pieces, but after this PR anyone can go down this road ;-)

index.dd
```diff
@@ -98,8 +96,31 @@ void main()
...
+        return Array.prototype.filter.call(el, fn)[0];
+    }
+    (function() {
+    // randomly pick one example and bootstrap the runnable editor
+    var examples = document.getElementsByClassName('your-code-here-extra');
+    var rouletteIndex = Math.floor(Math.random() * examples.length);
+    var rouletteChild = examples[rouletteIndex];
+
+    // step 1: select the .runnable-example <div>
+    var rouletteChildNode = findFirst(rouletteChild.children, function(e) { return e.className.indexOf("runnable-example") != -1; });
```
CyberShadow: Why not `rouletteChild.getElementsByClassName`?

index.dd
```diff
@@ -98,8 +96,31 @@ void main()
...
+    // step 2: select the first <pre> element (of the .runnable-example <div>)
+    var el = findFirst(rouletteChildNode.children, function(e) { return e.nodeName == "PRE";});
```
CyberShadow: Why not `getElementsByTagName`?

CyberShadow: Also, it looks like querySelector is good to use today:
http://caniuse.com/#feat=queryselector

CyberShadow: Also, I believe you can avoid traversing the element's children by setting a class on the `rouletteChild` element.

wilzbach:
Hmm, now that we don't do any layout modifications, I can simply drop the three lines in question - all we do is `style.block = "display" after all -> even tinier diff :)

wilzbach:
Btw if we already spent so much time on the examples, I really like how kotlin presents the first example:

image

https://kotlinlang.org

CyberShadow:
> all we do is `style.block = "display" after all -> even tinier diff :)

Awesome. So the coveted 5 lines of JS are possible after all :)

> Regarding the buttons: I won't be going for the container hack as imho the proper solution is to do the entire layout at once and not splitting it into pieces, but after this PR anyone can go down this road ;-)

Honestly, I don't think it's any more of a hack that specifying the dimensions of an image on the `` tag itself. I think it's a good idea in general for any page element that's not trivial to lay out.

wilzbach:
OT: Not so sure whether you noticed, but [someone pointed out on the NG](http://forum.dlang.org/post/stqrlgncmeootnwrlamf@forum.dlang.org) had the fair point that the examples lack explanation and explanatory links. Python.org puts a good bar here:

image
https://www.python.org/

There's also a lot more on this discussion, e.g. how complex the examples should be etc.

CyberShadow:
Yes, I saw that. Eye of the beholder etc. - the impression of any example depends on the reader's experience, and D has a pretty wide target audience. Not sure we can reasonably approach this without an overhaul of the front page.

wilzbach:
>   CyberShadow added the auto-merge label 2 minutes ago

I think you forgot to approve? Maybe we should add a warning to the bot?
-> https://github.com/dlang-bots/dlang-bot/issues/103

CyberShadow:
Could have sworn I did. Maybe I just forgot to submit it.

wilzbach:
>  the impression of any example depends on the reader's experience, and D has a pretty wide target audience.
> Not sure we can reasonably approach this without an overhaul of the front page.

CC @popen2 
IIRC you wanted to create a new, fancy frontpage in the next few weeks? :)

wilzbach:
FYI: the dlang-bot didn't receive the approval hook (they weren't sent). I have enabled this for all dlang repos now, but it needs a bit of handling on the dlang-bot side as well (though I need to wait a bit until a receive a relevant payload to put it into the testsuite).

See also: https://github.com/dlang-bots/dlang-bot/issues/104

PR #1776, opened 2017-06-27
css: Enforce input/textarea background only on dlang.org

CyberShadow:
Hotfix for forum.dlang.org.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -79,7 +79,10 @@ body > .container
     position: relative;
 }
 
-input, textarea
+body.doc input,
+body.doc textarea,
+body.std input,
+body.std textarea
 {
     background: white;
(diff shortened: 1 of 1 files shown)


wilzbach:
Interestingly this also affected the runnable examples, but I actually like the new CSS a lot more :)

image

image

Before:

image

CyberShadow:
Err. Hmm. CSS specficity shenanigans?

Note that it might look that way only in your browser.

wilzbach:
> Err. Hmm. CSS specficity shenanigans?

More "precise" selectors have a higher priority.

image

CyberShadow:
Yeah, specificity is actually the official term I think :) My guess was correct then?

PR #1777, opened 2017-06-27
Allow to select an example on the front page

wilzbach:
Marking this as "needs work" as I will to rebase this once we agree on #1775

image

image

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -505,7 +505,7 @@ Macros:
     D=<span class="d_inlinecode">$0</span>
     EXAMPLE=$(TAG2 a, id="a$1-control" class="example-control", )$(TAG2 div, id="a$1" class="example-box", $(RUNNABLE_EXAMPLE $2))
     EXTRA_EXAMPLE=<div class="your-code-here-extra" style="display:none">
-        <div class="your-code-here-title">$1</div>
+        $(DIVC your-code-here-title, $1)
         $(RUNNABLE_EXAMPLE $+)
     </div>
     _=
--- a/index.dd
+++ b/index.dd
@@ -19,6 +19,7 @@ $(DIVC intro, $(DIV, $(DIV,
         )
     )
     $(DIVID your-code-here,
+        $(DIVID your-code-here-select-example)
         $(DIVC tip,
             $(LINK2 https://forum.dlang.org/newpost/general?subject=%5Byour+code+here%5D, your code here)
             $(DIV,
@@ -28,14 +29,13 @@ $(DIVC intro, $(DIV, $(DIV,
                 $(P Upon approval it will be showcased here on a random schedule.)
             )
         )
-$(EXTRA_EXAMPLE
+$(EXTRA_EXAMPLE Compute average line length for stdin,
 $(RUNNABLE_EXAMPLE_STDIN
 The D programming language
 Modern convenience.
 Modeling power.
 Native efficiency.)
 ----
-// Compute average line length for stdin
 void main()
 {
     import std.range, std.stdio;
@@ -49,10 +49,9 @@ void main()
 }
 ----
 )
-$(EXTRA_EXAMPLE
+$(EXTRA_EXAMPLE Round floating point numbers,
 $(RUNNABLE_EXAMPLE_STDIN 2.4 plus 2.4 equals 5 for sufficiently large values of 2.)
 ----
-// Round floating point numbers
 import std.algorithm, std.conv, std.functional,
     std.math, std.regex, std.stdio;
 
@@ -71,7 +70,7 @@ void main()
 }
 ----
 )
-$(EXTRA_EXAMPLE
+$(EXTRA_EXAMPLE Sort lines,
 $(RUNNABLE_EXAMPLE_STDIN
 Mercury
 Venus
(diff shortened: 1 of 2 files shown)


wilzbach:
Alrighty, finally thanks to the minimal solution in #1775 this should be ready.

CyberShadow:
First: Diff looks great! Thanks for not giving up.

Dropdown looks a bit weird for me, any idea why?

wilzbach:
Okay, I think we can fix this by providing our own selector symbol from FontAwesome.
However, there are still UI differences:

Chrom(ium):

image

Firefox:

image

CyberShadow:
Alright, but why is that necessary? Why isn't the browser rendering the arrow?

wilzbach:
> Alright, but why is that necessary? Why isn't the browser rendering the arrow?

Without it looked really ugly in my Firefox:

image

CyberShadow:
> Without it looked really ugly in my Firefox:

That's because we're setting the background color. If we exclude that block from having its color and background color set, it should use the native styling... with the downside that it will look a little weird to people who have their browsers set to have a light-on-dark color theme by default. I don't think we need to worry about that, as long as the website is not unusable to them.

CyberShadow:
I guess either way works, but if possible, I think it would be better to not rely on font-awesome for the front page.

wilzbach:
>  If we exclude that block from having its color and background color set, it should use the native styling... with the downside that it will look a little weird to people who have their browsers set to have a light-on-dark color theme by default. I don't think we need to worry about that, as long as the website is not unusable to them.

That wouldn't look good:

image

I actually tried to be close to styling of the "Install X" boxes.

> I guess either way works, but if possible, I think it would be better to not rely on font-awesome for the front page.

FWIW we already rely on it:

image

AFAICT it's the only way to ensure a consistent experience across different browsers.

CyberShadow:
> That wouldn't look good:

No, that's not right. You're trying to undo the `background` rule with another `background` rule, right? That's not going to work. There must be no `background` rule in effect at all.

CyberShadow:
> FWIW we already rely on it:

Ah, that's right.

CyberShadow:
Seb do you want to try the no-CSS approach or should we go forward with the current FontAwesome version?

wilzbach:
> Seb do you want to try the no-CSS approach or should we go forward with the current FontAwesome version?

Hmm, the no-CSS approach would look like this:

image

image

>  or should we go forward with the current FontAwesome version?

I am fine with both ways (slight preference on FontAwesome because it keeps the styling consistent).

CyberShadow:
Needs some tweaking for mobile:

index.dd
```diff
@@ -481,7 +504,11 @@ Macros:
     TAG2=<$1 $2>$3</$1>
     D=<span class="d_inlinecode">$0</span>
     EXAMPLE=$(TAG2 a, id="a$1-control" class="example-control", )$(TAG2 div, id="a$1" class="example-box", $(RUNNABLE_EXAMPLE $2))
-    EXTRA_EXAMPLE=<div class="your-code-here-extra" style="display:none">$(RUNNABLE_EXAMPLE $0)</div>
+    EXTRA_EXAMPLE=<div class="your-code-here-extra" style="display:none">
+        <div class="your-code-here-title">$1</div>
```
andralex: DIVC?

wilzbach:
> Needs some tweaking for mobile:

Thanks. I came to the conclusion that tip box hasn't been used [much](http://forum.dlang.org/search?q=%22%5Byour+code+here%5D%22&search=Search) over the last years and especially users on their phone most likely won't post an example, so imho the best solution is also the simplest here: hiding the tip box on devices with a narrow layout.

CyberShadow:
Still broken for very narrow widths:

dtest dlang io-artifact-website-ec00ac7dd35faa307819ad634ab38bf91aabacb4-3f759d90a402ac3595f96e341659d3c6-web-index html

wilzbach:
> Still broken for very narrow widths:

I can't really reproduce this in Chrome or FF, but I added `max-width: 100%` which should fix things even for these tiny layouts:

image

@CyberShadow does it work for you now?

CyberShadow:
> @CyberShadow does it work for you now?

Looks perfect, thanks!

index.dd
```diff
@@ -110,6 +108,31 @@ $(SCRIPT
...
+        el.value = i;
+        el.innerText = title;
+        sel.appendChild(el);
+    });
+    sel.selectedIndex = rouletteIndex;
+    sel.addEventListener("change", function(e) {
+        examples[rouletteIndex].style.display = "none";
+        rouletteIndex = sel.selectedIndex;
+        examples[rouletteIndex].style.display = "block";
+    });
```
CyberShadow: The event handler could have been added later in the main JS file, but it's just 3 lines so no matter.

CyberShadow:
> tip box hasn't been used much over the last years

Forgot to post this earlier, but I think you simply found a bug in the forum search (or SQLite indexing). I know I fixed one yesterday (regarding deleted posts), but that search and its variations is producing inconsistent results with some variations of it (for one, the query needs to search the subject, not body).

PR #1780, opened 2017-06-28
Fix small makefile bug affecting string comparison in some shells

xi2:
Hi. Just getting started with D and decided to build a local copy of the docs for myself. I noticed a small makefile bug.

At least on my system (Debian), you need `=` and not `==` here. I believe some (but not all) shells accept both.

Thanks.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -620,13 +620,13 @@ $(ASSERT_WRITELN_BIN)_test: assert_writeln_magic.d $(DUB) $(STABLE_DMD)
 
 $(PHOBOS_FILES_GENERATED): $(PHOBOS_DIR_GENERATED)/%: $(PHOBOS_DIR)/% $(DUB) $(ASSERT_WRITELN_BIN)
 	@mkdir -p $(dir $@)
-	@if [ $(subst .,, $(suffix $@)) == "d" ] && [ "$@" != "$(PHOBOS_DIR_GENERATED)/index.d" ] ; then \
+	@if [ $(subst .,, $(suffix $@)) = "d" ] && [ "$@" != "$(PHOBOS_DIR_GENERATED)/index.d" ] ; then \
 		$(ASSERT_WRITELN_BIN) -i $< -o $@ ; \
 	else cp $< $@ ; fi
 
 $(PHOBOS_STABLE_FILES_GENERATED): $(PHOBOS_STABLE_DIR_GENERATED)/%: $(PHOBOS_STABLE_DIR)/% $(DUB) $(ASSERT_WRITELN_BIN)
 	@mkdir -p $(dir $@)
-	@if [ $(subst .,, $(suffix $@)) == "d" ] && [ "$@" != "$(PHOBOS_STABLE_DIR_GENERATED)/index.d" ] ; then \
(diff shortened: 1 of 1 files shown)


wilzbach:
> Hi. Just getting started with D 

Glad to hear this :)

> At least on my system (Debian), you need = and not == here. I believe some (but not all) shells accept both.

This is interesting. Thanks a lot for reporting & submitting the PR.

@CyberShadow what's your take on this? Should we fix the shell at the Makefile to avoid such errors?

```
SHELL := /bin/bash
```

CyberShadow:
Are these all the changes needed to successfully build dlang.org? I wouldn't be surprised if we depended on bashisms in a lot more places.

xi2:
Hi again. Thanks for the welcome. I didn't build all of dlang.org. I just ran the following, following the instructions on the wiki:

```
make -f posix.mak html phobos-release
```

The build completed actually without this PR. But I noticed some shell errors whizzing up the screen half way through and decided to investigate.

wilzbach:
> The build completed actually without this PR. But I noticed some shell errors whizzing up the screen

Hmm have you tried setting `SHELL := /bin/bash` in the beginning of the Makefile? If this works, it would be my preferred solution as then we would completely avoid such weird shell vs. bash issues in the future.

xi2:
hi. `SHELL := /bin/bash` at top of makefile does not fix the issue. For the record the errors look like this

```
/bin/sh: 1: [: d: unexpected operator
/bin/sh: 1: [: d: unexpected operator
/bin/sh: 1: [: d: unexpected operator
... etc
```

[This](https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html) part of the GNU make manual seemed to offer an explanation, suggesting you need `export SHELL:=/bin/bash`. However that didn't work either. Make seems very insistant on using `/bin/sh` it would seem :( Feel free to offer any more suggestions, and I'll try them.

xi2:
Sorry, my mistake. I got my log files mixed up. `SHELL:=/bin/bash` *does* fix the issue. No `export` needed. Apologies!

quickfur:
My 2 cents: instead of using hacks like `SHELL := /bin/bash`, what about just merging this PR to use the correct string comparison operator?

First of all, bash also understands it, so it doesn't hurt; second, this change increases the likelihood of correct interpretation by other shells. It's not all there yet (I know there are bashisms in the current makefiles), but IMO we should be moving _in_, rather than away from, the direction of better compatibilty with other shells.

Lastly, this PR is a tiny 2-line change, and it should not be held up like this(!).

CyberShadow:
> My 2 cents: instead of using hacks like SHELL := /bin/bash, what about just merging this PR to use the correct string comparison operator?

Well, for one thing, we don't test building anything with anything but the bash shell, so it's possible that setting `SHELL` may be a more complete solution. For another, we almost surely already use `bash` in other parts of the build process, so POSIX shell syntax compatibility may be irrelevant.

> Lastly, this PR is a tiny 2-line change, and it should not be held up like this(!).

The PR is uncontroversial, but is also not urgent, and it's something that does warrant some discussion. Open PRs are more visible in that regard.

But sure, let's merge this now.

PR #1781, opened 2017-06-29
Documentation examples: send JSON as text/plain to avoid an additional preflight OPTIONs request

wilzbach:
Putting this in the queue, s.t. once https://github.com/dlang-tour/core/pull/507 is deployed (there are currently some issues with the auto-deployment), we can simply flip the switch.

### DIFF
--- a/js/run.js
+++ b/js/run.js
@@ -92,7 +92,9 @@ var backends = {
   },
   tour: {
     url: "https://tour.dlang.org/api/v1/run",
-    contentType: "application/json; charset=utf-8",
+    // send json as text/plain to avoid an additional preflight OPTIONS request
+    // https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Preflighted_requests
+    contentType: "text/plain; charset=UTF-8",
     requestTransform: function(data) {
         return JSON.stringify({
             source: data.code


CyberShadow:
FWIW, I guess you could also use multipart/form-data with a application/json part (or just putting the stuff in the form data), then you wouldn't be technically be lying about the content type. But I guess it doesn't matter.

Please ping once this is unblocked.

wilzbach:
Alrighty - finally found time to trigger the manual deploy (not my authority, so I am trying to avoid this).

Currently:

image

Now:

image

My connection is quite undeterminstic, but the main point is that we can avoid the preflight request (between 100-600ms, depending on the latency of the connection).

Note that this is _without_ caching, with caching we should be able to serve the requests even faster :)
(-> https://github.com/dlang-tour/core/pull/545)

> FWIW, I guess you could also use multipart/form-data with a application/json part (or just putting the stuff in the form data), then you wouldn't be technically be lying about the content type. But I guess it doesn't matter.

Actually using "text/plain" makes the acceptance, quite easy (see .e.g https://github.com/dlang-tour/core/pull/507/files)
Btw there's also the option of adding `Access-Control-Max-Age`, but this won't help for the first request (and most browsers actively ignore the Max-Age value) and even GH accepts the `text/plain` lie for its API :)

CyberShadow:
Great stuff :)

PR #1782, opened 2017-06-29
Add check for broken Ddoc

wilzbach:
Follow-up automation to https://github.com/dlang/phobos/pull/5523

Can be enabled after https://github.com/dlang/phobos/pull/5526

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -639,6 +639,8 @@ test: $(ASSERT_WRITELN_BIN)_test all
 	@grep -n '[[:blank:]]$$' $$(find . -type f -name "*.dd") ; test $$? -eq 1
 	@echo "Searching for undefined macros"
 	@grep -n "UNDEFINED MACRO" $$(find $(DOC_OUTPUT_DIR) -type f -name "*.html" -not -path "$(DOC_OUTPUT_DIR)/phobos/*") ; test $$? -eq 1
+	@echo "Searching for undefined ddoc"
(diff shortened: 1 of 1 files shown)


CyberShadow:
```
grep: Unmatched [ or [^
```

quickfur:
Sigh. The shell is interpreting `$]` because you used double quotes rather than single quotes.

posix.mak
```diff
@@ -639,6 +639,8 @@ test: $(ASSERT_WRITELN_BIN)_test all
 	@grep -n '[[:blank:]]$$' $$(find . -type f -name "*.dd") ; test $$? -eq 1
 	@echo "Searching for undefined macros"
 	@grep -n "UNDEFINED MACRO" $$(find $(DOC_OUTPUT_DIR) -type f -name "*.html" -not -path "$(DOC_OUTPUT_DIR)/phobos/*") ; test $$? -eq 1
+	@echo "Searching for undefined ddoc"
+	@grep -rn "[$$](" $(DOC_OUTPUT_DIR)/phobos-prerelease ; test $$? -eq 1
```
quickfur: Change `"[$]("` to `'[$]('` to fix it.

quickfur: I'm not sure using `$$` is the right solution. That may just be another magic variable...

JackStouffer:
One would assume that this would be a feature in DMD.

quickfur:
File an enhancement request?

CyberShadow:
This needs to exclude .verbatim files, as those intentionally consist of DDoc macros.

wilzbach:
@CyberShadow  - this is finally green & passing :)

PR #1783, opened 2017-06-30
adjust .pitch's width to avoid excessive hyphenation

aG0aep6G:
Followup to #1765.

I don't know if it looks like this for everyone, but for me (Firefox, Ubuntu) the new text doesn't break nicely. Luckily, adding just 1em fixes it, and the box still fits in the layout.

Before/after:
spectacle c11827

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1372,7 +1372,7 @@ body#Home #content > .intro .pitch
     font-size: large;
     font-weight: 300;
     margin: auto;
-    width: 21em;
+    width: 22em;
 }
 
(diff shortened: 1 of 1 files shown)


wilzbach:
> I don't know if it looks like this for everyone, but for me

Well Chrom(ium) still doesn't like hyphenation:

image

CyberShadow:
Jeezus.

Just kill hyphenation for that block (or entirely ;)) and put a &shy; in "programming" or something. It's going to look broken *somewhere* with *some* window widths otherwise.

PR #1785, opened 2017-07-01
Update Areas of Usage

jansengb:
Remove link to std.experimental.ndslice.

### DIFF
--- a/areas-of-d-usage.dd
+++ b/areas-of-d-usage.dd
@@ -185,9 +185,8 @@ $(AREA_SECTION3 $(LNAME2 numerical, Numerical computing), $(ARTICLE_FA_ICON calc
     Additionally, due to its compile-time introspection capabilities, D enables
     further optimizations for superior performance. See Andrei's talk
     $(HTTPS www.youtube.com/watch?v=AxnotgLql0k, Fastware) on this topic.
-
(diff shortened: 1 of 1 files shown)


wilzbach:
I remember that there were some controversy about linking to mir, so leaving this to someone else to merge. Maybe @JackStouffer or @CyberShadow?

CyberShadow:
> I remember that there were some controversy about linking to mir

In the language spec. This is fine IMO.

CyberShadow:
Actually, this is not fine. The text now says that "D includes support for X in a third-party library". That's self-contradicting. Also, D supports a lot of various X-s through third-party libraries, no point in listing them all.

Just remove that paragraph, IMO.

jansengb:
Don't agree to not mentioning mir/ndslices just because there are many other third-party libraries. This library is fundamental for this usage area. How about "There is excellent support for ..."? I'll leave it to someone who can edit the page directly to fix it.

wilzbach:
@jansengb seems like you have trailing whitespace:

```
./areas-of-d-usage.dd:188:    There is excellent support for multidimensional arrays with third-party libraries  
posix.mak:639: recipe for target 'test' failed
```

wilzbach:
> @jansengb seems like you have trailing whitespace:

Removed the trailing whitespace and squashed, s.t. we can move forward here.

PR #1787, opened 2017-07-02
Improve qualifier documentation

andralex:
Add a section on qualifier combinations. Also added graph illustrations.


spec/const3.dd
```diff
@@ -397,27 +397,98 @@ $(H2 $(LNAME2 const_member_functions, Const Member Functions))
...
+    "shared T" -> "const shared T" [label="const"];
+    "shared T" -> "inout shared T" [label="inout"];
+    "const inout T" -> "const inout shared T" [label="shared"];
+    "const shared T" -> "const inout shared T" [label="inout"];
+    "inout shared T" -> "const inout shared T" [label="const"];
+}
+)
 
-$(H2 $(LNAME2 implicit_conversions, Implicit Conversions))
+<center>
```
wilzbach: FWIW
1 ) [`center` isn't supported in HTML5](https://www.w3schools.com/tags/tag_center.asp), there's `$(DIVC center)` but please see point (2)
2) The spec is also generated to PDF/ebook - there are already many issues with people using HTML (see [Bugzilla](https://issues.dlang.org/buglist.cgi?component=dlang.org&list_id=215655&product=D&resolution=---)). 

This is how it will look in the PDF:

image

Unfortunately there's no precedent of images in the spec and you most likely need to add the Ddoc macros for images yourself.

andralex: Great catch, thanks. Do we have a macro for "HTML only?" In that case I can insert the HTML-specific parts inside that macro. I did define LATEX to only work for LaTeX generation.

wilzbach: > Great catch, thanks. Do we have a macro for "HTML only?" 

Yes, see https://github.com/dlang/dlang.org/pull/1788
Though it be weird to reference to a non-existing graphic in the PDF spec ...

andralex:
@wilzbach awesome, thanks. Somewhat miraculously a simple solution exists for the LaTeX version as well - see the new commit.

wilzbach:
>  Somewhat miraculously a simple solution exists for the LaTeX version as well - see the new commit.

Not so sure - you would need to define `LATEX` in `dlang.org.ddoc`, due to many people using undefined macros, we check for them since a couple of weeks:

```
/dev/shm/dtest/work/repo/dlang.org/web/spec/const3.html:425:<!--UNDEFINED MACRO: "LATEX"-->
/dev/shm/dtest/work/repo/dlang.org/web/spec/const3.html:441:<!--UNDEFINED MACRO: "LATEX"-->
```

andralex:
@wilzbach oh, of course. I thought it was defined already - apparently it wasn't used anywhere yet.

CyberShadow:
Could you please add the GraphViz source files for the graphs? Just place them in the `images` directory alongside the renders.

andralex:
@CyberShadow cool, done

PR #1791, opened 2017-07-03
Fix auto-deploy due to missing osmodel.mak

wilzbach:
See also: https://github.com/dlang/phobos/pull/5534#issuecomment-312745191

CC @CyberShadow


CyberShadow:
Thing is that if this fix was needed, then DAutoTest should be failing. But it's not. Argh.

CyberShadow:
ARGH. I know what it is.

`git clean -fdx` ignores submodules, and we put an entire repo in there

CyberShadow:
Log looks good so far!

wilzbach:
> Log looks good so far!

:tada: - glad that it worked :)

CyberShadow:
Thanks for the quick fix!

andralex:
thanks for the frenetic pace you guys

PR #1794, opened 2017-07-04
Update struct initializer docs

ntrel:
* Add member initializer example.
* `memberName:expression` can be used for non-static variables too.
* Move non-static struct initialization paragraph to dynamic initialization.
* Minor tweaks.

### DIFF
--- a/spec/struct.dd
+++ b/spec/struct.dd
@@ -28,8 +28,8 @@ $(SPEC_S Structs and Unions,
         $(CHECK), $(UNCHECK), $(CHECK), $(CHECK))
         $(TROW static members, $(CHECK), $(CHECK), $(UNCHECK),
         $(CHECK), $(CHECK))
-        $(TROW default member initializers, $(CHECK), $(CHECK),
-        $(UNCHECK), $(UNCHECK), $(UNCHECK))
+        $(TROW $(RELATIVE_LINK2 static_struct_init, default member initializers),
+        $(CHECK), $(CHECK), $(UNCHECK), $(UNCHECK), $(UNCHECK))
         $(TROW bit fields, $(UNCHECK), $(UNCHECK), $(CHECK), $(CHECK),
         $(CHECK))
         $(TROW non-virtual member functions, $(CHECK), $(CHECK),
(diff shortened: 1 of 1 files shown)


spec/struct.dd
```diff
@@ -181,23 +189,12 @@ static S q = { 1, 2 }; // q.a = 1, q.b = 2, q.c = 0, q.d = 7
 ------
 
         $(P Struct literals can also be used to initialize statics, but
-        they must be evaluable at compile time.)
+        they must be evaluatable at compile time.)
```
ntrel: Changed for consistency with other pages on dlang.org.

andralex: We need to replace "evaluatable" (which is not a word) with "evaluable" (which is) throughout our docs:

https://www.merriam-webster.com/dictionary/evaluatable
https://www.merriam-webster.com/dictionary/evaluable

JohanEngelen:
Thanks.
What I am missing (what trigger my forum post) is explicit mention of these two kinds of examples:

```d
static S x = {1,  2}; // <-- member:value  syntax is not the only way for statics
S y = { a:1 }; // <-- member:value is also allowed for non-statics
```

spec/struct.dd
WalterBright: A PR should not insert gratuitous errors to be fixed in a later PR, regardless of whether other pages do it. Please fix.

WalterBright: And the other instances in the context lines here. These lines are being changed anyway.

WalterBright: The rest is good to go.

ntrel:
@JohanEngelen C-style static initialization `static S q = { 1, 2 };` was already in the spec. The last commit now has `S q = { 1, b:i };`.

@WalterBright Fixed using *evaluable*. I also added an example showing that C-style initialization and named member initialization can be combined. Also I reduced an uncalled `S.opCall(S)` in a later example to just `assert(0);` for clarity (this is part of the struct initialization docs too).

JohanEngelen:
Nice.

ntrel:
@wilzbach great, thanks 😊

PR #1797, opened 2017-07-05
Look ma, no $(P ...) - a test drive of <p>-less pages

andralex:
I've eliminated all $P macros in modules.dd and reflown the text a bit. No content changes. I don't plan to convert all of the spec right now, but this should serve as an example and field test.

### DIFF
--- a/spec/module.dd
+++ b/spec/module.dd
@@ -37,18 +37,14 @@ $(GNAME DeclDef):
     $(D ;)
 )
 
+Modules have a one-to-one correspondence with source files. The module name is,
+by default, the file name with the path and extension stripped off, and can be
+set explicitly with the module declaration.
 
-	$(P Modules have a one-to-one correspondence with source files.
-	The module name is, by default, the file name with the path and
-	extension stripped off, and can be set explicitly with the module
-	declaration.
-	)
+Modules automatically provide a namespace scope for their contents. Modules
+superficially resemble classes, but differ in that:
 
-	$(P Modules automatically provide a namespace scope for their contents.
-	Modules superficially resemble classes, but differ in that:
-	)
-
-	$(UL
+$(UL
 	$(LI There's only one instance of each module, and it is
 	statically allocated.)
 
@@ -62,14 +58,13 @@ $(GNAME DeclDef):
 
 	$(LI Modules are always compiled at global scope, and are unaffected
 	by surrounding attributes or other modifiers.)
-	)
+)
 
-	$(P Modules can be grouped together in hierarchies called $(I packages).
-	)
+Modules can be grouped together in hierarchies called $(I packages).
 
-	$(P Modules offer several guarantees:)
+Modules offer several guarantees:
 
-	$(UL
+$(UL
 
 	$(LI The order in which modules are imported does not affect the
 	semantics.)
@@ -80,14 +75,13 @@ $(GNAME DeclDef):
 	$(LI If a module C imports modules A and B, any modifications to B
 	will not silently change code in C that is dependent on A.)
 
-	)
+)
 
 $(H3 $(LNAME2 module_declaration, Module Declaration))
 
-	$(P The $(I ModuleDeclaration) sets the name of the module and what
-	package it belongs to. If absent, the module name is taken to be the
-	same name (stripped of path and extension) of the source file name.
-	)
+The $(I ModuleDeclaration) sets the name of the module and what package it
+belongs to. If absent, the module name is taken to be the same name (stripped of
+path and extension) of the source file name.
 
 $(GRAMMAR
 $(GNAME ModuleDeclaration):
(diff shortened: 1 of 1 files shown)


adamdruppe:
just fyi i did this like 18 months ago with adrdox (in fact, as you might recall, the horrible brokeness of ddoc's paragraph handling was actually the straw that broke the camel's back with my decision to start from scratch).

Here's a random page from the website with full conversion to my syntax:
http://dpldocs.info/experimental-docs/source/built_in_rationale.d

And the same one in an earlier stage with just the P ones removed, but the lists still done the traditional way:
http://dpldocs.info/experimental-docs/source/dlang.articles.built_in_rationale.d.html

These little things really do make a pretty big difference in ease of editing - just fewer random (yet often manual) reflows and stray parens to worry about.

andralex:
@adamdruppe thanks. Neat lists! Did you do something about tables as well?

CyberShadow:
@andralex I had a chat with Stefan on IRC yesterday and I asked him how hard would it be to implement an `EVAL` macro so that:

```
Ddoc
$(JOIN_WITH, _, a, b ,c )
Macros:
DDOC=$(BODY)
JOIN_WITH=$(EVAL (string[] args) { import std.array; return args[1..$].join(args[0]); })
```

results in `a_b_c`. He gave me an estimate of 20 minutes for a quick hack on the CTFE-side.

What do you think of the idea? On one hand, it would allow us to do *so much* more with just DDoc; on the other hand, abusing it would put alternative document generators in an unfair position.

(BTW, the above example seems like it should be easy to do with DDoc, but it's actually impossible due to the special case of one argument - AFAICS you can't implement it so that `$(JOIN_WITH, _, a)` returns exactly `a`! And never even mind the special case of zero arguments.)

wilzbach:
> What do you think of the idea? On one hand, it would allow us to do so much more with just DDoc; on the other hand, abusing it would put alternative document generators in an unfair position.

I love it! It would prevent _a lot_ of Ddoc hacks, maybe other documentation generators can precompile the EVAL macros like it is done for Diet templates?

CC @s-ludwig

adamdruppe:
On Wed, Jul 05, 2017 at 04:59:08AM -0700, Andrei Alexandrescu wrote:
> @adamdruppe thanks. Neat lists! Did you do something about tables as well?

Yeah. To test that, I did the std.allocator page:

http://dpldocs.info/experimental-docs/source/std.experimental.allocator.building_blocks.d.html

(though I am likely to revert that specific file just because I am sick of the merge conflicts from upstream!)

Syntax is described here:
http://dpldocs.info/experimental-docs/adrdox.syntax.html

tables here:
http://dpldocs.info/experimental-docs/adrdox.syntax.html#tables

s-ludwig:
> implement an EVAL macro

Ugh, that would make real-time generation of documentation, like I'm currently doing with the embedded variant of DDOX, a no-go due to the performance implications. To me this seems like a pretty huge hammer for a simple thing like documentation generation.

adamdruppe:
On Wed, Jul 05, 2017 at 04:13:55PM +0000, Sönke Ludwig wrote:
> To me this seems like a pretty huge hammer for a simple thing like documentation generation.

Yeah, I'm against EVAL too. Heck, in adrdox, I even removed user-defined macros. I sometimes regret that and have considered adding my script interpreter to it, but then I look back at the official dlang repository and see the confusing mess of macros defined all over the place with very little consistency and am reminded "oh yeah, THAT'S why I killed it".

With EVAL, you'd have the mess of macros AND whatever else comes out with a very complicated implementation. While there's potential to simplify and unify macros with D functions, I fear it would in reality make things even worse than it is now to read the sources AND break every third party doc gen.

CyberShadow:
We already run D code as part of the dlang.org build process, no going around this, and the amount of D code involved is only going to grow. Moving it into DDoc itself is not going to change whether we run D code or not during the build, but it will certainly make a lot of things much easier, and not require many little helper programs that try to parse HTML DOMs or use crude regex hacks to do things that DDoc itself can't.

adamdruppe:
On Wed, Jul 05, 2017 at 09:28:14AM -0700, Vladimir Panteleev wrote:
> Moving it into DDoc itself is not going to change whether we run D code or not during the build, but it will certainly make a lot of things much easier

Some things, sure, but it will also likely not be able to do everything you want anyway, meaning those external D programs are still necessary.

D code inside ddoc will still have ddoc's limitations, and probably CTFE's limitations too. That means you (probably) won't be able to download external files to construct the homepage, or read other pages to create a cross-reference or table of contents. Those things will still be done externally so the EVAL will fail to significantly simplify the website build while still complicating the ddoc implementation.

andralex:
@CyberShadow there is something disquieting about adding this much unstructured power to a macro engine. There's this nice article about it: http://www.cs.usfca.edu/~parrt/papers/mvc.templates.pdf.

CyberShadow:
Fair enough!

PR #1799, opened 2017-07-05
Only use DPaste as backend if absolutely necessary (example as STDIN or ARGS)

wilzbach:
image

I didn't account for the fact that DPaste would throw errors with the new invocation example.
Let's use the DTour then whenever possible.

The tour works fine: https://is.gd/DTFINJ

For reference STDIN/ARGS support for the tour is already in the queue: https://github.com/dlang-tour/core-rdmd/pull/9

### DIFF
--- a/js/run.js
+++ b/js/run.js
@@ -209,8 +209,10 @@ $(document).ready(function()
     $('textarea[class=d_code]').each(function(index) {
         var parent = $(this).parent();
         var outputDiv = parent.children("div.d_code_output");
+        var hasStdin = parent.children(".inputButton").length > 0;
+        var hasArgs  = parent.children(".argsButton").length > 0;
         setupTextarea(this, {parent: parent, outputDiv: outputDiv,
-                        stdin: true, args: true});
+                        stdin: hasStdin, args: hasArgs});
     });
 });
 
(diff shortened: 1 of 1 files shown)


CyberShadow:
BTW, I didn't really think about it in this way until now, but a front page example of invoking external programs is essentially an invitation for anyone to poke at the sandbox's security :)

I don't know much about how it's set up, but is it protected against e.g. denial of service attacks (exhaustion of some resource or other), as well as against being used to cause denial of service on other services?

wilzbach:
> I don't know much about how it's set up, 

Good question - @stonemaster configured the box and deployment, so I try to answer what I know:
- The examples are run as docker within docker (so I guess in terms of sandboxing we are safe - minus Docker bugs, of course)
- The entire application is on a dedicated VPS for the DTour

> but is it protected against e.g. denial of service attacks (exhaustion of some resource or other), as well as against being used to cause denial of service on other services?

I think the amount if docker instance that can be spawned is limited to 10 or 20, but I have no idea what happens on a Ddoc attack. I will look into setting up Nginx + simple rate-limiting

https://github.com/dlang-tour/core/issues/568

PR #1801, opened 2017-07-06
Add a version chooser to DDoc pages

wilzbach:
From the [NG](http://forum.dlang.org/post/niuatqnkrygvunacwjti@forum.dlang.org):

> IMO there should be a link under the "switch to pre-release." link in the current docs.
> Maybe replace it with a dropdown to select the version from 2.066 to pre-release?

CC @JackStouffer @CyberShadow 

image

image

image

It's not perfect yet (e.g. I need to detect the current version automatically and add the same for the Ddox build), but I wanted to get your opinion on this before I invest more work.

### DIFF
--- a/js/listanchors.js
+++ b/js/listanchors.js
@@ -88,6 +88,69 @@ function listanchors()
         id += key;
         var e = document.getElementById(id);
         e.innerHTML = newText;
+
+    }
+    function addVersionSelector() {
+      // Latest version offered by the archive builds
+      // This needs to be manually updated after new versions have been archived
+      var currentArchivedVersion = 74;
+      // build URLs for dlang.org: DDoc + Dox
+      var ddocModuleURL = document.body.id.replace(/[.]/g, "_") + ".html";
+      var ddoxModuleURL = document.body.id.replace(/[.]/g, "/") + ".html";
+
+      // build list of versions available in the docarchives
+      var archivedVersions = [];
+      while (currentArchivedVersion >= 66) {
+        archivedVersions.push("2.0" + currentArchivedVersion--);
+      }
+      archivedVersions = archivedVersions.map(function(e) {
+          return {
+            name: e,
+            url: "https://docarchives.dlang.io/v" + e + ".0/phobos/" + ddocModuleURL,
+            selected: false,
+          };
+      });
+
+      var rootURL = location.href.split(/\/(phobos|library)(-prerelease)?/)[0]
+      var onlineVersions = [{
+        name: "master",
+        url: rootURL + "/phobos-prerelease/" + ddocModuleURL,
+      },{
+        name: "master (ddox)",
+        url: rootURL + "/library-prerelease/" + ddoxModuleURL,
+      },{
+        name: "stable",
+        url: rootURL + "/phobos/" + ddocModuleURL,
+      },{
+        name: "stable (ddox)",
+        url: rootURL + "/library/" + ddoxModuleURL,
+      }];
+
+      // set the current URL as selected
+      var currentURL = location.href.split(/[#?]/)[0];
+      onlineVersions.forEach(function(v, i) {
+        onlineVersions[i].selected = v.url === currentURL;
+      });
+      // Don't show the option chooser if the page hasn't been recognized
+      // For example, Ddox symbol pages are currently not supported
+      if (onlineVersions.filter(function(v){return v.selected}).length === 0)
+        return;
+
+      // build select box of all versions and append to current DOM
+      var versions = onlineVersions.concat(archivedVersions);
+      var options = versions.map(function(e, i){
+        return "<option value='" + i + "'" + (e.selected ? "selected" : "") + ">" + e.name + "</option>";
+      });
+      $("h1").after("<div class='version-changer-container fa-select'><select id='version-changer'>" + options.join("") + "</select></div>");
+      // attach event listener to select box -> change URL
+      $("#version-changer").change(function(){
+        var selected = parseInt($(this).find("option:selected").val());
+        var option = versions[selected];
+        if (!option.selected) {
+          window.location.href = option.url;
+        }
+      });
     }
     addAnchors();
+    addVersionSelector();
 }
--- a/css/style.css
+++ b/css/style.css
@@ -1770,7 +1770,7 @@ div.openInEditorButton {
 #your-code-here-select-example {
     padding-bottom: 0.3em;
 }
-#your-code-here-select-example select {
+.fa-select select {
     background: white;
     border: 1px solid #CCC;
     border-radius: 4px;
@@ -1780,6 +1780,11 @@ div.openInEditorButton {
     -moz-appearance: none;
     appearance: none;
     max-width: 100%;
+}
+.fa-select::after {
+   font-family: FontAwesome;
+   content: "\f0d7";
+   margin-left: -15px;
 }
 /* Runnable-examples css -end */
 
--- a/css/style.css
+++ b/css/style.css
@@ -2226,3 +2226,19 @@ dt.d_decl:hover .decl_anchor {
     text-decoration: none;
     color: #333;
 }
+
+/**
+ * Version select box
+ */
+.version-changer-container {
+    float: right;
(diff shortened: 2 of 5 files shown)


CyberShadow:
The dropdown will be absent in your old builds, right? So it will only be present on dlang.org?

wilzbach:
> The dropdown will be absent in your old builds, right? So it will only be present on dlang.org?

For now, yes, though we could always patch the archived files.

CyberShadow:
Needs a bit of margin to the left:
 

Why place it inside the content and not to the right of the module title?

CyberShadow:
@andralex You've been asking for this since forever, what do you think of this implementation?

CyberShadow:
Being able to go to the *corresponding* page between DDoc/DDox is awesome! But it seems a little buggy: for http://dtest.dlang.io/artifact/website-191526224e06ec8685ef43831cbe43983d14ec25-b0341c56fbd877b803c357bf95312672/web/library/core/atomic/atomic_fence.html the dropdown has "stable" selected by default, and if I select ddox again, it goes to a 404.

andralex:
Hook, line, and sinker! But there should be no ddox vs ddoc distinction. Instead, each should have a version chooser. No? Thanks @wilzbach!!

wilzbach:
> Why place it inside the content and not to the right of the module title?

I did - just get a proper browser ;-)

> Being able to go to the corresponding page between DDoc/DDox is awesome! But it seems a little buggy: 

I never intended it to work on DDox - cleaned up the code and made it Ddoc/Ddox agnostic :)

> Hook, line, and sinker! But there should be no ddox vs ddoc distinction. Instead, each should have a version chooser. No? 

That's actually one of the feature that I have missed myself for a long line. It's extremely handy and raises the awareness of Ddox. OTOH even if we switch to Ddox as default, which we hopefully will do within the next months, people still might want to switch back to Ddoc.
However, I renamed it from "(ddox)" to "(beta)", s.t. people aren't that confused.

CyberShadow:
> I did - just get a proper browser ;-)

What counts for a proper browser, then?

wilzbach:
One that behaves as my local ones - or in other words "one that doesn't have bugs" :smiling_imp: 

Chrom(ium):

image

Firefox:

image

Maybe you have more luck with the latest version.

js/listanchors.js
```diff
@@ -88,6 +88,65 @@ function listanchors()
...
+      onlineVersions.forEach(function(v, i) {
+        onlineVersions[i].selected = v.url === currentURL;
+      });
+
+      // build select box of all versions and append to current DOM
+      var versions = onlineVersions.concat(archivedVersions);
+      var options = versions.map(function(e, i){
+        return "<option value='" + i + "'" + (e.selected ? "selected" : "") + ">" + e.name + "</option>";
+      });
+      $("h1").after("<div class='version-changer-container fa-select'><select id='version-changer'>" + options.join("") + "</select></div>");
```
CyberShadow: I'm pretty sure you're supposed to put the element you're floating *before* the element it is supposed to "overlap", not `after`.

Why does this work for you?

wilzbach: > I'm pretty sure you're supposed to put the element you're floating before the element it is supposed to "overlap", 

I tried `before` as well, but discarded the approach due as floating didn't look nice for mobile devices. Though of course, now I already added viewport-specific CSS.
For reference, currently it looks like this:

image

With `before` it could only look like this:

image

Though of course, we could simple hide the select box on mobile devices.

> Why does this work for you?

```css
.version-changer-container {
    margin-top: -4em;
}
```

The interesting question is why it doesn't work for ;-)
A couple of external tools show the same results:

https://tools.pingdom.com/#!/bpOIFJ/http://dtest.dlang.io/artifact/website-b29c592343b3eca4d1a2bb7774123c738e67b788-7991f163dc4768833cc161ca4595703d/web/phobos-prerelease/core_bitop.html
https://www.webpagetest.org/result/170707_0E_8F6/

Google even allows to preview the mobile viewport (this one has special CSS - no floating):

https://developers.google.com/speed/pagespeed/insights/?url=http%3A%2F%2Fdtest.dlang.io%2Fartifact%2Fwebsite-b29c592343b3eca4d1a2bb7774123c738e67b788-7991f163dc4768833cc161ca4595703d%2Fweb%2Fphobos-prerelease%2Fcore_bitop.html

CyberShadow:
Oh, you fixed it!

Old (broken): http://dtest.dlang.io/artifact/website-191526224e06ec8685ef43831cbe43983d14ec25-b0341c56fbd877b803c357bf95312672/web/library/core/atomic/atomic_fence.html
New (works): http://dtest.dlang.io/artifact/website-b29c592343b3eca4d1a2bb7774123c738e67b788-7991f163dc4768833cc161ca4595703d/web/library-prerelease/core/atomic/atomic_fence.html

CyberShadow:
Well that was confusing. Sneakily fixing it and then blaming my browser. How much is Google paying you, eh ;)

http://dtest.dlang.io/artifact/website-b29c592343b3eca4d1a2bb7774123c738e67b788-7991f163dc4768833cc161ca4595703d/web/library/std/algorithm/comparison/among.html doesn't recognize it's on a DDox page. Selecting the "beta" option takes you to the module's page.

I noticed you changed "DDox" to "master (beta)". I was going to say something about the DDox name being an implementation detail, but I figured that space is tight anyway and it might not hurt to let it build up a brand name... However "master (beta)" may be confusing because it's not clear that it's a beta of the documentation generator, and not of D itself.

wilzbach:
> Well that was confusing. Sneakily fixing it and then blaming my browser. How much is Google paying you, eh ;)

Not enough!
I [did state this](https://github.com/dlang/dlang.org/pull/1801#issuecomment-313567140):

>> Maybe you have more luck with the latest version.

> http://dtest.dlang.io/artifact/website-b29c592343b3eca4d1a2bb7774123c738e67b788-7991f163dc4768833cc161ca4595703d/web/library/std/algorithm/comparison/among.html doesn't recognize it's on a DDox page. Selecting the "beta" option takes you to the module's page.

Hmm symbol pages are currently not supported as I wanted to go for the the simplest, working solution for now. Detecting the current symbol shouldn't be that hard - it's essentially a one-line

```js
var symbol = (location.href.split(ddoxModuleURL.replace(".html", ""))[1] || "").replace("/", "").split(".html")[0];
```

But looking at the output it seems that more CSS tweaking would be required:

image

Hence, I would leave showing the select box on a symbol page for a future improvement and simply hide the select box if the detection logic failed.

> I noticed you changed "DDox" to "master (beta)". I was going to say something about the DDox name being an implementation detail, but I figured that space is tight anyway and it might not hurt to let it build up a brand name... However "master (beta)" may be confusing because it's not clear that it's a beta of the documentation generator, and not of D itself.

Fair point, [I only changed it](https://github.com/dlang/dlang.org/pull/1801#issuecomment-313563540) because Andrei didn't like "Ddox" at all. I wasn't a huge friend of "beta" -> changing it back then.

CyberShadow:
> Maybe you have more luck with the latest version.

Oops, I thought you meant the latest version of the web browser! I'm using Firefox Developer.

> Fair point, I only changed it because Andrei didn't like "Ddox" at all. I wasn't a huge friend of "beta" -> changing it back then.

Yeah, I can't think of better solutions right now.

I guess "beta" + a banner on DDox pages explaining why it's in beta and where to report issues would work, but it seems like a relatively minor issue with no obviously correct solution anyway.

wilzbach:
> I guess "beta" + a banner on DDox pages explaining why it's in beta and where to report issues would work, but it seems like a relatively minor issue with no obviously correct solution anyway.

Yup, especially because we haven't shown such a banner before, so I guess I leave finding a "correct solution" up to another PR.

So are we good to go here @andralex?

andralex:
That the space is tight doesn't make "ddox" any better. It's an internal name that tells nothing to the casual observer. I'll approve this but definitely we need to revisit the matter in the future.

CyberShadow:
@andralex I think we are all in agreement, just lacking a better solution.

PR #1802, opened 2017-07-06
chmgen stuff, part 1

CyberShadow:
Fixes broken links and prepares things for Phobos fixes and part 2.

### DIFF
--- a/changelog/2.006.dd
+++ b/changelog/2.006.dd
@@ -31,8 +31,7 @@ $(WHATSNEW
 	$(LI Added the
 	$(LINK2 $(ROOT_DIR)phobos/std_variant.html, std.variant) module.)
 	$(LI Incorporated many of the Tango GC structural differences (much more to go still).)
-	$(LI Added the
-	$(LINK2 $(ROOT_DIR)phobos/std_contracts.html, std.contracts) module.)
+	$(LI Added the $(D std.contracts) module.)
 	$(LI Breaking change: $(CODE std.stdio.writef) can now only accept a format as
 	its first argument.)
 )
--- a/chmgen.d
+++ b/chmgen.d
@@ -14,6 +14,7 @@ import std.regex;
 import std.path;
 
 string docRoot = `.`;
+string chmDir = "chm";
 
 // ********************************************************************
 
@@ -97,15 +98,16 @@ void main(string[] args)
     getopt(args,
         "only-tags", &onlyTags,
         "root", &docRoot,
+        "dir", &chmDir,
     );
 
     bool chm = !onlyTags;
 
     if (chm)
     {
-        if (exists(`chm`))
-            rmdirRecurse(`chm`);
-        mkdirRecurse(`chm/files`);
+        if (exists(chmDir))
+            rmdirRecurse(chmDir);
+        mkdirRecurse(chmDir ~ `/files`);
     }
 
     enforce(exists(docRoot ~ `/phobos/index.html`),
@@ -131,7 +133,7 @@ void main(string[] args)
 
         pages[fileName] = page;
 
-        auto outPath = `chm/files/` ~ fileName;
+        auto outPath = chmDir ~ `/files/` ~ fileName;
         if (chm)
             outPath.dirName().mkdirRecurse();
 
@@ -273,7 +275,7 @@ void loadNavigation()
                         return null;
                     }
                     else
-                    if (!exists(`chm/files/` ~ url))
+                    if (!exists(chmDir ~ `/files/` ~ url))
                     {
                         if (/*warn*/true)
                             stderr.writeln("Warning: Item in navigation does not exist: " ~ url);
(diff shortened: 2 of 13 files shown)


changelog/2.006.dd
```diff
@@ -31,8 +31,7 @@ $(WHATSNEW
 	$(LI Added the
 	$(LINK2 $(ROOT_DIR)phobos/std_variant.html, std.variant) module.)
 	$(LI Incorporated many of the Tango GC structural differences (much more to go still).)
-	$(LI Added the
-	$(LINK2 $(ROOT_DIR)phobos/std_contracts.html, std.contracts) module.)
+	$(LI Added the $(D std.contracts) module.)
```
ZombineDev: So, AFAIU, all links to now non-existing modules, should be changed to plain code font.

posix.mak
```diff
@@ -79,7 +79,9 @@ MOD_EXCLUDES_PRERELEASE=$(addprefix --ex=, gc. rt. core.internal. core.stdc.conf
 	std.algorithm.internal std.c. std.concurrencybase std.internal. std.regex.internal.  \
 	std.windows.iunknown std.windows.registry etc.linux.memoryerror	\
 	std.experimental.ndslice.internal std.stdiobase \
-	tk. msvc_dmc msvc_lib)
+	tk. msvc_dmc msvc_lib \
+	ddmd.libmach ddmd.libmscoff ddmd.objc_glue \
+	ddmd.scanmach ddmd.scanmscoff)
```
CyberShadow: I had a quick look at whether it was possible to enable them. At least in the case of the OS X stuff, it was importing stuff from Druntime, and those modules had a `version(OSX):` on top.

I'm not sure if that makes sense for things such as type declarations; I have a Druntime PR open to make the Windows headers usable on other platforms (for things such as .bmp and .exe/.dll file formats), but that grew the size of the built library file by a disproportionate amount due to all the `.init` / `TypeInfo` clutter...

changelog/2.006.dd
ZombineDev: > I guess we could link to Seb's new doc archive, but I'm not sure how stable those links would be.

If we can upstream Seb's doc archive, then this certainly sounds like a very good idea.

posix.mak
ZombineDev: > I had a quick look at whether it was possible to enable them. At least in the case of the OS X stuff, it was importing stuff from Druntime, and those modules had a `version(OSX):` on top.

Yeah, I expected as much.

> I'm not sure if that makes sense for things such as type declarations; I have a Druntime PR open to make the Windows headers usable on other platforms (for things such as .bmp and .exe/.dll file formats), but that grew the size of the built library file by a disproportionate amount due to all the .init / TypeInfo clutter...

Disregarding the bloat issue for a moment, I certainly find this valuable. (I also wrote my own bmp file reader/writer, though in my case writing the definitions by hand had high educational value :D) Also, I would love to be able to check dlang.org if a certain definition is present in `core.sys.windows`. OTOH, `core.sys.windows` being probably the largest D subpackage should better be excluded from the sidebar.

I faintly remember reading about the .init / TypeInfo code bloat issue on your PR. Is there a bug report on that? I would certainly expect those modules to behave like header-only libraries which don't get compiled.

changelog/2.006.dd
wilzbach: > If we can upstream Seb's doc archive, then this certainly sounds like a very good idea.

Do you mean moving the repo to `dlang`? Or moving it under the `dlang.org` domain? I wouldn't mind about both. After all, it's just a raw dump (deployed on netlify)

wilzbach: > I guess we could link to Seb's new doc archive, but I'm not sure how stable those links would be.

Well in this I think `std.contracts` was gone way before `2.066`. Of course, we could try to rebuild more history, but I am not sure whether that's worth it.

chmgen.d
```diff
@@ -63,6 +64,9 @@ void addKeyword(string keyword, string link, int confidence, bool isAnchor = tru
     if (link.endsWith("/"))
         link ~= "index.html";
 
+    if (prerelease && link.startsWith("phobos/"))
```
wilzbach: nit: could be `link.skipOver("phobos/")` - then you would save the `link[7 .. $]` below.

dlang.org.ddoc
```diff
@@ -214,6 +214,7 @@ MULTICOLS=$+
 MULTIROW_HEADER=<th rowspan="$1">$+</th>
 MREF=<a href="$(PHOBOS_PATH)$1$(UNDERSCORE_PREFIXED $+).html">$(D $1$(DOT_PREFIXED $+))</a>
 MREF_ALTTEXT=<a href="$(PHOBOS_PATH)$2$(UNDERSCORE_PREFIXED_SKIP $+).html">$1</a>
+MREF1=<a href="$(PHOBOS_PATH)$1.html">$(D $1)</a>
```
wilzbach: Urgh, not a huge friend of more Ddoc macros, but I see the Ddoc limitations (we should really push Ddox forward).

posix.mak
CyberShadow: > Is there a bug report on that?

This one is pretty close: https://issues.dlang.org/show_bug.cgi?id=15130

CyberShadow:
> If you want feel free to fix the nit, otherwise merge ahead ;-)

Took me all of 20 seconds :)

PR #1803, opened 2017-07-06
fix border-radius for "Why D?" boxes

aG0aep6G:
Just noticed this little defect. I guess it has been like this from the start. Wasn't meant that way.

Before/after (zoomed in):
spectacle n12854


wilzbach:
(tested this locally & it's very simply CSS, so giving this a go directly)

PR #1804, opened 2017-07-06
std-ddox: Define REF1 macro variants

CyberShadow:
Fixup for #1802 and dlang/phobos#5554.


wilzbach:
Impressed that the Ddox magic works and sorry for not catching this single instance before.

CyberShadow:
np :)

PR #1805, opened 2017-07-06
Insert the actual ddoc file containing predefined macros instead of d…

andralex:
…uplicating it

Automation FTW.


CyberShadow:
Duplicate of https://github.com/dlang/dlang.org/pull/1703 ?

andralex:
Looks like a superset. I'd say we either pull this, or remove the content from the doc page and replace it with a link to the actual file.

CyberShadow:
I think an earlier version of #1703 did pretty much what this does, but not the current diff.

wilzbach:
Btw if we go with #1688, we could easily insert the exact file from DMD here and avoid it getting outdated in the future.

andralex:
Push this and close https://github.com/dlang/dlang.org/pull/1703?

andralex:
Rebase. Let's take action on this soon lest it's liable to rot every time something changes in the ddoc default - which is what it's supposed to :)

andralex:
It doesn't preclude a more radical approach a la https://github.com/dlang/dlang.org/pull/1688 but it does make progress while we mull over that.

CyberShadow:
Essentially I don't think we should put that huge DDoc template in our specs. It's huge, likely to change (as it's stylistic in nature), and overshadows the definitions that one might actually look at.

I've said this before, and IMO the default DDoc template change wasn't done as well as it could have. The built-in definitions should have stayed unchanged; instead, the compiler could the include an additional `.ddoc` file by default, that's available separately.

spec/ddoc.src.dd
```diff
@@ -685,104 +685,7 @@ $(P
...
-DDOC_CONSTRAINT         = $(DDOC_CONSTRAINT) if ($0)
-DDOC_OVERLOAD_SEPARATOR =
-DDOC_TEMPLATE_PARAM_LIST= $0
-DDOC_TEMPLATE_PARAM     = $0
-DDOC_BACKQUOTED       = $(DOLLAR)(D_INLINECODE $(DOLLAR)0)
-
-ESCAPES = /$(LT)/$(AMP)lt;/
-          /$(GT)/$(AMP)gt;/
-          /$(AMP)/$(AMP)amp;/
+3fc0f638fb1ab1bad2e01f507b968324
```
jacob-carlborg: Where is this coming from? A Git hash or a random set of characters?

andralex: I generated it with mcookie. The idea is it's a unique string to be replaced with the actual text.

andralex:
I agree that the default macro definitions should be implementation-defined. I've reworked this diff to mention that, and also added the idea in https://github.com/dlang/dlang.org/pull/1703 to it.

CyberShadow:
> I agree that the default macro definitions should be implementation-defined.

Good call. LGTM.

spec/ddoc.src.dd
jacob-carlborg: I see, thanks.

PR #1806, opened 2017-07-07
Ddoc no hardcoded characters

andralex:
https://github.com/dlang/dmd/pull/6972 moves some stray line feeds from the ddoc engine into macro definition. This diff does the same for the dlang.org macros.


andralex:
How do I get rid of the stray commit from unrelated work? Thx!

quickfur:
How did you end up with a commit from unrelated work? Did you commit to the wrong branch?

Since only one commit is involved here, the simplest solution I can think of is to checkout a new branch from master, apply the diff, and push that instead of what you have here.

Also, on that note, in the future be careful of where you invoke `git checkout -b ...`, because it takes the *current branch* as the base for the newly-created branch. To be safe, for creating PRs, you should always precede that with `git checkout master` so that your new branch doesn't "accidentally" inherit commits from an unrelated branch.

andralex:
Superseded by https://github.com/dlang/dlang.org/pull/1809/

PR #1807, opened 2017-07-07
Use dmd-nightly for prerelease docs

wilzbach:
The DTour now allows to select the compiler (e.g. https://github.com/dlang-tour/core/pull/575, https://github.com/dlang-tour/core/pull/574, https://github.com/dlang-tour/core/issues/548).

This means that we can use `dmd-nightly` for the prerelease pages, e.g.

image

Details:
- The Docker images are built & updated daily
- There weren't many new symbols in Phobos since 2.074.0, but one is in `std.digest.crc`

### DIFF
--- a/js/run.js
+++ b/js/run.js
@@ -66,6 +66,9 @@ function safeVar(data, path)
     return res;
 }
 
+// compile the examples on the prerelease pages with dmd-nightly
+var dmdCompilerBranch = location.href.indexOf("-prerelease/") >= 0 ? "dmd-nightly" : "dmd";
+
 var backends = {
   dpaste: {
     url: "https://dpaste.dzfl.pl/request/",
@@ -97,7 +100,8 @@ var backends = {
     contentType: "text/plain; charset=UTF-8",
     requestTransform: function(data) {
         return JSON.stringify({
-            source: data.code
+            source: data.code,
+            compiler: dmdCompilerBranch
         });
     },
     parseOutput: function(data, opts) {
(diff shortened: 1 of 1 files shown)


CyberShadow:
Awesome!

js/run.js
```diff
@@ -97,7 +97,8 @@ var backends = {
     contentType: "text/plain; charset=UTF-8",
     requestTransform: function(data) {
         return JSON.stringify({
-            source: data.code
+            source: data.code,
+            compiler: location.href.indexOf("prerelease") >= 0 ? "dmd-nightly" : "dmd"
```
CyberShadow: It's just luck we don't have any symbols in Phobos or Druntime that have "prerelease" in the name.

Can't this at least be `indexOf("-prerelease/")` ?

CyberShadow:
Can the external editor button be fixed so that it chooses the nightly compiler automatically?

wilzbach:
> It's just luck we don't have any symbols in Phobos or Druntime that have "prerelease" in the name.

Or knowledge ;-)

> Can't this at least be `indexOf("-prerelease/")` ?

Sure.

Btw I would be hugely in favor of moving all these manual checks to a common place, s.t. I don't need to check the URL bar for everything.

CyberShadow:
> Btw I would be hugely in favor of moving all these manual checks to a common place, s.t. I don't need to check the URL bar for everything.

Be my guest :)

wilzbach:
> Can the external editor button be fixed so that it chooses the nightly compiler automatically?

Sure, I will do so later tonight - after all this feature is still hot :hot_pepper:  from the oven.

wilzbach:
> Sure, I will do so later tonight - after all this feature is still hot :hot_pepper:  from the oven.

Oh, the best way is doing it directly in this PR. Sorry.

CyberShadow:
> Oh, the best way is doing it directly in this PR. Sorry.

You mean just the dlang.org side, right? The tour side seems to ignore the parameter right now.

wilzbach:
> You mean just the dlang.org side, right? The tour side seems to ignore the parameter right now.

Yes, PR for the Tour is out: https://github.com/dlang-tour/core/pull/578

PR #1808, opened 2017-07-07
[Hotfix] - Fix DBlog regex

wilzbach:
Noticed that our frontpage got broken:

image

This will make the regex more error-tolerant - apparently the last time the article got edited and I wrongly made this part of the regex.

I will directly merge this as a hotfix frontpage. We can always talk later about better approaches.


wilzbach:
It's back to normal: 

image

PR #1810, opened 2017-07-08
Numbering the chapters/sections/subsections/subsubsections/subsubsubsections/paragraphs in the spec

andralex:
I'll do this in several commits to make reviewing easier. First commit is the counting mechanism.


spec/attribute.dd
```diff
@@ -817,4 +810,5 @@ $(SPEC_SUBNAV_PREV_NEXT property, Properties, pragma, Pragmas)
 )
 
 Macros:
+        CHAPTER=8
         TITLE=Attributes
```
wilzbach: FWIW you could use the `footer_gen` script to automatically generate/update this.

spec/spec.ddoc
```diff
@@ -3,6 +3,36 @@ ROOT = ..
 SEARCHDEFAULT_SPEC = selected
 DDOC_BLANKLINE = $(DIVC blankline)
 
+BODY_PREFIX =
```
wilzbach: This is a huge hack against Ddoc's capabilities. Maybe at some point in time we can switch to a proper template engine (or at least Ddox) ...

spec/attribute.dd
andralex: Thanks! I guess it's about the same amount of work because the assignment of chapter numbers is arbitrary. Putting them all in the script has the nice effect of keeping the mapping in a single place. I'll keep it in mind.

spec/spec.ddoc
andralex: I'm surprised. This is a straightforward application - the macro is there for being overridden, and that's what this does. Where is the hack?

spec/attribute.dd
andralex: @wilzbach just ran the script, got:

```
...
simd.dd: $(SPEC_SUBNAV_PREV_NEXT abi, Application Binary Interface, $1, $2)))
$1.dd: $(SPEC_SUBNAV_PREV_NEXT simd, Vector Extensions, $1, $2)))
std.file.FileException@std/file.d(350): /home/andrei/d/dlang.org/spec/$1.dd: No such file or directory
```

spec/spec.ddoc
```diff
@@ -5,6 +5,23 @@ DDOC_BLANKLINE = $(DIVC blankline)
...
+}
+h3:before {
+  counter-increment: h3;
+  content: "" counter(h3) ": ";
+  counter-reset: para;
+}
+h3 {
+  counter-reset: para;
+}
+</style>
```
CyberShadow: This is not the correct place for this. CSS rules are global, adding them inline doesn't make sense.

1. Wrap `SUBNAV_SPEC` in a `div` or something else that gives it a unique id or class (e.g. `id="subnav-spec"`).
2. In style.css, add these rules, but prefixed with `#subnav-spec` (i.e. `#subnav-spec p:before { ... }`).

spec/spec.ddoc
CyberShadow: Ditto, this is not the correct place for this either.

1. In spec.ddoc (this file), add `BODYCLASS=doc spec`
2. In style.css, add these rules, but prefixed with `body.spec` (i.e. `body.spec p:before { ... }`)

spec/spec.ddoc
```diff
@@ -35,23 +35,6 @@ $(T style,
...
-}
-h3:before {
-  counter-increment: h3;
-  content: "" counter(h3) ": ";
-  counter-reset: para;
-}
-h3 {
-  counter-reset: para;
-}
-</style>
```
CyberShadow: Please don't mix unrelated changes in the same commit. If a change in one commit negates the change in another commit, then please squash the irrelevant commit using an interactive rebase.

spec/intro.dd
```diff
@@ -8,15 +8,13 @@ $(SPEC_S Introduction,
...
-	not perturbed by the semantic analyzer. This separation of the passes
-	makes language tools like syntax
-	directed editors relatively easy to produce.
-	It also is possible to compress D source by storing it in
-	$(SINGLEQUOTE tokenized) form.
+$(P The process of compiling is divided into multiple phases. Each phase has no
+dependence on subsequent phases. For example, the scanner is not perturbed by
+the semantic analyzer. This separation of the passes makes language tools like
+syntax directed editors relatively easy to produce. It also is possible to
+compress D source by storing it in $(SINGLEQUOTE tokenized) form.)
```
CyberShadow: Please don't mix whitespace changes and non-whitespace changes in the same commit.

latex.ddoc
```diff
@@ -44,17 +44,15 @@ _=
...
 \lstset{language=C++}
 _=
 
 D=\lstinline|$0|
 _=
 
 D_CODE=\begin{lstlisting}
-$0
-\end{lstlisting}
+$0\end{lstlisting}
```
CyberShadow: Don't mix unrelated changes within the same commit.

spec/type.dd
```diff
@@ -216,20 +216,20 @@ ulong  u4 = cast(long)-1; // ok, -1 can be represented in a ulong
...
+$(CODE_PIPE) $(D ^) $(CODE_AMP)$(D =) $(CODE_PIPE)$(D =) $(D ^=) !
+$(CODE_AMP)$(CODE_AMP) $(CODE_PIPE)$(CODE_PIPE) $(D ?:).)
+
+$(P A `bool` value can be implicitly converted to any integral type, with
+`false` becoming 0 and `true` becoming 1.)
+
+$(P The numeric literals `0` and `1` can be implicitly converted to the `bool`
+values `false` and `true`, respectively. Casting an expression to `bool` means
+testing for `0` or `!=0` for arithmetic types, and `null` or `!=null` for
+pointers or references.)
```
CyberShadow: Don't mix whitespace changes and non-whitespace changes within the same commit.

spec/type.dd
```diff
@@ -216,20 +216,20 @@ ulong  u4 = cast(long)-1; // ok, -1 can be represented in a ulong
...
-    A $(D bool) value can be implicitly converted to any integral type,
-    with $(D_KEYWORD false) becoming 0 and $(D_KEYWORD true) becoming 1.
-    The numeric literals 0 and 1 can be implicitly
-    converted to the bool values $(D_KEYWORD false) and $(D_KEYWORD true),
-    respectively.
-    Casting an expression to bool means testing for 0 or !=0 for
-    arithmetic types, and $(D_KEYWORD null) or !=$(D_KEYWORD null)
-    for pointers or references.
-    )
+$(P The bool type is a byte-size type that can only hold the value `true` or
```
CyberShadow: This hunk changes both content and whitespace, meaning it's very difficult to see just the differences in content.

spec/spec.ddoc
wilzbach: @CyberShadow the `style.css` is growing constantly in size, which makes adds a lot of unnecessary ballast to sites that want to use the dlang.org theme (I am not innocent about this). If we would use a modular CSS like SASS or LESS (or simply multiple files instead of a single `css/style.css`) it wouldn't be a big problem, but here I think that this is the right approach for CSS that only affects the spec pages.
However, `EXTRA_HEADERS` like it's used on the main `index.dd` are better suited for this:

```
EXTRA_HEADERS=$(T style,
...
)
```

https://github.com/dlang/dlang.org/blob/990d61d36a77fc62da1dd862f9e97013b8fd1346/index.dd#L554

CyberShadow:
> the style.css is growing constantly in size, which makes adds a lot of unnecessary ballast to sites that want to use the dlang.org theme

I don't think this is an argument that should affect dlang.org PRs.

> (or simply multiple files instead of a single css/style.css

That should be fine, as long as they're concatenated into a single one during the build. But I don't think it's an issue at all as long as the changes are properly namespaced, e.g. using the `body.class` rules.

> but here I think that this is the right approach for CSS that only affects the spec pages

No, it is a pointless increase of the page sizes. Redundant information should be retrieved once, which is why we have external CSS files in the first place.

andralex:
>I am going to stop here because this pull request is close to being unreviewable, due to putting unrelated changes and mixing whitespace and content changes within the same commit.

@CyberShadow yah, I though you were going to say that :). I've kept changes to a minimum, given the following considerations:

* There is a bug in the listings package that I also ran into while writing TDPL: sometimes a listing straddling the end of a page will fail to compile, therefore rendering the pdf unbuildable. The process of numbering sections and paragraphs has reflown the pdf a bit, and therefore the bug reared its ugly head. That's why you'll see unexplainable content changes in a few places.

* I took the opportunity to make minor and obvious content improvements in the paragraphs I touched anyway to add or remove `$(P`.

* Some were comically indented by many spaces, I've reflown them (only if I touched them anyway).

* To keep matters as conservative as I could, I didn't take many nearby opportunities that I otherwise could. I agree I could have been even more conservative, but I think things are within reasonable limits.

If you find this annoying to review, you should know producing it has been mind-numbingly tedious. We're talking hours. If you point at a few specific matters to improve in the PR, I'll gladly do it. If you ask me to redo the entire PR for a slight optimization of reviewing experience, I doubt it's an effective use of time. Thanks for your consideration.

CyberShadow:
> That's why you'll see unexplainable content changes in a few places.

Git accommodates for this by allowing you to leave such hacks in your working tree and stage/commit only relevant changes, so that you can then put the hacks in a separate commit documenting the changes while doing so.

> I took the opportunity to make minor and obvious content improvements in the paragraphs I touched anyway to add or remove $(P.

This is fine, but needs to be done as separate commits.

> Some were comically indented by many spaces, I've reflown them (only if I touched them anyway).

These are fine too, but whitespace changes must be in separate commits.

> To keep matters as conservative as I could, I didn't take many nearby opportunities that I otherwise could.

I don't mind grouping quasi-related changes in a single pull request, since as you know, I too am against the tedium of managing a horde of inter-dependent GitHub PRs. However, all such changes should be done as separate commits.

> If you ask me to redo the entire PR for a slight optimization of reviewing experience

It's not a slight optimization - it is blatantly disproportional. It takes 10x more effort reviewing a commit that has both whitespace and non-whitespace changes in it than 2 commits that only have one kind of each. In the former case, you need to carefully look at every diff to see whether it contains any content changes; in the latter case, all changes are obvious at a glance. 

All of this could have been avoided by committing more frequently (even as often as you save), then wrapping it up with an interactive rebase to fold and rearrange your editing history into a series of logical changes. That's probably, like, 10% overhead, plus one final hurdle at the end to review your changes altogether, which is useful anyway to review the sum of your work.

Now, me sitting down to take apart these diffs and review this PR properly would not be the end of the world, but it is such a stupid waste of time. Furthermore, I'm worried that if I consent to this now, I will need to review a very similar PR in a month or few from today, so I'm really tempted to just reject this PR and let someone else review it. I would really rather spend some time to fix the problem at its root, i.e. help you use git more effectively. Martin mentioned that git illiteracy is hurting us in general, and I agree that it can occasionally become embarrassing.

You use Emacs, right? Do you use Magit? If not, then it's a must and I strongly suggest you get acquainted with it ASAP. Assuming `magit-status` is bound to `M-g`, you can:

- stage some changes: `M-g` (move cursor over file or hunk to stage) `s`
- commit staged changes: `M-g` `c` `c` (type commit message) `C-c` `C-c`
- commit all changes: `M-g` `S` `c` `c` (type commit message) `C-c` `C-c`
- add changes to the previous commit (amend): `M-g` `S` `c` `e`
- add changes to any previous commit (instant fixup): `M-g` `c` `F` (select commit with arrow keys) `C-c` `C-c`
- edit commit message of any previous commit (rebase to reword): `M-g` `c` `w` (select commit with arrow keys) `C-c` `C-c` (type commit message) `C-c` `C-c`
- interactive rebase: `M-g` `r` `i` (select first commit with arrow keys) `C-c` `C-c`, then:
  - arrow keys - move cursor as usual
  - `M-` arrow keys - reorder commits
  - `f` - mark commit for fixup (merge into previous commit)
  - `C-c` `C-c` - perform rebase

This is scratching the surface but should cover the basics. Essentially you should not be leaving Emacs to use git command line for anything other than `git init`, `git clone`, or complex stuff like `git filter-branch`. If you're having trouble, perhaps we can set up a screen share OSLT.

andralex:
Edits were done with Atom, although that shouldn't make a big difference. I agree things could have been done with even a finer granularity, although it does seem past a point we're becoming more Catholic than the Pope. I mean I think committing with every save is pushing it. That said, I'll experiment with more aggressive commits for the remainder of the commits that are necessary to get this PR to build.

It's your prerogative to forgo reviewing this, thanks for looking over it.

CyberShadow:
> Edits were done with Atom,

Sorry, I'm not very familiar with that editor.

> although that shouldn't make a big difference.

Fairly sure it should - people have switched to Emacs specifically so that they could use Magit: https://magit.vc/quotes/

> I agree things could have been done with even a finer granularity, although it does seem past a point we're becoming more Catholic than the Pope. I mean I think committing with every save is pushing it.

That wasn't quite the point I was trying to make - rather, committing with every save at least would allow to merge redundant commits afterwards. It's certainly not necessary. Splitting commits, especially those that mix various kinds of changes in the same hunks, is much more work in contrast.

> It's your prerogative to forgo reviewing this, thanks for looking over it.

I'll put it on my list. No promises :)

CyberShadow:
@andralex Do you have any uncommitted/unpushed changes? If so, please push them now and put this on hold, I will try to clean up this PR.

andralex:
@CyberShadow yah, spec/enum.dd which fortunately was in good shape already (only one line changed). Done.

CyberShadow:
@andralex Done. I've reordered the changes into commits grouped by the nature of the change. Only the git history was changed, the final diff remains the same.

If you wish to continue working from here, I'd appreciate if you grouped the changes into separate commits as they are grouped now. Either way (now or after you went over all spec files), it will probably make sense to submit these changes as a series of PRs, so that we can preview the impact of each kind of change on the HTML diff on its own.

wilzbach:
>  it will probably make sense to submit these changes as a series of PRs, so that we can preview the impact of each kind of change on the HTML diff on its own.

In case you want to automate things - I have a simple script that splits up a PR into files and submits a PR for each one:

https://github.com/wilzbach/splitup-prs

CyberShadow:
BTW, as an experiment I recorded my screen while working on the git refactoring above:
https://www.youtube.com/watch?v=2PQ6WFduyVk https://www.youtube.com/watch?v=5decasCJo2k
(The audio quality in the first video is pretty awful but it's much better in the second video.)
I'll delete this comment after posting as there is probably too much data in the videos to leave them public, and it's not that relevant to the discussion anyway.

CyberShadow:
@andralex I think you force-pushed and overwrote my commits? By accident?

andralex:
@CyberShadow oy that's awful. Sorry about that! I forgot to pull remote updates before force-pushing. How do we get from where we are to where we should be?

CyberShadow:
@andralex Restored the changes I've refactored + rebased your latest changes on top of them. From here you should fetch and reset your `spec-numbered` to `origin/spec-numbered`. Ping me on IRC if you need help :)

andralex:
@CyberShadow thanks. I'm back to working on this, please lmk whether the granularity of the last few commits is appropriate.

andralex:
Whew! I can't believe I got through everything. Ready for review.

CyberShadow:
> please lmk whether the granularity of the last few commits is appropriate

Yes, thanks!

If you feel like it, merging commits with the same theme but editing different files will shorten the commit list without any loss of granularity (as splitting things by file is trivial whether the changes are in one commit or many).

I'll have a closer look at this as I catch up with my D backlog. You may want to have a look at the test failure.

andralex:
@CyberShadow cool, I've merged a few CHAPTER=XXX commits together, will look into a few more consolidation opportunities.

andralex:
@CyberShadow I'm done with consolidating things.

andralex:
@CyberShadow ah, the build failure was because the TOC has no chapter number. It shouldn't! Fixed.

spec/function.dd
```diff
@@ -2054,14 +2054,12 @@ void func(T)(T x)
 {
     writeln(x);
 }
-
```
CyberShadow: This is troubling. I don't think we should be deleting empty lines from any spec examples because of Latex bugs. Can you put some dummy whitespace here?

CyberShadow: It looks like we can use macros inside DDoc example blocks, so if it comes to that, you could define a $(EMPTY_LINE) which expands to nothing in .tex and an empty line otherwise.

spec/attribute.dd
```diff
@@ -484,11 +484,9 @@ runtime error.)
 
 ---
 @disable void foo() { }
-
-void main()
-{
-    foo();   // error, foo is disabled
-}
+---
+---
```
CyberShadow: Same here. We shouldn't make the HTML documentation (which is likely going to be used 1000x more often than the PDF) uglier because of Latex bugs...

spec/function.dd
andralex: I don't see a pressing need. TDPL is full of such adjustments and they don't impede quality.

CyberShadow: Links:

http://dtest.dlang.io/artifact/website-5f8248914455d3d83945a3b59d88073e5ee6fc7c-82f02509ce078ef3f9b9e8813db20d59/web/spec/function.html

http://dtest.dlang.io/artifact/website-5f8248914455d3d83945a3b59d88073e5ee6fc7c-82f02509ce078ef3f9b9e8813db20d59/web/spec/attribute.html#disable

Kinda ugly, but if you say it's not worth bothering then so be it.

CyberShadow:
Any thoughts on whether we should try to keep the section orders stable from now on (so that "D spec, section 19.21" doesn't lose its meaning)?

spec/function.dd
andralex: I found where the problem lies an a workaround. Removing `\lstset{breaklines=true}` from latex.ddoc makes everything work nicely. However, that foregoes an important feature - long lines were nicely split.

Anyway, this is an ongoing annoyance that must be addresse. I'll explore more workarounds in a future PR.

andralex:
>Any thoughts on whether we should try to keep the section orders stable from now on (so that "D spec, section 19.21" doesn't lose its meaning)?

Thanks for asking. I'm thinking of the following concerns:

* The quality of documentation is highly variable so we need to invest quite a bit more work before we find the right standardese tone and paragraph cadence.
* AFAIK standards documents do not preserve section and para numbers across versions. Instead, they prescribe a version of the document itself, so people can refer to it as "D spec 2.083.1 Section 25.2 para 3".
* The above means we should distinguish the version of the standard in the URL, e.g. what's now https://dlang.org/spec/property.html should become https://dlang.org/spec-2.075.0/property.html. Also, https://dlang.org/dlangspec.pdf should become http://dlang.org/dlangspec-2.075.0.pdf.

So the next order of business is to put the spec numbering in the URLs. I recall @wilzbach has done some work on the phobos side of that (which is somewhat independent).

wilzbach:
> . I recall @wilzbach has done some work on the phobos side of that (which is somewhat independent).

The docarchives contain a snapshot of the entire dlang.org site (even with the spec as PDF/epub/mobi), e.g.

https://docarchives.dlang.io/v2.067.0/attribute.html

We could setup a http rewrite, but I'm not sure whether we should make the version explicit by default. After all, the spec hardly changes from version to version (which of course is also partly due to the fact that many language change are still merged without a doc update.)

andralex:
>We could setup a http rewrite, but I'm not sure whether we should make the version explicit by default. After all, the spec hardly changes from version to version (which of course is also partly due to the fact that many language change are still merged without a doc update.)

Probably we need to; this kind of stuff will become increasingly important going forward.

wilzbach:
> So the next order of business is to put the spec numbering in the URLs. I recall @wilzbach has done some work on the phobos side of that (which is somewhat independent).

Better late than never: https://github.com/dlang/dlang.org/pull/2045

PR #1811, opened 2017-07-09
add D_AVX and D_AVX2 to version identifier docs

MartinNowak:
also see https://github.com/dlang/dmd/pull/6977

Arguably the D_ prefix doesn't make much sense here, but that's how it was done for D_AVX and there is little point in changing that.


ibuclaw:
Though really, you really need to stop putting x86 versions under the global namespace.  CPU features should be prefixed by the CPU. This is also documented in the spec iirc.

PR #1812, opened 2017-07-09
Link to get involved

wilzbach:
Follow-up to https://github.com/dlang/phobos/pull/5582#issuecomment-313922823

> BTW, the dlang.org front page's "Contribute" section should link to a wiki page which lists entry-level tasks such as these.

I have started to rework https://wiki.dlang.org/Get_involved a bit, but, of course, more love is always welcome!


CyberShadow:
How about shorting it to: "There are $(LINK2 https://wiki.dlang.org/Get_involved, many other ways) to help, too!"

wilzbach:
> How about shorting it to: "There are $(LINK2 https://wiki.dlang.org/Get_involved, many other ways) to help, too!"

I like it!

PR #1814, opened 2017-07-10

MartinNowak:


wilzbach:
Btw did you already think about how we could preview the changelog for 2.076 before a new release?
Not a major issue, but I guess the solution would be as easy as using a different file than `changelog/2.075.0_pre.dd` either on your side or on the @CyberShadow's DAutoTest side.

MartinNowak:
See https://dlang.slack.com/archives/C59UPP30D/p1499683698318570 for discussions of changelog preview during beta.

MartinNowak:
Doesn't seem very pressing though a bit annoying.

PR #1817, opened 2017-07-11
Use _pending for the latest changelog entries

wilzbach:
Regarding https://github.com/dlang/dlang.org/pull/1814#issuecomment-313988447

A simple solution is to use `_pending` for the automatically generated changelog. (@MartinNowak) doesn't seem to use the Makefile target here anyways.

Regarding the changelog menu: I wasn't really sure how we should handle the case when a prerelease is progress, so I simply renamed them "Nightlies" as this actually is exactly what the page depicts.

CC @CyberShadow

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,4 +44,4 @@ modlist-release.ddoc
 /assert_writeln_magic
 
 # Generated changelogs
-changelog/*_pre.dd
+changelog/*_pending.dd
--- a/changelog/changelog.ddoc
+++ b/changelog/changelog.ddoc
@@ -63,10 +63,12 @@ PHOBOSPR = $(PULL_REQUEST phobos,$1)
 _=
 
 CHANGELOG_VERSION = $(LI <a id="$1" href="$1.html">$1</a><span class="hide-from-nav"> ($2, $3)</span>)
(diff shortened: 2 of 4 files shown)


CyberShadow:
Looks like we had the same idea: https://github.com/dlang/phobos/pull/5504#issuecomment-314165842

CyberShadow:
Thanks!

The _pre page still has nothing on it to indicate it's not yet released, except the future date and URL, sigh.

wilzbach:
Hmm any idea what could have went wrong with the deployment?
The changes were properly shown in DAutoTest, but aren't visible on dlang.org, e.g.

http://dlang.org/changelog/2.075.0_pending.html

MartinNowak:
What was the point of using `2.076.0_pending.html` instead of just `2.076.0.html` having pending somewhere clearly visible on the page?

CyberShadow:
I think because `update_nav.sh` infers the page type from the filename?

wilzbach:
>> What was the point of using 2.076.0_pending.html instead of just 2.076.0.html having pending somewhere clearly visible on the page?
> I think because update_nav.sh infers the page type from the filename?

Yes, and also it's not trivial to know the future release version of `master` as this would require detecting the beta tags and this `_pending` seemed a lot easier.

PR #1818, opened 2017-07-12
Fix spec for `out` contracts without identifier

ntrel:
* Non-void functions can use `out` without result variable too
* Replace HTML tags.
* Minor tweaks.

### DIFF
--- a/spec/contracts.dd
+++ b/spec/contracts.dd
@@ -100,7 +100,7 @@ in
 }
 out (result)
 {
-    assert((result * result) <= x && (result+1) * (result+1) > x);
+    assert(result ^^ 2 <= x && (result + 1) ^^ 2 > x);
 }
 do
 {
--- a/spec/contracts.dd
+++ b/spec/contracts.dd
@@ -84,14 +84,14 @@ do
     although $(D do) is preferred.)
 
 	$(P By definition, if a pre contract fails, then the function received bad
-	parameters.
-	An AssertError is thrown. If a post contract fails,
-	then there is a bug in the function. An AssertError is thrown.)
+	parameters. An `AssertError` is thrown. If a post contract fails,
+	then there is a bug in the function. An `AssertError` is thrown.)
 
-	$(P Either the <code>in</code> or the <code>out</code> clause can be omitted.
-	If the <code>out</code> clause is for a function
-	body, the variable <code>result</code> is declared and assigned the return
+	$(P Either the `in` or the `out` clause can be omitted.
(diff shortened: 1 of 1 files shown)


spec/contracts.dd
```diff
@@ -99,18 +97,22 @@ body
     return cast(long)std.math.sqrt(cast(real)x);
```
wilzbach: Btw shouldn't this page not also be updated in regards of [DIP1003](http://forum.dlang.org/post/esjclvbebspqkekjmwki@forum.dlang.org)?

spec/contracts.dd
```diff
@@ -122,8 +124,18 @@ body
     ...
 }
 ------
-	In an out statement, $(I result) is initialized and set to the
-	return value of the function.
+	$(P When the return type is not `void`, and no identifier is
+	declared, a variable named `__result` is defined by the compiler
+	for the `out` contract and initialized to the return value of the
+	function:)
```
wilzbach: Isn't this just the current behavior of DMD?

ntrel: OK, maybe `__result` shouldn't be documented then. @zachthemystic quoted use of it to me [here](https://github.com/dlang/DIPs/pull/79#discussion_r126827866), I assumed it was a language feature.

ntrel: Removed `__result` docs.

spec/contracts.dd
ntrel: Maybe @MetaLang wants to update contracts.html and grammar.html? Anyway, AIUI the current stable compiler doesn't support it yet.

MetaLang: I'll put it on my todo list.

spec/contracts.dd
```diff
@@ -122,8 +127,6 @@ body
     ...
 }
 ------
-	In an out statement, $(I result) is initialized and set to the
-	return value of the function.
```
ntrel: Removed as it repeats what's on line 84.

spec/contracts.dd
MetaLang: And here they are:
https://github.com/dlang/dlang.org/pull/1828
https://github.com/dlang/dlang.org/pull/1829

ntrel:
@CyberShadow Any idea why DAutoTest failed on the last commit?

CyberShadow:
> Any idea why DAutoTest failed on the last commit?

code.dlang.org was down.

ntrel:
@wilzbach thanks!

PR #1821, opened 2017-07-13
Show messages boxes for beta and nightly

wilzbach:
https://github.com/dlang/tools/pull/251 needs to be pulled first.

### DIFF
--- a/changelog/2.075.0_pre.dd
+++ b/changelog/2.075.0_pre.dd
@@ -2,7 +2,7 @@ Ddoc
 
 $(CHANGELOG_NAV_LAST 2.074.1)
 
-$(VERSION Jul 16, 2017, =================================================,
+$(BETA_VERSION Jul 16, 2017, =================================================,
 
 $(BUGSTITLE Compiler changes,
 
--- a/changelog/changelog.ddoc
+++ b/changelog/changelog.ddoc
@@ -15,7 +15,7 @@ SUBNAV=$(SUBNAV_TEMPLATE
 SEARCH_OPTIONS_EXTRA=<option value="dlang.org/changelog" selected>Change
 Log</option>
 
-VERSION=
+STABLE_VERSION=
 $(DIVC version,
 $(P
 $(B $(LARGE $(LINK2 http://downloads.dlang.org/releases/2.x/$(VER), Download D $(VER))))$(BR)
@@ -23,6 +23,7 @@ $(SMALL released $1, $2)
 )
 $4
 )
+VERSION=$(STABLE_VERSION $0)
 
 BETA_VERSION=
 $(DIVC version,
@@ -43,7 +44,7 @@ $(SMALL $1)
 $(P Please report bugs at $(LINK2 $(ROOT_DIR)bugstats.html, Bugzilla).)
 $(MESSAGE_BOX red, Unstable - nightly releases are not suited for production. Use stable.)
 )
-$4
+$3
 )
 
 BUGZILLA = <a href="https://issues.dlang.org/show_bug.cgi?id=$0">Bugzilla $0</a>
--- a/changelog/changelog.ddoc
+++ b/changelog/changelog.ddoc
@@ -24,11 +24,24 @@ $(SMALL released $1, $2)
 $4
 )
 
+BETA_VERSION=
+$(DIVC version,
+$(P
+$(B $(LARGE $(LINK2 http://downloads.dlang.org/releases/2.x/$(VER), Download D$(VER) (beta))))$(BR)
+$(SMALL to be released $1, $2)
+$(P Please report bugs at $(LINK2 $(ROOT_DIR)bugstats.html, Bugzilla).)
+$(MESSAGE_BOX orange, Experimental - beta releases may contain regressions. Test with care.)
+)
+$4
+)
+
 NIGHTLY_VERSION=
 $(DIVC version,
 $(P
-$(B $(LARGE $(LINK2 http://nightlies.dlang.org, Download D nightlies)))$(BR)
+$(B $(LARGE $(LINK2 http://nightlies.dlang.org, Download D (Nightly))))$(BR)
 $(SMALL $1)
+$(P Please report bugs at $(LINK2 $(ROOT_DIR)bugstats.html, Bugzilla).)
+$(MESSAGE_BOX red, Unstable - nightly releases are not suited for production. Use stable.)
 )
 $4
 )
--- a/css/style.css
+++ b/css/style.css
@@ -2243,3 +2243,26 @@ dt.d_decl:hover .decl_anchor {
         text-align: right;
     }
 }
+
+/**
+ * Message boxes - different color for each category
+ */
+.message-box {
+    padding: 1em 1em 1em 1em;
+    border-radius: 5px;
+}
+.message-box, .message-box a {
+    color: white;
+}
+.message-box-red {
+    background-color: #D60027;
+}
+.message-box-gray {
+    background-color: #797979;
+}
+.message-box-orange {
+    background-color:  #EC5315;
+}
+.message-box-green {
+    background-color: #4EBA0F;
+}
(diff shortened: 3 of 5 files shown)


changelog/2.075.0_pre.dd
```diff
@@ -2,7 +2,7 @@ Ddoc
 
 $(CHANGELOG_NAV_LAST 2.074.1)
 
-$(VERSION Jul 16, 2017, =================================================,
+$(BETA_VERSION Jul 16, 2017, =================================================,
```
wilzbach: To avoid doing this manually, there will be `--release-type=beta` (https://github.com/dlang/tools/pull/251)

changelog/changelog.ddoc
```diff
@@ -15,22 +15,36 @@ SUBNAV=$(SUBNAV_TEMPLATE
...
-VERSION=
+STABLE_VERSION=
 $(DIVC version,
 $(P
 $(B $(LARGE $(LINK2 http://downloads.dlang.org/releases/2.x/$(VER), Download D $(VER))))$(BR)
 $(SMALL released $1, $2)
 )
 $4
 )
+VERSION=$(STABLE_VERSION $0)
```
wilzbach: For compatibility, in a follow-up we could do a quick `sed`.

changelog/changelog.ddoc
```diff
@@ -15,22 +15,36 @@ SUBNAV=$(SUBNAV_TEMPLATE
...
 $(DIVC version,
 $(P
-$(B $(LARGE $(LINK2 http://nightlies.dlang.org, Download D nightlies)))$(BR)
+$(B $(LARGE $(LINK2 http://nightlies.dlang.org, Download D (Nightly))))$(BR)
 $(SMALL $1)
+$(P Please report bugs at $(LINK2 $(ROOT_DIR)bugstats.html, Bugzilla).)
+$(MESSAGE_BOX red, Unstable - nightly releases are not suited for production. Use stable.)
 )
-$4
+$3
```
wilzbach: In https://github.com/dlang/tools/pull/251 the format is unified, that's why the tools PR needs to be pulled first.

css/style.css
```diff
@@ -2243,3 +2243,26 @@ dt.d_decl:hover .decl_anchor {
         text-align: right;
     }
 }
+
+/**
+ * Message boxes - different color for each category
```
wilzbach: As it turns out even without https://github.com/dlang/dlang.org/pull/1778, the message boxes are universally useful.

dlang.org.ddoc
```diff
@@ -205,6 +205,7 @@ MENU = <li><a href='$1'><span>$+</span></a></li>
 MENU_W_SUBMENU = <li class='expand-container'><a class='expand-toggle' href='#'><span>$0</span></a>
 MENU_W_SUBMENU_LINK = <li class='expand-container'><a class='expand-toggle' href='$1'><span>$+</span></a>
 MENU_W_SUBMENU_END = </li>
+MESSAGE_BOX = $(DIVC message-box message-box-$1, $+)
```
wilzbach: We might define `RED_MESSAGE_BOX` or `ERROR_MESSAGE_BOX` etc. in the future, but for now I kept it as bloat-free as possible.

posix.mak
```diff
@@ -656,7 +656,7 @@ test: $(ASSERT_WRITELN_BIN)_test all
 
 changelog/${NEXT_VERSION}_pending.dd: | ${STABLE_DMD} ../tools ../installer
 	$(STABLE_RDMD) $(TOOLS_DIR)/changed.d $(CHANGELOG_VERSION_MASTER) -o $@ \
-	--version "${NEXT_VERSION} (pending)" --date "Pending" --nightly
+	--version "${NEXT_VERSION} (Nightly)" --date "To be released" --release-type "nightly"
```
wilzbach: Nightly is more meaningful than "pending"

wilzbach:
Not so sure how well this will work without the tools PR, hence two screenshots:

image

image

CyberShadow:
TBH the boxes seem too much to me. Not only are they really distracting (at least as they are in your screenshots), but the changelog is not the place for telling the user how (not) to use a version. I would suggest a much simpler, subtler change - enough that people are not going to be confused about the changelog of a yet-unreleased version.

wilzbach:
> TBH the boxes seem too much to me. Not only are they really distracting (at least as they are in your screenshots), but the changelog is not the place for telling the user how (not) to use a version. I would suggest a much simpler, subtler change - enough that people are not going to be confused about the changelog of a yet-unreleased version.

Sure, we can insert anything we want there once we have specific templates from the changed tool.

MartinNowak:
I like the idea to add extra information such as "report bugs to `issues.dlang.org`", but I doubt there is any individual in our target audience that doesn't have a basic understanding of what a beta and nightlies are.

CyberShadow:
> but I doubt there is any individual in our target audience that doesn't have a basic understanding of what a beta and nightlies are.

Yep, just to clarify - currently on http://dlang.org/changelog/2.075.0_pre.html the only clues that this is the pending changelog of a yet-unreleased version is the URL and the (for now) future date.

MartinNowak:
> Yep, just to clarify - currently on http://dlang.org/changelog/2.075.0_pre.html the only clues that this is the pending changelog of a yet-unreleased version is the URL and the (for now) future date.

That got broken with #1817 as it's using the generic make rule now, it used to say `to be released`.
Making this more visible makes sense, what about the stale URLs though?

CyberShadow:
> That got broken with #1817 as it's using the generic make rule now, it used to say to be released.

Aaaah. Sorry.

> what about the stale URLs though?

Hmm, that's a fair concern, though when the content of the page is bound to change, keeping the links working isn't as important, as an old link will not point to the version of the page at the time when the link was created anyway.

I think we can go with a `.htaccess` rule that redirects nonexisting _pre / _pending requests to the corresponding stable page.

wilzbach:
>> That got broken with #1817 as it's using the generic make rule now, it used to say to be released.
>Aaaah. Sorry.

Huh? @MartinNowak don't you commit the  `_pre` file [into the repository](https://github.com/dlang/dlang.org/blob/master/changelog/2.075.0_pre.dd)?
It's quite interesting how it worked before:

https://web.archive.org/web/20170703054803/https://dlang.org/changelog/2.075.0.html -> "to be released Jul 9, 2017"
https://web.archive.org/web/20170703033505/http://dlang.org/changelog/2.075.0_pre.html -> "released Jul 9, 2017"

We could, of course, always generate `_pre` as well and avoid the need to push it into the repos, but I think you currently push it because of the release date?
With https://github.com/dlang/tools/pull/251 _pre would have its own ddoc template, so you wouldn't need to adjust the generated file anymore for this ;-)

MartinNowak:
> Huh? @MartinNowak don't you commit the _pre file into the repository?
> It's quite interesting how it worked before:

Read the [Makefile](https://github.com/dlang/dlang.org/blob/40e16ab350a8f7d3b7dca54fb0431cfef456fb46/posix.mak#L295) ;), we were using different macros for those changelogs. IMO a better mechanism than hardcoding differences into the changed tool.
The added NEXT_VERSION always used _pre.html for whatever reason, hence the beta and the pending changelog coexisted until #1823 (under confusing URLs).

There are two parts that need to happen for a proper pending changelog during beta.

We need to get a proper NEXT_VERSION. I can write the current beta and the latest version to the dlang.org repo, or we might use dmd's VERSION file which is now updated like `git describe` and contains the latest beta/release tag.
See https://gist.github.com/MartinNowak/a471fe7ddbfeef205cdf04c93a94c6d0/962d62ce7d0f608f352fbdec051cc809cfa79cde#file-release-sh-L75 for more info.
We should definitely fix the terminology and use consistent file names (version, stable version, latest version..., #1824 looks confusing to me).

We need to purge the changelog in time. I'd suggest we extend the changed tool so that it can append new loose changelog entries to a committed `changelog/2.075.0_pre.dd` file. That way I can purge the changelogs on stable and merge that back fairly soon. Also avoids the duplication which sometimes leads to people editing the wrong file.
This should still work with multiple release branches and backporting as loose changelog entries are part of PRs (or can be picked as well).

@wilzbach, @CyberShadow

MartinNowak:
Ping on this, just pulling of the NEXT_VERSION based on dmd/VERSION should be fairly straightforward.
Enhancing the changed tool is a bit more effort, but not necessary to get the changelog preview running, just to make it accurate.

MartinNowak:
BTW, quite a nice changelog.
https://jenkins.io/changelog/

Might be a nice idea to combine all the changelog TOCs on the index page.
https://dlang.org/changelog/index

Not sure how they produce the overall points, maybe some fancy burndown planning, but that doesn't carry too much value IMO.

wilzbach:
Fair enough, let's close this then.

PR #1823, opened 2017-07-15
Add pre_changelog and release_changelog commands

wilzbach:
Follow-up to https://github.com/dlang/dlang.org/pull/1817 and https://github.com/dlang/dlang.org/pull/1821

The idea is the more logic we move to the same spot, the easier it gets to update it and improve it. 
Also if we can try out the script that Martin runs, the less likely it is for us to break them in the future ;-)

Eventually the plan is to generate the `_pre.dd` pages automatically, but this depends on https://github.com/dlang/tools/pull/251

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -652,17 +652,35 @@ test: $(ASSERT_WRITELN_BIN)_test all
 
 ################################################################################
 # Changelog generation
+#
+# pending:		$(LATEST)..upstream/master		(all changes of the next _major_ release)
+# pre:			$(LATEST)..upstream/stable 		(all changes of the next release)
+# release:		$(PREVIOUS_LATEST)..$(LATEST)   (used by the release maintainer after updating $(LATEST))
 ################################################################################
 
 changelog/${NEXT_VERSION}_pending.dd: | ${STABLE_DMD} ../tools ../installer
 	$(STABLE_RDMD) $(TOOLS_DIR)/changed.d $(CHANGELOG_VERSION_MASTER) -o $@ \
 	--version "${NEXT_VERSION} (pending)" --date "Pending" --nightly
 
-changelog/${NEXT_VERSION}_pre.dd: | ${STABLE_DMD} ../tools ../installer
-	$(STABLE_RDMD) $(TOOLS_DIR)/changed.d $(CHANGELOG_VERSION_STABLE) -o $@ \
-		--version "${NEXT_VERSION}"
-
 pending_changelog: changelog/${NEXT_VERSION}_pending.dd html
 	@echo "Please open file:///$(shell pwd)/web/changelog/${NEXT_VERSION}_pending.html in your browser"
+
+changelog/${NEXT_VERSION}_pre.dd: | ${STABLE_DMD} ../tools ../installer
+	$(STABLE_RDMD) $(TOOLS_DIR)/changed.d $(CHANGELOG_VERSION_STABLE) -o $@ \
+		--version "${NEXT_VERSION} (beta)"  --date "To be released"
+
+pre_changelog: changelog/${NEXT_VERSION}_pre.dd html
+	@echo "Please open file:///$(shell pwd)/web/changelog/${NEXT_VERSION}_pre.html in your browser"
+
+# --no-text is used for point-releases like 2.074.1 to exclude plain-text files in changelog/* folders
+#  which are intended for major releases only.
+changelog/${LATEST}.dd: | ${STABLE_DMD} ../tools ../installer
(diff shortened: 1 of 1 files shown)


posix.mak
```diff
@@ -652,17 +652,35 @@ test: $(ASSERT_WRITELN_BIN)_test all
...
+
+# --no-text is used for point-releases like 2.074.1 to exclude plain-text files in changelog/* folders
+#  which are inteded for major releases only.
+changelog/${LATEST}.dd: | ${STABLE_DMD} ../tools ../installer
+	previous_latest=$$(git tag | grep -vE "(b|rc)" | tail -n2 | head -n1) && \
+	changelog_flags=$$(LATEST=$(LATEST) && [ "$${LATEST: -1}" -gt 0 ] && echo "--no-text") && \
+	$(STABLE_RDMD) $(TOOLS_DIR)/changed.d "$$(echo $${previous_latest}..v$(LATEST))" $$changelog_flags \
+		-o "changelog/$(LATEST).dd" --version "$(LATEST)" --date "$$(date +'%B %d, %Y')"
+
+release_changelog: changelog/${LATEST}.dd html $(DOC_OUTPUT_DIR)/changelog/$(LATEST).html
```
wilzbach: @MartinNowak `release_changelog` should be a drop-in replacement for whatever you are currently using. It's a bit different:

```diff
diff --git a/changelog/2.074.1.dd b/changelog/2.074.1.dd
index f1e38b17..f497c8ce 100644
--- a/changelog/2.074.1.dd
+++ b/changelog/2.074.1.dd
@@ -2,17 +2,13 @@ Ddoc
 
 $(CHANGELOG_NAV_LAST 2.074.0)
 
-$(VERSION May 30, 2017, =================================================,
+$(VERSION July 15, 2017, =================================================,
 
-$(BR)$(BIG $(RELATIVE_LINK2 bugfix-list, List of all bug fixes and enhancements in D $(VER).))
-
-$(HR)
-
-$(BR)$(BIG $(LNAME2 bugfix-list, List of all bug fixes and enhancements in D $(VER):))
+$(BR)$(BIG List of all bug fixes and enhancements in D $(VER).)
 
 $(BUGSTITLE DMD Compiler regressions,
 
-$(LI $(BUGZILLA 17407): [REG] __traits$(LPAREN)compiles$(RPAREN) triggers assertion failure)
+$(LI $(BUGZILLA 17407): [REG2.067] __traits$(LPAREN)compiles$(RPAREN) triggers assertion failure)
 )
 $(BUGSTITLE DMD Compiler bugs,
 
@@ -20,6 +16,7 @@ $(LI $(BUGZILLA 17289): With Xcode 8.3 linker, warnings of "pointer not aligned"
 )
 $(BUGSTITLE Phobos regressions,
 
+$(LI $(BUGZILLA 17264): [REG2.073] uniq fails with const elements)
 $(LI $(BUGZILLA 17330): [REG 2.072] BigInt's constructor used to be pure)
 $(LI $(BUGZILLA 17340): [REG 2.074.0] isNumeric!bool should not be true)
 )
```

posix.mak
```diff
@@ -652,17 +652,35 @@ test: $(ASSERT_WRITELN_BIN)_test all
...
-		--version "${NEXT_VERSION}"
+		--version "${NEXT_VERSION} (beta)"  --date "To be released"
 
-pending_changelog: changelog/${NEXT_VERSION}_pending.dd html
-	@echo "Please open file:///$(shell pwd)/web/changelog/${NEXT_VERSION}_pending.html in your browser"
+pre_changelog: changelog/${NEXT_VERSION}_pre.dd html
+	@echo "Please open file:///$(shell pwd)/web/changelog/${NEXT_VERSION}_pre.html in your browser"
+
+# --no-text is used for point-releases like 2.074.1 to exclude plain-text files in changelog/* folders
+#  which are inteded for major releases only.
```
CyberShadow: intended*

MartinNowak:
In my experience any integration into dlang.org's Makefile was always a pain to work with as it's taking a role as central build script, thus doesn't integrate well with an outer build script (e.g. starts to clone and/or build random random stuff).
Here is how I currently generate the changelogs which looks fine.
https://gist.github.com/MartinNowak/a471fe7ddbfeef205cdf04c93a94c6d0#file-release-sh-L61
Might make sense to add this as makefile rule, but it's already `release_changelog -> html -> $(ALL_FILES) -> dmd` again, not the most stable approach for (semi-)automation.

MartinNowak:
I'd rather revert #1817 for now, even though it contained some useful changes, and integrate the discussion and all related changes into https://github.com/dlang/dlang.org/pull/1821#issuecomment-315603750. Also slightly reducing the confusion of half a dozen of PRs.

PR #1824, opened 2017-07-15
Rename VERSION => STABLE_VERSION for the unification at the changed tool

wilzbach:
Small-scale version of https://github.com/dlang/dlang.org/pull/1821, s.t. we can move step by step.

### DIFF
--- a/changelog/changelog.ddoc
+++ b/changelog/changelog.ddoc
@@ -15,7 +15,7 @@ SUBNAV=$(SUBNAV_TEMPLATE
 SEARCH_OPTIONS_EXTRA=<option value="dlang.org/changelog" selected>Change
 Log</option>
 
-VERSION=
+STABLE_VERSION=
 $(DIVC version,
 $(P
 $(B $(LARGE $(LINK2 http://downloads.dlang.org/releases/2.x/$(VER), Download D $(VER))))$(BR)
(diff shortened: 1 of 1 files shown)


wilzbach:
As mentioned before, if `change.d` would be part of this repo I wouldn't need to go through these hops to please DAutoTest)...

changelog/changelog.ddoc
```diff
@@ -15,14 +15,15 @@ SUBNAV=$(SUBNAV_TEMPLATE
 SEARCH_OPTIONS_EXTRA=<option value="dlang.org/changelog" selected>Change
 Log</option>
 
-VERSION=
+STABLE_VERSION=
```
wilzbach: To be fair: it wouldn't be too hard to continue to generate `VERSION` at the `changed.d` script, but imho this helps clarifying things.

MartinNowak:
Let's wait a sec and properly think this through before churning more code.
https://github.com/dlang/dlang.org/pull/1821#issuecomment-315603750

changelog/changelog.ddoc
MartinNowak: Not sure I have the same understanding of STABLE_VERSION as you, so far we used that term to pin tools/compilers, but not for anything in our docs/targets.

PR #1827, opened 2017-07-17
Fix Issue 17649 - CONTRIBUTING.md instructions failed (no ../druntime…

wilzbach:


CyberShadow:
This CI failure is baffling.

wilzbach:
>Root package assert_writeln_magic references unknown package libdparse

This just means that the DUB registry is down. Sadly this happens way too often :/
Will force-push to restart.

posix.mak
```diff
@@ -46,6 +46,10 @@ DMD_STABLE=$(DMD_STABLE_DIR)/generated/$(OS)/release/$(MODEL)/dmd
 DRUNTIME_STABLE_DIR=${DRUNTIME_DIR}-${LATEST}
 PHOBOS_STABLE_DIR=${PHOBOS_DIR}-${LATEST}
 
+# Auto-closing missing directories
```
CyberShadow: cloning*

wilzbach:
@CyberShadow I am not sure whether you noticed that I changed quite a bit of the last commit in the final rebase. I forgot to mention, but aa we auto-clone all required repos, the contributing guide can be simplified even further.

CyberShadow:
I don't see anything objectionable either way. Haven't really been paying much attention to developer documentation other than that it doesn't get too long (because then that means we have problems elsewhere) ;)

CONTRIBUTING.md
```diff
@@ -11,36 +11,34 @@ Make sure you have these prerequisites working:
...
-* If you're running OSX, make sure you have some version of `libevent` installed
+* If you're running OSX, make sure you have some version of `libevent` installed (needed for Ddox)
 
 ## Getting the code
 
-Create a working directory for the D language, e.g. `~/code/d`. The remainder
-of this document calls that directory henceforth `$R` from "Root". To get the
-code:
+If you already have a [working directory for the D language](https://wiki.dlang.org/Building_under_Posix#Fetch_repositories_from_GitHub),
+e.g. `~/dlang`, change to it.
```
ZombineDev: Not sure that it is a good idea to suggest the same working directory as the one install.sh manages.

wilzbach: It use ~/dlang for a long time now and since we moved the checkouts into `.generated`, I haven't had any problems. Though it would have been nice for the install.sh script to support an alternate directory - see https://github.com/dlang/installer/pull/274

CyberShadow: > It use ~/dlang for a long time now

and I recall you said it caused problems due to conflicts between what the installer tool does and what the makefiles did.

I agree, we should change this, probably in the documentation.

wilzbach: Sorry for the typo: s/It/I/

> and I recall you said it caused problems due to conflicts between what the installer tool does and what the makefiles did.

Yep, it used to. However, we fixed that with https://github.com/dlang/dlang.org/pull/1889

> I agree, we should change this, probably in the documentation.

This was an outdated comment I saw by chance and accidentally never replied, so I thought it's a good idea to leave an answer here, but I don't see any need for action anymore.

CyberShadow: > Yep, it used to. However, we fixed that with #1889

That's just one problem we know about.

> This was an outdated comment I saw by chance and accidentally never replied, so I thought it's a good idea to leave an answer here, but I don't see any need for action anymore.

This is a minor issue, but otherwise I disagree. We have to places in the official material that refer to the same directory for different purposes. They don't conflict *now*, but it's still an unnecessary source of confusion and potential problems, for the same reason you don't use one variable for two different purposes.

PR #1830, opened 2017-07-19

MartinNowak:
Conflicts:
	changelog/changelog.ddoc


MartinNowak:
I've seen this error a couple of times, no idea why suddenly reappears here though.
```d
find .generated/phobos-release -name '*.d' | \
  sed -e /unittest.d/d -e /windows/d | sort >> .generated/.release-files.txt
../dmd-2.075.0/generated/linux/release/64/dmd -J../dmd-2.075.0/res -J../dmd-2.075.0/generated/linux/release/64/ -c -o- -version=CoreDdoc \
  -version=MARS -version=CoreDdoc -version=StdDdoc -Df.generated/.release-dummy.html \
  -Xf.generated/docs.json -I.generated/phobos-release @.generated/.release-files.txt
../dmd-2.075.0/src/ddmd/mars.d(47): Error: module objc is in file 'ddmd/objc.d' which cannot be read
import path[0] = .generated/phobos-release
```

wilzbach:
> I've seen this error a couple of times, no idea why suddenly reappears here though.

Because we build the documentation of dmd and you switched the `VERSION`, which leads to a newer dmd to be built. The solution is to do the same as the `prerelease.json below.

MartinNowak:
Thanks :).
The objc modules should just use matching module and filenames instead :).

wilzbach:
> The objc modules should just use matching module and filenames instead :).

There was also a discussion on the ML this year: http://forum.dlang.org/post/2D143DD8-A978-4C60-9C78-488E19802B88@me.com

wilzbach:
FYI: 2.075.0 is now live on dlang.org ;-)

jacob-carlborg:
> The objc modules should just use matching module and filenames instead :).

They are matching [1], or am I missing something? The glue part has not been refactored yet as the frontend part and therefore the `objc_glue_stubs.d` still exists.

[1] https://github.com/dlang/dmd/blob/master/src/ddmd/objc.d#L11

wilzbach:
> The glue part has not been refactored yet as the frontend part and therefore the objc_glue_stubs.d still exists.

Yep that's what Martin was talking about - `objc_glue_stubs.d` doesn't match the `objc.d` filename.

jacob-carlborg:
Ok, I'll try to get to the refactoring of the glue layer as well. But the slow progress of getting PRs accepted is not really encouraging.

PR #1834, opened 2017-07-19
Ignore generated PDF files

wilzbach:
Files in question:

```
images/qualifier-combinations-eps-converted-to.pdf
images/qualifier-conversions-eps-converted-to.pdf
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ dlangspec.out
 dlangspec.tex
 dlangspec.mobi
 dlangspec.toc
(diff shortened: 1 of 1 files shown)


CyberShadow:
General policy is that instead of adding more things to .gitignore, to move the files to .generated instead. Would that be trivial here?

wilzbach:
> Would that be trivial here?

It's the LaTeX pipeline that generates these files, so I am not sure whether it would be trivial.

PR #1837, opened 2017-07-21
Regenerate changelog on new changes

wilzbach:
From https://github.com/dlang/phobos/pull/5617#issuecomment-316623709

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -664,8 +664,13 @@ test: $(ASSERT_WRITELN_BIN)_test all
 ################################################################################
 # Changelog generation
 ################################################################################
+CHANGELOG_FILES=$(wildcard $(DMD_DIR)/changelog/*.dd) \
+				$(wildcard $(DRUNTIME_DIR)/changelog/*.dd) \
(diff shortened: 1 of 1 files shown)


posix.mak
```diff
@@ -653,16 +653,19 @@ test: $(ASSERT_WRITELN_BIN)_test all
 ################################################################################
 # Changelog generation
 ################################################################################
+CHANGELOG_FILES=$(wildcard $(DMD_DIR)/changelog/*.dd) \
+				$(wildcard $(DRUNTIME_DIR)/changelog/*.dd) \
+				$(wildcard $(PHOBOS_DIR)/changelog/*.dd)
```
wilzbach: Should we add `tools` here too? It has only been used once so far.

ZombineDev: Yes, ideally all repos that can have changelog entries - `installer` also looks like a good candidate. Though, I'm not sure about dub.

wilzbach:
Alrighty. Added tools + installer & rebased.

PR #1848, opened 2017-07-28
Add RPN calculator front-page example

CyberShadow:
Initial submission by Max Haughton, minimized by Timon Gehr.

http://forum.dlang.org/post/ol9ocn$13rs$2@digitalmars.com

Updated a bit to conform to D Style and horizontal space.

Blocked by general availability of static foreach in released compilers.

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -216,7 +216,7 @@ void main()
             // Generate operator switch cases statically
             foreach (c; AliasSeq!('+', '-', '*', '/'))
                 case [c]:
-                    return stack[0 .. $-2] ~
+                    return stack[0 .. $ - 2] ~
                         mixin("stack[$ - 2] " ~ c ~
                             " stack[$ - 1]");
             default: return stack ~ op.to!real;
--- a/index.dd
+++ b/index.dd
@@ -200,6 +200,30 @@ void main()
 }
 ----
 )
+$(EXTRA_EXAMPLE_RUNNABLE Tiny RPN calculator,
+$(RUNNABLE_EXAMPLE_STDIN 2 3 3 4 + * *)
+----
+void main()
+{
+    import std.stdio, std.string, std.algorithm, std.conv;
+
+    // Reduce the RPN expression using a stack
+    readln.split.fold!((stack, op)
+    {
+        switch (op)
+        {
+            // Generate operator switch cases statically
+            static foreach (c; "+-*/")
+                case [c]:
+                    return stack[0 .. $-2] ~
+                        mixin("stack[$ - 2] " ~ c ~
+                            " stack[$ - 1]");
+            default: return stack ~ op.to!real;
+        }
+    })((real[]).init).writeln;
+}
+----
+)
 
 ) $(COMMENT your-code-here)
 ))) $(COMMENT intro, div, div)
--- a/index.dd
+++ b/index.dd
@@ -205,7 +205,8 @@ $(RUNNABLE_EXAMPLE_STDIN 2 3 3 4 + * *)
 ----
 void main()
 {
-    import std.stdio, std.string, std.algorithm, std.conv;
+    import std.stdio, std.string,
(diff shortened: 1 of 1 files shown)


CyberShadow:
I guess we can use `AliasSeq` instead of `static foreach` for now.

Pushed that as a separate commit, so that we can easily revert it later.

index.dd
```diff
@@ -200,6 +200,31 @@ void main()
...
+
+    // Reduce the RPN expression using a stack
+    readln.split.fold!((stack, op)
+    {
+        switch (op)
+        {
+            // Generate operator switch cases statically
+            foreach (c; AliasSeq!('+', '-', '*', '/'))
+                case [c]:
+                    return stack[0 .. $-2] ~
```
tgehr: I guess the "D style" requires whitespace around "-"?

index.dd
```diff
@@ -200,6 +200,31 @@ void main()
...
+$(EXTRA_EXAMPLE_RUNNABLE Tiny RPN calculator,
+$(RUNNABLE_EXAMPLE_STDIN 2 3 3 4 + * *)
+----
+void main()
+{
+    import std.stdio, std.string,
+        std.algorithm, std.conv, std.meta;
+
+    // Reduce the RPN expression using a stack
+    readln.split.fold!((stack, op)
```
wilzbach: Q1: Why do we need `readln` here?
Putting the string directly has the following advantages:
- People don't need to switch tabs to browse stdin
- We can take advantage of the fast, cached run.dlang.io backend

Q2: Why not the non-allocating `splitter`?

index.dd
```diff
@@ -200,6 +200,31 @@ void main()
...
+    {
+        switch (op)
+        {
+            // Generate operator switch cases statically
+            foreach (c; AliasSeq!('+', '-', '*', '/'))
+                case [c]:
+                    return stack[0 .. $-2] ~
+                        mixin("stack[$ - 2] " ~ c ~
+                            " stack[$ - 1]");
+            default: return stack ~ op.to!real;
```
wilzbach: Q: Why not `double`?
Reasoning:
- `real` is usually a lot slower to compute
- people coming from other languages are more with `double` (exists in almost every other language)

index.dd
```diff
@@ -200,6 +200,31 @@ void main()
...
+        }
+    })((real[]).init).writeln;
```
wilzbach: Q: Why not `.back` or `[$ - 1]` here?
After all, we are only interested in the final value.

index.dd
CyberShadow: > Why do we need readln here?

Because a calculator program isn't very useful if it can calculate one single hard-coded expression.

> People don't need to switch tabs to browse stdin

I think this is an UI issue.

> We can take advantage of the fast, cached run.dlang.io backend

And this is definitely an implementation issue :)

index.dd
CyberShadow: `real` is just fine for a user-facing calculator app.

Not many languages support extended precision, I think D supporting it is good advertising.

It's also 2 characters shorter :)

index.dd
CyberShadow: Not necessarily. With an RPN calculator, you can calculate several values with one expression. Also, if the number of operators and operands is accidentally unbalanced (more operands than operators), the mistake will be silently ignored.

index.dd
CyberShadow: Fixed :)

index.dd
CyberShadow: > real is usually a lot slower to compute

BTW, this goes against my understanding of x86 FPU. I thought everything is calculated in 80-bit precision (because that's the precision used by the FPU stack)?

tgehr: For 64-bit executables, `double` will use SSE instructions. The performance difference for serial computations is not huge though, and the precision is actually better.

PR #1844, opened 2017-07-26
Use dub from the release binary

wilzbach:

PR #1850, opened 2017-07-29
Use .tar.xz balls if supported

wilzbach:
Discussion from https://github.com/dlang/dlang.org/pull/1844

> me: The `tar.xz` balls has about half the size of the zip archive and apart from making the CI generate less traffic on the AWS bucket, it's quite noticeably faster on slower connections.
If there's consensus about this, I would propose to do the same for the DMD bootstrapping.

> @CyberShadow: I wonder, how likely is it that `xz` won't be available on users' machines (and it would be too onerous to install it)?

> me: I thought about the same, but then I guessed that if I didn't include the fallback, chances of complaint / rejection would be high. I actually think that `zip` is less frequent than xz-utils

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -574,11 +574,16 @@ dpl-docs: ${DUB} ${STABLE_DMD}
 	DFLAGS="$(DPL_DOCS_DFLAGS)" ${DUB} build --root=${DPL_DOCS_PATH} \
 		--compiler=${STABLE_DMD}
 
+# .tar.xz's archives are smaller (and don't need a temporary dir) -> prefer if available
 ${STABLE_DMD_ROOT}/.downloaded:
-	mkdir -p ${STABLE_DMD_ROOT}
-	TMPFILE=$$(mktemp ${TMP}/dmd-download-deleteme.XXXXXXXX) && curl -fsSL ${STABLE_DMD_URL} > $${TMPFILE}.zip && \
-		unzip -qd ${STABLE_DMD_ROOT} $${TMPFILE}.zip && rm $${TMPFILE}.zip
-	touch $@
+	@mkdir -p $(dir $@)
+	@if command -v xz >/dev/null 2>&1 ; then \
+		curl -fSL --retry 3 $(subst .zip,.tar.xz,$(STABLE_DMD_URL)) | tar -Jxf - -C $(dir $@); \
+	else \
+		TMPFILE=$$(mktemp deleteme.XXXXXXXX) && curl -fsSL ${STABLE_DMD_URL} > ${TMP}/$${TMPFILE}.zip && \
+			unzip -qd ${STABLE_DMD_ROOT} ${TMP}/$${TMPFILE}.zip && rm ${TMP}/$${TMPFILE}.zip; \
+	fi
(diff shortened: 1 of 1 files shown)


CyberShadow:
@MartinNowak As release manager I'm guessing you're more familiar with the file formats situation, any insight you can offer?

posix.mak
```diff
@@ -578,10 +578,19 @@ dpl-docs: ${DUB} ${STABLE_DMD}
 	DFLAGS="$(DPL_DOCS_DFLAGS)" ${DUB} build --root=${DPL_DOCS_PATH} \
 		--compiler=${STABLE_DMD}
 
-${STABLE_DMD}:
-	mkdir -p ${STABLE_DMD_ROOT}
+# .tar.xz is archives are smaller (and don't need a temporary dir) -> prefer if available
+HAS_XZ := $(shell command -v xz 2> /dev/null)
```
MartinNowak: It was a bit more tricky to detect tar.xz support on older OSX versions, not sure if xz is available standalone by default nowadays.
https://github.com/dlang/installer/blob/139995801f53095af70c4393270ca86de1942e6b/script/install.sh#L110

posix.mak
```diff
@@ -578,10 +578,19 @@ dpl-docs: ${DUB} ${STABLE_DMD}
...
+${STABLE_DMD_ROOT}/.downloaded:
+	@mkdir -p $(dir $@)
+ifdef HAS_XZ
```
MartinNowak: I'd say that always running a shell command to define a global make variable (HAS_XZ) is a bit of an anti-pattern.
Might be nicer to just do the if/else in the shell, also provide more helpful output.
```make
    if command -v xz 2>/dev/null; then \
        curl -fSL --retry 3 $(subst .zip,.tar.xz,$(STABLE_DMD_URL)) | tar -Jxf - -C $(dir $@); \
    else \
        TMPFILE=$$(mktemp deleteme.XXXXXXXX) && curl -fsSL ${STABLE_DMD_URL} > ${TMP}/$${TMPFILE}.zip && \
            unzip -qd ${STABLE_DMD_ROOT} ${TMP}/$${TMPFILE}.zip && rm ${TMP}/$${TMPFILE}.zip; \
    fi
```
Just make sure to use semicolons (`;`) to end commands as `\` in make will not result in a newline, so the shell sees the command in a single line and needs `;` separators `if exp; then exp; else exp; fi`.

wilzbach:
> It was a bit more tricky to detect tar.xz support on older OSX versions, not sure if xz is available standalone by default nowadays.

No idea either, but falling back to `zip` doesn't hurt either for now.
I don't have a OSX machine, so if someone likes this, it's probably better for him/her to do it directly.

> I'd say that always running a shell command to define a global make variable (HAS_XZ) is a bit of an anti-pattern.
> Might be nicer to just do the if/else in the shell, also provide more helpful output.

Good idea. Thanks.

wilzbach:
I was just working on a very slow mobile connection and I could have really used the smaller .tar.gz checkout, thus I'm going to move forward with this, s.t. other people on slow connections don't end in the same misery as I just did.

PR #1853, opened 2017-08-02
Document the proper return value of IsExpression

MetaLang:
The spec says that an `IsExpression` returns an int with value 1 or 0, but it actually returns a boolean with value true or false.


CyberShadow:
Looks like this changed back in 2.004:

http://dlang.org/changelog/2.004.html

> * Changed result type of IsExpression from int to bool.

ZombineDev:
Thanks @MetaLang

PR #1854, opened 2017-08-02
Add hex viewer runnable example

CyberShadow:
Adapted from
https://forum.dlang.org/post/dvjltobyaaxoihhqywhp@forum.dlang.org

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -129,11 +129,14 @@ void main()
 {
     import std.algorithm, std.stdio, std.file;
     enum cols = 16;
+    // Split file into 16-byte chunks per row
     thisExePath.File("rb").byChunk(cols).each!(chunk =>
+        // Use range formatting to format the
+        // hexadecimal part and align the text part
         writefln!"%(%02X %)%*s  %s"(
             chunk,
-            3 * (cols - chunk.length), "",
-            chunk.map!(c =>
+            3 * (cols - chunk.length), "", // Padding
+            chunk.map!(c => // Replace non-printable
                 c < 0x20 || c > 0x7E ? '.' : char(c))));
 }
 ----
--- a/index.dd
+++ b/index.dd
@@ -123,6 +123,21 @@ void main()
 }
 ----
 )
+$(EXTRA_EXAMPLE_RUNNABLE Print hex dump,
+----
+void main()
+{
+    import std.algorithm, std.stdio, std.file;
+    enum cols = 16;
+    thisExePath.File("rb").byChunk(cols).each!(chunk =>
+        writefln!"%(%02X %)%*s  %s"(
+            chunk,
+            3 * (cols - chunk.length), "",
+            chunk.map!(c =>
+                c < 0x20 || c > 0x7E ? '.' : char(c))));
+}
+----
+)
 $(EXTRA_EXAMPLE Start a minimal web server,
 ----
 #!/usr/bin/env dub


index.dd
```diff
@@ -123,6 +123,21 @@ void main()
...
 ----
 )
+$(EXTRA_EXAMPLE_RUNNABLE Print hex dump,
+----
+void main()
+{
+    import std.algorithm, std.stdio, std.file;
+    enum cols = 16;
+    thisExePath.File("rb").byChunk(cols).each!(chunk =>
+        writefln!"%(%02X %)%*s  %s"(
```
dgileadi: This code depends on the magic of formatted write, which is obscure and arcane. It could use some comments and/or a link to the writefln docs.

ZombineDev:
The example looks great, though running it via dlang-tour yields:
> Program's output exceeds limit of 4096 bytes.

4096 is quite restrictive, though increasing the limit into the MB range (probably what's needed fit the binary) sounds like an easy DoS attack vector. For this example, I suggest limiting the output to 8 lines (by adding `.take(16)` after `.byChunk(cols)`), as more than this doesn't fit the editor/output window.

Anyway, @CyberShadow I would welcome your expertise on what would be a good and practical limit.

I agree with @dgileadi that adding 2-3 lines of explanation would help a great deal to people inexperienced with language / library. On the contrary I think that this is examplary use of `std.format`, but I may be biased as used it quite often.

CyberShadow:
I think truncating the output to 5-10 KB would be more reasonable than not showing any of it. @wilzbach

CyberShadow:
Any suggestions for the comments? Space is a bit tight.

CyberShadow:
We can use `__FILE__` instead of `thisExePath`:

https://run.dlang.io/is/0RPmys

dgileadi:
You're right, it's hard with the short space. Maybe something like:

```
// format: an array of hex numbers, then * spaces, then content
// see https://dlang.org/phobos/std_format.html
```

wilzbach:
> I think truncating the output to 5-10 KB would be more reasonable than not showing any of it. @wilzbach

Sounds reasonable to me as well.
For the record - I didn't set this limit. This is part of the initial DTour core and it just happen to be the only one (with @ZombineDev) who continues to maintain and develop it from time to time ;-)
I am a bit busy for the next two days, but I will have time on Sunday to have a look at changing this limit.

ZombineDev:
How about adding a link below each example pointing to an article with detailed explanation about each technique that's been used? We can use the wiki or add a section on dlang-tour.

dgileadi:
I love that idea—it lets us show off some of the features that make D so great, in the guise of explaining an example.

I still think that we need to make our examples short and simple, and comment bits that are confusing. First impressions are important, and these code examples are our first impression.

index.dd
MetaLang: Another suggestion is to use the version of `writefln` that takes a runtime since all C and C++ programmers should understand those.

CyberShadow: Not sure that would be much of an improvement.

- C doesn't have ranges or range formatting syntax in printf
- The fmt!formatString is pretty similar to Rust - C programmers interested in D will almost certainly also be interested in Rust
- Compile-time format string parsing is awesome - D didn't have support for it for a long time, which meant mistakes in the format string weren't caught until runtime.

CyberShadow:
One (rather far-fetched) wish is for runnable examples to have interactive links/popups for all symbols, i.e. hovering over a Phobos identifier would show a popup with a small help text, and clicking on it would go to its full documentation page in the Phobos docs.

CyberShadow:
Added comments.

JinShil:
LGTM

wilzbach:
> One (rather far-fetched) wish is for runnable examples to have interactive links/popups for all symbols, i.e. hovering over a Phobos identifier would show a popup with a small help text, and clicking on it would go to its full documentation page in the Phobos docs.

FYI: This shouldn't be too difficult. Ddox already uses an index of all symbol. The only work left is stamping the code into words and looking each up in the index.

PR #1855, opened 2017-08-03
Add Netflix and Symmetry to Organisations using D

Laeeth:
@wilzbach hope it is okay.  I'm not so good with ddoc.


wilzbach:
> @wilzbach hope it is okay

No worries, thanks a  lot for submitting and pushing this!
I will fix the details later, but we need the permission from Netflix to be listed here.
Maybe @rostyboost can help and confirm that Netflix is okay with being listed on the organizations using D page?

Laeeth:
Could we maybe add Symmetry now and Netflix when they agree?

rostyboost:
You can go ahead and list Netflix. However please use the following [logo](https://media.netflix.com/downloadCorporateAsset?assetId=11608). Thanks

ZombineDev:
I've updated @Laeeth's branch to address @rostyboost's request.

PR #1857, opened 2017-08-05
Need to distinguish buint-in and user-defined op=

Bolpat:
That is not only a spec thing. In the real world, most `op=` overloads return `void`, while built-ins return the result of `a op b`.

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -79,7 +79,7 @@ $(GNAME AssignExpression):
 
 $(H3 $(LNAME2 assignment_operator_expressions, Assignment Operator Expressions))
 
-    $(P Assignment operator expressions, such as:)
+    $(P For argumets of built-in types, assignment operator expressions such as)
 
         --------------
         a op= b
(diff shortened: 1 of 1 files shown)


spec/expression.dd
```diff
@@ -79,7 +79,7 @@ $(GNAME AssignExpression):
 
 $(H3 $(LNAME2 assignment_operator_expressions, Assignment Operator Expressions))
 
-    $(P Assignment operator expressions, such as:)
+    $(P For argumets of built-in types, assignment operator expressions such as)
```
schveiguy: arguments

spec/expression.dd
```diff
@@ -91,12 +91,16 @@ $(H3 $(LNAME2 assignment_operator_expressions, Assignment Operator Expressions))
...
-        $(LI operand $(D a) is only evaluated once)
-        $(LI overloading $(I op) uses a different function than overloading $(I op)= does)
-        $(LI the left operand of $(D >>>=) does not undergo integral promotions before shifting)
+        $(LI operand $(D a) is only evaluated once,)
+        $(LI overloading $(I op) uses a different function than overloading $(I op)= does, and)
+        $(LI the left operand of $(D >>>=) does not undergo integral promotions before shifting.)
+    )
+    
+    $(P For user-defined types, assignment operator expressions can be overloaded separately from
+        the binary operator. Still the left operand must be an lvalue.
```
schveiguy: I think actually, it *must* be overloaded separately. I tried just defining opBinary!"+", and += didn't work. i.e. there is no equivalent rewrite for user types.

spec/expression.dd
```diff
@@ -91,12 +91,16 @@ $(H3 $(LNAME2 assignment_operator_expressions, Assignment Operator Expressions))
...
 
     $(UL
-        $(LI operand $(D a) is only evaluated once)
-        $(LI overloading $(I op) uses a different function than overloading $(I op)= does)
-        $(LI the left operand of $(D >>>=) does not undergo integral promotions before shifting)
+        $(LI operand $(D a) is only evaluated once,)
+        $(LI overloading $(I op) uses a different function than overloading $(I op)= does, and)
+        $(LI the left operand of $(D >>>=) does not undergo integral promotions before shifting.)
+    )
+
```
schveiguy: Apparently there is trailing whitespace on this line.

spec/expression.dd
Bolpat: Right. I wonder why there is no mixin template in Phobos for doing these correctly and en masse.
With *can* I meant there could be an overload that does not behave as the example does.
I'll fix the wording.

Bolpat:
@schveiguy: Anything left to do?

schveiguy:
Sorry, was waiting until the docs were built, and I got pulled into other things :)

PR #1861, opened 2017-08-09
Fix error

Bolpat:
When using the spec as-is, the compiler gives the error message
```
AA key type Key does not have 'bool opEquals(ref const Key) const'
```
Probably the compiler is right.


MetaLang:
@andralex @WalterBright

PR #1864, opened 2017-08-11
Wrong references removed

drug007:
Deleted references to documentation not related to modules.

### DIFF
--- a/spec/module.dd
+++ b/spec/module.dd
@@ -10,31 +10,6 @@ $(GNAME Module):
 $(GNAME DeclDefs):
     $(GLINK DeclDef)
     $(GLINK DeclDef) $(I DeclDefs)
-
-$(GNAME DeclDef):
-    $(GLINK2 attribute, AttributeSpecifier)
-    $(GLINK2 declaration, Declaration)
-    $(GLINK2 class, Constructor)
-    $(GLINK2 class, Destructor)
-    $(GLINK2 struct, Postblit)
-    $(GLINK2 class, Allocator)
-    $(GLINK2 class, Deallocator)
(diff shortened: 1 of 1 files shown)


drug007:
Probably, don't know how to verify. There are wrong references in [Modules documentation](http://dlang.org/spec/module.html)
Second `DeclDef:` provides references to other documentation that is confusing for me.

aG0aep6G:
> Probably, don't know how to verify. There are wrong references in Modules documentation

As far as I see, the references are correct. There's certainly no reason to just delete that part of the grammar.

> Second DeclDef: provides references to other documentation that is confusing for me.

There's only one `DeclDef:`. It links to other parts of the documentation, because that's where the symbols are defined.

A `DeclDef` can be an `AttributeSpecifier`, or a `Declaration`, or a `Constructor`, etc.
`AttributeSpecifier` is defined in attribute.html, so that's where the link goes. Same with the others.

drug007:
Hmm, definitly today is not my day) I used `wrong` meaning that presence of these references was confusing for me when I was looking for the information about modules. I had impression that modules had `allocators` etc so I decided to improve the docs. But probably I'm wrong of course.

aG0aep6G:
> I had impression that modules had allocators etc so I decided to improve the docs.

I see. Constructors, destructors, and various others from the list can't actually occur at the module level. But the grammar says they can.

If you're up for it, you could try to modify the grammar so that it doesn't allow those things at the module level, but still allows them in [`AggregateBody`](http://dlang.org/spec/grammar.html#AggregateBody).

drug007:
Thanks for comments. I'll try to touch the grammar.

PR #1869, opened 2017-08-18
Remove wrong changelog entry

Calrama:
The fix for [ssue 14246](https://issues.dlang.org/show_bug.cgi?id=14246) was reverted, it is not included in 2.075.0
(see bugzilla page for details), and the issue has been reopened.

### DIFF
--- a/changelog/2.075.0.dd
+++ b/changelog/2.075.0.dd
@@ -426,7 +426,6 @@ $(BUGSTITLE DMD Compiler bugs,
 
 $(LI $(BUGZILLA 7016): local import does not create -deps dependency)
 $(LI $(BUGZILLA 13331): naked asm functions are broken when compiling with -profile)
-$(LI $(BUGZILLA 14246): RAII - proper destruction of partially constructed objects/structs)
 $(LI $(BUGZILLA 15896): private ignored when import bindings are used)
 $(LI $(BUGZILLA 15945): sizeof on an invalid type seems to compile.)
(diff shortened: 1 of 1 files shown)


ZombineDev:
This PR needs to target the `stable` branch. As a general rule of thumb, all bug fixes need to target `stable` branch of the respective dlang repo.

ZombineDev:
It turns you can now change the target branch of pull request after it has been opened and I just did that. So the only thing left to do is to rebase your branch on top of `dlang.org/stable` to get rid of the unnecessary merge commits.

Calrama:
Thank you, rebased upon stable branch. Interestingly, it required a force push.

wilzbach:
> This PR needs to target the stable branch

FYI master is deployed on dlang.org, not stable

ZombineDev:
> FYI master is deployed on dlang.org, not stable

Yes, but `stable` is what goes into the release archive, right? So the only way to fix both is to merge the PR into `stable` and then merging `stable` into `master`, otherwise fixes to the changelog would be delayed to the 2.077.0 release archive.

wilzbach:
> Yes, but stable is what goes into the release archive, right? 

Do you mean the offline copy of dlang.org that is shipped in the official release? Yes that one is built from stable, but I highly doubt that anyway looks at that.
Btw this changelog was and will be generated automatically from Bugzilla again, so I wouldn't worry about the stable branch.

ZombineDev:
> Btw this changelog was and will be generated automatically from Bugzilla again

Are you sure? AFAIK, once a version is released, its changelog will never be regenerated. (This PR is for a past release, not for the recently branched 2.076.0-beta1.)

PR #1871, opened 2017-08-20
document -betterC switch better

WalterBright:


JinShil:
@WalterBright [You said](https://github.com/dlang/dmd/pull/6918#issuecomment-309576639) "...betterC is a work in progress, I don't know yet what all it will need to do."  That's perfectly fine, but let's make that known in the documentation.  

I suggest adding something like "-betterC is an unstable, experimental feature, and is subject to changes without notice."

PR #1872, opened 2017-08-20
Clarification on @property functions

Bolpat:
[This question](https://stackoverflow.com/questions/44887077/d-properties-with-two-arguments) on stackoverflow and my personal struggling with the sentence suggests to add clarification.
I've done some tests with static member functions.
```d
@property bool parity(int x) { return x & 1; }
@property void parity(ref int x, bool p) { (x |= 1) &= p; }
```
work well. For static member functions like these
```d
struct Test
{
  static:
    @property bool parity(int x) { return x & 1; }
    @property void parity(ref int x, bool p) { (x |= 1) &= p; }
    
    void test()
    {
        int x = 1;
        bool p = parity(x); // (1)
        int y = 2;
        parity(y, p); // (2)
    }
}
```
the `@property` has no effect: Replace `(1)` and `(2)` by `bool p = x.parity;` and `y.parity = p;` respectively, you get the error *no property 'parity' for type 'int'*.

I don't know if this is intensional or a bug.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -559,7 +559,8 @@ $(H2 $(LNAME2 property-functions, Property Functions))
 
     $(UL
     $(LI $(D @property) functions cannot be overloaded with non-$(D @property) functions with the same name.)
-    $(LI $(D @property) functions can only have zero, one or two parameters.)
+    $(LI $(D @property) functions can only have zero or one parameter if they are non-static members.)
+    $(LI $(D @property) functions can only have one or two parameters if they are non-members.)
     $(LI $(D @property) functions cannot have variadic parameters.)
     $(LI For the expression $(D typeof(exp)) where $(D exp) is an $(D @property) function,
     the type is the return type of the function, rather than the type of the function.)


spec/function.dd
```diff
@@ -559,7 +559,8 @@ $(H2 $(LNAME2 property-functions, Property Functions))
 
     $(UL
     $(LI $(D @property) functions cannot be overloaded with non-$(D @property) functions with the same name.)
-    $(LI $(D @property) functions can only have zero, one or two parameters.)
+    $(LI $(D @property) functions can only have zero or one parameter if they are non-static members.)
+    $(LI $(D @property) functions can only have one or two parameters if they are non-members.)
```
jmdavis: What does `static` have to do with anything? Whether the member function is static or not, if it's a getter or a setter that returns by `ref`, it has no parameters, and if it's a setter that does not return by `ref` it has one parameter. If it's a free function, then a getter has zero parameters and is used by itself, or it has one parameter and is used with UFCS, and if it's a setter, then it either has one parameter if it's used without UFCS and two if it's used with UFCS. `static` has no effect on any of this.

jmdavis:
Okay. You clearly don't understand how `@property` works with a static function. It has to use either the name of the struct/class with the `@property` function or use an instance. e.g. std.datetime.date.Date has `Date.max`.

Neither `bool p = x.parity;` nor `y.parity = p;` is attempting to call a static function. In order to call the static functions on `Test`, you'd need `Test` - e.g. `Test.parity` or `Test.parity = something`. I suppose that the `bool` overload of `parity` could work as a setter that returned `bool`, as bizarre as that would be, but neither works as a getter, and the other overload is not a valid `@property` function, because it has too many parameters.

I don't think that there's anything wrong with what the spec currently says. It's just that it doesn't necessarily go into enough detail. If we wanted to make it clearer, we'd probably need to provide examples of non-static member functions, member functions, and free functions with getters, setters, and functions that return by `ref` so that they're both getters and setters. I don't think that adjusting this particular piece of text is really going to do much by itself.

jmdavis:
The other issue here is that none of this really has much to do with `@property`, in spite of the fact that it was originally supposed to. All of these rules with regards to the number of parameters and whatnot relate to whether a function can be called as a property function (which does not require `@property`), and the compiler clearly does not consider it an error to mark a function with `@property` when it can't be used as a property. AFAIK, all `@property` really does at this point is make it so that you can't overload the function with another function that isn't marked with `@property` and affect the type of `is(foo.propFunc)` so that if it's a getter property, the result is the return type and not the type of the function (I'm not sure what it does with a setter property). So, this whole section of the spec should probably be rewritten.

PR #1874, opened 2017-08-23
Purge the .sort and .reverse array properties

kuettler:
Those are gone in 2.075.0

### DIFF
--- a/spec/arrays.dd
+++ b/spec/arrays.dd
@@ -423,10 +423,6 @@ Returns an array literal with each element of the literal being the $(D .init) p
         $(TROW $(D .ptr), Returns a pointer to the first element of the array.)
         $(TROW $(D .dup), Create a dynamic array of the same size and copy the contents of the array into it. The copy will have any immutability or const stripped. If this conversion is invalid the call will not compile.)
         $(TROW $(D .idup), Create a dynamic array of the same size and copy the contents of the array into it. The copy is typed as being immutable. If this conversion is invalid the call will not compile.)
-        $(TROW $(D .reverse), Reverses in place the order of the elements in the array.
-        Returns the array.)
-        $(TROW $(D .sort), Sorts in place the order of the elements in
-        the array. Returns the array.)
         )
 
         $(P Dynamic array properties are:)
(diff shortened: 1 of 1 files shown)


schveiguy:
This should have gone into stable, no?

wilzbach:
> This should have gone into stable, no?

Master is automatically deployed. Maybe we should get rid of master here to about the constant confusion. What do you think @CyberShadow?

CyberShadow:
Sorry, I don't understand what you mean. "get rid of master"?

wilzbach:
> Sorry, I don't understand what you mean. "get rid of master"?

Sorry for the sloppy language. I was referring to removing the master branch and setting stable as default

ZombineDev:
I guess that would be the same as removing the `stable` branch. @MartinNowak do we really need stable here? For the release archives, can we just bundle the latest master?

CyberShadow:
I think that would cause more problems than it would solve.

Essentially the `stable` branch should only be targeted for changes that should make it into the included HTML documentation of point-releases. `master` affects the documentation of DMD/Phobos `master`. One branch would mean that either we're completely unable to update the documentation of point-releases, or changes in git `master` would end up in the documentation of point-releases.

MetaLang:
@schveiguy you're right, I need to start paying better attention to which branch is being targeted. I can make a PR tonight to to add this to stable as well.

schveiguy:
> Master is automatically deployed.

To dlang.org? That's not right, it should be stable that's automatically deployed!

Vladimir is right, stable should match stable branches of other repos, master should match master.

ZombineDev:
@schveiguy

> That's not right, it should be stable that's automatically deployed!

Meaning that either:
* We should wait 3 months between updates to the website for new features (e.g. runnable examples, new articles, etc.)
* All web development should happen in stable

schveiguy:
No, what I mean is that dlang.org master should track the master branch of the various tools, stable should track the current release (including the current compiler release).

If master of dlang.org is what is deployed on the actual web site, then the docs don't match what people actually download.

PR #1883, opened 2017-09-01
changelog/2.076.0.dd: Fix trailing whitespace

CyberShadow:
Needed to unbreak dlang.org tests.


rainers:
Tried that, too, in #1882, but it fails due to white spaces in 2.077.0_pre.dd. I don't find that file, though.

CyberShadow:
Thanks. The _pre will be fixed when https://github.com/dlang/dmd/pull/7115 is merged.

PR #1884, opened 2017-09-04
fix Issue 17798 - [2.076] "static foreach" not documented

tgehr:
Note: I was not able to build stable with `make -fposix.mak`. Therefore this is _untested_.

### DIFF
--- a/spec/grammar.dd
+++ b/spec/grammar.dd
@@ -579,6 +579,7 @@ $(GNAME NonEmptyStatementNoCaseNoDefault):
     $(GLINK MixinStatement)
     $(GLINK ForeachRangeStatement)
     $(GLINK ConditionalStatement)
+    $(GLINK StaticForeachStatement)
     $(GLINK StaticAssert)
     $(GLINK TemplateMixin)
     $(GLINK ImportDeclaration)
@@ -666,8 +667,11 @@ $(GNAME Increment):
 )
 
 $(GRAMMAR
+$(GNAME AggregateForeach):
+    $(I Foreach) $(D $(LPAREN)) $(GLINK ForeachTypeList) $(D ;) $(GLINK ForeachAggregate) $(D $(RPAREN))
+
 $(GNAME ForeachStatement):
-    $(I Foreach) $(D $(LPAREN)) $(GLINK ForeachTypeList) $(D ;) $(GLINK ForeachAggregate) $(D $(RPAREN)) $(PS0)
+    $(GLINK AggregateForeach) $(PS0)
 
 $(GNAME Foreach):
     $(D foreach)
@@ -680,6 +684,7 @@ $(GNAME ForeachTypeList):
 $(GNAME ForeachType):
     $(GLINK ForeachTypeAttributes)$(OPT) $(GLINK2 declaration, BasicType) $(GLINK2 declaration, Declarator)
     $(GLINK ForeachTypeAttributes)$(OPT) $(I Identifier)
+    $(GLINK ForeachTypeAttributes)$(OPT) $(D alias) $(I Identifier)
 
 $(GNAME ForeachTypeAttributes)
     $(GLINK ForeachTypeAttribute)
@@ -688,20 +693,24 @@ $(GNAME ForeachTypeAttributes)
 $(GNAME ForeachTypeAttribute):
     $(D ref)
     $(GLINK2 declaration, TypeCtor)
+    $(D enum)
 
 $(GNAME ForeachAggregate):
     $(EXPRESSION)
 )
 
 $(GRAMMAR
-$(GNAME ForeachRangeStatement):
-    $(GLINK Foreach) $(D $(LPAREN)) $(GLINK ForeachType) $(D ;) $(GLINK LwrExpression) $(D ..) $(GLINK UprExpression) $(D $(RPAREN)) $(PSSCOPE)
+$(GNAME RangeForeach):
+    $(GLINK Foreach) $(D $(LPAREN)) $(GLINK ForeachType) $(D ;) $(GLINK LwrExpression) $(D ..) $(GLINK UprExpression) $(D $(RPAREN))
 
 $(GNAME LwrExpression):
     $(EXPRESSION)
 
 $(GNAME UprExpression):
     $(EXPRESSION)
+
+$(GNAME ForeachRangeStatement):
+    $(GLINK RangeForeach) $(PSSCOPE)
 )
 
 $(GRAMMAR
@@ -1016,6 +1025,9 @@ $(GNAME Declaration):
     $(GLINK AggregateDeclaration)
     $(GLINK EnumDeclaration)
     $(GLINK ImportDeclaration)
+    $(GLINK2 version, ConditionalDeclaration)
+    $(GLINK2 version, StaticForeachDeclaration)
+    $(GLINK2 version, StaticAssert)
 )
 
 $(GRAMMAR
@@ -1729,6 +1741,19 @@ $(GNAME VersionSpecification):
 $(GNAME DebugSpecification):
     $(D debug =) $(I Identifier) $(D ;)
     $(D debug =) $(GLINK_LEX IntegerLiteral) $(D ;)
+)
+
+$(GRAMMAR
+$(GNAME StaticForeach):
+    $(D static) $(GLINK2 statement, AggregateForeach)
+    $(D static) $(GLINK2 statement, RangeForeach)
+
+$(GNAME StaticForeachDeclaration):
+    $(GLINK StaticForeach) $(GLINK2 attribute, DeclarationBlock)
+    $(GLINK StaticForeach) $(D :) $(GLINK2 module, DeclDefs)$(OPT)
+
+$(GNAME StaticForeachStatement):
+    $(GLINK StaticForeach) $(GLINK2 statement, NoScopeNonEmptyStatement)
 )
 
 $(GRAMMAR
--- a/spec/version.dd
+++ b/spec/version.dd
@@ -108,7 +108,7 @@ else
 }
 ------
 
-$(P The $(D version(unittest)) is satisfied if and only if the code is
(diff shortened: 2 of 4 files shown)


MetaLang:
`fatal: 'refs/ae-sys-d-cache/website-865504ca7123aa6b75c71a5f783b02a5e26dbcae-7bad98e9d56422f52983962526ed878a' - not a valid ref`

If I recall correctly, this is a problem with Github sometimes sending the wrong commit hash. I can't remember what the workaround is... I think closing and re-opening this PR will re-run it.

tgehr:
@BBasile @MetaLang Thanks!

BBasile:
okay thanks, looks good but i want to see the build now, it's not up to you @tgehr, you 've done the job, it's up to the build to success.

BBasile:
@tgehr you need to rebase and it will work. Or close reopen.

tgehr:
@BBasile There are no new commits in stable. Am I missing something?

BBasile:
@tgehr yes, indeed you are missing https://github.com/dlang/dlang.org/pull/1885 which should solve the failure.

tgehr:
@BBasile: That pull request went to master, not stable. I guess this was accidental?

CyberShadow:
> That pull request went to master, not stable. I guess this was accidental?

Argh. Oops.

BBasile:
@tgehr I think that the rebase makes more sense now. Sorry for the previous failure but since i 'm the reporter on bugzilla i feel implicated in a way.

spec/version.dd
```diff
@@ -537,6 +538,127 @@ INT!(17) c;  // error, static assert trips
 	)
 	)
 
+$(H2 $(LNAME2 staticforeach, Static Foreach))
```
BBasile: there's an error here i think because of the case

BBasile: `StaticForeach`

tgehr: There is also this line in the file, from which it is derived.
`$(H2 $(LNAME2 staticif, Static If Condition))`

I don't think whatever error there is is actually in this line.

BBasile: Yeah, i may have misinterpreted the log, sorry.

spec/grammar.dd
```diff
@@ -680,6 +684,7 @@ $(GNAME ForeachTypeList):
 $(GNAME ForeachType):
     $(GLINK ForeachTypeAttributes)$(OPT) $(GLINK2 declaration, BasicType) $(GLINK2 declaration, Declarator)
     $(GLINK ForeachTypeAttributes)$(OPT) $(I Identifier)
+    $(GLINK ForeachTypeAttributes)$(OPT) $(D alias) $(I Identifier)
```
ZombineDev: It looks like `static foreach (enum ref alias T; TypeList)` is valid syntax. Is this correct?
Ditto for `foreach (enum ref x; someRange)`.

ZombineDev: (Should `ForeachTypeAttributes` be allowed before `alias`?)

spec/grammar.dd
```diff
@@ -688,20 +693,24 @@ $(GNAME ForeachTypeAttributes)
 $(GNAME ForeachTypeAttribute):
     $(D ref)
     $(GLINK2 declaration, TypeCtor)
+    $(D enum)
```
ZombineDev: I wonder if `enum` should not be a `ForeachTypeAttribute`, but instead listed as a separate `ForeachType` entry, similar to `alias`.

spec/grammar.dd
```diff
@@ -1732,6 +1744,19 @@ $(GNAME DebugSpecification):
 )
 
 $(GRAMMAR
+$(GNAME StaticForeach):
+    $(D static) $(GLINK2 statement, AggregateForeach)
+    $(D static) $(GLINK2 statement, RangeForeach)
+
+$(GNAME StaticForeachDeclaration):
+    $(GLINK StaticForeach) $(GLINK2 attribute, DeclarationBlock)
+    $(GLINK StaticForeach) $(D :) $(GLINK2 module, DeclDefs)$(OPT)
```
ZombineDev: Can you give an example for this grammar rule? Since `static if` doesn't support trailing colon, I'm wondering if `static foreach` should not support it too.

spec/version.dd
```diff
@@ -537,6 +538,127 @@ INT!(17) c;  // error, static assert trips
...
+    $(GLINK StaticForeach) $(GLINK2 attribute, DeclarationBlock)
+    $(GLINK StaticForeach) $(D :) $(GLINK2 module, DeclDefs)$(OPT)
+
+$(GNAME StaticForeachStatement):
+    $(GLINK StaticForeach) $(GLINK2 statement, NoScopeNonEmptyStatement)
+)
+
+	$(P The aggregate/range bounds are evaluated at compile time and
+	turned into a sequence of compile-time entities by evaluating
+	corresponding code with a $(GLINK2 statement
```
ZombineDev: Missing comma at the end of the line (the link to `ForeachStatement` is broken).

spec/version.dd
```diff
@@ -537,6 +538,127 @@ INT!(17) c;  // error, static assert trips
...
+    $(D static) $(GLINK2 statement, RangeForeach)
+
+$(GNAME StaticForeachDeclaration):
+    $(GLINK StaticForeach) $(GLINK2 attribute, DeclarationBlock)
+    $(GLINK StaticForeach) $(D :) $(GLINK2 module, DeclDefs)$(OPT)
+
+$(GNAME StaticForeachStatement):
+    $(GLINK StaticForeach) $(GLINK2 statement, NoScopeNonEmptyStatement)
+)
+
```
ZombineDev: Perhaps you should specify what happens with Ddoc when `StaticForeachDeclaration` declares functions preceded by a ddoc comment. Is documentation generated for each function, or is there only once instance. With `static if` I think there's a trick that goes along the lines of `static if (condition || generatingDoc) /** This function is available iff ... */ void foo();`. Another question is whether you can generate documentation and mix it in in for each declaration (e.g. declare an overload set where each function has the same documentation, save for the parameters section).

spec/version.dd
```diff
@@ -537,6 +538,127 @@ INT!(17) c;  // error, static assert trips
...
+	used to generate declarations:
+	)
+
+------
+import std.range : iota;
+import std.algorithm : map;
+import std.conv : text;
+static foreach(i; iota(0, 3).map!text)
+{
+    mixin(`enum x`~i~`=i;`);
```
ZombineDev: Currently the coding style of the spec is all over the place, but AFAIK the goal is for new code to be formatted according to DStyle (where applicable), so in this case you would need to add a space around operators (https://dlang.org/dstyle.html#phobos_whitespace).

spec/version.dd
```diff
@@ -537,6 +538,127 @@ INT!(17) c;  // error, static assert trips
...
+    return r;
+}
+
+static foreach(i;0..200)
+{
+    static assert(test(i) == (i<100 ? i : -1));
+}
+-------
+
+	$(P An explicit $(D break)/$(D continue) label can be used to avoid this limitation.
```
ZombineDev: Maybe also worth making explicit is that you can't (at least you shouldn't be able to?) put a label before `static foreach` or any of the other conditional statement / declarations (e.g. `static if`).

ZombineDev:
@tgehr FYI, now that CyberShadow's DAutoTest is green you can see preview of your changes by clicking on the respective "Details" button.

spec/grammar.dd
tgehr: `static if` supports trailing semicolon.

spec/grammar.dd
tgehr: `enum` is a storage class, so it would be inconsistent not to allow it.

spec/grammar.dd
tgehr: I don't think that separating grammar rules to complicate the parser makes sense. It's good for the parser to be lenient. Semantic analysis should catch it (the resulting error messages are better).

It might however be sensible to not allow `ForeachTypeAttributes` before alias (the reasoning was that global `alias` allows (some) storage classes, but it turns out that local declarations also do not allow storage classes before `alias`), but this pull request documents existing behaviour. You might want to open an enhancement request.

spec/grammar.dd
ZombineDev: Hmm, I missed that even though I searched for it, nevermind then.

spec/grammar.dd
ZombineDev: I don't think we need to change the compiler if we restructure the grammar so it is a bit harder to produce semantically invalid code.
Yes, you can (and most likely you already do) catch this during semantic analysis, however I feel it would be better for the reader of the spec if the grammar rules match more closely what is semantically valid.
Presently, the reader wouldn't have any clue that `foreach (shared ref enum inout alias x; aggregate)` should be rejected by the compiler - the intended semantics are not specified.

MetaLang:
Does this need to be squashed before merge? I remember someone saying that commits are squashed for this repo on merge but I don't know if that's actually true.

wilzbach:
>  I remember someone saying that commits are squashed for this repo on merge but I don't know if that's actually true.

You can select the merge type once on merging (press the tiny arrow next to the merge button to switch the method) - and FWIW this is enabled for D all repos.

MetaLang:
Right, but then what's the point of the auto-merge tag? Just to make sure nobody merges something broken?

wilzbach:
> Right, but then what's the point of the auto-merge tag? Just to make sure nobody merges something broken?

Well you can't merge anything broken (GitHub won't allow this). The point of auto-merge is that you don't need to wait until the CIs have run. There's also been auto-merge-squash, but due to "abuse" this has been temporarily removed.

Squashed it an rebased to `master` (that's the branch which gets auto-deployed on dlang.org - it is a constant point of confusion :S).

@ZombineDev you are still on "Requesting changes" - anything from your side that blocks this PR or can we move ahead?

ZombineDev:
Let me check again

PR #1888, opened 2017-09-05
tone down spec heading/paragraph numbers a bit

aG0aep6G:
Just a little visual tweak.

Exemplary before/after:
spectacle cs7269

### DIFF
--- a/spec/spec.ddoc
+++ b/spec/spec.ddoc
@@ -30,6 +30,12 @@ $(T style,
         counter-increment: p;
         content: counter(p) ". ";
     }
+    h1::before, h2::before, h3::before, h4::before, p::before
+    {
+        color: #999;
+        font-size: 80%;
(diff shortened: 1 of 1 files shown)


spec/spec.ddoc
```diff
@@ -30,6 +30,12 @@ $(T style,
         counter-increment: p;
         content: counter(p) ". ";
     }
+    h1::before, h2::before, h3::before, h4::before, p::before
+    {
+        color: #999;
+        font-size: 80%;
```
wilzbach: I'm no visual expert or designer, but I'm not so sure whether reducing the font looks good.

ZombineDev:
TBH, I also prefer the current font size.

JinShil:
For whatever it's worth, I actually prefer the change in this PR.

PR #1889, opened 2017-09-09
Checkout stable repositories into .generated

wilzbach:
My main D directory got spammed with:

```
dmd
dmd-2.072.0
dmd-2.073.0
dmd-2.074.0
....
phobos-2.075.0
```

Thus, as dlang.org only needs those repositories, it would be very nice to keep things clean and put every into `$(GENERATED)`
Also currently the `dmd-$(VERSION)` folders conflict with the archives downloaded by the DLang installer.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -152,12 +152,14 @@ DPL_DOCS_PATH=dpl-docs
 DPL_DOCS=$(DPL_DOCS_PATH)/dpl-docs
 REMOTE_DIR=d-programming@digitalmars.com:data
 TMP?=/tmp
+GENERATED=.generated
+G=$(GENERATED)
 
 # Last released versions
-DMD_LATEST_DIR=${DMD_DIR}-${LATEST}
+DMD_LATEST_DIR=$G/${DMD_DIR}-${LATEST}
 DMD_LATEST=$(DMD_LATEST_DIR)/generated/$(OS)/release/$(MODEL)/dmd
-DRUNTIME_LATEST_DIR=${DRUNTIME_DIR}-${LATEST}
-PHOBOS_LATEST_DIR=${PHOBOS_DIR}-${LATEST}
+DRUNTIME_LATEST_DIR=$G/${DRUNTIME_DIR}-${LATEST}
+PHOBOS_LATEST_DIR=$G/${PHOBOS_DIR}-${LATEST}
 
 # Auto-cloning missing directories
 $(shell [ ! -d $(DMD_DIR) ] && git clone --depth=1 ${GIT_HOME}/dmd $(DMD_DIR))
@@ -165,8 +167,6 @@ $(shell [ ! -d $(DRUNTIME_DIR) ] && git clone --depth=1 ${GIT_HOME}/druntime $(D
 
 ################################################################################
 # Automatically generated directories
-GENERATED=.generated
-G=$(GENERATED)
 PHOBOS_DIR_GENERATED=$(GENERATED)/phobos-prerelease
 PHOBOS_LATEST_DIR_GENERATED=$(GENERATED)/phobos-latest
 # The assert_writeln_magic tool transforms all source files from Phobos. Hence
@@ -574,9 +574,11 @@ $G/twid_latest.ddoc:
 # Git rules
 ################################################################################
 
-../%-${LATEST} :
+# Clone snapshots of the latest official release of all main D repositories
+$G/%-${LATEST} :
 	git clone -b v${LATEST} --depth=1 ${GIT_HOME}/$(notdir $*) $@
 
+# Clone all main D repositories
 ${DMD_DIR} ${DRUNTIME_DIR} ${PHOBOS_DIR} ${TOOLS_DIR} ${INSTALLER_DIR}:
 	git clone --depth=1 ${GIT_HOME}/$(notdir $(@F)) $@
 
--- a/posix.mak
+++ b/posix.mak
@@ -847,7 +847,7 @@ test_dspec: dspec_tester.d $(STABLE_DMD)
 
 test: $(ASSERT_WRITELN_BIN)_test test_dspec test/next_version.sh all
 	@echo "Searching for trailing whitespace"
-	@grep -n '[[:blank:]]$$' $$(find . -type f -name "*.dd") ; test $$? -eq 1
+	@grep -n '[[:blank:]]$$' $$(find . -type f -name "*.dd" | grep -v .generated) ; test $$? -eq 1
 	@echo "Searching for undefined macros"
 	@grep -n "UNDEFINED MACRO" $$(find $W -type f -name "*.html" -not -path "$W/phobos/*") ; test $$? -eq 1
 	@echo "Searching for undefined ddoc"


ZombineDev:
I'm very much in favor of moving all the downloading to `.generated/`, the only problem remaining is that we also need to keep the ability to specify existing dmd, druntime and phobos for the case where you want to checkout your personal branch of either of those projects.

wilzbach:
CC @CyberShadow @ZombineDev - this is finally green :tada: 
It was failing because the `test` target which is checking for trailing whitespace in ddoc files correctly found trailing whitespaces in https://github.com/dlang/dmd/blob/v2.076.0/changelog/staticforeach.dd (i.e. in `.generated/dmd-2.076.0`)
As we can't change this fact for now, I opted to ignore `.generated` for this check temporarily. In the future we should probably extend the check to all D repos, s.t. such things don't sneak in...

CyberShadow:
Some things to be aware of:

- `git clean` will skip over repositories, so we need to be careful to not leave stale files inside these repositories. IIRC we ran into this with Digger/DAutoTest where a build was succeeding only because there was a successfully-built file hidden in such a subrepository. With a separate repository it's at least clear that there is some external state not being cleaned up by anything, here `git clean` may only appear to succeed.
- Depending on how things are done on the user side, placing them in `.generated` means they may get cloned more often (pointless work and a break-prone task, depending on network connectivity). Is wiping `.generated` part of the `clean` target?

Overall I'm not entirely convinced this is an improvement, as it looks like trading off tidiness with some potential pitfalls.

> Also currently the dmd-$(VERSION) folders conflict with the archives downloaded by the DLang installer.

How would they conflict? Does the installer unpack things to some parent directory of itself?

wilzbach:
> How would they conflict? Does the installer unpack things to some parent directory of itself?

Yes the installer installs the dmd into `~/dlang/dmd-$(VERSION)` - if you happen to use have `~/dlang/dlang.org` they both conflict: `~/dlang/dmd-2.074.0` could either by the release bundle OR a git clone from dlang/dmd.

> Is wiping .generated part of the clean target?

Yes.

> git clean will skip over repositories, so we need to be careful to not leave stale files inside these repositories.

Well but don't you run the `clean` Makefile in the beginning of DAutoTest?

> Overall I'm not entirely convinced this is an improvement, as it looks like trading off tidiness with some potential pitfalls.

Yes, an alternative would be to have a permanent setting file for the installer, s.t. a custom download path can be specified permanently. If I find time, I might have a look into this, until then I will keep the PR open as a reminder and maybe other arguments pop up ...

CyberShadow:
> if you happen to use have ~/dlang/dlang.org they both conflict

And how would one happen to use have that?

Perhaps the installer should use ~/.cache OSLT...

> Well but don't you run the clean Makefile in the beginning of DAutoTest?

We can do anything we want with DAutoTest, that's not a problem and not what I was talking about. However, it was a surprise for me when we started having .git repositories in DAutoTest, and it might be a surprise to others.

wilzbach:
> Perhaps the installer should use ~/.cache OSLT...

-> https://github.com/dlang/installer/pull/274

> However, it was a surprise for me when we started having .git repositories in DAutoTest, and it might be a surprise to others.

Yes, but we still checkout the git repos, it's just that now we clone them into `.generated`, s.t. it's neatly organized, doesn't clog the `~/dlang` directory and gets removed on `make -f posix.mak clean`.
(I just rebased this)

wilzbach:
Ping @CyberShadow - this is finally passing. Can we move on with this or do you still have concerns?
It would really help to keep my `~/dlang` directory clean and organized. I clean it every now and then, but it still adds up:

```
> ls ~/dlang | grep -E "(dmd|druntime|phobos)"
dmd/
dmd-2.062/
dmd-2.064/
dmd-2.065.0/
dmd-2.066.0/
dmd-2.067.0/
dmd-2.070.2/
dmd-2.071.1/
dmd-2.072.0/
dmd-2.072.2/
dmd-2.073.0/
dmd-2.073.1/
dmd-2.073.2/
dmd-2.074.0/
dmd-2.074.1/
dmd-2.075.0/
dmd-2.075.0-b2/
dmd-2.076.0/
dmd-2.076.1/
dmd-2.077.0/
dmd-2.077.1/
druntime/
druntime-2.073.1/
druntime-2.074.0/
druntime-2.074.1/
druntime-2.075.0/
druntime-2.076.0/
druntime-2.076.1/
druntime-2.077.0/
druntime-2.077.1/
phobos/
phobos-2.074.0/
phobos-2.074.1/
phobos-2.075.0/
phobos-2.076.0/
phobos-2.076.1/
phobos-2.077.0/
phobos-2.077.1/
``

(a couple of the DMD folders are from the installer, which as mentioned is even more annoying because it creates conflicts)

**Comment by CyberShadow** [comment]
Well, I don't like it purely subjectively - I kind of enjoyed the stable version checkouts being where they were, and sometimes made use of them since they were already there. If other D components were to need stable checkouts, they will now not be reusable. I'm also a little sad about going away from avoiding pointless work - all the redundant cloning is just wasted time and network traffic.

So, I'm not sure I can judge this objectively. There is no immediate technical reason not to merge it, but I also don't see an immediate technically-motivated urgency in merging it either.

**Comment by ZombineDev** [comment]
> If other D components were to need stable checkouts, they will now not be reusable.

@CyberShadow You can still reuse them by setting them the `*_LATEST_DIR` makefile variables manually. Is this enough to cover your concerns about this PR?

**Comment by CyberShadow** [comment]
Yes, go ahead and merge it if you agree with the change.

File: posix.mak [hunk 1]
```diff
@@ -152,21 +152,21 @@ DPL_DOCS_PATH=dpl-docs
 DPL_DOCS=$(DPL_DOCS_PATH)/dpl-docs
 REMOTE_DIR=d-programming@digitalmars.com:data
 TMP?=/tmp
+GENERATED=.generated
+G=$(GENERATED)
```
andralex: Shall we just use G and delete GENERATED?

wilzbach: I don't have a strong opinion about this, having it like this allows people to overwrite the more expressive `GENERATED` variable and doesn't cost us too much.
Note that I just moved these two lines up, s.t. the variables are in order (i.e. the are defined before used).

If you or someone else has a stronger opinion about this, let me know and then I will make a PR ;-)

ZombineDev:
@wilzbach the build failed downloading dmd 2.072.1 - you need to rebase this PR to restart it.

wilzbach:
@ZombineDev thanks a lot! I realized that I made a small mistake and the repositories are currently checked out into the root folder -> https://github.com/dlang/dlang.org/pull/2016

MartinNowak:
Fundamentally all of those latest release builds are a mess. We should build /spec, /phobos, and /library for a specific release once (actually with the release), upload it to a server and never touch them again.
Rebuilding the docs from the latest release with every change on master is fragile and modifying already released docs is not wanted.
https://docarchives.dlang.io is a good first step, but not fully embedded in dlang.org which **does** receive daily updates from master.

wilzbach:
> Rebuilding the docs from the latest release with every change on master is fragile

But ensures that our documentation doesn't look too horrible.

> modifying already released docs is not wanted.

Hmm, and I wanted to ask you whether we can build the docs from stable instead of the fixed release, because it often happens that it takes a month for a simple UI fix (typo, Ddoc bug, ...) to reach dlang.org

PR #1891, opened 2017-09-10
Upgrade ddox

wilzbach:
This is WIP and at least gets dpl-docs compiling with ddox 0.16.
I opened this PR to (ab)use DAutoTest a bit - sorry for the noise.

### DIFF
--- a/dpl-docs/views/ddox.inc.module-tree.dt
+++ b/dpl-docs/views/ddox.inc.module-tree.dt
@@ -1,11 +1,5 @@
 - import ddox.entities;
 
-- Module getPackageModule(Package pack)
-	- import std.algorithm : find;
-	- if (auto pp = cast(Package)pack.parent)
-		- return pp.lookup!Module(pack.name);
-	- return null;
-
 - void moduleNavTree(Package pack, Module pack_mod)
 	- string pack_class = "expand-container";
 	- foreach(i, p; pack.packages)
--- a/dpl-docs/views/ddox.layout.dt
+++ b/dpl-docs/views/ddox.layout.dt
@@ -5,6 +5,7 @@ block title
 	block ddox.title
 
 block navigation
+	include ddox.inc.utils
 	include ddox.inc.module-tree
 
 block body
--- a/dpl-docs/views/ddox.overview.dt
+++ b/dpl-docs/views/ddox.overview.dt
@@ -7,6 +7,7 @@ block ddox.title
 	- title = "API documentation";
 
 block ddox.members
+	- import std.conv : to;
 
 	- void moduleInfoRec(Package pack)
 		- foreach( p; pack.packages )
@@ -16,7 +17,7 @@ block ddox.members
 				tr
 					td
 						a(href="#{info.linkTo(m)}")
-							code!= m.qualifiedName.replace(".", "<wbr/>.")
+							code!= m.qualifiedName.to!string.replace(".", "<wbr/>.")
 					td
 						- if( m.docGroup )
 							|!= info.formatDoc(m.docGroup, 0, sec => sec == "$Short")
@@ -27,7 +28,7 @@ block ddox.members
 		- foreach( m; pack.modules )
 			- if (m.qualifiedName.startsWith(prefix))
 				a(href="#{info.linkTo(m)}")
-					code!= m.qualifiedName.replace(".", "<wbr/>.")
+					code!= m.qualifiedName.to!string.replace(".", "<wbr/>.")
 
 	table
 		col.caption
--- a/dpl-docs/views/inc.disqus.dt
+++ b/dpl-docs/views/inc.disqus.dt
@@ -1,8 +1,9 @@
+- import std.conv : text;
 div#disqus_thread
 script(language="javascript").
     var disqus_shortname = 'vibe-d'; // required: replace example with your forum shortname
     var disqus_developer = 1;
-    var disqus_identifier = #{"\\\"phobos-" ~ info.node.qualifiedName ~ "\\\""};
+    var disqus_identifier = #{text(`"phobos-`, info.node.qualifiedName.text, `"`)};
 
 
     /* * * DON'T EDIT BELOW THIS LINE * * */
--- a/dpl-docs/views/layout.dt
+++ b/dpl-docs/views/layout.dt
@@ -7,6 +7,7 @@ html(lang='en-US')
     | All Rights Reserved.
     | https://dlang.org/foundation.html
   - import std.process : environment;
+  - import std.array : replace;
   - string version_id = environment["GIT_TARGET"];
   - bool noExactSourceCodeLinks = environment["NO_EXACT_SOURCE_CODE_LINKS"] == "1";
   - bool haveVersion = version_id.startsWith("v");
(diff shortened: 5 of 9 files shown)


wilzbach:
> This is WIP and at least gets dpl-docs compiling with ddox 0.16.

Finally got even the search working again. Should be good to go now - let's see what DAutoTest has to say.

dpl-docs/views/ddox.inc.module-tree.dt
```diff
@@ -1,11 +1,5 @@
 - import ddox.entities;
 
-- Module getPackageModule(Package pack)
-	- import std.algorithm : find;
-	- if (auto pp = cast(Package)pack.parent)
-		- return pp.lookup!Module(pack.name);
-	- return null;
-
```
ZombineDev: Why remove?

dpl-docs/views/ddox.overview.dt
```diff
@@ -7,16 +7,18 @@ block ddox.title
...
 
 block ddox.members
+	- import std.conv : to;
 
 	- void moduleInfoRec(Package pack)
 		- foreach( p; pack.packages )
 			- moduleInfoRec(p);
 		- foreach( m; pack.modules )
-			- if (!m.qualifiedName.startsWith("std.c.") && !m.qualifiedName.startsWith("core.stdc.") && !m.qualifiedName.startsWith("core.sys."))
+			- auto qualifiedName = m.qualifiedName.to!string;
```
ZombineDev: Why is `.to!string` necessary? `.startsWith` should work with any string.

dpl-docs/views/ddox.inc.module-tree.dt
wilzbach: Because it's defined in [`ddox.inc.utils.dt`](https://github.com/rejectedsoftware/ddox/blob/8ceb3815552765630e9fa0dece63f35837d416ee/views/ddox.inc.utils.dt#L19) which is now in this scope too.

dpl-docs/views/layout.dt
```diff
@@ -51,7 +52,7 @@ html(lang='en-US')
                   span Documentation
                 ul
                   li
-                    a(href="#{root_dir}spec/spec.html")Language Reference
+                    a(href="#{root_dir}spec/spec.html") Language Reference
```
wilzbach: Newer Diet versions enforce a whitespace before the text.

dpl-docs/views/inc.disqus.dt
```diff
@@ -1,8 +1,9 @@
+- import std.conv : text;
 div#disqus_thread
 script(language="javascript").
     var disqus_shortname = 'vibe-d'; // required: replace example with your forum shortname
     var disqus_developer = 1;
-    var disqus_identifier = #{"\\\"phobos-" ~ info.node.qualifiedName ~ "\\\""};
+    var disqus_identifier = #{text(`"phobos-`, info.node.qualifiedName.text, `"`)};
```
wilzbach: `info.node.qualifiedName` was converted to a range (and isn't a string anymore)

dpl-docs/views/ddox.layout.dt
```diff
@@ -5,6 +5,7 @@ block title
 	block ddox.title
 
 block navigation
+	include ddox.inc.utils
```
wilzbach: `heading` has been moved in this auxiliary file: https://github.com/rejectedsoftware/ddox/blob/8ceb3815552765630e9fa0dece63f35837d416ee/views/ddox.inc.utils.dt

wilzbach:
Sadly due to a change in diet-ng, ddox now produces compact HTML, so the entire diff is unusable.
As far as I can see, we have two options:
- manually check a few page & merge
- wait for https://github.com/rejectedsoftware/ddox/pull/166

wilzbach:
> wait for rejectedsoftware/ddox#166

Thanks to @s-ludwig's fast reaction time, the PR got in and has been applied for this PR, s.t. the diff hopefully will be small.
Afterwards in a follow-up PR, I will then switch from pretty-printing to compact in a new separate PR.

wilzbach:
> , s.t. the diff hopefully will be small.

Sadly it's not due to other changes in the rendering of diet-ng
We could do the upgrade in yet another PR or simply move on with this.
@CyberShadow @ZombineDev what do you prefer?

wilzbach:
> @CyberShadow @ZombineDev what do you prefer?

Friendly ping ;-)

CyberShadow:
> Friendly ping ;-)

IIRC the CI was failing at the time of your previous comment.

Still, this does not look right:
```diff
+									<a id="MemoryOrder" class="[&quot;public&quot;]" href="../core/atomic/memory_order.html">MemoryOrder</a>
```

CyberShadow:
There are a lot of whitespace changes, which make it difficult to find the other kinds.

Is it possible to split this into two or three PRs, one of which bumps past just the commits which introduce the whitespace changes? Or, more generally, one PR per ddox change which affects generated content would be even better.

wilzbach:
> Is it possible to split this into two or three PRs, one of which bumps past just the commits which introduce the whitespace changes? Or, more generally, one PR per ddox change which affects generated content would be even better.

Tried to only bump diet-ng or vibe-d -> no changes.
However, when I bump ddox 0.15.18 to 0.16.0, we get the full load again:

https://github.com/dlang/dlang.org/pull/1902

(this PR doesn't include the `compactHTML - false` change).

Any other ideas?
CC @s-ludwig

s-ludwig:
We could of course make a PR that runs the output through some HTML formatter (e.g. "tidy", which is included in the Debian repositories), once for the current version and once for the new one.

> `class="[&quot;public&quot;]"`

Good catch, that is a silent behavior change in diet-ng vs the old implementation. I kind of lean towards keeping the semantics and fixing it in Ddox, tough (https://github.com/rejectedsoftware/ddox/issues/171). I'll try to get to fixing the recently discovered issues ASAP.

wilzbach:
> We could of course make a PR that runs the output through some HTML formatter (e.g. "tidy", which is included in the Debian repositories), once for the current version and once for the new one.

-> https://github.com/dlang/dlang.org/pull/1940 (but it would be a lot easier if we find the problems directly).

> Good catch, that is a silent behavior change in diet-ng vs the old implementation. I kind of lean towards keeping the semantics and fixing it in Ddox, tough (rejectedsoftware/ddox#171).

I submitted both options as PRs: https://github.com/rejectedsoftware/diet-ng/pull/41 and https://github.com/rejectedsoftware/ddox/pull/187.

wilzbach:
>> Good catch, that is a silent behavior change in diet-ng vs the old implementation. I kind of lean towards keeping the semantics and fixing it in Ddox, tough (rejectedsoftware/ddox#171).
> I submitted both options as PRs: rejectedsoftware/diet-ng#41 and rejectedsoftware/ddox#187.

@s-ludwig we are still blocked on the Ddox regression for this update. Which PR do you prefer?

wilzbach:
Okay I took one file in the DDox, applied `tidy` to both versions and then compared it. I observed the following changes:

- the order of attributes changed in `diet-ng` (e.g. before: `<body id="core.atomic" class="std">`, after: `<body class="std" id="core.atomic">`)
- arguments are now displayed (e.g. `<span class="decoration">(here,
ifThis, writeThis)</span>`) - this is a very nice improvement

So imho everything looks good now and I think we are finally ready to upgrade.
@CyberShadow what do you think?

MartinNowak:
Nice, thanks for the effort. Gave it a shot myself, but it was quite tedious.

PR #1893, opened 2017-09-11
Add a simple overview of all contributors

wilzbach:
Add initial list of all contributors. This is just as a start to explore options on how we could utilize this information from the git log.

Depends on https://github.com/dlang/tools/pull/253

### DIFF
--- a/contributors.dd
+++ b/contributors.dd
@@ -1,0 +1,17 @@
+Ddoc
+
+$(D_S $(TITLE),
+
+$(P
+A list of all the awesome people that made D possible.
+)
+
+$(UL
+    $(D_CONTRIBUTORS)
+)
+
+)
+
+Macros:
+    TITLE=Contributors ($(NR_D_CONTRIBUTORS))
+    D_CONTRIBUTOR=$(LI $1)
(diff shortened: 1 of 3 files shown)


wilzbach:
Rebased and reworked, s.t. it can be used with the now merged (https://github.com/dlang/tools/pull/253) contributor query script.
Preview:

image

The number is a bit off as there probably are still duplicates. See also: https://github.com/dlang/tools/pull/270

CyberShadow:
Why the CircleCI failure?

```
fatal: Cannot change to '/home/ubuntu/tools/../installer': No such file or directory
object.Exception@../tools/contributors.d(78): Failed to execute '"git" "-c" "mailmap.file=/home/ubuntu/tools/.mailmap" "-C" "/home/ubuntu/tools/../installer" "log" "--use-mailmap" "--pretty=format:%aN|%aE" "master"'.
```

wilzbach:
> Looks like it's trying to get installer contributions but it's not being cloned.

Yeah seems like I didn't add it as order-only dependency. Done now.

CyberShadow:
Should this be in the navigation? "Community" menu, above "Foundation" maybe?

wilzbach:
> Should this be in the navigation? "Community" menu, above "Foundation" maybe?

Done & resolved the merge conflict.

MartinNowak:
Using shallow clones is a common optimization in build and CI systems, so querying the full git history of branch that might not even be available doesn't work too well.
We could check-in the generated contributors.dd and sort+merge+dedup it with the available history.
Any other ideas?

CyberShadow:
> Any other ideas?

Query the GitHub API or set up a self-hosted service which keeps the full history.

MartinNowak:
Apparently there is already http://contribs.dlang.io/ for phobos docs, though it's blocked by Privacy Badger.

MartinNowak:
Let's try https://github.com/dlang/dlang.org/pull/2015 so I can finally build the beta.

PR #1896, opened 2017-09-19
Add class (de)allocators to 'Deprecated features' list

Calrama:


deprecate.dd
```diff
@@ -10,6 +10,7 @@ $(SPEC_S Deprecated Features,
 
     $(TABLE2 Deprecated Features,
         $(THEAD Feature,                                                          Spec,  Dep,    Error,  Gone)
+        $(TROW $(DEPLINK Class allocators and deallocators),                      ?,      ?,     &nbsp;, &nbsp;)
```
ZombineDev: Similar to `delete` a couple of lines below, this should read `future, &nbsp;, &nbsp;, &nbsp;)` as this feature was neither removed from the spec (just a [note](https://dlang.org/spec/class.html#allocators) saying that it is "deprecated in D2"), nor the compiler emits any warnings if you use them.

Apart from that LGTM, we need to guide the users to better ways of using D.

Calrama:
@ZombineDev Thank you for pointing out the correct line. Didn't know a force push would also remove your review comment (didn't want to add another commit for such a small change, my bad).

Calrama:
> Yeah would be awesome to start with this!

Yes, it would be really nice to start the process of removing those things (amongst others) from D.

ZombineDev:
> Yes, it would be really nice to start the process of removing those things (amongst others) from D.

We can't remove them yet, because we first need to make `object.destroy(someClass)` (which `std.experimental.allocator.dispose(someClass)` uses internally) work without `TypeInfo` and also fix class destructors, so that they respect attribute covariance, similar to class virtual functions. Currently you can inherit a class with `@nogc ~this()` and add a non-`@nogc` destructor.

Currently can use a class with `@nogc` class allocator and deallocator just fine:
```D
import core.stdc.stdlib, core.stdc.stdio;

class C
{   
    @nogc:
    new(size_t s)
    {
        printf("new(size_t s)\n");
        return malloc(s);
    }
    
    delete(void* a)
    {
        printf("delete(void* a)\n");
        free(a);
    }
}

void main(string[] args) @nogc
{
    // The code below compiles only if the class (de)allocator functions are @nogc
    auto c = new C();
    delete c;

    c.destroy(); // doesn't compile - object.destroy!(C).destroy is not @nogc
}
```

So we should wait till we have a good enough replacement.

ZombineDev:
@Calrama Np, `git commit --amend` + `git push -f` is indeed most appropriate in this situation. Just make sure to ping people (like you did) when you push code, since github doesn't send notifications.

PR #1897, opened 2017-09-22
fix issue 17848 - Example of floating point literals in the documenta…

aG0aep6G:
…tion is invalid

'd' is not a valid float suffix. It's not in the grammar, and dmd doesn't accept it. ~~Changing to 'L' suffix.~~

### DIFF
--- a/spec/lex.dd
+++ b/spec/lex.dd
@@ -915,8 +915,6 @@ $(GNAME LeadingDecimal):
 ---------
 1f; // OK
 1.f; // forbidden
-1d; // OK
-1.d; // forbidden
 1.; // OK, int
 ---------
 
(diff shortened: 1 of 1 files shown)


schveiguy:
There are a couple of problems here.

exhibit a: `1L` is actually not a floating point, it's a long.
exhibit b: `1.` says "OK, int". No, it's actually a double.

You may want to make a note that the `L` suffix requires a decimal and subsequent digit to be considered a `real`.

aG0aep6G:
> exhibit a: 1L is actually not a floating point, it's a long.

Right. Instead of trying to fix the example, I'm just going to remove it. It doesn't really add anything over the 'f' example anyway. Amended the commit.

> exhibit b: 1. says "OK, int". No, it's actually a double.

Pre-existing, unrelated to issue 17848.

PR #1899, opened 2017-09-25
Add example for __traits(identifier).

quickfur:
The description of `__traits(identifier)` alone is rather difficult to understand if you don't already know what it means.

### DIFF
--- a/spec/traits.dd
+++ b/spec/traits.dd
@@ -403,6 +403,15 @@ $(H2 $(GNAME identifier))
 	$(P Takes one argument, a symbol. Returns the identifier
 	for that symbol as a string literal.
 	)
+---
+import std.stdio;
+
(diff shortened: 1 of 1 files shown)


spec/traits.dd
```diff
@@ -403,6 +403,16 @@ $(H2 $(GNAME identifier))
 	$(P Takes one argument, a symbol. Returns the identifier
 	for that symbol as a string literal.
 	)
+---
+string fun() { return "abc"; }
+
+pragma(msg, fun); // "abc"
```
MetaLang: I think this line should be removed, or at least explained, because if you're new to D it'll be very confusing as to why it prints "abc" and not "fun".

quickfur:
Cooked up a different, better example instead.

PR #1908, opened 2017-10-09
add release build

MartinNowak:
- replaces patch in build script
- relies on `ifdef RELEASE` to workaround the phobos-latest cloning


posix.mak
```diff
@@ -8,6 +8,12 @@
 #
 # make -f posix.mak rsync
 #
+#
+# This makefile supports 3 different builds latest, prerelease, and release. The
```
CyberShadow: Missing punctuation?

posix.mak
```diff
@@ -8,6 +8,12 @@
...
+# first 2 are live on the website and build from the latest released version or
```
CyberShadow: s/build/built/

posix.mak
```diff
@@ -8,6 +8,12 @@
...
+# master (prerelease). The release build is used for the documentation that is
+# shipped with binary releases. For the latter the LATEST version is not yet
+# published at build time, hence a few things differ from a prerelease build.
```
CyberShadow: How about mentioning how to activate each build? Might be worth mentioning that latest/prerelease are not actually different targets but are just built separately from the default target, and release is activated by setting an option.

CyberShadow:
This is breaking the -prerelease DDoc build: http://dtest.dlang.io/artifact/website-a3969403905ee92ff9918dd70fae1e2f4ef25e9b-9cc12822dc7bea2dd875841e1f1f3a12/web/phobos-prerelease/ddmd_access.html

Looks like it's now using the default DDoc template.

MartinNowak:
> Looks like it's now using the default DDoc template.

Fixed, oversight of an STD_DDOC_PRE in DDOC_HTML_VARS.

CyberShadow:
Thanks!

I believe this needs a corresponding tools PR to update the `d-tags.json` usage there. Should I make one or will you?

MartinNowak:
Here is the tools PR https://github.com/dlang/tools/pull/265, get's rid of another build script patch.

posix.mak
```diff
@@ -265,10 +296,15 @@ ${GENERATED}/${LATEST}.ddoc :
 
 ${GENERATED}/modlist-${LATEST}.ddoc : modlist.d ${STABLE_DMD} $(DRUNTIME_LATEST_DIR) $(PHOBOS_LATEST_DIR) $(DMD_LATEST_DIR)
 	mkdir -p $(dir $@)
-	$(STABLE_RDMD) modlist.d $(DRUNTIME_LATEST_DIR) $(PHOBOS_LATEST_DIR) $(DMD_LATEST_DIR) $(MOD_EXCLUDES_RELEASE) \
+	$(STABLE_RDMD) modlist.d $(DRUNTIME_LATEST_DIR) $(PHOBOS_LATEST_DIR) $(DMD_LATEST_DIR) $(MOD_EXCLUDES_LATEST) \
+		$(addprefix --dump , object std etc core $(DMD_SRC_NAME)) >$@
+
+${GENERATED}/modlist-release.ddoc : modlist.d ${STABLE_DMD} $(DRUNTIME_DIR) $(PHOBOS_DIR) $(DMD_DIR)
```
wilzbach: I guess we might want to rename these to `DRUNTIME_PRERELEASE_DIR` and `DRUNTIME_RELEASE_DIR` too?

posix.mak
```diff
@@ -141,7 +165,7 @@ DDOC_VARS=\
 
 DDOC_VARS_HTML=$(DDOC_VARS) \
```
wilzbach: Let's better rename this to `DDOC_VARS_PRERELEASE_HTML` for consistency

wilzbach:
Removed the empty commit, rebased and force-pushed, s.t. we are sure that this is still passing on DAutoTest.

PR #1916, opened 2017-10-24
Update traits.dd

RazvanN7:
After [1] is merged and **@future** is documented, this addition should be the next step.

[1] https://github.com/dlang/dmd/pull/7232

### DIFF
--- a/spec/traits.dd
+++ b/spec/traits.dd
@@ -22,6 +22,7 @@ $(GNAME TraitsKeyword):
     $(GBLINK isFinalClass)
     $(GBLINK isPOD)
     $(GBLINK isNested)
+    $(GBLINK isFuture)
     $(GBLINK isFloating)
     $(GBLINK isIntegral)
     $(GBLINK isScalar)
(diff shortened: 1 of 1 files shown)


RazvanN7:
@andralex all the other true/false occurrences are highlighted with $D. Anyway, it's not important...

andralex:
Yah, most of the docs predate the introduction of the backticks feature. This gave me the idea of https://github.com/dlang/phobos/pull/5801.

spec/traits.dd
```diff
@@ -192,6 +193,12 @@ $(H2 $(GNAME isNested))
     $(DDSUBLINK spec/struct, nested, structs), and
     $(DDSUBLINK spec/function, variadicnested, functions).)
 
+$(H2 $(GNAME isFuture))
+
+    $(P Takes one argument. It returns `true` if the argument is a symbol
+    marked with the `@future` keyword, otherwise `false`. Currently, only
+    functions and variable declarations have support for the `@future` keyword.)
```
leandro-lucarella-sociomantic: I think it's worth noting that this is an experimental feature. To the point that it's not even `@future` yet, but `@__future` (IIRC). Also, I think we should strongly state that this is not intended for user code yet, and link to the DIP.

I think this feature is having too much visibility and too little clarity for an experimental feature...

PR #1917, opened 2017-10-25
Update grammar.dd to support Elvis operator

RazvanN7:
After (and if) [1] is merged.

[1] https://github.com/dlang/dmd/pull/7242


jacob-carlborg:
I think this [1] should be updated as well.

[1] https://dlang.org/spec/expression.html#conditional_expressions

RazvanN7:
@jacob-carlborg There it is. I made a separate PR since it was easier to just select "Improve this page", then manually make a PR.

andralex:
Closing pending DIP if anyone takes it.

PR #1924, opened 2017-11-02
[2.077] Remove runnable examples blacklist

wilzbach:
With 2.077 all public examples in Phobos are runnable and thus we can
finally remove the blacklist.

As 2.077 is around the corner, this can be merged fairly soon :)

### DIFF
--- a/js/run_examples.js
+++ b/js/run_examples.js
@@ -44,14 +44,6 @@ $(document).ready(function()
     if (!$('body').hasClass("std"))
         return;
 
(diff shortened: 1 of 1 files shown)


wilzbach:
For reference, the final PR to make all public examples in Phobos runnable as standalone was https://github.com/dlang/phobos/pull/5607

wilzbach:
As 2.077.0 has been published three weeks ago, this is more than ready to be merged.
CC @JackStouffer @ZombineDev

PR #1925, opened 2017-11-04
Bump docarchives to 2.077.0

wilzbach:


MetaLang:
Urgh, can we automate this eventually?

wilzbach:
> Urgh, can we automate this eventually?

Well I need to do the snapshot of dlang.org manually, thus I do this manually for now to prevent 404s. Better ideas?

MetaLang:
Not really. Just the fact that this number needs to be manually incremented rubs me the wrong way.

PR #1927, opened 2017-11-04
Mention D_BetterC predefined version

wilzbach:
See also: https://github.com/dlang/dmd/pull/7248


MetaLang:
Unfortunately the rendered result doesn't look like it's correct:

image

wilzbach:
> Unfortunately the rendered result doesn't look like it's correct:

Fixed:

image

PR #1929, opened 2017-11-20
fix issue 17925 - Mention the old body keyword

BBasile:
New comers might be confused by the specs, not matching the D code they can read here and there so a note about the syntax explains the situation.

### DIFF
--- a/spec/contracts.dd
+++ b/spec/contracts.dd
@@ -59,7 +59,7 @@ $(H2 $(LNAME2 pre_post_contracts, Pre and Post Contracts))
 	typical use of this would be in validating the parameters to a function. The post
 	contracts validate the result of the statement. The most typical use of this
 	would be in validating the return value of a function and of any side effects it has.
-	The syntax is:
+	The syntax is:)
 
 ------
 in
@@ -76,12 +76,11 @@ do
 }
 ------
 
-    The actual function body starts with <code>do</code>.
-    In the past, <code>body</code> was used, and could still be encountered in old code bases.
-    In the long term, <code>body</code> may be deprecated, but for now it's still allowed
+    $(P The actual function body starts with $(D do).
+    In the past, $(D body) was used, and could still be encountered in old code bases.
+    In the long term, $(D body) may be deprecated, but for now it's still allowed
     as a keyword in this context and as an identifier in the function body,
-    although <code>do</code> is preferred.
-)
+    although $(D do) is preferred.)
 
 	$(P By definition, if a pre contract fails, then the function received bad
 	parameters.
--- a/spec/contracts.dd
+++ b/spec/contracts.dd
@@ -76,7 +76,7 @@ do
 }
 ------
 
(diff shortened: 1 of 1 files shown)


spec/contracts.dd
```diff
@@ -75,6 +75,14 @@ do
     ...code...
 }
 ------
+
+    The actual function body starts with the <code>do</code> keyword. 
+    In the past, <code>body</code> was used and it will be still encountered in old code bases. 
+    On the long term, <code>body</code> could be completely unsupported but for now it's still allowed,
+    as a keyword in this context and as an identifier in the function body 
+    although <code>do</code> must be prefered.
+)
```
JinShil: Thanks for the PR.  Please consider this wordsmithing:

The actual function body starts with <code>do</code>. 
In the past, <code>body</code> was used, and could still be encountered in old code bases. 
In the long term, <code>body</code> may be deprecated, but for now it's still allowed
as a keyword in this context and as an identifier in the function body, although <code>do</code> is preferred.

JinShil:
Output doesn't look right.  Maybe missing a parentheses or parentheses in the wrong place.

image

BBasile:
Yeah i need to add a paragraph but that's not ideal. It looks like because of the style the paragraph will have its own number.

BBasile:
I think this is good now but actually this is for the stable branch. The problem was also that the change was not in the changelog.

spec/contracts.dd
```diff
@@ -75,6 +75,13 @@ do
     ...code...
 }
 ------
+
+    $(P The actual function body starts with $(D do).
+    In the past, $(D body) was used, and could still be encountered in old code bases.
+    In the long term, $(D body) may be deprecated, but for now it's still allowed
```
wilzbach: Remove 'still'

wilzbach: Why don't you state that it will be deprecated. After all, the regarding DIP has been approved.

BBasile: Yeah, but the problem is not the compiler change, the problem is the documentation that has been brutally changed without a changelog entry, that's the point of the bugzilla report. Even in phobos you still use `body`.

wilzbach:
> Even in phobos you still use body.

Excellent point -> https://github.com/dlang/phobos/pull/5869

ZombineDev:
@wilzbach I'm not sure, but I don't think DScanner has support for the `do` keyword (in place of `body`) yet.

ZombineDev:
> ZombineDev removed the auto-merge label 35 seconds ago

I just want to check the final website output before merging.

BBasile:
yeah, let's wait, i'm not comfy with ddoc, i'm not sure if the changes will be okay.

BBasile:
[I think we're good now ?](http://dtest.dlang.io/artifact/website-28a6fafd74edf65dc5458f5a6644a0edd06d79e6-15bd7c630afee927fea8c339288e18ac/web/spec/contracts.html)

ZombineDev:
> I think we're good now ?

Yes, thanks!

PR #1931, opened 2017-11-21
Update download.dd

cbcoutinho:
Append installation instructions for OpenSUSE Tumbleweed (rolling release distribution of OpenSUSE). I'm not sure if this is applicable for other versions of OpenSUSE. I'm assuming for the stable release (OpenSUSE 42.3) you would need to add a development repository, but I haven't confirmed.

If dmd is not available in the basic repos of stable releases, then users would probably need to add the dlang:devel repo pertaining to their release version:

https://software.opensuse.org/download.html?project=devel%3Alanguages%3AD&package=dlang-tools

### DIFF
--- a/download.dd
+++ b/download.dd
@@ -150,6 +150,8 @@ sudo apt-get update && sudo apt-get -y --allow-unauthenticated install --reinsta
 sudo apt-get update && sudo apt-get install dmd-bin)
 )
 
+$(DOWNLOAD $(OPENSUSE), OpenSUSE Tumbleweed, $(CONSOLE sudo zypper install dmd))
+
 $(H2 Other Downloads)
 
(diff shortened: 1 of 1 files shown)


JackStouffer:
Do we know if the zypper DMD version is maintained? We don't want to link to something outdated.

cbcoutinho:
I'm not sure, I'll leave the definition of 'maintained' up to you. This is the version `dmd` I downloaded using zypper:

```
$ dmd --version
DMD64 D Compiler v2.076.1
Copyright (c) 1999-2017 by Digital Mars written by Walter Bright
```

JackStouffer:
Well, it is a version behind. But someone seems to be maintaining it.

PR #1934, opened 2017-11-27
Clarify WithStatement scope resolution rules

JinShil:
This PR was created in response to comments on https://github.com/dlang/dmd/pull/7356.  However, it is not dependent on that PR.  This PR is not introducing any new rules; it is only clarifying the existing implementation.

https://github.com/dlang/dmd/pull/7356 was submitted to bring `opDispatch` in compliance with these rules and the existing implementation.

### DIFF
--- a/spec/statement.dd
+++ b/spec/statement.dd
@@ -1413,6 +1413,69 @@ void main()
 }
 ---
 
+        $(P In nested $(I WithStatement)s, the inner-most scope takes precedence.  If
+        a symbol cannot be resolved at the inner-most scope, resolution is forwarded
+        incrementally up the scope hierarchy.)
+---
+import std.stdio;
+
+struct Foo
+{
+    void f() { writeln("Foo.f"); }
+}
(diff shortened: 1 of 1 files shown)


MetaLang:
Since @andralex seemed to have reservations on the other PR, I think we should get confirmation on this behaviour.

JinShil:
FYI: @andralex has approved dlang/dmd#7356

ZombineDev:
Looks great, though I have one request: can you reformat the comments, so that the example fits in 80-100 columns? For the most part, examples in the docs are short enough to fit on a mobile screen, though there are many exceptions and we should strive to improve the situation.
For tour.dlang.org we enforce a much stricter limit on the line length - 48.

JinShil:
> can you reformat the comments, so that the example fits in 80-100 columns?

Done

wilzbach:
Fixed a typo in the commit message (Width -> With)

PR #1936, opened 2017-12-02
Fix Issue 17998 - Document Options for install.sh

wilzbach:
While this isn't perfect, it should provide a good basis for continuous improvements ;-)

I wasn't sure whether `~/dlang/install.sh` or `./install.sh` should be the recommended invocation.

CC @MartinNowak @ZombineDev

### DIFF
--- a/install.dd
+++ b/install.dd
@@ -143,7 +143,8 @@ $(CONSOLE
 source ~/dlang/&lt;installed-compiler&gt;/activate
 )
 
-This will setup `PATH`, `LIBRARY_PATH`, `LD_LIBRARY_PATH`, `DMD`, `DC`, and `PS1`.
+This will setup the `PATH`, `LIBRARY_PATH`, `LD_LIBRARY_PATH`, `DMD`, `DC`,
+and `PS1` environment variables.
 It's also possible to combine this into one command:
 
 $(CONSOLE
@@ -174,7 +175,7 @@ source ~/dlang/&lt;installed-compiler&gt;/activate.fish
 
 $(H3 $(LNAME2 uninstall, Uninstall))
 
-Uninstall a D compiler
+Uninstall a D compiler.
 
 $(CONSOLE
 ~/dlang/install.sh uninstall &lt;compiler&gt;
--- a/install.dd
+++ b/install.dd
@@ -57,7 +57,7 @@ $(CONSOLE
 wget https://dlang.org/install.sh
 )
 
-If you want, you can directly invoke it:
+Alternatively, the script can be invoked directly:
 
 $(CONSOLE
 curl https://dlang.org/install.sh | bash -s
@@ -131,7 +131,7 @@ $(EXAMPLES
 
 $(P Once a compiler can be activated for the current session:)
 
-Run the respective `activate` script in your shell to use a compiler by default:
+Run the respective `activate` script in a shell to use a compiler by default:
 
 $(CONSOLE
 source ~/dlang/&lt;installed-compiler&gt;/activate
--- a/install.dd
+++ b/install.dd
@@ -5,14 +5,17 @@ $(D_S $(TITLE),
 
 $(H2 Introduction)
 
-The $(LINK2 https://dlang.org/install.sh, `dlang.org/install.sh`) script is the official D version manager for DMD, GDC and LDC on FreeBSD, Linux and macOS and provides a convenient way to fetch and install a D compiler without requiring administrative privileges nor external dependencies.
+The $(LINK2 https://dlang.org/install.sh, `dlang.org/install.sh`) script is the
+official D version manager for DMD, GDC and LDC on FreeBSD, Linux and macOS.
+It provides a convenient way to fetch and install a D compiler
+without requiring administrative privileges nor external dependencies.
 Its features include:
 
 $(UL
  $(LI Maintaining one or more portable compiler installations in the user's home directory (`~/dlang`))
- $(LI Maintaining the latest stable version of $(LINK2 http://code.dlang.org/docs/commandline, `dub`))
+ $(LI Maintaining the latest stable version of the D package manager, $(LINK2 http://code.dlang.org/docs/commandline, `dub`))
  $(LI Self-updating via `~/dlang/install.sh update`)
- $(LI Supporting nested activation of compilers. Activating a compiler, will setup the `PATH`, `LIBRARY_PATH`, `LD_LIBRARY_PATH`, `DMD`, `DC`, and `PS1` environment variables. After activation, user's can revert those variables to their old values by invoking the `deactivate` command, added to current shell session, by the activation command.)
+ $(LI Supporting nested activation of compilers. Activating a compiler, will setup the `PATH`, `LIBRARY_PATH`, `LD_LIBRARY_PATH`, `DMD`, `DC`, and `PS1` environment variables, and a `deactivate` command, which reverts the modified variables to their old values.)
  $(LI Supporting Fish shell, in addition to Bash-compatible shells.)
  $(LI Verifying binary signatures via the $(LINK2 $(ROOT_DIR)gpg_keys.html,D keyring) (automatic if `gpg` is installed))
  $(LI For `dmd`, installation of the following versions is supported:
@@ -63,9 +66,9 @@ $(CONSOLE
 curl https://dlang.org/install.sh | bash -s
 )
 
-$(P If no argument are provided, the latest DMD compiler will be installed.)
+$(P If no arguments are provided, the latest DMD compiler will be installed.)
 
-$(P The installer will also install a copy of itself to `~/dlang/install.sh`.)
+$(P The installer also makes a copy of itself at `~/dlang/install.sh`.)
 
 $(H2 $(LNAME2 usage, Usage))
 
@@ -116,7 +119,7 @@ $(DL
             Installs the latest beta version of a compiler)
         $(SWITCH $(SWNAME dmd-nightly),
             Installs DMD nightly)
-        $(SWITCH $(SWNAME dmd-20170-02-10),
+        $(SWITCH $(SWNAME dmd-2017-02-10),
             Installs specific DMD nightly)
     )
 )
@@ -131,8 +134,7 @@ $(EXAMPLES
 ~/dlang/install.sh install dmd-nightly
 )
 
-$(P Once a compiler can be activated for the current session:)
-
+Once a compiler is installed, it can be activated for the current session.
 Run the respective `activate` script in a shell to use a compiler by default:
 
 $(CONSOLE
--- a/install.dd
+++ b/install.dd
@@ -127,6 +127,8 @@ $(EXAMPLES
 ~/dlang/install.sh install dmd
 ~/dlang/install.sh install dmd-2.071.1
 ~/dlang/install.sh install ldc-1.1.0-beta2
+~/dlang/install.sh install dmd-beta
+~/dlang/install.sh install dmd-nightly
 )
 
 $(P Once a compiler can be activated for the current session:)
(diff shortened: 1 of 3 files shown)


ZombineDev:
Thanks, this was on my to do list as well ;)

install.dd
```diff
@@ -0,0 +1,163 @@
...
+)
+
+$(P If no argument are provided, the latest DMD compiler will be installed.)
+
+$(P The installer will also install a copy of itself to `~/dlang/install.sh`.)
+
+$(H2 $(LNAME2 usage, Usage))
+
+$(CONSOLE
+./install.sh [&lt;command&gt;] [&lt;args&gt;]
```
ZombineDev: Always use `~/dlang/install.sh`. Avoid using relative paths, especially in documentation.

wilzbach:
> What do you think about an introductory paragraph, along the lines of:

I like it. I converted your text to Ddoc and made a few changes. Please feel free to improve upon them ;-)

install.dd
```diff
@@ -0,0 +1,200 @@
...
+$(H2 $(LNAME2 get, Downloading the installer))
+
+$(CONSOLE
+wget https://dlang.org/install.sh
+)
+
+If you want, you can directly invoke it:
+
+$(CONSOLE
+curl https://dlang.org/install.sh | bash -s
```
andralex: I recall someone (@CyberShadow?) quoted a scathing article about this practice. Perhaps it would be best to not mention it.

install.dd
```diff
@@ -0,0 +1,200 @@
...
+    )
+)
+
+$(H2 $(LNAME2 get, Downloading the installer))
+
+$(CONSOLE
+wget https://dlang.org/install.sh
+)
+
+If you want, you can directly invoke it:
```
andralex: Oh, also please avoid the use of "you" in documentation.

s/If you want, you can directly invoke it/Alternatively, the script can be invoked directly/

install.dd
```diff
@@ -0,0 +1,200 @@
...
+~/dlang/install.sh
+~/dlang/install.sh dmd
+~/dlang/install.sh install dmd
+~/dlang/install.sh install dmd-2.071.1
+~/dlang/install.sh install ldc-1.1.0-beta2
+)
+
+$(P Once a compiler can be activated for the current session:)
+
+Run the respective `activate` script in your shell to use a compiler by default:
```
andralex: s/your shell/the shell/

install.dd
CyberShadow: AFAIK, it's fine if done right:
https://sandstorm.io/news/2015-09-24-is-curl-bash-insecure-pgp-verified-install

wilzbach: Well we do mention it on the downloads page too:

image

wilzbach:
Okay addressed the feedback.
I also switched to use the [recently introduced](https://github.com/dlang/dlang.org/pull/1697) "info on hover" button on the downloads page to link to the script documentation (instead of simply using "Explain").

install.dd
```diff
@@ -0,0 +1,200 @@
...
+Uninstall a D compiler
+
+$(CONSOLE
+~/dlang/install.sh uninstall &lt;compiler&gt;
+)
+
+$(EXAMPLES
+~/dlang/install.sh uninstall dmd
+/install.sh uninstall dmd-2.071.1
+/install.sh uninstall ldc-1.1.0-beta2
```
ZombineDev: Use `~/dlang/install.sh` here too. It looks a bit confusing, otherwise.

install.dd
```diff
@@ -0,0 +1,200 @@
...
+~/dlang/install.sh install dmd-2.071.1
+~/dlang/install.sh install ldc-1.1.0-beta2
+)
+
+$(P Once a compiler can be activated for the current session:)
+
+Run the respective `activate` script in a shell to use a compiler by default:
+
+$(CONSOLE
+source ~/dlang/&lt;installed-compiler&gt;/activate
```
ZombineDev: You should also mention that for `fish` shell the command is:
```
source ~/dlang/&lt;installed-compiler&gt;/activate.fish
```

install.dd
```diff
@@ -0,0 +1,200 @@
...
+            Installs specific DMD nightly)
+    )
+)
+
+$(EXAMPLES
+~/dlang/install.sh
+~/dlang/install.sh dmd
+~/dlang/install.sh install dmd
+~/dlang/install.sh install dmd-2.071.1
+~/dlang/install.sh install ldc-1.1.0-beta2
```
ZombineDev: Just for completeness (and easy copy-pasting) can you mention also:
```
~/dlang/install.sh install dmd-beta
~/dlang/install.sh install dmd-nightly
```

install.dd
```diff
@@ -0,0 +1,200 @@
...
+            Only prints the path to the activate script)
+        $(SWITCH $(SWNAME dmd|ldc|gdc),
+            Installs the latest version of a compiler)
+        $(SWITCH $(SWNAME dmd|ldc|gdc-&lt;version&gt;),
+            Installs a specific version of a compiler (e.g. dmd-2.071.1, ldc-1.1.0-beta2))
+        $(SWITCH $(SWNAME dmd|ldc-beta),
+            Installs the latest beta version of a compiler)
+        $(SWITCH $(SWNAME dmd-nightly),
+            Installs DMD nightly)
+        $(SWITCH $(SWNAME dmd-20170-02-10),
```
joakim-noah: 2017-

install.dd
```diff
@@ -0,0 +1,200 @@
...
+
+$(EXAMPLES
+~/dlang/install.sh
+~/dlang/install.sh dmd
+~/dlang/install.sh install dmd
+~/dlang/install.sh install dmd-2.071.1
+~/dlang/install.sh install ldc-1.1.0-beta2
+)
+
+$(P Once a compiler can be activated for the current session:)
```
joakim-noah: Why put this in its own paragraph?  Merge it with the next sentence instead `session, run the`

install.dd
```diff
@@ -0,0 +1,200 @@
...
+$(H2 Introduction)
+
+The $(LINK2 https://dlang.org/install.sh, `dlang.org/install.sh`) script is the official D version manager for DMD, GDC and LDC on FreeBSD, Linux and macOS and provides a convenient way to fetch and install a D compiler without requiring administrative privileges nor external dependencies.
+Its features include:
+
+$(UL
+ $(LI Maintaining one or more portable compiler installations in the user's home directory (`~/dlang`))
+ $(LI Maintaining the latest stable version of $(LINK2 http://code.dlang.org/docs/commandline, `dub`))
+ $(LI Self-updating via `~/dlang/install.sh update`)
+ $(LI Supporting nested activation of compilers. Activating a compiler, will setup the `PATH`, `LIBRARY_PATH`, `LD_LIBRARY_PATH`, `DMD`, `DC`, and `PS1` environment variables. After activation, user's can revert those variables to their old values by invoking the `deactivate` command, added to current shell session, by the activation command.)
```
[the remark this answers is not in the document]
wilzbach: > Can you activate a compiler inside another one's session?

Yes it's a stack.

install.dd
wilzbach: Ok. done

wilzbach:
@ZombineDev @joakim-noah thanks a lot for your feedback! Addressed.

install.dd
```diff
@@ -0,0 +1,204 @@
...
+        $(LI $(RELATIVE_LINK2 uninstall, Uninstallation))
+        $(LI $(RELATIVE_LINK2 list, Listing available compilers))
+        $(LI $(RELATIVE_LINK2 update, Update))
+    )
+)
+
+$(H2 $(LNAME2 get, Downloading the installer))
+
+$(CONSOLE
+wget https://dlang.org/install.sh
```
wilzbach: How about:

```shell
mkdir -p ~/dlang && wget https://dlang.org/install.sh -O ~/dlang/install.sh
```

install.dd
```diff
@@ -0,0 +1,210 @@
...
+Alternatively, it is also possible to directly use a compiler by calling its
+binary directly.
+Note that the exact path varies between the compiler vendors:
+
+$(CONSOLE
+~/dlang/&lt;installed-dmd-compiler&gt;/bin64/dmd
+~/dlang/&lt;installed-ldc-compiler&gt;/ldc2
+)
+
+Users of the $(LINK2 https://fishshell.com, Fish shell), can `activate.fish`:
```
wilzbach: I put the note about the FISH shell at the end of this section as AFAICT it's not a very popular shell and it stops the reading flow awkwardly if put in-between.

install.dd
```diff
@@ -0,0 +1,210 @@
...
+official D version manager for DMD, GDC and LDC on FreeBSD, Linux and macOS.
+It provides a convenient way to fetch and install a D compiler
+without requiring administrative privileges nor external dependencies.
+Its features include:
+
+$(UL
+ $(LI Maintaining one or more portable compiler installations in the user's home directory (`~/dlang`))
+ $(LI Maintaining the latest stable version of the D package manager, $(LINK2 http://code.dlang.org/docs/commandline, `dub`))
+ $(LI Self-updating via `~/dlang/install.sh update`)
+ $(LI Supporting nested activation of compilers. Activating a compiler, will setup the `PATH`, `LIBRARY_PATH`, `LD_LIBRARY_PATH`, `DMD`, `DC`, and `PS1` environment variables, and a `deactivate` command, which reverts the modified variables to their old values.)
```
joakim-noah: No comma needed after `compiler,` and `variables,`

install.dd
```diff
@@ -0,0 +1,210 @@
...
+ $(LI Supporting Fish shell, in addition to Bash-compatible shells.)
```
joakim-noah: `the Fish`

install.dd
joakim-noah: Yes, better to have a consistent install location, if the curl method installs it differently.

install.dd
joakim-noah: No comma after `shell),`

install.dd
```diff
@@ -0,0 +1,210 @@
...
+
+The activated compiler can be removed from the current session by restoring
+the previous environment:
+
+$(CONSOLE
+deactivate
+)
+
+Alternatively, it is also possible to directly use a compiler by calling its
+binary directly.
```
joakim-noah: Replace end with `possible to call the compiler binary directly`.

install.dd
wilzbach: You are absolutely right about the first occurrence, but isn't the second one the Oxford comma?
I reworded it slightly:

> will setup the `PATH`, `LIBRARY_PATH`, .. `PS1` environment variables,  and a `deactivate` command

JinShil:
This is really good, but we are introducing a maintenance burden between install.sh and this file.  Would it not be better to add this documentation to the `--help` output of install.sh and maybe generate this file as part of the build?

wilzbach:
> but we are introducing a maintenance burden and a fragile synchronization between this file and install.sh. Would it not be better to add this documentation to the --help output of install.sh and maybe generate this file as part of the build?

Of course it would be better, but that's pretty difficult to do all the custom text, explanations and formatting.
I don't think that this file changes much (the installer has about 2-3 pull requests per month and they are mostly just upgrades to the release build system), so that it's worth looking into the effort of generating the page automatically.

OTOH the compiler help page changes quite a lot and we are incredibly bad in keeping it up to date. So if we want to invest time in generating something automatically, I think the `dmd-*.html` pages are a better candidate.

wilzbach:
> maybe linux shouldn't be capitalized

Hmm, I have only seen it written capitalized except for DMD's inconsistent version file.

(Also I am terribly sorry for forgetting to squash this. We really need to re-enable auto-merge-squash.)

MartinNowak:
> Would it not be better to add this documentation to the --help output of install.sh and maybe generate this file as part of the build?

Indeed figuring out how we can base those things on a common man-page like document would be helpful, but not that urgent as it's rarely changing.
https://github.com/fish-shell/fish-shell/blob/f563262cfda72aa291dcca4cd703b2ccba3331a9/share/tools/create_manpage_completions.py
What could be done fairly easily is dumping the outputs of `./install.sh --help, ./install.sh <cmd> --help` verbatim into a code block, but it's less versatile than the docs in *this* PR.

PR #1937, opened 2017-12-02
Fix Issue 17723 - Replace Facebook on the front page with Weka.io

wilzbach:
Before:

image

After:

image


JackStouffer:
To echo the sentiment in the big report, I’d also like Netflix on the front page if possible. If you’re willing to wait on this I’d wait until they reply, but otherwise this can be merged immediately.

wilzbach:
I would suggest to merge this now. As soon as we get a positive reply from Netflix, we can always easily change it, but then again I don't have a strong opinion here and wouldn't mind waiting one or two weeks. Your call ;-)

PR #1938, opened 2017-12-04
Fix Issue 18032 - [Home] Print hex dump example doesn't run

wilzbach:
I also increased the output limit at the DTour server (see e.g. https://github.com/dlang-tour/core-exec/pull/11), but imho it doesn't make sense to dump everything to the user anyhow.
I reduced it to 14 columns because I had issues with scrolling.

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -127,10 +127,10 @@ $(EXTRA_EXAMPLE_RUNNABLE Print hex dump,
 ----
 void main()
 {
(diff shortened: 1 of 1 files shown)


MetaLang:
Why doesn't the former version run?

wilzbach:
It does work, but as until today displays no output  run/tour.dlang.io limited the output to 4096 bytes and only showed an error message if that limit has been exceeded.

PR #1940, opened 2017-12-05
Reformat HTML with Tidy

wilzbach:
Trying to push https://github.com/dlang/dlang.org/pull/1891 forward:

> We could of course make a PR that runs the output through some HTML formatter (e.g. "tidy", which is included in the Debian repositories), once for the current version and once for the new one.

Let's see how good this works - locally I get a lot of warnings and errors with tidy :/

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -235,7 +235,7 @@ ALL_FILES = $(ALL_FILES_BUT_SITEMAP) $(DOC_OUTPUT_DIR)/sitemap.html
 # Rulez
 ################################################################################
 
-all : docs html
+all : docs html tidy
 
 docs-release: dmd-release druntime-release phobos-release apidocs-release
 docs-prerelease: dmd-prerelease druntime-prerelease phobos-prerelease apidocs-prerelease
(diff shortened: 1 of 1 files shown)


CyberShadow:
Guess I'll install tidy on `k3` then.

wilzbach:
> Edit: Oh, this builds it from source. We probably don't want to do that for every website build though.

Yep, I didn't know how responsive you were and also by building from source it should build on other Linux machines without interaction as well.
The idea was just to use Tidy during the ddox-0.15 -> ddox-0.16 transition, s.t. we can be sure that the diff is equal and that there are no hidden changes, but it looks like tidy isn't the best tool to work with the dirty HTML that DDoc and ddox produce:

```
Tidy found 88 warnings and 5163 errors! Not all warnings/errors were shown.

This document has errors that must be fixed before
using HTML Tidy to generate a tidied up version.
```

(see the log for more details)

wilzbach:
> but it looks like tidy isn't the best tool to work with the dirty HTML that DDoc and ddox produce Tidy found 88 warnings and 5163 errors!

I guess this approach isn't going anywhere. Let's focus on fixing the bug in ddox (-> https://github.com/dlang/dlang.org/pull/1891#issuecomment-349155913).

PR #1907, opened 2017-10-09
[WIP] compute next version from ../dmd/VERSION

MartinNowak:
- allows to build changelog preview during beta

PR #1944, opened 2017-12-07
compute next version from ../dmd/VERSION

wilzbach:
Extracted from https://github.com/dlang/dlang.org/pull/1907, s.t. we can move forward bit by bit.

Moving the `next_version` computation to a script seemed like sth. that could easily be a separate PR, s.t. the complexity of #1907 is reduced.

For now I kept things simple and used `changelog/pending.dd` instead of `changelog/${NEXT_VERSION}_pre.dd` (as Martin will do in #1907 too). As this page isn't linked to anywhere we can use any name and not using the version in the filename makes things easier as we now (1) don't depend on `../dmd/VERSION` and thus cloning `dmd` every time and (2) don't need to call the `next_version.sh` on every Make invocation.

### DIFF
--- a/changelog/next_version.sh
+++ b/changelog/next_version.sh
@@ -2,6 +2,8 @@
 
 set -ueo pipefail
 
+# cd and pwd to account for relative paths/symlinks. See also:
+# https://stackoverflow.com/questions/59895/getting-the-source-directory-of-a-bash-script-from-within
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 VERSION=$(cat "${1:-$DIR/../../dmd/VERSION}")
--- a/changelog/next_version.sh
+++ b/changelog/next_version.sh
@@ -1,0 +1,14 @@
+#!/usr/bin/env bash
+
+set -ueo pipefail
+
+VERSION=$(cat "$1")
+# v2.076.1-b1 -> 2.076.1
+VERSION=${VERSION:1:7}
+# 2.076.1 -> (2 076 1)
+PARTS=(${VERSION//./ })
+# use 10#076 prefix to read octal as base10 int
+PARTS[1]=0$((10#${PARTS[1]} + 1))
+PARTS[2]=0
+# 2 077 0 -> 2.077.0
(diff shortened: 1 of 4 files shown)


changelog/next_version.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+set -ueo pipefail
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
andralex: Why not just

```bash
DIR=`dirname ${BASH_SOURCE[0]}`
```

changelog/next_version.sh
```diff
@@ -0,0 +1,16 @@
...
+VERSION=$(cat "${1:-$DIR/../../dmd/VERSION}")
+# v2.076.1-b1 -> 2.076.1
+VERSION=${VERSION:1:7}
+# 2.076.1 -> (2 076 1)
+PARTS=(${VERSION//./ })
+# use 10#076 prefix to read octal as base10 int
+PARTS[1]=0$((10#${PARTS[1]} + 1))
+PARTS[2]=0
+# 2 077 0 -> 2.077.0
+echo "${PARTS[0]}.${PARTS[1]}.${PARTS[2]}"
```
andralex: A check for a correct version would be great here, e.g. make sure VERSION matches a regex.

changelog/next_version.sh
wilzbach: `BASH_SOURCE` might be relative or having relative symlinks. See also https://stackoverflow.com/questions/59895/getting-the-source-directory-of-a-bash-script-from-within for a full discussion.

changelog/next_version.sh
wilzbach: We can even do better - I added a small testsuite.

changelog/next_version.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+set -ueo pipefail
```
andralex: Also use `set -e`?

wilzbach: It's already there: `set -ueo` includes `-e` :)

andralex: oh ok thx!

PR #1946, opened 2017-12-08
"d-apt" deb package name

jordisayol:
Update to the new "dmd-compiler" deb package name.


jordisayol:
On d-apt repository, I can update as soon as a new dmd release appears. On Debian upstream, this is not possible. It takes a long time to be available.

wilzbach:
> On Debian upstream, this is not possible. It takes a long time to be available.

Yes I am aware, but AFAICT there is quite a list of people who prefer to use Debian upstream despite the packages being old, so imho getting DMD into the official Debian upstream would be worthwhile. Anyhow, I was just asking on the status and judging by your reply, there hasn't been progress on this one yet.

BTW thanks a lot for the hard work of maintaining this important distribution channel. Thanks!

PR #1838, opened 2017-07-21
Add macros for replacement with _pending

wilzbach:
See https://github.com/dlang/dlang.org/pull/1821#issuecomment-315603750 and https://github.com/dlang/tools/pull/251#issuecomment-315604776

Step 1) Add new macros
- as suggested no-hardcoded difference is generated by the changelog script
- similarly to `prerelease.ddoc`, a ddoc file is used to overwrite the definitions

Step 2) Replace full text with macros at the `change.d` tool (https://github.com/dlang/tools/pull/252)

Step 3) Add option to parse `BUGSTITLE_TEXT_BODY` and generate a "incremental" changelog

PR #1949, opened 2017-12-09
Show all contributors of the upcoming release

wilzbach:
An initial start at showing this information in the generated changelog.

image
...

This is proof of concept as I would prefer to go without the ugly perl replace hack, i.e. by having clearer templates in the changelog generator (https://github.com/dlang/tools/pull/252).
OTOH this is ready now, works and can be shown on the next changelog. We can always improve it once https://github.com/dlang/tools/pull/252 is in.

This currently only adds the list of contributors to the pending changelog file, but as the changelog generation will be changed soon (e.g. https://github.com/dlang/dlang.org/pull/1907), imho it's better to start small.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -733,12 +733,25 @@ LOOSE_CHANGELOG_FILES:=$(wildcard $(DMD_DIR)/changelog/*.dd) \
 changelog/next-version: ${DMD_DIR}/VERSION
 	$(eval NEXT_VERSION:=$(shell changelog/next_version.sh ${DMD_DIR}/VERSION))
 
-changelog/pending.dd: changelog/next-version | ${STABLE_DMD} ../tools ../installer
-	[ -f changelog/pending.dd ] || $(STABLE_RDMD) $(TOOLS_DIR)/changed.d \
-		$(CHANGELOG_VERSION_LATEST) -o changelog/pending.dd --version "${NEXT_VERSION}" \
-		--date "To be released" --nightly
+changelog/pending.dd: $(TOOLS_DIR)/contributors.d $(TOOLS_DIR)/changed.d $(LOOSE_CHANGELOG_FILES) changelog/next-version | ${STABLE_DMD} $(TOOLS_DIR) $(INSTALLER_DIR)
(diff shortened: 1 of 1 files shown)


wilzbach:
For the script used to query the contributors, see https://github.com/dlang/tools/pull/253

posix.mak
```diff
@@ -733,12 +733,25 @@ LOOSE_CHANGELOG_FILES:=$(wildcard $(DMD_DIR)/changelog/*.dd) \
 changelog/next-version: ${DMD_DIR}/VERSION
 	$(eval NEXT_VERSION:=$(shell changelog/next_version.sh ${DMD_DIR}/VERSION))
 
-changelog/pending.dd: changelog/next-version | ${STABLE_DMD} ../tools ../installer
-	[ -f changelog/pending.dd ] || $(STABLE_RDMD) $(TOOLS_DIR)/changed.d \
-		$(CHANGELOG_VERSION_LATEST) -o changelog/pending.dd --version "${NEXT_VERSION}" \
-		--date "To be released" --nightly
+changelog/pending.dd: $(TOOLS_DIR)/contributors.d $(TOOLS_DIR)/changed.d $(CHANGELOG_FILES) changelog/next-version | ${STABLE_DMD} $(TOOLS_DIR) $(INSTALLER_DIR)
```
CyberShadow:
```
make: *** No rule to make target 'changelog/2.033', needed by 'changelog/pending.dd'.  Stop.
```

`$(CHANGELOG_FILES)` is not suitable as a target/dependency name directly, because it does not contain file extensions (they are being stripped by `basename`).

wilzbach: Oh that error is due to Martin renaming it one day ago (-> https://github.com/dlang/dlang.org/pull/1953/files) and myself forgetting about it in the rebase.

JackStouffer:
Not necessary for this PR but it would be awesome to have a section noting any completely new contributors.

wilzbach:
Btw if you don't want to have the `perl` replacements in there for a short while -> https://github.com/dlang/dlang.org/pull/1838 -> https://github.com/dlang/tools/pull/252

wilzbach:
> Btw if you don't want to have the perl replacements in there for a short while -> #1838 -> dlang/tools#252

-> https://github.com/dlang/dlang.org/pull/1977

PR #1951, opened 2017-12-12
[TRIVIAL] [PDF] Tweak empty lines to avoid listings bug

andralex:
Package listings has a bug whereby sometimes it generates invalid LaTeX code when a listing straddles a page boundary. I could not find a fix for it yet, so from time to time the text needs tweaking. Any helpful info, please lmk.

### DIFF
--- a/spec/attribute.dd
+++ b/spec/attribute.dd
@@ -219,7 +219,6 @@ struct S
     int b;    // placed at offset 4
     long c;   // placed at offset 8
 }
-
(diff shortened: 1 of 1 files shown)


andralex:
I assume pdf is not built by default because someone ran into this problem.

wilzbach:
> I assume pdf is not built by default because someone ran into this problem.

No it's, but there are a few pages where the layout has degraded over time.

PR #1945, opened 2017-12-07
Add druntime options for code coverage

leandro-lucarella-sociomantic:

PR #1952, opened 2017-12-12
Disable cache for CircleCi

wilzbach:
Looks like this is the reason for CircleCI failing at https://github.com/dlang/dlang.org/pull/1945

```
+ git remote add upstream https://github.com/dlang/dlang.org.git
fatal: remote upstream already exists.
```

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -3,8 +3,6 @@ dependencies:
     - ./circleci.sh install-deps
     - ./circleci.sh install-make
     - ./circleci.sh setup-repos
-  cache_directories:
-    - "~/dlang"
 
 test:
   override:
--- a/circleci.sh
+++ b/circleci.sh
@@ -77,6 +77,8 @@ setup_repos()
     # Merge upstream branch with changes, s.t. we check with the latest changes
     if [ -n "${CIRCLE_PR_NUMBER:-}" ]; then
         local current_branch=$(git rev-parse --abbrev-ref HEAD)
+        # work around weird CircleCi bug, see https://github.com/dlang/dlang.org/pull/1952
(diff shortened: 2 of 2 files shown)


circle.yml
```diff
@@ -3,8 +3,6 @@ dependencies:
     - ./circleci.sh install-deps
     - ./circleci.sh install-make
     - ./circleci.sh setup-repos
-  cache_directories:
-    - "~/dlang"
```
JinShil: What was the original purpose of having this cache?

wilzbach: Faster build times - we use the [same for Phobos](https://github.com/dlang/phobos/blob/master/circle.yml) and DRuntime and we didn't had problems with it so far :/

circleci.sh
```diff
@@ -77,6 +77,8 @@ setup_repos()
     # Merge upstream branch with changes, s.t. we check with the latest changes
     if [ -n "${CIRCLE_PR_NUMBER:-}" ]; then
         local current_branch=$(git rev-parse --abbrev-ref HEAD)
+        # work around weird CircleCi bug, see https://github.com/dlang/dlang.org/pull/1952
+        git remote remove upstream || true
```
wilzbach: Well, it seems like this workaround did do his deed :)

circle.yml
JinShil: Then why is it only a problem in this repository?

wilzbach: Hehe, it's actually not an issue here either. The second commit fixes the CI, this was just my initial assumption and I left the commit, because we don't save that much with the cache here anyways (`install-deps` takes 3 seconds) and a cache can always be a potential error source.

PR #1953, opened 2017-12-12
fix chmgen tag generation

MartinNowak:
- sitemap contained invalid links to ../dmd/changelog/PR7356.dd.html
  caused by conflicting CHANGELOG_FILES variable

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -228,7 +228,7 @@ SPEC_ROOT=$(addprefix spec/, \
 	abi simd betterc)
 SPEC_DD=$(addsuffix .dd,$(SPEC_ROOT))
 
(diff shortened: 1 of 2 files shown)


posix.mak
```diff
@@ -228,7 +228,7 @@ SPEC_ROOT=$(addprefix spec/, \
 	abi simd betterc)
 SPEC_DD=$(addsuffix .dd,$(SPEC_ROOT))
 
-CHANGELOG_FILES=$(basename $(subst _pre.dd,.dd,$(wildcard changelog/*.dd)))
+CHANGELOG_FILES:=$(basename $(subst _pre.dd,.dd,$(wildcard changelog/*.dd)))
```
MartinNowak: It's often a good use early binding in make to avoid surprised by later reassignments :o.

PR #1957, opened 2017-12-12
fixup chmgen release build on win32

MartinNowak:


MartinNowak:
> Well if we really want to actively support building dlang.org on Windows, we should add a CI for it. See e.g. dlang/tools#269

No, we've abandoned doc builds on Windows, it's only the chm generation that's left.

PR #1963, opened 2017-12-13
[TRIVIAL] [PDF] Fix table Overloadable Index Unary Operators

andralex:
before: 
a

### DIFF
--- a/latex.ddoc
+++ b/latex.ddoc
@@ -260,6 +260,7 @@ SINGLEQUOTE=`$0'
 SMALL=\small{$0}
 SPEC_S=$(LAYOUT , $1, $(ARGS $+))
 SUB=\textsubscript{$0}
+SUBSCRIPT=\textsubscript{$0}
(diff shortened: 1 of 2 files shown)


andralex:
after: 
b

latex.ddoc
```diff
@@ -260,6 +260,7 @@ SINGLEQUOTE=`$0'
 SMALL=\small{$0}
 SPEC_S=$(LAYOUT , $1, $(ARGS $+))
 SUB=\textsubscript{$0}
+SUBSCRIPT=\textsubscript{$0}
```
wilzbach: This is exactly similar to the existing, unused `SUB` macro. Which one do you prefer? (let's remove the other)

wilzbach:
(found a matching Bugzilla issue again)

latex.ddoc
andralex: SUB is already "subset", so it should be deleted. I'll do this in a moment.

PR #1965, opened 2017-12-13
Update the main page to reflect the new slogan

mdparker:


wilzbach:
@mdparker I rebased your PR to latest, greatest master, s.t. CircleCi isn't failing.

quickfur:
Wow. With a slogan like that, we'd better get our act together on the slow parts of D, like the GC... ;-)

index.dd
```diff
@@ -8,8 +8,9 @@ $(DIVC intro, $(DIV, $(DIV,
     $(DIV,
         $(TC p, pitch, $(B D) is a general-purpose programming language with
         static typing, systems-level access, and C-like syntax.
-        It combines efficiency, control and modeling power with safety
-        and programmer productivity.)
+        With the $(B D Programming Language), write fast, 
+        read fast, and run fast.
```
wilzbach:
```
./index.dd:11:        With the $(B D Programming Language), write fast, 
posix.mak:714: recipe for target 'test' failed
```

wilzbach: I quickly fixed this.

wilzbach:
Before:

image


After:

image

quickfur:
I'm also worried about the omnipresent word ***fast***.  The way it's repeatedly emphasized comes off as somewhat phony, or indicating something of the nature of "get it done fast, get short-term benefits, long-term considerations be damned", which I don't think reflects what D is about.

But hey, if this is the kind of image @WalterBright and @andralex want to convey, what can I say.

JinShil:
I'm sorry I didn't respond in time, but I don't like this slogan at all.  I think it is disingenuous and completely ignores a modern trend in software engineering that I thought was supposed to be a pillar of D: safety.

adamdruppe:
What I like about D is I never (well, ok, rarely) feel like the language is getting in my way. I think about the actual problem instead of how to contort the language to do what I need it to do. I guess that's "write fast" but meh.

andralex:
@JinShil noted. We've taken to the new slogan after tinkering with a few variants for a while. (One problem with the old slogan is it was too long. So long in fact that even Walter and I had difficulty reciting it.) It is understood no slogan will be liked by everyone, but at some point we've got to take some risk. We'll follow feedback and if negatives are mounting we can always change it again.

PR #1969, opened 2017-12-13
Place all generated latex files in .generated

wilzbach:
Store all temporary generated files of PDF and mobi targets in the `.generated` folder.

If done correctly, zero changes should appear on DAutoTest.


wilzbach:
@CyberShadow oh so you seem to have hardcoded the `dlangspec.html` and `dlangspec.tex`:

```d
if (config.build.components.website.diffable)
	make(["all", "verbatim", "pdf", "dlangspec.html"]);
else
	make(["all", "verbatim", "pdf", "kindle"]);
```

```
foreach (item; ["web", "dlangspec.tex", "dlangspec.html"])
		cp(
			buildPath(sourceDir, item),
			buildPath(stageDir , item),
		);
```

https://github.com/CyberShadow/ae/blob/1028e7dea59f97c4b723e19a6b796d7586bf8958/sys/d/manager.d#L1751

I added them back as simple `cp` targets, but wouldn't it be better if they are added as dependencies to `pdf` and the special casing on your side dropped?

CyberShadow:
We can't not have the location of the `.tex` and `.html` file in DAutoTest, because they need to be included in the diffed output.

wilzbach:
> We can't not have the location of the .tex and .html file in DAutoTest, because they need to be included in the diffed output.

Yes, but why can't we do sth. like this:

```make
$W/dlangspec.tex: $G/dlangspec.tex
   cp $< $@

pdf: $W/dlangspec.pdf $W/dlangspec.tex
```

`$W=web`

CyberShadow:
Yep, we could do that.

Alternatively, there can be an option or target that copies the `.tex` / `.html` files to the `web/` directory, e.g. `diffable-intermediaries`. Then DAutoTest won't need to do anything, only detect the presence of the target and disable manual copying.

wilzbach:
> Alternatively, there can be an option or target that copies the .tex / .html files to the web/ directory, e.g. diffable-intermediaries. Then DAutoTest won't need to do anything, only detect the presence of the target and disable manual copying.

Okay I added both, s.t. it's finally passing and once merged, you can remove the `dlangspec.html` target and the respective copy-overs from DAutoTest without breaking anything.

posix.mak
```diff
@@ -277,7 +277,14 @@ verbatim : $(addprefix $(DOC_OUTPUT_DIR)/, $(addsuffix .verbatim,$(PAGES_ROOT)))
 
 kindle : ${DOC_OUTPUT_DIR}/dlangspec.mobi
 
+ifeq (1,$(DIFFABLE))
+# Remove /dlangspec.{tex,html} once DAutoTest has been updated
+# https://github.com/dlang/dlang.org/pull/1969
+pdf : ${DOC_OUTPUT_DIR}/dlangspec.pdf ${DOC_OUTPUT_DIR}/dlangspec.tex ${DOC_OUTPUT_DIR}/dlangspec.html \
+		dlangspec.tex dlangspec.html
```
wilzbach: At the moment it has to build `dlangspec.{tex,html}` by default too as otherwise DAutotest will complain that it can't copy non-existent files.

posix.mak
```diff
@@ -380,54 +385,69 @@ $(DOC_OUTPUT_DIR)/dmd-%.html : %.ddoc dcompiler.dd $(DDOC) $(DMD)
 $(DOC_OUTPUT_DIR)/dmd-%.verbatim : %.ddoc dcompiler.dd verbatim.ddoc $(DMD)
 	$(DMD) -c -o- -Df$@ verbatim.ddoc dcompiler.dd $<
 
+$(DOC_OUTPUT_DIR):
+	mkdir -p $@
```
wilzbach: When I locally ran just `pdf`, the `web` folder was never created and thus the targets failed. Hence I added `web` as order-only dependency to the respective targets.

posix.mak
```diff
@@ -380,54 +385,69 @@ $(DOC_OUTPUT_DIR)/dmd-%.html : %.ddoc dcompiler.dd $(DDOC) $(DMD)
...
-dlangspec.html : $(DDOC) ebook.ddoc dlangspec.d $(DMD)
-	$(DMD) -conf= $(DDOC) ebook.ddoc dlangspec.d
+$G/dlangspec.html : $(DDOC) ebook.ddoc $G/dlangspec.d $(DMD)
+	$(DMD) -conf= -Df$@ $(DDOC) ebook.ddoc $G/dlangspec.d
 
-dlangspec.zip : dlangspec.html ebook.css
+$G/dlangspec.zip : $G/dlangspec.html ebook.css
 	rm -f $@
-	zip $@ dlangspec.html ebook.css
+	zip --junk-paths $@ $G/dlangspec.html ebook.css
```
wilzbach: Without `--junk-paths` the zip would have `.generated/dlangspec.html` and not just `dlangspec.html`

ebook.ddoc
```diff
@@ -56,4 +56,7 @@ NOT_EBOOK=
 RUNNABLE_EXAMPLE=$0
 RUNNABLE_EXAMPLE_STDIN=
 RUNNABLE_EXAMPLE_ARGS=
+SPEC_SUBNAV_NEXT=
+SPEC_SUBNAV_PREV_NEXT=
+SPEC_SUBNAV_PREV=
```
wilzbach: That's why there are a few changes on the diff. By putting `dlangspec.tex` into `web` before running `test`, the undefined macros were detected.

CyberShadow:
@wilzbach This broke auto-deploy:

```
 Output written on .generated/dlangspec.pdf (588 pages, 1423030 bytes).
Transcript written on .generated/dlangspec.log.
mv .generated/dlangspec.pdf /dev/shm/dlang/work/repo/dlang.org/web/dlangspec.pdf
make: *** No rule to make target '.generated/dlangspec.opf', needed by '/dev/shm/dlang/work/repo/dlang.org/web/dlangspec.mobi'.  Stop.
digger: Not caching website dependency build failure.
```

Another missing dependency?

CyberShadow:
> Okay I added both

What's the target name? I don't see one.

posix.mak
CyberShadow: Sorry, this is not at all what I meant.

As I wrote in a previous comment, I think the best way would be a new target which copies the intermediaries to the output directory.

This way is more complicated than necessary, and is causing duplicated files to be copied.

CyberShadow: This is also technically incorrect because `dlangspec.html` does not participate in building the PDF, only the eBook.

CyberShadow: -> https://github.com/dlang/dlang.org/pull/1988

wilzbach: Oh I got errors otherwise because DAutoTest was explicitly trying to copy `dlangspec.{tex,html}`
Anyhow I'm really sorry for breaking DAutoTest and thanks a lot for your quick fix. Really appreciated!

CyberShadow: Sorry if I miscommunicated. What I had in mind was you adding a target and me updating DAutoTest to use it. All's well now anyway.

CyberShadow: > Anyhow I'm really sorry for breaking DAutoTest

Auto-deploy is still broken... didn't look into that yet.

wilzbach: I know why. Fix incoming.

wilzbach: > Auto-deploy is still broken... didn't look into that yet.

-> https://github.com/dlang/dlang.org/pull/1990

PR #1970, opened 2017-12-13
Use $W instead of DOC_OUTPUT_DIR for better visibility

wilzbach:
Inspired by the usage of `$G` and `$D` in the DMD Makefile (we already use `$G` here)

The idea here is that this puts emphasis on the important bits, e.g. compare `$(DOC_OUTPUT_DIR)/sitemap.html` with `$W/sitemap.html`


andralex:
rebase

JinShil:
Would it be too much to ask for something like `$OD` instead?

andralex:
One-letter vars don't need parens, so they're vastly easier on the fingers/eyes.

PR #1975, opened 2017-12-13
Add an extensive Makefile documentation header

wilzbach:
While the Makefile is growing over time, the documentation isn't.

This is an effort to improve the dlang.org's Makefile documentation.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -74,7 +74,6 @@
 #
 #   	DIFFABLE=1			Removes inclusion of all dynamic content and timestamps
 #   	RELEASE=1			Release build (needs to be set for the `release` target)
-#   	LATEST=2.072.1		Specify the LATEST version manually. Avoids network traffic.
 #   	CSS_MINIFY=1		Minify the CSS via an online service
 #   	DOC_OUTPUT_DIR		Folder to build the documentation (default: `web`)
 #
--- a/posix.mak
+++ b/posix.mak
@@ -110,7 +110,7 @@
 #	--------------------------------
 #
 #	You can force this Makefile to use the generated DMD executable in `../dmd`
-#	instead of download a stable DMD compiler:
+#	instead of downloading a stable DMD compiler:
 #
 #		make -f posix.mak VERSION=2.076.0
 #
--- a/posix.mak
+++ b/posix.mak
@@ -87,7 +87,7 @@
 #   	pdf					Generates the D specification as a PDF
 #   	mobi				Generates the D specification as an ebook (Amazon mobi)
 #   	verbatim			Copies the Ddoc plaintext files to .verbatim files (i.e. doesn't run Ddoc on them)
-#   	rsync				Publishes the website (not used anymore)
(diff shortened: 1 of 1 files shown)


posix.mak
```diff
@@ -1,25 +1,129 @@
...
+#
+#    The respective local repositories are used.
+#   	This is very useful for testing local changes.
+#   	Individual targets include:
+#
+#    	dmd-prerelease
+#    	druntime-prerelease
+#    	phobos-prerelease
+#    	apidocs-prerelease		Ddox documentation
+#    	apidocs-serve			Starts a development Ddox webserver
```
CyberShadow: Perhaps clarify which of these are actually dependencies of `docs-prerelease`

posix.mak
```diff
@@ -1,25 +1,129 @@
...
+#   	apidocs-latest			Ddox documentation
+#
+#  Options
+#  -------
+#
+#  Most commonly used options include:
+#
+#   	DIFFABLE=1			Removes inclusion of all dynamic content and timestamps
+#   	RELEASE=1			Release build (needs to be set for the `release` target)
+#   	LATEST=2.072.1		Specify the LATEST version manually. Avoids network traffic.
```
CyberShadow: LATEST no longer (thankfully) requires network traffic to get. It's specified in the VERSION file. I'm not sure if there is any reason to override it on the command line any more.

posix.mak
```diff
@@ -1,25 +1,129 @@
...
+#  -------------
+#
+#   	html				Builds all HTML files and static content
+#   	pending_changelog	Collects and assembles the changelog for the next version
+#   						(This is based on references Bugzilla issues and files in the `/changelog` folders)
+#   	rebase				Rebase all DLang repos to upstream/master
+#   	pdf					Generates the D specification as a PDF
+#   	mobi				Generates the D specification as an ebook (Amazon mobi)
+#   	verbatim			Copies the Ddoc plaintext files to .verbatim files (i.e. doesn't run Ddoc on them)
+#   	rsync				Publishes the website (not used anymore)
```
CyberShadow: I think Martin/Andrei still use rsync when there's a problem with the auto-deploy.

If a target is actually no longer used, it probably ought to be removed.

posix.mak
```diff
@@ -1,25 +1,129 @@
...
+#
+#   For more documentation on Ddox, see https://github.com/rejectedsoftware/ddox
+#   For more information and current blocking points of the Ddoc -> Ddox tranisition,
+#   see https://github.com/dlang/dlang.org/pull/1526
+#
+#	Building without network traffic
+#	--------------------------------
+#
+#	You can force this Makefile to use the generated DMD executable in `../dmd`
+#	instead of download a stable DMD compiler:
```
CyberShadow: "instead of downloading"?

posix.mak
wilzbach: AFAICT they haven't done this in a long time. Your auto-deployment is pretty amazing!

PR #1976, opened 2017-12-13
Fix Issue 16392 - Remove deprecated Windows targets

wilzbach:
Support for Windows has been abandoned for more than a year.
It's dangerous to leave unmaintained scripts in the official repository
as it can easily lead to a frustrating experience.

This leaves the CHM generation as that's the sole part still being
actively used.

See also: https://github.com/dlang/dlang.org/pull/1957#issuecomment-351302277


MetaLang:
Has building the docs on Windows been officially deprecated or is it just the fact that nobody has been maintaining the doc build on Windows?

wilzbach:
> Has building the docs on Windows been officially deprecated

From https://github.com/dlang/dlang.org/pull/1957#issuecomment-351302277

> No, we've abandoned doc builds on Windows, it's only the chm generation that's left.

PR #1977, opened 2017-12-13
Show all contributors of the upcoming release

wilzbach:
An alternative to https://github.com/dlang/dlang.org/pull/1949 using https://github.com/dlang/tools/pull/273


wilzbach:
Thinking about it, it's probably better if we include the CONTRIBUTORS directly in the HTML file, s.t. they can be archived nicely.

wilzbach:
> Thinking about it, it's probably better if we include the CONTRIBUTORS directly in the HTML file, s.t. they can be archived nicely.

-> https://github.com/dlang/tools/pull/274 & https://github.com/dlang/dlang.org/pull/1979

PR #1981, opened 2017-12-14
Show colored error messages for the run.dlang.io backend

wilzbach:
Pretty horrible JS magic as most of the previous output from DPaste isn't supported, but it's not a huge deal as the user is either just interested in the normal output or the error message.

In a follow-up, I might also add support for colored error messages like run.dlang.io has


wilzbach:
> In a follow-up, I might also add support for colored error messages like run.dlang.io has

Turns out it's not that hard with AnsiUp (that's the lib we used for run.dlang.io):

image

Frontpage

image

More colors:

image

The colors are CSS classes and can be tweaked.

JinShil:
That cyan looks pretty bad on white.  Maybe a blue instead.  Doesn't really matter to me, as long as it has more contrast.

CyberShadow:
Do we really need to add another HTTP request for that CSS file?

wilzbach:
> Do we really need to add another HTTP request for that CSS file?

No. Included into `style.css`.
Should I also inline the regarding JS code?

CyberShadow:
Thanks. Keeping things in separate files would be more amenable if they were concatenated during the build, but even so, I'm not sure it's worth it.

CyberShadow:
> Should I also inline the regarding JS code?

Honestly, I think doing the ANSI->HTML translation ought to be done by the server. There is no value in sending raw ANSI over the wire, as far as I can see.

wilzbach:
> There is no value in sending raw ANSI over the wire, as far as I can see.

- Security: sending HTML means that it easily holes for injecting arbitrary HTML
- Simplicity: I didn't know of any D ANSI 2 HTML implementation. There are plenty in JS (main point)

But I do agree that adding the ansi JS increases the size of the sources needed for dlang.org :/

wilzbach:
On a second thought, I'm not planning on adding this HTML output to the backend for the reasons from above (security:holes for injecting arbitrary HTML + simplicity: I didn't know of any D ANSI 2 HTML implementation. There are plenty in JS).
I rebased this to include a uglified version of AnsiUp. It's 10K without compression and and 2.3K with compression. Imho that's more than worth it for nice, colorized error messages without investing a lot of time to reinvent the wheel.

CyberShadow:
I'm not going to block this PR, but I still don't like this.

DMD uses a very tiny subset of ANSI, parsing which should be absolutely trivial by hand. This both puts under the question of including an entire library someone else has written, and the argument that doing it on the server is too hard.

PR #1986, opened 2017-12-14
[TRIVIAL] [PDF] Replace &lt;/&gt; with $(LT)/$(GT)

andralex:


spec/cpp_interface.dd
```diff
@@ -815,7 +815,7 @@ $(CPPLISTING
 // Yes:
 // ptr to const int
 const int* p;
-int* q = const_cast&lt;int*&gt;p; //ok
+int* q = const_cast$(LT)int*$(GT)p; //ok
```
wilzbach: Before:

image


After:

image

PR #1990, opened 2017-12-15
Fix 'kindle' target

wilzbach:
- This target is apparently not run on DAutoTest (?) -> how about defining a target `dautotest`, s.t. we don't have to update DAutoTest manually?
- Uses a trap to avoid residue intermediate files
- Moving this into `$G` seems to be tricky as `dlangspec.opf ` etc. reference the files directly


wilzbach:
@CyberShadow: Did this fix the auto-deployment?

(It still reads "Dec 14" on the bottom of dlang.org)

CyberShadow:
Yes, it looks like it's fixed now.

PR #1983, opened 2017-12-14
[TRIVIAL] [PDF] Fix escaped dollar sign in code

andralex:

PR #1994, opened 2017-12-15
Fix Issue 18081 - dlangspec.pdf: don't escape dollars in code examples

wilzbach:
Before:

image

After:

image

-----

Before:

image

After:

image

----

See also: https://github.com/dlang/dlang.org/pull/1983


JinShil:
Is it just the font, or are those backticks actually single quotes?

wilzbach:
> Is it just the font, or are those backticks actually single quotes? -- must be the font, the code looks right.

Ah yes, we shouldn't execute the macro here nor use the `$(DOLLAR)`.
-> https://github.com/dlang/dlang.org/pull/1996

PR #1996, opened 2017-12-15
[DLANGSPEC] Use a raw dollar in inline code

wilzbach:
Before:

image

After:

image

(different PDF readers were used)

### DIFF
--- a/spec/ddoc.dd
+++ b/spec/ddoc.dd
@@ -496,13 +496,13 @@ $(P
 
 $(P
     Text inside these sections will be escaped according to the rules described above,
-    then wrapped in a `$(DDOC_BACKQUOTED)` macro. By default, this macro expands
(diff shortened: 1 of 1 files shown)


spec/ddoc.dd
```diff
@@ -965,3 +965,4 @@ Macros:
     TITLE=Documentation Generator
     _= Remove the following line when 2.067.0 is used to compile the site.
     BACKTIP=`
```
JinShil: What is a "BACKTIP"?  Should that be "BACKTICK"?

wilzbach:
> What is a "BACKTIP"? Should that be "BACKTICK"?

The line above explains it all:

>  Remove the following line when 2.067.0 is used to compile the site.

-> removed these two lines (there were no occurrences on dlang.org).

PR #1998, opened 2017-12-15
[PoC] Start to test the D Language specification automatically

wilzbach:
This PR adds a simple tool that searches the entire specification for annotated examples and compiles them.
If we know that they compile, we can expose also make them runnable :)
For now I just started with one page.

Supersedes: https://github.com/dlang/dlang.org/pull/1754

As in https://github.com/dlang/dlang.org/pull/1754, we probably want to have similarly to the DMD testsuite these categories

- `SPEC_EXAMPLE_RUNNABLE`
- `SPEC_EXAMPLE_COMPILABLE`
- `SPEC_EXAMPLE_FAIL_COMPILATION`

and these blocks in front:

- `SPEC_EXAMPLE_ARGS`
- `SPEC_EXAMPLE_STDIN`

If this experiment is successful, we can go for more page and additionally also check text examples included e.g. in Phobos or DRuntime.

Future
---------

1) Execute examples

At the examples are only compiled (`-c`) and not executed.
An improvement would be to invoke them and check whether the output matches a pre-stored output.

```
$(SPEC_EXAMPLE_RUNNABLE_STDOUT
a
b
)
$(SPEC_EXAMPLE_RUNNABLE
---
void main() {
   "a\nb".writeln;
}
---
```

Alternative: parse the output from the comment, e.g. `// output: `
This is probably too simplistic.
Sometimes the output is given in a `$(CONSOLE)` block.

2) Check examples which fail compilation

It shouldn't be though hard to check whether only the lines annotated with `// error` have an error.

```
$(SPEC_EXAMPLE_FAIL_COMPILATION
void main() {
   a++; // error: a is undefined
}
)
```

DScanner uses this pattern quite successfully: https://github.com/dlang-community/D-Scanner/blob/master/src/analysis/has_public_example.d#L313

As an alternative one could store the error output above the example in an `SPEC_EXAMPLE_STDERR` macro.


spec/function.dd
```diff
@@ -1923,24 +1986,28 @@ void bar()
...
 
+$(SPEC_RUNNABLE_EXAMPLE
 ------
 int abc(int x) { return x + 1; }
 int def(int y) { return y + 1; }
 
-int function() fp1 = &abc;
-int function() fp2 = &def;
+int delegate(int) fp1 = &abc;
+int delegate(int) fp2 = &def;
```
wilzbach: The first error which this script found in the specs!

CyberShadow:
As with all integrations between components, it's important to keep in mind that there should be a procedure when an interdependency cycle needs to be temporarily broken.

In this case, it would be unwrapping the example out of the `SPEC_RUNNABLE_EXAMPLE` macro. Hopefully it will be obvious enough should the time come.

CyberShadow:
Seems like actually making the examples runnable needs some more work, as they all complain that there is an undefined reference to main currently.

wilzbach:
> Seems like actually making the examples runnable needs some more work, as they all complain that there is an undefined reference to main currently.

Yeah, but it wasn't that much.

js/run.js
```diff
@@ -174,6 +174,24 @@ function parseOutput(res, o, oTitle)
         o.text(output);
 }
 
+// wraps a unittest into a runnable script
+function wrapIntoMain(code) {
```
CyberShadow: A bit annoying that this logic is duplicated in the JS code and in `dspec_tester.d`, but I guess that's not easily avoidable her.

PR #2000, opened 2017-12-16
Fix Issue 18082 - Ubuntu/Debian repository installation should mention dub

wilzbach:
As we are already at the topic:
- there's an [internal discussion](https://github.com/orgs/dlang-community/teams/all/discussions/2) going on about moving the d-apt repository to dlang-community
- it was discussed on the [NG](http://forum.dlang.org/post/mailman.704.1512753538.9493.digitalmars-d-announce@puremagic.com) that `rdmd` will be moved into `dmd-compiler`
- even though DMD is now fully open source since May this year, no one has yet stepped up & submitted the package to Debian upstream

CC @leandro-lucarella-sociomantic @jordisayol

### DIFF
--- a/download.dd
+++ b/download.dd
@@ -183,7 +183,7 @@ $(DOWNLOAD_OTHER $(HOMEBREW), Homebrew, $(CONSOLE brew install dmd))
 $(DOWNLOAD_OTHER $(UBUNTU) $(DEBIAN), Ubuntu/Debian, $(LINK2 http://d-apt.sourceforge.net/, APT repository)
 $(CONSOLE sudo wget http://master.dl.sourceforge.net/project/d-apt/files/d-apt.list -O /etc/apt/sources.list.d/d-apt.list
 sudo apt-get update && sudo apt-get -y --allow-unauthenticated install --reinstall d-apt-keyring
-sudo apt-get update && sudo apt-get install dmd-compiler)
+sudo apt-get update && sudo apt-get install dmd-compiler dub)
 )
 
 $(DOWNLOAD_OTHER $(OPENSUSE), OpenSUSE Tumbleweed, $(CONSOLE sudo zypper install dmd))


leandro-lucarella-sociomantic:
I am by no means a Debian Developer, but I have friends that are, so if anyone wants to take on this I can probably get the support from Debian people.

quickfur:
I'm a DD, but haven't had enough time to keep up with things, so I probably can't do the work myself.  But I can help sponsor packages if someone is willing to iron out the wrinkles.

Keep in mind, though, that there was a discussion with other DDs interested in putting dmd into Debian in the D general newsgroup a while back, and the general consensus was to stick with `ldc` as the official D compiler for Debian packages, and to offer dmd as an alternative, standalone package. See: http://forum.dlang.org/thread/hhefnnighbowonxsnbdy@forum.dlang.org, also the part starting here: http://forum.dlang.org/post/mailman.1524.1491845869.31550.digitalmars-d@puremagic.com

The primary concerns are that dmd only supports a very small subset of platforms supported by Debian, and also that libraries built using dmd isn't compatible with libraries built using other D compilers.  Debian is pretty strong about being cross-platform, and supports quite a large number of platforms that most may consider as marginal, to the point where a new package almost needs to justify why it doesn't work on a given platform.  Also, compatibility across libraries is pretty important, so the fact that dmd ABI isn't compatible with, say, gdc ABI, means dmd cannot be used as the standard D compiler in Debian (for building official Debian packages -- though there is no such restriction for providing dmd as a standalone package that users can install and use to their hearts' content).

Nevertheless, there is great value to be had in having the official reference compiler as part of the Debian archive, as long as it isn't used as the standard D compiler for archive packages. So I'd be happy to sponsor such packages, if someone can step up to working out the details to make it work.

leandro-lucarella-sociomantic:
Thanks for the clarifications!

PR #1040, opened 2015-07-21
Update Core Language specification

9rnsr:
Points:
- Update outdated descriptions.
- Follow the latest compiler behaviors.
- Add inter-page links to associate related sections.

PR #2001, opened 2017-12-16
Revive #1040 - spec/{enum,expression}.d + dcompiler.d

wilzbach:
Part of the revival of #1040.


spec/enum.dd
```diff
@@ -260,6 +260,27 @@ enum size = __traits(classInstanceSize, Foo);  // evaluated at compile-time
...
+}
+
+template Bar(T)
+{
+    // Better, the manifest constant has no runtime location in the executable.
+    enum size_t size = T.sizeof;        // evaluated at compile-time
+
+    // ... use of 'size' at compile time ...
+
+    // If you take Foo's address, it would also go into exe file.
```
andralex: s/Foo's address/the address of `Foo!T.size`/

PR #2003, opened 2017-12-16
Revive #1040 - spec/attributes.d

wilzbach:
Part of the revival of #1040.

### DIFF
--- a/spec/attribute.dd
+++ b/spec/attribute.dd
@@ -37,12 +37,13 @@ $(GNAME Property):
     $(GLINK UserDefinedAttribute)
 
 $(GNAME PropertyIdentifier):
-    $(RELATIVE_LINK2 property, $(D property))
-    $(D safe)
-    $(D trusted)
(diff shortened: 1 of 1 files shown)


spec/attribute.dd
```diff
@@ -722,7 +685,13 @@ $(GNAME UserDefinedAttribute):
...
 )
 
-And looks like:
+    $(P
+        User Defined Attributes (UDA) are compile-time expressions that can be attached
+        to a declaration. These attributes can then be queried, extracted, and manipulated
+        at compile time. There is no runtime component to them.
+    )
+
+A user-defined attribute looks like:
```
wilzbach: @andralex [suggested](https://github.com/dlang/dlang.org/pull/1040/files#r39329490) to use "user-defined attributes" .
What's the official spelling: with dash or without?

andralex: dash please

PR #2005, opened 2017-12-16
Revive #1040 - spec/{interface,template}.d + glossary.d

wilzbach:
Part of the revival of #1040.

### DIFF
--- a/spec/template.dd
+++ b/spec/template.dd
@@ -966,6 +966,18 @@ $(H2 $(LNAME2 function-templates, Function Templates))
         }
         ------
 
+    $(P It is lowered to:)
+
+        ------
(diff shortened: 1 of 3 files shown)


spec/template.dd
```diff
@@ -966,6 +966,18 @@ $(H2 $(LNAME2 function-templates, Function Templates))
         }
         ------
 
+    $(P It is lowered to:)
+
+        ------
+        template $(CODE_HIGHLIGHT Square)(T)
+        {
+            T $(CODE_HIGHLIGHT Square)(T t)
```
wilzbach: Reminder @ self: check the PDF output of this. The same pattern is used in the code block above, though.

PR #2010, opened 2017-12-18
Use changelog/next-version for prerelease doc builds

wilzbach:
AFAICT this isn't even used during the doc build, hence I want to check the diff.

As a next stage there are two things planned:
- display the current version on the top (for the prerelease docs)

image

- use https://github.com/dlang/phobos/blob/v2.077.1/std/meta.d as a source link instead of https://github.com/dlang/phobos/blob/master/std/meta.d (Ddoc already does this correctly).

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -267,7 +267,7 @@ DDOC_VARS_PRERELEASE= \
 	DMD_DIR="$(abspath ${DMD_DIR})" \
 	DRUNTIME_PATH="$(abspath ${DRUNTIME_DIR})" \
 	DOCSRC="$(PWD)" \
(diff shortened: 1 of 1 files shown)


wilzbach:
Great, no changes on the diff :)

> As a next stage there are two things planned:

FYI: I was planning on doing this in a separate PR.

ZombineDev:
The build fails with:
```
make: *** No rule to make target 'changelog/next-version', needed by 'dmd-prerelease'.  Stop.
```

wilzbach:
> The build fails with:

This was due to a different PR being merged. Fixed and thanks a for the ping!

PR #2011, opened 2017-12-18
Cleanup root directory by moving the page files to pages

wilzbach:
Ideally only the "Improve this page" link should change on the diff.


CyberShadow:
This probably needs Andrei's sign-off, and he has historically been against moving things around under the hood...

wilzbach:
> This probably needs Andrei's sign-off, and he has historically been against moving things around under the hood...

Okay. I will refrain from fixing the CIs until then (shouldn't be too difficult).
My main motivation was that the root directory is so messy that e.g. [on GitHub](https://github.com/dlang/dlang.org) one has to scroll six pages (on my laptop) to reach the `README.md`. This is quite scaring and off-putting for new contributors.
(Git handles renames quite well and the history is preserved.)

andralex:
Not a fan of this, and that "pages" is a generic uninformative name is a symptom. I'd go instead for a more logical cleanup, e.g. "ddoc", "articles", "scripts", "samples" etc.

PR #2012, opened 2017-12-18
Move DDoc files to /ddoc

wilzbach:
Similarly to https://github.com/dlang/dlang.org/pull/2011


wilzbach:
(copied motivation from #2011)

> I will refrain from fixing the CIs until then (shouldn't be too difficult).
> My main motivation was that the root directory is so messy that e.g. on GitHub one has to scroll six pages (on my laptop) to reach the README.md. This is quite scaring and off-putting for new contributors.
> (Git handles renames quite well and the history is preserved.)

PR #2014, opened 2017-12-18
Parse dub{-registry} + dconf.org for the contributors list too

wilzbach:
Needs https://github.com/dlang/tools/pull/278

So while I was doing this, I realized that this is maybe not the best idea for a default built as:
- `dconf.org` is 400M big
- these three repositories now would be cloned by default when someone runs the default target or e.g. `html`

Other ideas? Should add a FLAG like `DEPLOYMENT_BUILD` which adds these repos?
CC @CyberShadow


CyberShadow:
I don't think this change is at all necessary.

wilzbach:
Fair enough. I don't like the prospect of cloning the giant dconf.org repo either

PR #2015, opened 2017-12-18
check-in generated contributors list

MartinNowak:
- add update_contributors make target
- avoid failured with shallow clones


wilzbach:
Hmm, this would mean that DAutoTest test has to run `update_contributors` explicitly. Also it won't build the newest list by default on a fresh clone or after a `clean`.

How about making the contributors script skip repositories which don't have a full history? I presume you don't need this page for your local documentation build?

wilzbach:
> avoid failured with shallow clones

Tested this locally (removed `../tools` and `../installer`) and everything ran fine. In fact I believe that DAutoTest also only has shallow clones of `tools` and the `installer` (-> https://github.com/dlang/dlang.org/pull/2018).
What kind of failure do you get?

MartinNowak:
> How about making the contributors script skip repositories which don't have a full history? I presume you don't need this page for your local documentation build?

It's mostly relevant for release and CI builds.
I can update this information when preparing releases ([Orchestration for dlang release building](https://gist.github.com/MartinNowak/a471fe7ddbfeef205cdf04c93a94c6d0)).
Maybe we shouldn't have another tool in our live doc build system for such a rarely changing information. Updating this list from git is simple and stable enough to not warrant too many tests.

> What kind of failure do you get?

There is no master branch :/, and with shallow clones the information is incorrect.

I'll move forward with this for now, so I can finally build the beta. Feel free to come up with a different approach.

MartinNowak:
Here is how other people implement this, but [KISS](https://en.wikipedia.org/wiki/KISS_principle).
- https://contributors.rubyonrails.org
  https://github.com/rails/rails-contributors
- https://thanks.rust-lang.org/ (also https://.rust-lang.org/)
  https://words.steveklabnik.com/announcing-rust-contributors
  https://github.com/steveklabnik/contributors

posix.mak
```diff
@@ -926,12 +927,15 @@ prerelease_changelog: changelog/prerelease.dd html
...
+.PHONY: update_contributors
+update_contributors:
+	rm -f contributors_list.ddoc
+	$(MAKE) -f $(MAKEFILE) contributors_list.ddoc
 
-$G/contributors.ddoc: $G/contributors.dd
-	@echo "NR_D_CONTRIBUTORS=$$(wc -l < $<)" > $@
-	@echo "D_CONTRIBUTORS=" >> $@
-	@cat $< >> $@
+contributors_list.ddoc:  | $(STABLE_RDMD) $(TOOLS_DIR) $(INSTALLER_DIR)
```
wilzbach: I just realized that this generated file should have been put in `$G`

wilzbach:
> Here is how other people implement this, but KISS.

I'm was aware of both projects before (see https://github.com/dlang/tools/pull/253#issue-244820083), but they both run a full-blown web service (heck, they have schema migrations for PSQL) and aren't nicely integrated into the changelog [1], which btw it would be worth a consideration to run the new `prerelease_changelog` target (or modify it to your needs) instead of manually calling the changelog script ...


[1] https://dlang.org/changelog/pending.html

> There is no master branch :/, and with shallow clones the information is incorrect.

There's now: https://github.com/dlang/dlang.org/pull/2018

> Feel free to come up with a different approach.

I would simply disable the page for the release build. I'm pretty sure no one needs this file locally.
Would that be OK with you?

https://github.com/dlang/dlang.org/pull/2023

> Maybe we shouldn't have another tool in our live doc build system for such a rarely changing information. 

It actually changes quite frequently (there have been three new contributors since you checked the file in), hence it makes sense to update it at least nightly or simply with every build.

PR #2017, opened 2017-12-19
Don't build the overall contributors list on DAutoTest

wilzbach:
This avoids unrelated changes appearing on the DAutoTest diff.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -347,7 +347,7 @@ endif
 PAGES_ROOT=$(SPEC_ROOT) 404 acknowledgements areas-of-d-usage \
 	articles ascii-table bugstats builtin \
 	$(CHANGELOG_FILES) code_coverage community comparison concepts \
-	const-faq contributors cppcontracts cpptod ctarguments ctod donate \
+	const-faq cppcontracts cpptod ctarguments ctod donate \
 	D1toD2 d-array-article d-floating-point deprecate dlangupb-scholarship dll-linux dmd \
 	dmd-freebsd dmd-linux dmd-osx dmd-windows documentation download dstyle \
 	exception-safe faq forum-template foundation gpg_keys glossary \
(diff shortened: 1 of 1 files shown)


posix.mak
```diff
@@ -337,7 +337,7 @@ SPEC_ROOT=$(addprefix spec/, \
 SPEC_DD=$(addsuffix .dd,$(SPEC_ROOT))
 
 CHANGELOG_FILES:=$(basename $(subst _pre.dd,.dd,$(wildcard changelog/*.dd)))
-ifndef RELEASE
+ifneq (1,$(RELEASE))
```
wilzbach: This allows a user to do sth. like `RELEASE=$(MY_FANCY_RELEAS_VAR)`.
The other macro ifs also use this style.

CyberShadow:
> Documentation OK (no changes) 

Is this right?

wilzbach:
> Is this right?

Nope. Let's try again ...

Btw (off-topic): we finally bumped up to $(YEAR) with the PR numbers :)

wilzbach:
> Nope. Let's try again ...

Looks better now:

```
Documentation OK (704 deletions)
```

wilzbach:
BTW I see that you hard-coded `sitemap.html` and `dlangspec.pdf` to be ignored in the output diff summary. We could also ignore `sitemap.html` from being generated in the first place, AFAICT [this site](http://dtest.dlang.io/artifact/website-23d243caf947cc8cfc546223c9e186eee9cfe615-9e6c729b82b1bd712390ff4f450098a8/web/sitemap.html) is not very helpful for reviewers anyhow?

CyberShadow:
> Did you ever investigate this further?

Yes, I remember looking into it a while ago, but I don't remember if I followed up on it.

Sorry, I don't have a lot of time to work on D stuff for the remainder of the year.

MartinNowak:
How does this relate to https://github.com/dlang/dlang.org/pull/2015 @wilzbach?

wilzbach:
> How does this relate to #2015 @wilzbach?

`DIFFABLE=1` creates a reproducible `web` folder (no network activity, no timestamps, ...), not sure whether it's useful for you. Apart from that no relation to #2015 except that at some point I would prefer to switch back to automatically generate the contributors file instead of using the stored one...

PR #2019, opened 2017-12-19
Build a reproducible PDF on DAutoTest

wilzbach:
From https://github.com/dlang/dlang.org/pull/2017#issuecomment-352737769

> Maybe it's as simple as setting a few variables: https://tex.stackexchange.com/questions/229605/reproducible-latex-builds-compile-to-a-file-which-always-hashes-to-the-same-va

And it really was that simple and works:

```
> diff -s spec.pdf web/dlangspec.pdf
Files spec.pdf and web/dlangspec.pdf are identical
```

Of course, this needs to be merged first, before it will take effect on DAutoTest.

### DIFF
--- a/latex.ddoc
+++ b/latex.ddoc
@@ -123,6 +123,7 @@ LAYOUT=
 _=
 
 LATEX=$0
+LATEX_NO_DATE=
 LCURL=$\{$
 LEGACY_LNAME2=$(LNAME2 $+)
 LI=\item $0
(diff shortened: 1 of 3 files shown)


CyberShadow:
Thanks!

wilzbach:
Looks like this worked, e.g. on https://github.com/dlang/dlang.org/pull/2021

image

PR #2021, opened 2017-12-20
Fix missing letter

MrSmith33:


JinShil:
Thank you!

wilzbach:
FYI: the changelog is automatically generated by Martin, so the 2.078.0 release won't contain this fix.
Hence, https://github.com/dlang/dmd/pull/7474

PR #2024, opened 2017-12-21
Cleanup ddmd -> dmd transition

wilzbach:
Can be merged safely if no changes appear on the DAutoTest diff.


andralex:
is there a chicken-and-egg problem with the testers?

wilzbach:
> is there a chicken-and-egg problem with the testers?

No I forgot that we have to wait until `stable` is officially released, but that should happen fairly soon :)

wilzbach:
With 2.078.0 being released today, this is unblocked :)

andralex:
awesome!

PR #2027, opened 2017-12-26
Clarify deprecation of `scope` as a type constraint

JinShil:
`scope` for allocating classes on the stack is listed as a deprecated feature.  However, no work to actually deprecate it has ever been done, and now with DIP1000, [it appears it is no longer deprecated](http://forum.dlang.org/post/np1fll$ast$1@digitalmars.com).  In fact it is currently used in DMD's source code.

Here are just a few examples
 - https://github.com/dlang/dmd/blob/9da5c2dcac4f096dc90551fc92a97b810389a9b4/src/dmd/argtypes.d#L480
 - https://github.com/dlang/dmd/blob/9da5c2dcac4f096dc90551fc92a97b810389a9b4/src/dmd/arrayop.d#L245
 - https://github.com/dlang/dmd/blob/9da5c2dcac4f096dc90551fc92a97b810389a9b4/src/dmd/blockexit.d#L542

[It's very confusing](https://forum.dlang.org/post/ddqbairrkurpimbsmtwp@forum.dlang.org), and should either be removed from the deprecations page, or sufficiently clarified.  This PR assumes the former is where the leadership is going with this, but I'm not entirely sure.

@WalterBright and @andralex, if you could chime in on this PR and clarify where you intend to go with `scope` and DIP1000, it would be most welcome.

### DIFF
--- a/deprecate.dd
+++ b/deprecate.dd
@@ -16,7 +16,7 @@ $(SPEC_S Deprecated Features,
         $(TROW $(DEPLINK Implicit string concatenation),                          2.072,  2.072, &nbsp;, &nbsp;)
         $(TROW $(DEPLINK Using the result of a comma expression),                 2.072,  2.072,  2.079, &nbsp;)
         $(TROW $(DEPLINK delete),                                                 &nbsp;,  2.079, &nbsp;, &nbsp;)
-        $(TROW $(DEPLINK scope for allocating classes on the stack),              future, &nbsp;, &nbsp;, &nbsp;)
+        $(TROW $(DEPLINK scope as a type constraint),                             future, &nbsp;, &nbsp;, &nbsp;)
         $(TROW $(DEPLINK Imaginary and complex types),                            future, &nbsp;, &nbsp;, &nbsp;)
         $(TROW $(DEPLINK Implicit catch statement),                               2.072, 2.072,  future, future )
         $(TROW $(DEPLINK .sort and .reverse properties for arrays),               ?,     2.072,  &nbsp;, 2.075)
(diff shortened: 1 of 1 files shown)


JinShil:
Note, also, that we need to clarify what is actually being deprecated or not deprecated.  There are actually two uses of `scope

1.  `scope` as a type modifier

```D
scope class C { }
```

2.  `scope` as a storage class

```D
scope C c = new C()
```

The [deprecated features page](https://dlang.org/deprecate.html#scope%20for%20allocating%20classes%20on%20the%20stack) says...

> Note that `scope` for other usages (e.g. scoped variables) is unrelated to this feature and will not be deprecated.

... but all the examples given on the deprecations page use `scope` to create _scoped variables_.

So, the whole thing is a colossal mess.  I ask whoever is acting steward of the design and roadmap of D to please provide ample clarification.

JinShil:
Ok, @andralex  has removed the "Needs Approval" and asked for clarification in the spec.  Based on his response, I'm going to assume that we are not deprecating either of the following uses of scope.

1.  `scope` as a type modifier

```D
scope class C { }
```

2.  `scope` as a storage class

```D
scope C c = new C()
```

Rather, we just need to clarify that using it without `-dip1000` is a undefined behavior.

I'll update this PR shortly with clarifications in the spec, and then I think this will be ready to go.

andralex:
@JinShil thanks. You may want to add as many clarifying examples as possible.

JinShil:
FYI.  After working through some examples, I discovered some inconsistencies with the interplay of `scope class`, `@safe`, and `-dip1000`, forcing me to have to make some assumptions and do more research about the language authors' intent.  I'm going to need a little more time on this.

DmitryOlshansky:
> Rather, we just need to clarify that using scope in this way without -dip1000 is undefined behavior.

Let's be more careful with such claims. We are not C++ so let's not introduce UB everywhere for little gain.

If anything it's rather that leaking a scope allocated class past it's scope (pardon the pun) is undefined and should be compile-time checked with dip1000.

MartinNowak:
As per https://github.com/dlang/dmd/pull/6826 there are 2 different things to consider here, scoped class references that cannot be escaped with `-dip1000`, and the compiler optimization to allocate a class on the stack if the initialized variable is scope.
For the latter [scoped](https://dlang.org/library/std/typecons/scoped.html) was intended as replacement.
Both aspects, checking that the replacement works without issues and considering how prevalent the usage of stack allocated classes is, should have been considered when deprecating the stack allocated class optimization. Not sure if anything changed that warrants a rediscussion.

> scope as a type modifier

That doesn't really exist, it's a hacky mechanism to enforce stack only allocation. Scope is a storage class and shouldn't modify types.

MartinNowak:
> FYI. After working through some examples, I discovered some inconsistencies with the interplay of scope class, @safe, and -dip1000, forcing me to have to make some assumptions and do more research about the language authors' intent. I'm going to need a little more time on this.

Let me know if you have any specific question.

MartinNowak:
Timing for this is somewhat unfortunate as we're currently working on `@safe` manual memory management, limited lifetimes, which particularly includes ref-counted and uniq classes.
As those heavily rely on the new semantics for `scope` for DIP1000, it seems like an unclever move to strengthen existing usages of `scope` before all the details are worked out.
I tried to came up with sensible semantics in https://github.com/dlang/dmd/pull/6826, but wouldn't want to guarantee anything for DIP1000 yet.

JinShil:
> Let me know if you have any specific question.

@MartinNowak, my only question is about what the plan is.  `scope` is a necessity in the design and implementation of `-dip1000` and `-dip25`.  `scope` as a pseudo-type-modifier is quite simple; it just requires all variables of that class to be `scope`, and I don't see the problem with it.  If you think it is unprecedented, well, much of what D has implemented in the past 10 years has been unprecedented, so much that other languages seem to be copying it.  I thought this matter was settled with @andralex's comments above.

@andralex, Please see @MartinNowak's comments above.  I can't continue work on safeD until we remove the ambiguity about where we're headed.  Please consider @MartinNowak's comments above and clarify, unambiguously, what the plan is with `scope` as a storage class, and `scope` as a pseudo-type-modifier.  Thanks.

JinShil:
> Timing for this is somewhat unfortunate as we're currently working on @safe manual memory management

That's why we need to clarify and document what is being deprecated, what isn't being deprecated, and make sure our implementation matches what is documented (e.g. https://github.com/dlang/dmd/pull/7771).

MartinNowak:
> scope as a pseudo-type-modifier is quite simple; it just requires all variables of that class to be scope

Which will make such classes somewhat incompatible with safe RC/Uniq.
As said earlier, there is hardly a strong use-case to force every user of a class to stack-only allocated instances, and it remains problematic to commit to certain semantics of the `scope` keyword while we're in a major rework of the semantics of exactly that keyword.
I'm currently preparing a DIP and a proposal for dconf on limited lifetimes, but progress depends somewhat on the time I don't have to spent maintaining the wider ecosystem.

> I can't continue work on safeD until we remove the ambiguity about where we're headed.

What are your plans there? We should likely chat/talk if you plan to do some major work in this area.

WalterBright:
My view is:

1. Deprecate `scope class C { }`. It's kind of a bizarre, neglected quirk. It forces all allocations of C to go on the stack. It's hard to see what the point of it is these days.

2. Retain `scope C c = new C();`. The `C` gets allocated on the stack instead of the heap, but pedantically this is an optimization. The -dip1000 checks ensure it works in that `c` is not allowed to escape the lifetime of its stack frame.

JinShil:
Thank you @WalterBright for clarifying.  I've updated this PR to reflect your view.

This PR is ready for additional scrutiny and potentially a merge.

JinShil:
Ping!

deprecate.dd
```diff
@@ -281,54 +281,43 @@ $(H4 Rationale)
...
+    $(P
+        `scope` as a type constraint was a bizarre quirk in the language without a compelling
+        use case.
     )
     $(P
-        Note that $(D scope) for other usages (e.g. scoped variables) is unrelated
-        to this feature and will not be deprecated.
+        Note that this deprecation only affects the usage of `scope` as a type constraint
+        attributed to a class declaration.  `scope` as a storage class attributed to variables,
+        function parameters, etc.. is not deprecated.
```
wilzbach: Double dots here.

andralex:
Thanks @JinShil and sorry for the time taken.

PR #2029, opened 2017-12-27
Remove old items from the .gitignore

wilzbach:


.gitignore
```diff
@@ -19,9 +19,6 @@ Thumbs.db
 images/*.pdf
 
 /web
-Makefile
-css/cssmenu.css
```
wilzbach: Doesn't exist anymore

.gitignore
```diff
@@ -19,9 +19,6 @@ Thumbs.db
...
-*.min.*
```
wilzbach: We currently don't generate any `min` files, but if we ever start to do so again, they should be moved into `.generated` anyways.

PR #1926, opened 2017-11-04
Bump stable DMD and DUB version

wilzbach:
With DUB 1.6.0 (part of the DMD 2.077.0 release), DUB will retry
downloads from the DUB registry and alternatively use fallback mirrors.

PR #2030, opened 2017-12-27
Use STABLE_DMD only for dpl-docs

wilzbach:
Follow-up to https://github.com/dlang/dlang.org/pull/1926.
If we ever need `rdmd`, we can trivially built it from `tools`, which is already available.
Next step would be to replace the `STABLE_DMD` download with only a download of the DUB binary.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -410,23 +410,23 @@ ${GENERATED}/${LATEST}.ddoc :
 	mkdir -p $(dir $@)
 	echo "LATEST=${LATEST}" >$@
 
-${GENERATED}/modlist-${LATEST}.ddoc : modlist.d ${STABLE_DMD} $(DRUNTIME_LATEST_DIR) $(PHOBOS_LATEST_DIR) $(DMD_LATEST_DIR)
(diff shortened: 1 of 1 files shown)


ZombineDev:
```
../dmd/generated/linux/release/64/dmd -run modlist.d ../druntime ../phobos ../dmd  \
	--dump object --dump std --dump etc --dump core --dump "${DMD_SRC_NAME}" >.generated/modlist-release.ddoc
.generated/stable_dmd-2.072.2/dmd2/linux/bin64/dub build --single --compiler=../dmd/generated/linux/release/64/dmd assert_writeln_magic.d
Error: cannot find source code for runtime library file 'object.d'
       dmd might not be correctly installed. Run 'dmd -man' for installation instructions.
       config file: /home/circleci/dmd/generated/linux/release/64/dmd.conf
import path[0] = ../dmd/generated/linux/release/64/../../../../../druntime/import
import path[1] = ../dmd/generated/linux/release/64/../../../../../phobos
posix.mak:421: recipe for target '.generated/modlist-release.ddoc' failed
```

Do we really have druntime and phobos placed on those locations?

PR #2032, opened 2017-12-28
Automatically update the install.sh script

wilzbach:
I have no idea how `install.sh` is updated at the moment. I would guess that
@MartinNowak does this by hand?

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -368,7 +368,7 @@ endif
 TARGETS=$(addsuffix .html,$(PAGES_ROOT))
 
 ALL_FILES_BUT_SITEMAP = $(addprefix $W/, $(TARGETS) \
-$(PREMADE) $(STYLES) $(IMAGES) $(JAVASCRIPT))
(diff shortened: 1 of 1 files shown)


posix.mak
```diff
@@ -506,6 +506,9 @@ $W/dmd-%.html : %.ddoc dcompiler.dd $(DDOC) $(DMD)
 $W/dmd-%.verbatim : %.ddoc dcompiler.dd verbatim.ddoc $(DMD)
 	$(DMD) -c -o- -Df$@ verbatim.ddoc dcompiler.dd $<
 
+$W/install.sh:
+	wget https://raw.githubusercontent.com/dlang/installer/stable/script/install.sh -O web/install.sh
```
wilzbach: I think #2033 would be better. The git clone is cheap anyways and more reliable. This is already done for the changelog script anyways.

ZombineDev: Me too, though obviously we need @MartinNowak to confirm.

wilzbach:
The install script is signed, so we can't do this automatically easily :/

PR #2033, opened 2017-12-28
Automatically update the install.sh script (git clone)

wilzbach:
Alternative to #2032


wilzbach:
The install script is signed, so we can't do this automatically easily :/

PR #2038, opened 2017-12-30

MartinNowak:
Conflicts:
    changelog/2.078.0_pre.dd


MartinNowak:
I'm not entirely certain how that utterly broken makefile ever worked and why it stopped exactly now, but it's fixed now and separating `target` and `doc` for druntime is much cleaner anyhow.

PR #2039, opened 2017-12-31
Remove `objc_glue.d` from the set of excluded files

jacob-carlborg:
After the refactoring of the Objective-C glue layer in the compiler the `objc_glue.d` file is now used by all platforms. See https://github.com/dlang/dmd/pull/7513 for reference.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -205,9 +205,9 @@ MOD_EXCLUDES_PRERELEASE=$(addprefix --ex=, \
 	std.experimental.ndslice.internal std.stdiobase \
 	std.typetuple \
 	tk. msvc_dmc msvc_lib \
-	ddmd.libmach ddmd.libmscoff ddmd.objc_glue \
+	ddmd.libmach ddmd.libmscoff \
 	ddmd.scanmach ddmd.scanmscoff \
-	dmd.libmach dmd.libmscoff dmd.objc_glue \
+	dmd.libmach dmd.libmscoff \
 	dmd.scanmach dmd.scanmscoff)
 
 MOD_EXCLUDES_LATEST=$(MOD_EXCLUDES_PRERELEASE)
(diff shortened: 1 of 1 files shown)


posix.mak
```diff
@@ -205,9 +205,9 @@ MOD_EXCLUDES_PRERELEASE=$(addprefix --ex=, \
 	std.experimental.ndslice.internal std.stdiobase \
 	std.typetuple \
 	tk. msvc_dmc msvc_lib \
-	ddmd.libmach ddmd.libmscoff ddmd.objc_glue \
+	ddmd.libmach ddmd.libmscoff \
 	ddmd.scanmach ddmd.scanmscoff \
-	dmd.libmach dmd.libmscoff dmd.objc_glue \
+	dmd.libmach dmd.libmscoff \
 	dmd.scanmach dmd.scanmscoff)
```
wilzbach: This is just for the documentation menu.

posix.mak
```diff
@@ -728,7 +728,7 @@ endif
 $G/docs-latest.json : ${DMD_LATEST} ${DMD_LATEST_DIR} \
 			${DRUNTIME_LATEST_DIR} ${PHOBOS_LATEST_FILES_GENERATED} | dpl-docs
 	find ${DMD_LATEST_DIR}/src -name '*.d' | \
-		sed -e /mscoff/d -e /objc_glue.d/d ${DMD_EXCLUDE} \
```
wilzbach: I think you have hit a chicken-egg problem here.
I guess you have to do sth. like:

```make
// remove me after https://github.com/dlang/dmd/pull/7513 has been merged
ifneq (,$(wildcard $(DMD_DIR)/src/dmd/objc_glue_stubs.d))
   DMD_EXCLUDE_PRERELEASE += -e /objc_glue.d/d 
endif
```

Note that this needs to be a separate variable for each `.json` file as otherwise this will case problems on the next master-stable branchoff.

jacob-carlborg: Done. I'm not sure if this is correct.

PR #2041, opened 2018-01-01
Fix DDox exclusion for ObjC

wilzbach:
Required for https://github.com/dlang/dmd/pull/7513

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -725,16 +725,14 @@ else
  DMD_EXCLUDE += -e /scanmach/d -e /libmach/d
 endif
 
-# remove this after https://github.com/dlang/dmd/pull/7513 has been merged
-DMD_EXCLUDE_LATEST =
-ifneq (,$(wildcard $(DMD_DIR)/src/dmd/objc_glue_stubs.d))
-   DMD_EXCLUDE_LATEST += -e /objc_glue.d/d
-endif
-
 $G/docs-latest.json : ${DMD_LATEST} ${DMD_LATEST_DIR} \
 			${DRUNTIME_LATEST_DIR} ${PHOBOS_LATEST_FILES_GENERATED} | dpl-docs
+	# remove this after https://github.com/dlang/dmd/pull/7513 has been merged
+	if [ -f $(DMD_LATEST_DIR)/src/*/objc_glue_stubs.d ] ; then \
+	   DMD_EXCLUDE_LATEST="-e /objc_glue.d/d"; \
+	fi; \
 	find ${DMD_LATEST_DIR}/src -name '*.d' | \
-		sed -e /mscoff/d ${DMD_EXCLUDE_LATEST} ${DMD_EXCLUDE} \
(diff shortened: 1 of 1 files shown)


jacob-carlborg:
@wilzbach seems this didn't fix the issue.

wilzbach:
> @wilzbach seems this didn't fix the issue.

So the problem was that we still use `ddmd` for the latest release, hence `src/dmd` is never found for the latest release.

wilzbach:
Oh and we clone `dmd-2.077.1` _after_ we do the "ifeq" check, hence I had to move the checks into the target.

posix.mak
```diff
@@ -725,16 +725,14 @@ else
...
-
 $G/docs-latest.json : ${DMD_LATEST} ${DMD_LATEST_DIR} \
 			${DRUNTIME_LATEST_DIR} ${PHOBOS_LATEST_FILES_GENERATED} | dpl-docs
+	# remove this after https://github.com/dlang/dmd/pull/7513 has been merged
+	if [ -f $(DMD_LATEST_DIR)/src/*/objc_glue_stubs.d ] ; then \
+	   DMD_EXCLUDE_LATEST="-e /objc_glue.d/d"; \
+	fi; \
 	find ${DMD_LATEST_DIR}/src -name '*.d' | \
-		sed -e /mscoff/d ${DMD_EXCLUDE_LATEST} ${DMD_EXCLUDE} \
+		sed -e /mscoff/d $${DMD_EXCLUDE_LATEST} ${DMD_EXCLUDE} \
```
jacob-carlborg: Should there be two `$`?

posix.mak
```diff
@@ -747,19 +745,17 @@ $G/docs-latest.json : ${DMD_LATEST} ${DMD_LATEST_DIR} \
...
-   DMD_EXCLUDE_PRERELEASE += -e /objc_glue.d/d
-endif
-
 # DDox tries to generate the docs for all `.d` files. However for dmd this is tricky,
 # because the `{mach, elf, mscoff}` are platform dependent.
 # Thus the need to exclude these files.
 $G/docs-prerelease.json : ${DMD} ${DMD_DIR} ${DRUNTIME_DIR} \
 		${PHOBOS_FILES_GENERATED} | dpl-docs
+	# remove this after https://github.com/dlang/dmd/pull/7513 has been merged
+	if [ -f $(DMD)/src/*/objc_glue_stubs.d ] ; then \
```
jacob-carlborg: This should be `$(DMD_DIR)` not `$(DMD)`.

posix.mak
wilzbach: Yes it's now a shell variable and `$$` escapes the `$`.

jacob-carlborg: Ok.

jacob-carlborg:
😃.

jacob-carlborg:
Can we get this merged?

wilzbach:
> Note: you can test for file existence in make:

Hehe, I have been there, but this doesn't work for `DMD_LATEST` as the file _never_ exists - it's cloned later. Hence, the need for doing this in bash.

PR #2044, opened 2018-01-02
Convert "Const, Immutable and Shared" to a SECTION2

wilzbach:
See also: http://dtest.dlang.io/artifact/website-e6672ddd5f4efc6e60942fd38836a7bb2abc1361-6511322dd0ebeb894df6eea5bc0109e9/web/spec/class.html

image

From https://github.com/dlang/dlang.org/pull/2043

### DIFF
--- a/spec/class.dd
+++ b/spec/class.dd
@@ -1317,7 +1317,7 @@ $(D new) $(D $(LPAREN))$(I ArgumentList)$(D $(RPAREN)) $(I Identifier) $(D $(LPA
 
 where $(I Identifier) is the name generated for the anonymous nested class.
 
-$(SECTION3 $(LEGACY_LNAME2 ConstClass, const-class, Const, Immutable and Shared Classes),
+$(SECTION2 $(LEGACY_LNAME2 ConstClass, const-class, $(ARGS Const, Immutable and Shared Classes)),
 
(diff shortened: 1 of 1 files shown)


spec/class.dd
```diff
@@ -1317,7 +1317,7 @@ $(D new) $(D $(LPAREN))$(I ArgumentList)$(D $(RPAREN)) $(I Identifier) $(D $(LPA
 
 where $(I Identifier) is the name generated for the anonymous nested class.
 
-$(SECTION3 $(LEGACY_LNAME2 ConstClass, const-class, Const, Immutable and Shared Classes),
+$(SECTION2 $(LEGACY_LNAME2 ConstClass, const-class, Const, Immutable and Shared Classes),
```
andralex: The doc is weird, only shows "Const". I think you need: `$(SECTION2 $(LEGACY_LNAME2 ConstClass, const-class, $(ARGS Const, Immutable and Shared Classes)),`

wilzbach:
> The doc is weird,

Yeah this is still a small issue with the TOC generator, but that's unrelated to this PR.
`SECTION2` works like `SECTION3`:

image

andralex:
@wilzbach I figure but can you please do both improvements here? it's quick and easy.

wilzbach:
> @wilzbach I figure but can you please do both improvements here? it's quick and easy.

Done.

PR #2045, opened 2018-01-02
Show a version chooser for the spec pages

wilzbach:
Quite similar to the way we display the button for DDoc / Ddox.

From https://github.com/dlang/dlang.org/pull/1810#issuecomment-320342675

> So the next order of business is to put the spec numbering in the URLs. I recall @wilzbach has done some work on the phobos side of that (which is somewhat independent).

### DIFF
--- a/js/listanchors.js
+++ b/js/listanchors.js
@@ -30,6 +30,91 @@ function addAnchors()
     }
 }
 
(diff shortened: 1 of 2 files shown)


andralex:
That's big! Thanks!

PR #2046, opened 2018-01-02
Improve dspec_tester.d

wilzbach:
- Use CLI flags instead of environment variables for dspec_tester.d
- Annotate spec/arrays.d
- Introduce three SpecTypes: COMPILE, RUN, FAIL
- Use multiple spec types
- Add CompileConfig to the source code executor
- Remove unused import
- Move findExamples in a separate lambda
- move code block search to a separate function
- Use own state machine instead of a regex

I only annotated to show that the newly introduced macros work.
Once this is merged, the other pages can be annotated too.


wilzbach:
So anything blocking this?
I would love to continue annotating the spec completely and thus making a little testsuite out of it.

For reference, here's the recent PR in which I introduced the spec tester: https://github.com/dlang/dlang.org/pull/1998

dspec_tester.d
```diff
@@ -24,21 +24,83 @@ struct Config {
     string dmdBinPath = "dmd";
     bool printLineNumbers; // whether line numbers should be shown on errors
 }
-Config config;
+__gshared Config config;
```
andralex: Do we need the low-level `__gshared` for a casual script?

dspec_tester.d
```diff
@@ -24,21 +24,83 @@ struct Config {
...
+
+// a range until the next ')', nested () are ignored
+auto untilClosingParentheses(R)(R rs)
+{
+    struct State
+    {
+        size_t rightParensCount = 1;
```
andralex: `uint` (i.e. it's not the size of an object, just a counter)

dspec_tester.d
```diff
@@ -24,21 +24,83 @@ struct Config {
...
+        bool inCodeBlock;
+        size_t dashCount;
```
andralex: ditto

dspec_tester.d
```diff
@@ -24,21 +24,83 @@ struct Config {
...
+    return rs.cumulativeFold!((state, r){
+        if (r == '-')
+        {
+            state.dashCount++;
+        }
+        else
+        {
+            if (state.dashCount >= 3)
+                state.inCodeBlock = !state.inCodeBlock;
+
```
andralex: empty line is surprising

dspec_tester.d
wilzbach: Yeah, we use `parallel` below, so a TLS variable doesn't work (that was the reason for the failures on DAutoTest)
`shared` isn't fun to work with and here I would end up casting whenever accessing `config` :/

jmdavis: Annoying or not, I'd be inclined to argue that using `__gshared` for anything other than binding to a C global is a bug. It's incredibly error-prone to do so, because the compile treats it as thread-local even though it isn't. I know that a lot of folks seem to want to use `__gshared` because they either don't like having to use casts with `shared` or they don't understand how to use it correctly, but IMHO, the fact that `__gshared` gets used as often as it does is a serious problem.

wilzbach:
> Do we need the low-level __gshared for a casual script?

OK removed the `__gshared` in favor of `shared`.

dspec_tester.d
```diff
@@ -24,21 +24,82 @@ struct Config {
     string dmdBinPath = "dmd";
     bool printLineNumbers; // whether line numbers should be shown on errors
 }
-Config config;
+shared Config config;
```
ZombineDev: A cleaner solution would be:
```D
static immutable(Config)* config;
```

That way you can initialize the config and runtime and afterwards treat it as immutable from the parallel section of the code.

PR #2049, opened 2018-01-03
Final class methods are final by default

jmh530:


wilzbach:
> The only comment is that the indentation is unnecessary; the other paragraph in this section that uses the same indent seems to be an exception, since the surrounding sections all have their paragraphs unindented. But this is a minor nitpick, and shouldn't block this PR.

If you extend the diff, the `$(P` before seems to use the same indentation, so I think we are fine here.

PR #2050, opened 2018-01-04
Clarify that pointer assignment is allowed in @safe.

quickfur:
Based on ambiguity [pointed out in the learn forum](http://forum.dlang.org/post/bmriawepbztadaxemvyk@forum.dlang.org).


schveiguy:
Hm... really, this isn't enough. 

```D
void foo(int *p) @safe {}

void bar(int *p) @safe { foo(p + 1); }
```

This doesn't technically modify any pointer values, but still is unsafe. What it should really say is that pointer arithmetic, including indexing, is not allowed.

quickfur:
You're right. Here's a better wording: #2050

PR #2053, opened 2018-01-04
Remove version switcher for Ddox navigation bar + show upcoming version in the Ddoc navbar

wilzbach:


dpl-docs/views/layout.dt
```diff
@@ -147,12 +147,6 @@ html(lang='en-US')
           span.smallprint
             - if( haveVersion )
               | version #{version_name}
-              br
-              a(href='#{root_dir}library-prerelease/index.html') switch to pre-release version
-            - else
-              | pre-release version
-              br
-              a(href='#{root_dir}library/index.html') switch to latest release version
```
wilzbach: Motivation: there's now the version switcher on the top right

PR #2058, opened 2018-01-05
gray download/install info symbol

MartinNowak:
- cause red looks more like an alert

### DIFF
--- a/download.dd
+++ b/download.dd
@@ -276,6 +276,9 @@ Macros:
         .download_paragraph:hover > span {
             visibility: visible;
         }
+        .download_paragraph a > i.fa {
+            color: #999;
+        }
     )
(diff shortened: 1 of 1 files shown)


MartinNowak:
They don't look particularly clickable atm., so it might be more confusing without a hover effect.
Also we don't ship dmd for phones yet ;).

Some redirect or a toast after the download would be another option, though a bit invasive.
They would have the benefit, that we could use the for the short-cut downloads on the front page.

https://codeseven.github.io/toastr/demo.html
https://stackoverflow.com/questions/17723164/show-an-android-style-toast-notification-using-html-css-javascript

But then again the https://dlang.org/dmd-<os>.html pages are somewhat outdated and not that user-friendly to send starters there.

PR #2060, opened 2018-01-08
Add header TOC generation

wilzbach:
Alternative to https://github.com/dlang/dlang.org/pull/2043 with a custom Ddoc wrapper that allows injecting macros on-the-fly.


ddoc.d
```diff
@@ -0,0 +1,180 @@
...
+    auto pos = fileText.representation.countUntil(ddocKey);
+    if (pos < 0)
+        return fileText;
+    const ddocStartLength = ddocKey.representation.until('(', No.openRight).count;
+    auto len = fileText[pos .. $].representation.drop(ddocStartLength).untilClosingParentheses.walkLength;
+    return fileText.replace(fileText[pos .. pos + len + ddocStartLength + 1], newContent);
+}
+
+// a range until the next ')', nested () are ignored
+auto untilClosingParentheses(R)(R rs)
```
ntrel: This reminded me of:
https://dlang.org/phobos-prerelease/std_algorithm_searching.html#.balancedParens

But that only returns a `bool`, perhaps another function could return a range of remaining elements from the last closing bracket, or empty if there isn't one.

PR #2063, opened 2018-01-08
Added Donation Call to Action to the Front Page

JackStouffer:
There needs to be an obvious call to action for donation on both the front page and on the foundation page if we want people to know about it. Currently, the only way people know about donating to the Foundation is if they happen to go to the Foundation's page.

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -665,6 +665,19 @@ consistency. $(LINK2 spec/memory-safe-d.html, Read more).)
 ) $(COMMENT SECTION1)
 ) $(COMMENT D_S)
 
+$(DIVC call_to_donate,
+    <h2>Support D</h2>
+    <p class="center">D is made possible through the hard work and dedication of many volunteers,
+        with the coordination and outreach of the D Language Foundation (501(c)(3)).
+        You can help further the development of the D language and help grow our
+        community by supporting the Foundation.
+    </p>
+    <p class="center">
+        <a href="$(ROOT_DIR)donate.html" class="btn call_to_donate_btn">Donate</a>
+        <a href="$(ROOT_DIR)foundation.html" class="btn subdued">Learn More About The Foundation</a>
+    </p>
+)
+
 Macros:
     TITLE=Home
     TAG=<$1>$+</$1>
@@ -721,6 +734,38 @@ Macros:
             text-align: right;
             padding-right: 0.7em;
             margin-bottom: -0.5em;
+        }
+        .call_to_donate
+        {
+            padding: 40px 140px;
+        }
+        .call_to_donate h2
+        {
+            text-align: center;
+            margin: 0 0 20px 0;
+        }
+        .call_to_donate p
+        {
+            font-size: 14px;
+            max-width: 1000px;
+            margin: 20px auto;
+        }
+        .call_to_donate .btn
+        {
+            font-size: 16px;
+            padding: 0.6em 1em;
+        }
+        a.call_to_donate_btn
+        {
+            background-color: #B03931;
+            border-color: #98312A;
+            color: white;
+        }
+        a.call_to_donate_btn:hover
+        {
+            background-color: #742620;
+            border-color: #98312A;
+            color: white;
         }
     )
     $(COMMENT index.html only for now, until we can figure out
--- a/index.dd
+++ b/index.dd
@@ -698,42 +698,12 @@ Macros:
     TOUR=$(DIVC item, $(SECTION4 $(TC i, fa fa-$1 big-icon)$2, $3))
     WHY_D_ICON=$(TC i, fa fa-$1 why-d-icon)
     _=
-        FRONTPAGE_ORG_IMG=$(DIVC frontpage-orgs-img-wrapper vcontainer-box, <img class="vcontainer-element" src="$(ROOT_DIR)images/orgs-using-d/$0" />)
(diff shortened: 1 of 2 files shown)


JackStouffer:
See https://www.python.org for a good example of this.

wilzbach:
As DAutoTest is down atm - preview:

image

index.dd
```diff
@@ -662,9 +662,19 @@ $(DIVC frontpage-orgs center,
     </div>
 )))
 
+)) $(COMMENT SECTION2 Why D?, DIVC whyd)
 
+$(DIVC call_to_donate,
+    <h1>D Language Foundation</h1>
```
wilzbach: There shall only be one `h1` on a page (otherwise search engines will hate you).
In any case it's a bit too large, so how about `h2`?

wilzbach:
The red background breaks the design a lot. How about?

image

```
$(DIVC section, $(DIV, $(SECTION3 $(WHY_D_ICON life-ring) D Language Foundation,
    $(P D is made possible through the hard work and dedication of many volunteers,
        with the coordination and outreach of the D Language Foundation (501(c)(3)).
        You can help further the development of the D language and help grow our
        community by supporting the Foundation.
    )
    <a href="$(ROOT_DIR)donate.html" class="btn">Donate</a>
    <a href="$(ROOT_DIR)foundation.html" class="btn">Learn more</a>
)))
```

JackStouffer:
I can make it a bit more subdued with a lighter red, but I wanted it to really stand out from the rest of the page and catch people's attention, like the one on the Python home page.

Another good example on the Nim page

screen shot 2018-01-08 at 1 46 37 pm

The call to action for the top of the page is delineated by being the first large button the users sees, and it has a red outline for added measure. The last call to action on the page needs to stand out from the elements above it because at this point on the page, people are probably skimming. 

Here's a slightly different design concept

screen shot 2018-01-08 at 1 42 21 pm

JackStouffer:
Same thing but without the boxshadow and h2
screen shot 2018-01-08 at 2 00 45 pm

JackStouffer:
@wilzbach How about this?

screen shot 2018-01-09 at 2 46 11 pm

JackStouffer:
tbh I'm tempted to make that a part of the footer on every page 😈

wilzbach:
> @wilzbach How about this?

Looks a lot better, but I'm not fully sold yet.
BTW I just talked with @andralex about this today. The gist:

1) We need to ensure that @WalterBright, @acehreli  and @mdparker are on board with this
2) Currently a donor doesn't receive anything as a thank you except for maybe eternal glory . Andrei suggested to set up a shop with franchised items in D branding, e.g. donate 100$ and get a D T-shirts. This isn't a requirement for this banner PR though. I remember that Sociomantic wanted to help setting up such a store (CC @dylan-cromwell-sociomantic).
3) We should start to offer membership packages, but I'm not sure whether there's anyone with free time. @acehreli maybe?

JackStouffer:
We should continue the membership package/store discussion in slack or the forums.

>We need to ensure that @WalterBright, @acehreli and @mdparker are on board with this

> Also CCing @CyberShadow as this is a big change to our front site.

Not that I'm complaining, but it seems like somethings off when this change needs to be signed off by five different people, two of which are the head honchos of the language. I thought we were trying to get Andrei and Walter less involved in monotony.

CyberShadow:
> Also CCing @CyberShadow as this is a big change to our landing page.

Heh, nobody asked me about the change that put the word "Fast" 12 times on the front page (talk about semantic saturation, sheesh).

This is foundation stuff which I don't participate in, so fully deferring to Walter/Andrei.

andralex:
It's a good initiative, let me discuss it a bit with the other officers. Thanks!

quickfur:
@CyberShadow Yeah, to be frank I'm no fan of the current "fast" slogan.  But hey, I have no say in the matter, and the world continues to spin. *shrug*

andralex:
We're in agreement we can experiment with a Donate button. Please let's get this going. Thanks!

CyberShadow:
In the current version, it's not obvious (to me at least) that "Learn More About The Foundation" is a link. Perhaps give it a border, like the "Other downloads" button?

JackStouffer:
@CyberShadow Done

screen shot 2018-01-11 at 9 15 21 am

wilzbach:
Any reason why you didn't restrict the block to a maximal width?

image

I think it would look better if it would be in the `.more` div and thus stick to `76em`:

JackStouffer:
I'll restrict the `max-width` of the `p` but I kept it out of `more` to have it centered on the entire page.

JackStouffer:
Fixed

wilzbach:
Thanks! Looks great now:

image

JackStouffer:
sure

wilzbach:
@CyberShadow are you okay with this change too? I think everyone else is ;-)

CyberShadow:
I already answered this above.

> This is foundation stuff which I don't participate in, so fully deferring to Walter/Andrei.

wilzbach:
> I already answered this above.

Oh I was talking about the "implementation" (e.g. layout and design), because you made a comment about it before.

> This is foundation stuff which I don't participate in, so fully deferring to Walter/Andrei.

We're fine on this side, quoting an internal mail:

> We're all in agreement we can try this. Will pull the PR. Thanks! -- Andrei

andralex:
Yes we're on!

PR #2064, opened 2018-01-08
Added call to action for donations to foundation page

JackStouffer:


css/style.css
```diff
@@ -627,6 +627,20 @@ a.btn:hover
     color: #333;
 }
 
+.btn.donate-large
+{
+    font-size: 22px;
+    padding: 0.4em 1em;
+    background-color: #D8645E;
```
wilzbach: Why not `#B03931`?

Yours:

image

---

image

css/style.css
```diff
@@ -627,6 +627,20 @@ a.btn:hover
...
+    color: white;
+    border: #D8645E;
+}
+.btn.donate-large:hover
+{
+    background-color: #F47069;
```
wilzbach: Why not `#98312A` (similar to the menu)?

JackStouffer:
No particular reason. I saw it in the CSS and thought it looked better. I can change it if you want.

wilzbach:
> No particular reason. I saw it in the CSS and thought it looked better. I can change it if you want.

Changed it myself as a separate commit (so it's easy to revert).
The CSS header says:

```
Reds:

#B03931 - our main red
#742620 - focus/hover

#D8645E - on dark background
#F47069 - focus/hover on dark background
```

So I would stick to the "main red".

PR #2065, opened 2018-01-08
Convert tabs to whitespace + add a test for it

wilzbach:
~~~`sed $'s/\t/    /' -i **/*.dd`~~~


CyberShadow:
> sed $'s/\t/ /' -i **/*.dd

Not sure if it makes a difference here, but that's not strictly correct. There is a `detab` program in tools, maybe use that?

wilzbach:
> There is a detab program in tools, maybe use that?

Good idea. Done. Ouput looks similar though.

PR #2068, opened 2018-01-09
Generate CLI pages automatically

wilzbach:
TODO:

- [x] translate `TargetOS` to Ddoc macros
- [x] https://github.com/dlang/dmd/pull/7703


wilzbach:
Rebased again. How perfect does this initial version need to be?

CyberShadow:
The HTML is looking pretty broken. In the new version, e.g. on the FreeBSD page, the "Requirements and Downloads" section, has a numbered list that's cut off and broken by the "Files" section.

wilzbach:
Found the problem - it was a missing brace in the first commit.

CyberShadow:
HTML output is still broken (on the other pages now):

http://dtest.dlang.io/artifact/website-ed4a19f9027a9bd9639b312e16b1ddb1c5612713-7b211ecb7c48d907acb12b71f596eb49/web/regular-expression.html

CyberShadow:
Can we catch such breakages in the test phase?

wilzbach:
> HTML output is still broken (on the other pages now):

Fixed.

> Can we catch such breakages in the test phase?

Sure, two steps forward:

https://github.com/dlang/dlang.org/pull/2175 (<-- would have catched this)
https://github.com/dlang/dlang.org/pull/2176

Though a big step towards catching more breakage would be to detect invalid links ;-)

CyberShadow:
Some options, like `-betterC` and `-bounscheck`, have much less documentation in the new version. What's up with that? Should that extra documentation be moved to the man pages?

wilzbach:
> What's up with that?

Argh I didn't use the newly added `ddoxText`, but always used the man description text `helpText`.

CyberShadow:
HTML has duplicate IDs for switches like `-debug` and `-debug=foo`

dcompiler.dd
```diff
@@ -37,7 +37,7 @@ $(H2 $(LNAME2 requirements, Requirements and Downloads))
     $(LINUX
     $(LI 32 bit x86 and 64 bit x86-64 Linux operating system)
 
-    $(LI Gnu C compiler (gcc))
+    $(LI Gnu C compiler (gcc) or LLVM C compiler (clang))
```
CyberShadow: Can `clang` be used for `/bin/cc` for linking, then?

wilzbach: Yes, works fine on my machine:

image

(though it would be nice to have a switch to change the linker)

wilzbach: Try it yourself with:

```
echo 'void main(){import std.stdio; "hello".writeln;}' | CC=/usr/bin/clang dmd -v -run
```

wilzbach:
> HTML has duplicate IDs for switches like -debug and -debug=foo

Noted, but the browser will simply fallback to the first one:

http://dtest.dlang.io/artifact/website-8d6682e19fae73ba0d3a11bd12b7cd9a529fcc41-10ff4c70d2a4cff2151d5b48d7aa7eec/web/dmd-linux.html#switch-debug

Can I do this in a quick follow-up? @marler8997 is depending on this to add the documentation for his new `-i` flag...

CyberShadow:
Sorry, I didn't notice the "Blocking other work" label.

wilzbach:
> HTML has duplicate IDs for switches like -debug and -debug=foo

https://github.com/dlang/dlang.org/pull/2193

PR #2073, opened 2018-01-10
Reduced the size of images using tinypng

JackStouffer:
Average 60% file size reduction.


wilzbach:
> 100644 → 100755

Any reason for this?

JackStouffer:
No, just how the images came out of the program. I can change it back if that's necessary.

wilzbach:
> I can change it back if that's necessary.

No worries, I could quickly do that as well.
Anyways back to my main question (as I have never used tinypng before), do we risk loosing quality?

> Excellent question! When you upload a PNG (Portable Network Graphics) file, similar colors in your image are combined. This technique is called “quantization”. By reducing the number of colors, 24-bit PNG files can be converted to much smaller 8-bit indexed color images. All unnecessary metadata is stripped too. The result: better PNG files with 100% support for transparency. Have your cake and eat it too!
(from https://tinypng.com)

It looks ok and I definitely like smaller images (big win!), but I would like to wait for a second opinion here.
CC @CyberShadow

JackStouffer:
@wilzbach When I uploaded this I realized that Github has a cool new feature for images where you can swipe between the two on the diff page. I can see no differences between the two versions.

JackStouffer:
Could you be specific on which images you see banding in? Because I don't see anything.

CyberShadow:
Images with gradients. E.g. dlogo_opengraph.png, the Funkwerk logo.

If you are OK with lossy compression, why not switch to JPEG?

JackStouffer:
> Images with gradients. E.g. dlogo_opengraph.png, the Funkwerk logo.

Ah, I see it in the D logo now. With the Funkwerk logo it's only ever displayed on the website in small sizes so I doubt anyone will notice. But if it's still an issue I could just leave those specific ones untouched.

>If you are OK with lossy compression, why not switch to JPEG?

Some of these images are jpegs, and the ones that aren't need transparent backgrounds.

CyberShadow:
How about preserving the originals with a `_hq` suffix, like it's already done with some of the organization logos?

> Some of these images are jpegs, and the ones that aren't need transparent backgrounds.

I think most images that have transparent bits won't actually need transparent backgrounds, because we control what background they will be displayed on.

JackStouffer:
@wilzbach @CyberShadow Converting to jpg and using OSS optimizers or tinyjpg gives way less of savings than using pngs. For example, compiler-dmd.png

Original size: 16 KB
compiler-dmd

Tinypng: 8.1KB
compiler-dmd

Non-lossy JPG: 14KB
compiler-dmd

Lossy JPG: 8 KB
compiler-dmd


The lossy JPG looks way worse than the PNG.

I'll change this to remove the ones with banding and to keep originals with a `_hq` suffix. But, I think we should stick with PNGs.

CyberShadow:
Agreed, the PNG looks much better with the line art ones. If there are any photo-like images, JPEG may perform better there.

JackStouffer:
Used JPEGOptim to get lossless compression along with tinypng. Added hq versions for everything but icons.

CyberShadow:
Looks good!

CyberShadow:
Thanks!

PR #2074, opened 2018-01-10
Change Front Page to Focus on Call to Action and Social Proof

JackStouffer:
Changed the design of the front page to leverage social proof to increase conversions (reducing bounce and increasing clicking on "download" and "learn").

## Rationale

Social proof is an extremely powerful way to convince people to spend time on the web or to build trust, where 70% of people say they trust the recommendation of a random stranger online in things like reviews and 92% trust a recommendation from anyone they know [1]. In our case, we don't have strangers, but a brands like Netflix or eBay which the visitor might trust already. This gives close to the same effect as someone hearing that a friend uses something.

Herein lies the problem: most people don't see our social proof. Only 22% of people scroll down to the bottom of webpages on average [2], right where the "Organizations Using D" is currently located.

Further, the first ten seconds are the most important in retaining users on a page, with a focus on the first two seconds [3]. So combined with the scrolling data we have, it's prudent to use your most convincing arguments "above the fold". Keeping the social proof above the 600px mark hits the user right when they are trying to decide if they want to stay on the site.

With that in mind, here's a preview of the changes:

screen shot 2018-01-10 at 1 15 32 pm

In order to best capture the attention of the user in these ten seconds, this design works off of the "Z pattern", which puts the most important things on the page in a Z above the fold [4]. The reasoning is that from eye tracking software, it was discovered that people's eyes skim the entire screen when they first open a page. They either go in a F across the screen, or a Z [5]. The Z pattern puts the most important things on the four points of a Z, in our page, this would be the download button (which is now emphasized), the code example, and the organizations across the bottom.

screen shot 2018-01-09 at 4 52 05 pm

This change also has the added benefit of helping to alleviate one of the perceived shortcomings of D is its adoption by the industry [6]. This claim is both true and false. It's not up to the same adoption of say, Go, but it's still being used. This change focuses on adoption better.

## What Exactly Has Been Changed

* Moved "Orgs using D" to the top of the page under the intro
* Changed which organizations are displayed and the order to show the most well known brands first
* Increased whitespace on intro to declutter first impression so the user sees just the intro and the social proof
* Changed the color of the download button to focus on the call to action

## The Scientific Approach

While I believe this will increase conversions, you don't have to take my word for it. If we were to integrate Google Analytics (even temporarily just for this experiment if people don't like it), then we can use it to A/B test this change.

Essentially what happens is you have a control page (the current front page) and a modified page (these changes). When a user loads dlang.org they are randomly assigned one or the other and given a cookie to make sure they don't see the other version on reload.  GA then tracks the differences in behavior between the control group and the experiment group over a period of time (say three weeks). At the end of the experiment, we could see the differences in initial bounce or going to the documentation or downloads page.

This would give us an objective measure to decide if we should adopt this or not.

### Sources

[1] http://www.nielsen.com/us/en/insights/news/2012/trust-in-advertising--paid-owned-and-earned.html
[2] https://www.clicktale.com/resources/blog/resourcesbloginvest-in-customer-experience-software/
[3] https://www.nngroup.com/articles/how-long-do-users-stay-on-web-pages/
[4] https://instapage.com/blog/z-pattern-layout
[5] https://99designs.com/blog/tips/visual-hierarchy-landing-page-designs/
[6] https://www.quora.com/Which-language-has-the-brightest-future-in-replacement-of-C-between-D-Go-and-Rust-And-Why/answer/Andrei-Alexandrescu


CyberShadow:
Looks good! I think @andralex wants to OK all index changes though.

wilzbach:
FYI @andralex is on a conference for this week, so you might not here back from him until early next week.

andralex:
Well let's see!

CyberShadow:
Perhaps the org list ought to have a `max-width`:

JackStouffer:
@andralex Thanks!

@CyberShadow ok I'll submit a followup PR. 

BTW what do you think of the A/B testing idea?

CyberShadow:
> ok I'll submit a followup PR.

If you want to. It's not a common use case, and doesn't break too badly.

> BTW what do you think of the A/B testing idea?

Could be worthwhile, but I would be cautious about:
- adding more dependencies / failure points
- adding more complexity and opportunities for things to break (e.g. Google is banned in China)
- slowing down the page load
- giving Google more data (even if we don't care, some of our users might)

BBasile:
use this comment to vote for keep or revert.

PR #1840, opened 2017-07-24
add function input and output parameters...

fkromer:
to "Programming in D for C++ Programmers"

PR #2075, opened 2018-01-11
cpptod: add function input and output parameters.

carun:
First attempt to renew PR #1840.

I respect and appreciate the work of the original author. However I think that this PR explains a trivial difference between C++ and D. Please consider if this is really required to be merged.

C++ to D article ought to contain the important differences that eases the transition of a C++ programmer (from his preconceived notions) to D. It should be packed with information. Trivial topics could dilute the focus of this article. Things like struct, RAII, are good places for improvement, I believe.

Just my 2 cents.

### DIFF
--- a/cpptod.dd
+++ b/cpptod.dd
@@ -26,6 +26,8 @@ $(UL
         $(LI $(RELATIVE_LINK2 metatemplates, Meta Templates))
         $(LI $(RELATIVE_LINK2 typetraits, Type Traits))
         $(LI $(RELATIVE_LINK2 interfaces, Interfaces))
+        $(LI $(RELATIVE_LINK2 functioninputparameters, Function Input Parameters))
+        $(LI $(RELATIVE_LINK2 functionoutputparameters, Function Output Parameters))
 )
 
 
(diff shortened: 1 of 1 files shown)


cpptod.dd
```diff
@@ -913,6 +915,130 @@ SecondVariant.method();
...
+// inputArgumentPassedByValue = 42
+
+int inputArgumentPassedByReference = 42;
+foo(inputArgumentPassedByReference);
+// inputArgumentPassedByReference = 42
+)
+
+$(H4 The D Way)
+
+        D supports explicit input parameters.
```
wilzbach: `.` is used before examples above and below, I think it would be the best to the same here for the sake of consistency.

cpptod.dd
```diff
@@ -913,6 +915,130 @@ SecondVariant.method();
...
+        is wether the parameter is an output only or an input and output parameter.
+
+$(CPPCODE
+void foo(int& inputAndOutputParameter)
+{
+    inputAndOutputParameter = 42;
+}
+)
+
+The function is used as follows ($(I call-by-reference)):
```
wilzbach: This sounds weird. Ideas:
- This function is called as follows:
- Calling by reference is done as follows

cpptod.dd
```diff
@@ -913,6 +915,130 @@ SecondVariant.method();
...
+
+$(H3 $(LNAME2 functioninputparameters, Function Input Parameters))
+
+$(H4 The C++ Way)
+
+        $(B C++ does not support explicit input parameters.) If you use a
+        $(I pass-by-value) parameter $(D void function(int inputOnlyImplicitly)) it can just
+        be an input parameter implicitly. If you use a $(I pass-by-reference) parameter
+        usually $(D const) is applied to the parameter
+        $(D void foo(const int& inputOnlyParameterImplicitly)) that a forbidden
```
wilzbach: so that?

cpptod.dd
```diff
@@ -913,6 +915,130 @@ SecondVariant.method();
...
+}
+
+void foo(int& implicitInputParameterPassedByReference)
+{
+    // some code dependent on implicitInputParameterPassedByReference
+    ...
+}
+)
+
+The functions are used as follows:
```
wilzbach: s/are/can be/

cpptod.dd
```diff
@@ -913,6 +915,130 @@ SecondVariant.method();
...
+
+------
+void foo(in int inputParameter)
+{
+    // some code dependent on inputParameter
+    ...
+}
+------
+
+The function is used as follows:
```
wilzbach: s/is/can be/

cpptod.dd
```diff
@@ -913,6 +915,130 @@ SecondVariant.method();
...
+
+$(CPPCODE
+int outputArgument = 0;
+foo(outputArgument);
+// outputArgument = 42
+)
+
+$(H4 The D Way)
+
+        D supports explicit output parameters.
```
wilzbach: Same point about inconsistent usage of `.` vs `:`

cpptod.dd
carun: Sorry, I didn't understand. Can you elaborate?

wilzbach: Have a look at the examples above and below - they use a colon (`:`)

carun:
I think that this PR needs to be even more accurate.

C++ has only `const` and `volatile` (which is a different subject) as qualifiers for function parameters. D improves over C++ by introducing many new qualifiers to the function parameters (`in`, `out`, `const`, `immutable`, ` inout`, `ref` and many more, but I'm stating those that are applicable for programmers transitioning from C++) that lets the reader clearly infer what the function does by just look at it's signature. We don't even have to explain what `void getArea(in int width, in int height, out int area);` does (without even knowing the pass by reference semantics). The function signature is self explanatory.

Both C++ and D support input and output parameters. C++ supports it in the form of pass-by-reference (and pointer). In addition, D improves over C++ by differentiating between value types and reference types and also though the function parameter qualifiers. So in my opinion this PR needs a complete rewrite.

PR #2076, opened 2018-01-11
Add naming convention about UDAs to the style guide

jacob-carlborg:


dstyle.dd
```diff
@@ -135,7 +135,22 @@ ubyte asciiChar;
 -------------------------------
     )
 
+    $(DT $(LNAME2 naming_udas, User Defined Attributes))
```
wilzbach: User-defined, see https://github.com/dlang/dlang.org/pull/2003#discussion_r157333463

dstyle.dd
```diff
@@ -135,7 +135,22 @@ ubyte asciiChar;
...
+    $(DD For symbols that are only to be used as user defined attributes the
```
wilzbach: user-defined

dstyle.dd
jacob-carlborg: It's called "User Defined" everywhere in the spec https://dlang.org/spec/attribute.html#uda.

jacob-carlborg: Updated to `User-Defined` since it's in a title.

dstyle.dd
jacob-carlborg: Done.

dstyle.dd
wilzbach: Yeah, we should probably update the spec -> https://github.com/dlang/dlang.org/pull/2077

dstyle.dd
```diff
@@ -135,7 +135,22 @@ ubyte asciiChar;
...
+    $(DD For symbols that are only to be used as user-defined attributes the
+         names should be camelCased, so their first letter is lowercase. This
+         convention takes precedence over any previously mentioned conventions.
+         This matches conventions of the built in attributes like `@safe`,
+         `@nogc` and the special compiler recognized UDA `@selector`.
 
+-------------------------------
+struct Foo {} // this struct follows the regular naming conventions
+
+// this struct is only intended to by used as an UDA and therefore overrides the
```
carun: typo in the comment.

to by used as an -> to **be** used as an

jacob-carlborg: Fixed.

dstyle.dd
```diff
@@ -135,7 +135,22 @@ ubyte asciiChar;
 -------------------------------
     )
 
+    $(DT $(LNAME2 naming_udas, User-Defined Attributes))
+    $(DD For symbols that are only to be used as user-defined attributes the
+         names should be camelCased, so their first letter is lowercase. This
+         convention takes precedence over any previously mentioned conventions.
+         This matches conventions of the built in attributes like `@safe`,
+         `@nogc` and the special compiler recognized UDA `@selector`.
```
klickverbot: Well, `@nogc` should actually be `@noGc`. ;)

jmdavis: Or really, `@noGC`, since `GC` is an acronym, not a word. :)

But it's not like this is the first time that we've been inconsistent with naming, and technically, `@nogc` isn't a UDA.

quickfur:
ping @andralex

andralex:
thx!

PR #2077, opened 2018-01-11
Fix spelling: s/user defined/user-defined/

wilzbach:
See https://github.com/dlang/dlang.org/pull/2003#discussion_r157333463 for the motivation (it has been requested by @andralex).

### DIFF
--- a/spec/traits.dd
+++ b/spec/traits.dd
@@ -438,12 +438,12 @@ $(H2 $(GNAME getAliasThis))
 
 $(SECTION2 $(GNAME getAttributes),
     $(P
-        Takes one argument, a symbol. Returns a tuple of all attached user defined attributes.
+        Takes one argument, a symbol. Returns a tuple of all attached user-defined attributes.
         If no UDA's exist it will return an empty tuple.
     )
 
     $(P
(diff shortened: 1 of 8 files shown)


spec/traits.dd
```diff
@@ -438,12 +438,12 @@ $(H2 $(GNAME getAliasThis))
...
 $(SECTION2 $(GNAME getAttributes),
     $(P
-        Takes one argument, a symbol. Returns a tuple of all attached user defined attributes.
+        Takes one argument, a symbol. Returns a tuple of all attached user-defined attributes.
         If no UDA's exist it will return an empty tuple.
     )
 
     $(P
-        For more information, see: $(DDSUBLINK spec/attribute, uda, User Defined Attributes)
+        For more information, see: $(DDSUBLINK spec/attribute, uda, User-defined Attributes)
```
quickfur: Probably should capitalize `defined` here, to be consistent with other places where it's capitalized?

jacob-carlborg: Yeah, it's in a title. I think standard English rules should apply.

wilzbach: OK. changed.

PR #2078, opened 2018-01-12
Fixed orgs using d on large screens

JackStouffer:
Also added prefixed css rules

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1425,12 +1425,15 @@ body#Home #content > .intro .download .btn.action:hover
 
 .org_list
 {
+    max-width: 1400px;
+    margin: 0 auto;
+    padding: 10px 0;
+
     display: flex;
     align-items: center;
     flex-direction: row;
     justify-content: space-evenly;
     flex-wrap: wrap;
-    padding: 10px 0;
 }
 
 .org_list a img


wilzbach:
Before:

image


After:

image

css/style.css
```diff
@@ -1425,12 +1425,29 @@ body#Home #content > .intro .download .btn.action:hover
 
 .org_list
 {
+    max-width: 1400px;
+    margin: 0 auto;
+    padding: 10px 0;
+
+    display: -webkit-box;
+    display: -ms-flexbox;
     display: flex;
```
wilzbach: `flex` is a standard:

https://developer.mozilla.org/en-US/docs/Web/CSS/flex
https://www.w3.org/TR/css-flexbox-1

css/style.css
```diff
@@ -1425,12 +1425,29 @@ body#Home #content > .intro .download .btn.action:hover
...
+
+    -webkit-box-align: center;
+    -ms-flex-align: center;
     align-items: center;
+
+    -webkit-box-orient: horizontal;
+    -webkit-box-direction: normal;
+    -ms-flex-direction: row;
```
wilzbach: It's a standard now: https://developer.mozilla.org/en-US/docs/Web/CSS/flex-direction

css/style.css
```diff
@@ -1425,12 +1425,29 @@ body#Home #content > .intro .download .btn.action:hover
...
     flex-direction: row;
+
+    -webkit-box-pack: space-evenly;
+    -ms-flex-pack: space-evenly;
     justify-content: space-evenly;
+
+    -ms-flex-wrap: wrap;
```
wilzbach: Standard: https://www.w3.org/TR/css-flexbox-1/#propdef-flex-wrap

JackStouffer:
We hit an extra 10% of the browser market with the prefixes https://caniuse.com/#feat=flexbox

wilzbach:
Here's what I see:

image

Those are all platforms we don't care about.
Also we use `display: flex` at other places too without worrying about it, e.g. 

https://github.com/dlang/dlang.org/pull/1437 (since 1 1/2 years)

JackStouffer:
Fixed

PR #2079, opened 2018-01-12
Use FA_ICONs for the front page

wilzbach:
Before:

image

After:

image

CC @JackStouffer

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -321,8 +321,8 @@ $(DIVC org_list,
 
 $(DIVC org_list_footer,
     <p class="center">
-        $(LINK2 $(ROOT_DIR)orgs-using-d.html, Testimonials And More Organizations)
-        $(LINK2 $(ROOT_DIR)areas-of-d-usage.html, Current D use)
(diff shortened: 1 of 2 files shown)


index.dd
```diff
@@ -321,8 +321,8 @@ $(DIVC org_list,
 
 $(DIVC org_list_footer,
     <p class="center">
-        $(LINK2 $(ROOT_DIR)orgs-using-d.html, Testimonials And More Organizations)
-        $(LINK2 $(ROOT_DIR)areas-of-d-usage.html, Current D use)
+        $(FA_ICON quote-left) $(LINK2 $(ROOT_DIR)orgs-using-d.html, Testimonials and more organizations)
```
wilzbach: I'm not sure on the last change anymore. Looking at both screenshots both options seem fine, so I'm happy to revert this case style change.

JackStouffer: Gramatical nitpick: As this is a title of a link and not a sentence, it makes sense to use title case here and not sentence case.

wilzbach: Ok reverted this change.

PR #2081, opened 2018-01-12
[trivial]  Fix some dcompiler.dd formatting, add -dip1000 

ntrel:
See full commit message.

### DIFF
--- a/dcompiler.dd
+++ b/dcompiler.dd
@@ -523,7 +523,7 @@ $(WINDOWS
         )
       )
 
-      $(SWITCH $(SWNAME -conf)=$(I path)
+      $(SWITCH $(SWNAME -conf)=$(I path),
         use config file at $(I path)
       )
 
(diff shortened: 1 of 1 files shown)


dcompiler.dd
```diff
@@ -603,6 +603,10 @@ dmd -cov -unittest myprog.d
         implement http://wiki.dlang.org/DIP25 (experimental)
       )
 
+      $(SWITCH $(SWNAME -dip1000),
+        implement http://wiki.dlang.org/DIP1000 (experimental)
```
wilzbach: that should be sth. like `$(LINK2 https://github.com/dlang/DIPs/blob/master/DIPs/DIP1000.md, DIP1000)`

wilzbach: Fixed it myself.

ntrel: Thanks - I got the wrong link from stable `dmd --help` (fixed in master) :-/. Also do you know what's happening about ddoc automatically recognising links? (The `-dip25` link is plain text too).

PR #2072, opened 2018-01-10
Generalize SPEC_HEADERNAV to HEADERNAV

wilzbach:
With the DDoc wrapper it's now pretty easy to do cool things without big Makefile hacks.
This PR now uses the Ddoc wrapper for all Ddoc pages and as a motivation starts to add HEADERNAV to a few pages.

PR #2069, opened 2018-01-09
Remove the assert->writeln pipeline in favor of the DDoc wrapper

wilzbach:
Idea: use a custom Ddoc wrapper to pre-process the source files and avoid the
need for creating a temporary directory with the processed sources.

PR #2082, opened 2018-01-14
Rename ddoc.d -> ddoc_preprocessor.d

wilzbach:
@ZombineDev pointed out that `ddoc.d` is a rather uninformative name.


andralex:
I find this more confusing than before. 

* Before: "ddoc.d -> aha, that is the file doing the ddoc stuff"
* After: "ddoc_preprocessor.d -> wait, is there some postprocessing or some other ddoc work as well that I need to worry about?"

Restructuring should add net value, this is at the very best a lateral step.

wilzbach:
FYI: I opened this PR, s.t. https://github.com/dlang/dlang.org/pull/2072 doesn't get blocked.

For a bit of history:
- https://github.com/dlang/dlang.org/pull/2060 (automatic TOC generation with `ddoc.d` - merged - PR that introduced `ddoc.d`)
- https://github.com/dlang/dlang.org/pull/2069 (moves the assert -> writeln step into `ddoc.d` - blocked)

andralex:
So it seems we added automation that requires us to rename modules in our code. Seems the right place to change things is in the automation, no?

wilzbach:
> So it seems we added automation that requires us to rename modules in our code.

I'm not sure I can follow. To what "automation" are you referring to? Do you mean the assert->writeln transformation?
I added the rewriting of [`assert`'s in the runnable examples](https://dlang.org/blog/2017/03/08/editable-and-runnable-doc-examples-on-dlang-org) last year. However, it was always pretty ugly because (1) it requires making a full copy of the Phobos repository and (2) it requires the Phobos repository to be checked-out and thus always does the existence check (even on `make clean`). That's why @CyberShadow is so happy that removing that extra step in #2069, but I'm not sure how that's related to this PR which is just about renaming the tool.

I don't care how the Ddoc wrapper tool is named as long as the name bike-shedding doesn't block my PRs. Hence, I moved this in a separate PR.
(I picked `ddoc.d` myself, so I obviously liked it.)

CyberShadow:
> (I picked ddoc.d myself, so I obviously liked it.)

I liked `ddoc` too. But it really doesn't matter, at all.

PR #2084, opened 2018-01-15
Show runtime docs in the menu on dlang.org (prerelease docs only)

wilzbach:
See also: https://github.com/dlang/druntime/pull/2042

image

image

For the release docs, we will have to wait until 2.079.0
Of course this is still a bit unpolished, but it won't change if we don't expose it.


wilzbach:
@andralex Thanks! FYI: dlang/druntime#2042 is needed for this too.

PR #2085, opened 2018-01-15
Run sub make evocations with -j$(nproc)

wilzbach:
Currently building dlang.org doesn't work well with `-j` due to not
fully correct specified dependencies on the target definitions.
However, there's no reson why we can't use parallelism for the
individual sub makes.


CyberShadow:
IIRC this is the wrong way to do it, wasn't there something like `MAKEFLAGS` better suited for this?

CyberShadow:
I'm uneasy about this. Why not fix the dependencies?

wilzbach:
> Why not fix the dependencies?

Because I don't know where they are broken. Everything works fine on my machine - even with `-j50`, maybe we have already fixed this?

wilzbach:
Hmm, how about just adding `-j$(NPROC)` to the big guy - building DMD?
It would still result in faster build time as DMD builds fine with `-j`.

PR #2086, opened 2018-01-15
Replace TWID with the second latest post from the DBlog

wilzbach:
@adamdruppe: I have noticed that TWID doesn't get much attention lately: http://arsdnet.net/this-week-in-d/2017-nov-26.html
I suppose this is due to the success of the DBlog and you more or less discontinued TWID?
FWIW I liked your in-depth articles, so it would be great to see them appear on the DBlog instead!

Here's how it currently looks - I'm not too happy about the formatting. Suggestions are welcome.

image

CC @mdparker


dlang.org.ddoc
```diff
@@ -145,8 +145,12 @@ _=
 
 DBLOG_LATEST_AUTHOR=
 DBLOG_LATEST_DATE=
-DBLOG_LATEST_LINK=
+DBLOG_LATEST_LINK=https://dlang.org/blog
```
wilzbach: This is required for builds with `DIFFABLE=0` and the `chmgen` URL checking:

```
Error while processing file: /home/circleci/dlang.org/web/index.html
object.Exception@chmgen.d(577): Empty URL
----------------
```

https://circleci.com/gh/dlang/dlang.org/487?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

adamdruppe:
Oh yes, indeed, I have generated many pages but not written anything original for a while. I do have a bit of material but haven't had the time. Heck, the twid template doesn't even work with new dmd - I had to keep an old copy of dmd around for it! And I've been wanting to do this differently forever but again it is just kinda low priority in life and keeps getting bumped by other things.

I kinda want to either switch it to more of a blog format or maybe just take what's left on my todo list and send it to Mike for that blog.

andralex:
@adamdruppe anything we can do, let us know. FWIW @mdparker is great at taking rough material and converting it to finalized blog posts!

PR #2089, opened 2018-01-16
Cherry-pick 1891 + 1926 from master to stable

wilzbach:
- https://github.com/dlang/dlang.org/pull/1891
- https://github.com/dlang/dlang.org/pull/1926

2.077.1 comes with DUB mirrors which should avoid the problems due to
code.dlang.org being down which we are seeing on doc builds against `stable`.

See also: https://github.com/dlang/phobos/pull/6035#issuecomment-357824158


quickfur:
Let's just do it!

wilzbach:
> At least one approved review is required by reviewers with write access.

@quickfur - you need to use the formal "Approve" ;-)

PR #2091, opened 2018-01-17
add spec macros to dlang.org.ddoc

WalterBright:
Add `BEST_PRACTICE`, `IMPLEMENTATION_DEFINED` and `UNDEFINED_BEHAVIOR` to the spec macros so that the display of these will be consistent and cross-referencable.

### DIFF
--- a/dlang.org.ddoc
+++ b/dlang.org.ddoc
@@ -4,6 +4,9 @@ ACRONYM = <acronym title="$+">$1</acronym> ($+)
 ASSIGNEXPRESSION = $(GLINK2 expression, AssignExpression)
 _=
 
+$(COMMENT While not part of the spec, best practices offers advice on how to best use a feature)
+BEST_PRACTICE= $(P $(B Best Practices:) $0)
+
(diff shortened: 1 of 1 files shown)


WalterBright:
Blocking https://github.com/dlang/dlang.org/pull/2092

wilzbach:
Wouldn't [spec.ddoc](https://github.com/dlang/dlang.org/blob/master/spec/spec.ddoc) be a better place for these macros?

WalterBright:
I don't know. I don't think it particularly matters, as the .ddoc files are quite a mess.

wilzbach:
> I don't think it particularly matters,

It does - at least if you want to use something nicer than a paragraph for the LaTeX PDF?

>  as the .ddoc files are quite a mess.

A first step at working against the mess and making it clearer for everyone which `.ddoc` files are used for the spec:

https://github.com/dlang/dlang.org/pull/2093
https://github.com/dlang/dlang.org/pull/2094

BTW the thing I worry about is that this adds three more "magic" macros to the mess.

PR #2092, opened 2018-01-17
AssertExp: make it a bit more formal and correct

WalterBright:
Correct omissions in the original, and clearly mark implementation defined behavior, undefined behavior, and best practices while avoiding compiler-specific details.

Relies on new macros in https://github.com/dlang/dlang.org/pull/2091

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -1623,44 +1623,54 @@ $(GNAME AssertExpression):
     $(D assert $(LPAREN)) $(GLINK AssignExpression) $(D ,) $(GLINK AssignExpression) $(D ,)$(OPT) $(D $(RPAREN))
 )
 
-    $(P The assert expression is used to declare conditions that the
-        programmer asserts must hold at that point in the program if the
-        program logic has been correctly implemented. It can be used both as a
-        debugging tool and as a way of communicating to the compiler facts
-        about the code that it may employ to produce more efficient code.
+    $(P The first $(I AssignExpression) must evaluate to true. If it does not, an $(I Assert Failure)
+    has occurred and the program enters an $(I Invalid State).
     )
 
-    $(P Programs for which $(I AssignExpression) is false are invalid.
-        Subsequent to such a false result, the program is in an invalid,
-        non-recoverable state.
+    $(P If the first $(I AssignExpression) consists entirely of compile time constants,
+    and evaluates to false, it is a special case; it
+    signifies that it is unreachable code.
+    Compile Time Function Execution (CTFE) is not attempted.
     )
 
-    $(P As a debugging tool, the compiler may insert checks to verify that
-        the condition indeed holds by evaluating $(I AssignExpression) at
-        runtime.  If it evaluates to a non-null class reference, the class
-        invariant is run.  Otherwise, if it evaluates to a non-null pointer to
-        a struct, the struct invariant is run.  Otherwise, if the result is
-        false, an $(D AssertError) is thrown.  If the result is true, then no
-        exception is thrown. In this way, if a bug in the code causes the
-        assertion to fail, execution is aborted, prompting the programmer to
-        fix the problem.
-    )
-
-    $(P It is implementation defined whether the $(I AssignExpression) is
-        evaluated at run time or not. Programs that rely on side effects of $(I
-        AssignExpression) are invalid.
-    )
-
-    $(P The result type of an assert expression is $(D void).  Asserts are
-        a fundamental part of the $(DDLINK spec/contracts, Contract Programming,
-        Contract Programming) support in D.
+    $(P $(I AssertExpression) has different semantics if it is in a
+    $(DDLINK spec/unittest, Unit Tests, $(D unittest)) or
+    $(DDLINK spec/contracts, Contract Programming, $(D in) contract).
     )
 
     $(P The second $(I AssignExpression), if present, must be implicitly
         convertible to type $(D const(char)[]).
-        It is evaluated if the
-        result is false, and the string result is appended to the
-        $(D AssertError)'s message.)
+    )
+
+    $(P If the first $(I AssignExpression) is a reference to a class instance for
+    which a $(DDSUBLINK spec/class, invariants, Class Invariant) exists, the $(I Class Invariant) must hold.
+    )
+
+    $(P If the first $(I AssignExpression) is a pointer to a struct instance for
+    which a $(I Struct Invariant) exists, the $(I Struct Invariant) must hold.
+    )
+
+    $(P The type of an $(I AssertExpression) is $(D void).
+    )
+
+    $(UNDEFINED_BEHAVIOR Once in an $(I Invalid State) the behavior of the continuing execution
+    of the program is undefined.)
+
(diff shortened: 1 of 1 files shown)


spec/expression.dd
```diff
@@ -1671,19 +1671,21 @@ $(GNAME AssertExpression):
...
-    $(P The expression $(D assert(0)) is a special case; it
-        signifies that it is unreachable code.
-        Without optimizations, an $(D AssertError) is thrown at runtime.
-        In release mode ($(LINK2 $(ROOT_DIR)/dmd.html#switch-release, `-release`))
-        the execution is halted
-        (on the x86 processor, a $(D HLT) instruction can be used to halt
-        execution).
-        The optimization and code generation phases of compilation may
-        assume that it is unreachable code.
+    The implementation may handle $(D assert(0))  differently in that in release mode
```
wilzbach: double space

spec/expression.dd
```diff
@@ -1623,44 +1623,44 @@ $(GNAME AssertExpression):
...
+    any side effects specified by the $(I AssertExpression) may not occur.
+    The behavior if the first $(I AssertExpression) is evaluated and is false
+    is also typically set with a compiler switch and may include these options:
+    $(OL
+	$(LI continuing execution)
+	$(LI immediately halting via execution of a special CPU instruction)
+	$(LI aborting the program)
+	$(LI calling the assert failure function in the corresponding C
+	runtime library)
+	$(LI throwing the $(D AssertError) exception in the D runtime library)
```
wilzbach: Don't use tabs.

wilzbach:
DAutoTest was failing due to:

```
> Searching for undefined macros
/dev/shm/dtest/work/repo/dlang.org/web/dlangspec.html:7265:    <!--UNDEFINED MACRO: "UNDEFINED_BEHAVIOR"-->
/dev/shm/dtest/work/repo/dlang.org/web/dlangspec.html:7267:    <!--UNDEFINED MACRO: "IMPLEMENTATION_DEFINED"-->
/dev/shm/dtest/work/repo/dlang.org/web/dlangspec.html:7269:    <!--UNDEFINED MACRO: "BEST_PRACTICE"-->
/dev/shm/dtest/work/repo/dlang.org/web/spec/expression.html:1752:    <!--UNDEFINED MACRO: "UNDEFINED_BEHAVIOR"-->
/dev/shm/dtest/work/repo/dlang.org/web/spec/expression.html:1754:    <!--UNDEFINED MACRO: "IMPLEMENTATION_DEFINED"-->
/dev/shm/dtest/work/repo/dlang.org/web/spec/expression.html:1756:    <!--UNDEFINED MACRO: "BEST_PRACTICE"-->
```

I rebased your PR to rekick DAutoTest -> should be passing now.

spec/expression.dd
```diff
@@ -1623,44 +1623,44 @@ $(GNAME AssertExpression):
...
-        debugging tool and as a way of communicating to the compiler facts
-        about the code that it may employ to produce more efficient code.
+    $(P The first $(I AssignExpression) must evaluate to true. If it does not, an $(I Assert Failure)
+    has occurred and the program enters an $(I Invalid State).
     )
 
-    $(P Programs for which $(I AssignExpression) is false are invalid.
-        Subsequent to such a false result, the program is in an invalid,
-        non-recoverable state.
+    $(P The expression $(D assert(0)) is a special case; it
```
andralex: We need to clarify this further. Are these also special-cased?

```d
assert(1 - 1);
assert(10 * 10 - 100 == 0);
enum a = 0;
assert(a);
enum b = sqrt(100);
assert(b == 10);
assert(false);
assert(null);
assert([]);
```

In other words, to what extent does the compiler go to evaluate the expression during compilation? I don't know the answer.

spec/expression.dd
```diff
@@ -1623,44 +1623,44 @@ $(GNAME AssertExpression):
...
+    signifies that it is unreachable code.
```
andralex: This paragraph should also discuss `assert(0, message);` If I remember correctly, the message is displayed in debug mode and not in release mode (which btw is irritating).

WalterBright:
> different semantics inside unittest blocks and contracts, but it's not defined exactly what these different semantics are.

Right, and I plan to address that in the unittest and contract sections, which will be rewritten as well.

spec/expression.dd
```diff
@@ -1671,19 +1672,21 @@ $(GNAME AssertExpression):
...
     $(P When compiled and run, it will produce the message:)
 
-$(CONSOLE
-core.exception.AssertError@test.d(3) an error message
-)
+    $(CONSOLE core.exception.AssertError@test.d(3) an error message)
+
+    The implementation may handle $(D assert(0)) differently in that in release mode
+    it may simply generate a $(D HLT) instruction or equivalent, even
+    in release mode.
```
aG0aep6G: Repetition of "in release mode".

andralex: @aG0aep6G you may want to make obvious/noncontroversial fixes like this by editing the PR and committing to it directly. It works real nice... props for the github team.

andralex:
The phrases italicized and title cased, as in $(I Assert Failure), would be only meaningful if defined elsewhere.

spec/expression.dd
aG0aep6G: > @aG0aep6G you may want to make obvious/noncontroversial fixes like this by editing the PR and committing to it directly. It works real nice... props for the github team.

I don't think I have the powers to do that. I'm not a member of Team D or anything.

andralex: @aG0aep6G if you go to "files", do you see something like a crayon in the decoration at the top of each file changed?

aG0aep6G: > @aG0aep6G if you go to "files", do you see something like a crayon in the decoration at the top of each file changed?

Yeah. It's not clickable and the tooltip says: "You must be signed in and have push access to the AssertExp branch to make changes." I don't have that kind of access to the D repositories.

quickfur:
@andralex For obvious reasons, the PR editing function is only available to people who have commit access to the repo (other than the PR submitter himself, of course).

WalterBright:
@andralex I made some changes to address the comments.(but not the italicized one)

andralex:
My changes have been overwritten, sigh. @WalterBright please avoid force pushing...

WalterBright:
Generally incomprehensible errors and warnings from DAutoTest.

spec/expression.dd
```diff
@@ -1623,44 +1623,54 @@ $(GNAME AssertExpression):
...
+    which a $(I Struct Invariant) exists, the $(I Struct Invariant) must hold.
+    )
+
+    $(P The type of an $(I AssertExpression) is $(D void).
+    )
+
+    $(UNDEFINED_BEHAVIOR Once in an $(I Invalid State) the behavior of the continuing execution
+    of the program is undefined.)
+
+    $(IMPLEMENTATION_DEFINED Whether the first $(I AssertExpression) is evaluated
```
wilzbach: You never close this ... any decent macro language would have made it _a lot_ easier to spot the error ;-)
Also your PR needs to be rebased against `upstream/master`, s.t. it includes the new macros.

WalterBright: thanks!

PR #2093, opened 2018-01-17
Move Latex spec Ddoc file to /spec

wilzbach:
Makes it clearer that this file is only used for the specification.

Motivation: https://github.com/dlang/dlang.org/pull/2091#issuecomment-358200790


WalterBright:
Adding comments in the makefile are also helpful.

PR #2095, opened 2018-01-17
move macros to spec/spec.ddoc

WalterBright:
Since they only apply to the spec.

### DIFF
--- a/dlang.org.ddoc
+++ b/dlang.org.ddoc
@@ -4,9 +4,6 @@ ACRONYM = <acronym title="$+">$1</acronym> ($+)
 ASSIGNEXPRESSION = $(GLINK2 expression, AssignExpression)
 _=
 
-$(COMMENT While not part of the spec, best practices offers advice on how to best use a feature)
-BEST_PRACTICE= $(P $(B Best Practices:) $0)
-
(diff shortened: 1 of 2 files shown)


wilzbach:
Sadly there's a common macro file that is used by all three spec builds (html, pdf, epub), but they all use different ones:

https://github.com/dlang/dlang.org/blob/0c475b6499fd5fdbcfb1961b6513f0a8af232c56/posix.mak#L470-L471

https://github.com/dlang/dlang.org/blob/0c475b6499fd5fdbcfb1961b6513f0a8af232c56/posix.mak#L522-L523

https://github.com/dlang/dlang.org/blob/0c475b6499fd5fdbcfb1961b6513f0a8af232c56/posix.mak#L545-L546

Should I quickly add a `spec/macros.dd` which is shared by all three or do you have a better idea?

WalterBright:
I said it was a disaster :-)

The reason for 3 .ddoc files is there's no way to do conditionals with macros. So for different expansions, different .ddoc files are needed. However, the identical macros should be merged into a common file.

Feel free to reorganize it. Just be sure and add comments to the makefile about what is going on!

PR #2096, opened 2018-01-17
put class invariant spec in class.dd

WalterBright:
Because `contracts.dd` is an article, not a specification.

### DIFF
--- a/spec/class.dd
+++ b/spec/class.dd
@@ -836,9 +836,99 @@ $(GNAME Invariant):
     $(D invariant) $(GLINK2 statement, BlockStatement)
 )
 
-$(P Class invariants are used to specify characteristics of a class that always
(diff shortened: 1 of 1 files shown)


wilzbach:
> Because contracts.dd is an article, not a specification.

It's part of the spec ... no?
https://dlang.org/spec/contracts.html

WalterBright:
But it shouldn't be. It was written as an article, and thrown into the spec long ago because I was in a hurry. It doesn't read like a spec, and is unsuitable to be a spec. This is the first of a series of PRs to fix that.

andralex:
Reminder: need to put invariant stuff in structs too.

WalterBright:
@andralex I know, but I wanted this approved first, as the struct invariants will be a cut&paste job.

PR #2099, opened 2018-01-18
Remove floating point NCEG operators from the spec

wilzbach:
- They have been deprecated since 2.067 and error since 2.072:

https://dlang.org/deprecate.html#Floating%20point%20NCEG%20operators

- I removed the bit about the exceptions as that's no true anymore:

https://run.dlang.io/is/vkwgyZ

I also rewrote the entire table to make it less confusing to humans.


JinShil:
Needs an update to the deprecations table.

PR #2103, opened 2018-01-19
remove comparisons to C++

WalterBright:
Comparisons to other languages do not belong in a language specification.

### DIFF
--- a/spec/contracts.dd
+++ b/spec/contracts.dd
@@ -7,9 +7,7 @@ $(HEADERNAV_TOC)
         $(P Contracts are a breakthrough technique to reduce the programming effort
         for large projects. Contracts are the concept of preconditions, postconditions,
         errors, and invariants.
-        Contracts can be done in C++ without modification to the language,
-        but the result is
-        clumsy and inconsistent.)
+        )
 
         $(P Building contract support into the language makes for:)
 
--- a/spec/template.dd
+++ b/spec/template.dd
@@ -4,9 +4,6 @@ $(SPEC_S Templates,
 
 $(HEADERNAV_TOC)
 
-$(BLOCKQUOTE_BY Richard Deyman,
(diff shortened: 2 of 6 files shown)


spec/template.dd
```diff
@@ -5,7 +5,7 @@ $(SPEC_S Templates,
 $(HEADERNAV_TOC)
 
 $(BLOCKQUOTE_BY Richard Deyman,
-I think that I can safely say that nobody understands C++ template mechanics.)
+I think that I can safely say that nobody understands template mechanics.)
```
JackStouffer: Since this one is a quote you might want to leave this one.

spec/contracts.dd
```diff
@@ -7,9 +7,7 @@ $(HEADERNAV_TOC)
         $(P Contracts are a breakthrough technique to reduce the programming effort
         for large projects. Contracts are the concept of preconditions, postconditions,
         errors, and invariants.
-        Contracts can be done in C++ without modification to the language,
-        but the result is
-        clumsy and inconsistent.)
+	)
```
wilzbach:
```
Searching for tabs
./spec/contracts.dd:10:	)
```

There are a couple of tools that simply things a bit. For example, all D repos contain a `[.editorconfig`](https://github.com/dlang/dlang.org/blob/master/.editorconfig) file which specifies indent, tabs <-> whitespace etc.
Most IDEs/Editors support this automatically - even Emacs:
https://github.com/editorconfig/editorconfig-emacs

spec/template.dd
wilzbach: This looks more like a "joke quote" based on the famous:

> I think I can safely say that nobody understands quantum mechanics.

-- Richard Feynman

http://www-history.mcs.st-andrews.ac.uk/Quotations/Feynman.html

andralex: I'd say just remove.

PR #2106, opened 2018-01-19
Use Smaller Netflix Image on Orgs Using D Page

JackStouffer:
Saves a couple of KBs

### DIFF
--- a/orgs-using-d.dd
+++ b/orgs-using-d.dd
@@ -154,7 +154,7 @@ $(DIVC orgs-using-d center,
             $(FA_HIRING www.microline.eu/work-with-us/?lang=en)
         )
     )
-    $(DORG Netflix, https://www.netflix.com, netflix.jpg,
+    $(DORG Netflix, https://www.netflix.com, netflix_small.png,
         Entertainment company,
         $(FA_QUOTE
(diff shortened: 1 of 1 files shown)


CyberShadow:
Not sure if this is making it more or less consistent... besides, it's a household brand, nothing wrong with it standing out a bit.

JackStouffer:
@CyberShadow Not sure what you're referring to. All this PR does is change the Netflix image from the  800px wide version to the 400px wide version because its only ever displayed at a 178px width.

CyberShadow:
There is a visible difference in DAutoTest's HTML preview. Was that not intended, then?

CyberShadow:
> All this PR does is change the Netflix image from the 800px wide version to the 400px wide version because its only ever displayed at a 178px width.

Only on regular DPI displays... mobile devices and high-DPI displays benefit from the extra pixels. Ideally we'd be using SVGs or imgsets here.

JackStouffer:
>Only on regular DPI displays... mobile devices and high-DPI displays benefit from the extra pixels. Ideally we'd be using SVGs or imgsets here.

Ah, ok. When I have time I'll look into getting an SVG

PR #2108, opened 2018-01-19
Minor fixes for ctod.dd

luismarques:
- Fix C array looping examples by adding the missing `size_t`;
- Avoid Windows/platform specific uses of the names "word" and "dword" in the alignment examples;
- Fix and tweak typos, grammar, etc.

Also, I would review the use of "in praxis". I had never seen that. Even if it is correct, maybe the more common "in practice" would sound more natural to most readers?

### DIFF
--- a/ctod.dd
+++ b/ctod.dd
@@ -244,7 +244,7 @@ $(P
 $(CCODE
 #define ARRAY_LENGTH        17
 int array[ARRAY_LENGTH];
-for (i = 0; i &lt; ARRAY_LENGTH; i++)
+for (size_t i = 0; i &lt; ARRAY_LENGTH; i++)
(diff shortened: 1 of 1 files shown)


luismarques:
I amended the commit to include the change "in praxis" -> "in practice"

quickfur:
Alright, let's do it!

luismarques:
@quickfur I missed a size_t in the first array example ("Initializing all elements of an array"). Github's direct edit feature is pretty crappy... Can you fix that one?

luismarques:
@quickfur also, there's another "codeunit" -> "code unit". Urg.

wilzbach:
I think the easiest is to edit the file again and submit a new PR.

PR #2109, opened 2018-01-19
Document the new -i dmd compiler option

marler8997:


marler8997:
Pinging @JinShil @wilzbach @timotheecour @ibuclaw (who are the most familiar with https://github.com/dlang/dmd/pull/7099) to review.

JinShil:
We might want to wait until the fate of https://github.com/dlang/dmd/pull/7703 is determined.  We're trying to make it so that the documentation all comes from "one source of truth" so we don't have so many inconsistencies between website, man page, and the actual implementation.

marler8997:
So do we want to just put this documentation in dmd or is it going to be some time before dlang.org uses that?

JinShil:
I think we need one more PR:  https://github.com/dlang/dlang.org/pull/2068  Then you should be able to add the documentation to DMD.  I suppose you can add it now, but I don't think we'll see the docs rendered until #2068 is in.

wilzbach:
>  I suppose you can add it now,

dlang/dmd#7703 has been merged, so you can safely add it to `cli.d` 
You can already see preview the change on the diff for the MAN pages at DAutotest.

> but I don't think we'll see the docs rendered until #2068 is in.

#2068 is its final stage, so hopefully it will be merged soon.
In any case, you can just add it to `cli.d` and it will appear on dlang.org once #2068 is in.

marler8997:
Work moved to dmd PR here (https://github.com/dlang/dmd/pull/7777)

PR #2110, opened 2018-01-19
Upgraded to jQuery 3.2.1

JackStouffer:
Google's auditing tool in chrome notified me that jquery 1.7.2 has two known XSS vulns and this marked the page down (it probably marks down our Page Rank as well). I've upgraded us to the latest and greatest. I did a cursory check of our existing JS code to see if anything was broken. I didn't see anything, and we weren't doing anything complicated with jQuery anyway.

Moved to jQuery's recommended CDN which supports integrity checks.


JackStouffer:
jQuery vulnerabilities: https://snyk.io/vuln/npm:jquery#lh@1.7.2

quickfur:
+1, we definitely need to get this in.  Security vulnerabilities are Not Good, both in the sense of security, and also from a marketing / PR perspective.

CyberShadow:
We're *probably* not affected but it's still an obviously good idea to update.

However, this updates across *two* major versions, which IIRC have breaking changes. We should either test thoroughly that all dynamic content on the site still works, or just elect to update to the latest minor version (1.12.14, it seems).

CyberShadow:
You need to update posix.mak BTW.

JackStouffer:
btw @wilzbach the certificate for https://contribs.dlang.io seems to be out of date.

JackStouffer:
Closing. No longer interested

PR #2112, opened 2018-01-21
clarify CTFE spec a bit

aG0aep6G:
Motivation: http://forum.dlang.org/post/p2o666$r4h$1@digitalmars.com

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -2233,7 +2233,7 @@ void foo()
     $(LI Non-portable casts (eg, from $(D int[]) to $(D float[])), including
         casts which depend on endianness, are not permitted.
         Casts between signed and unsigned types are permitted)
-    $(LI Reinterpretation of overlapped fields in a Union.)
+    $(LI Reinterpretation of overlapped fields in a Union is not permitted.)
     )
 
     $(P Pointers are permitted in CTFE, provided they are used safely:)
@@ -2355,11 +2355,11 @@ int foo(char[] s)
 const int x = foo("1");
 ---
 
-        $(P is illegal, because the runtime code for foo() cannot be
+        $(P is illegal, because the runtime code for `foo` cannot be
         generated. A function template would be the appropriate
         method to implement this sort of thing.)
 
-$(H3 $(LNAME2 nogc-functions, No-GC Functions))
+$(H2 $(LNAME2 nogc-functions, No-GC Functions))
 
         $(P No-GC functions are functions marked with the $(D @nogc) attribute.
             Those functions do not allocate memory on the GC heap,
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -2181,15 +2181,14 @@ void main()
 $(H2 $(LNAME2 interpretation, Compile Time Function Execution (CTFE)))
 
     $(P Functions which are both portable and free of global side-effects can be
-    executed at compile time. This is useful when constant folding
-    algorithms need to include recursion and looping. Compile time function
-    execution is subject to the following restrictions:
+    executed in Compile Time Function Execution. CTFE is subject to the
+    following restrictions:
     )
 
     $(OL
     $(LI The function source code must be available to the compiler. Functions
         which exist in the source code only as $(D_KEYWORD extern) declarations
-        cannot be executed at compile time.)
+        cannot be executed in CTFE.)
 
     $(LI Executed expressions may not reference any global or local
         static variables.)
@@ -2271,15 +2270,15 @@ static assert(countTen(6) == 6);    // OK
 static assert(countTen(12) == 12);  // invalid, modifies y.
 ---
     $(P The $(D __ctfe) boolean pseudo-variable, which evaluates to $(D_KEYWORD true)
-        at compile time, but $(D_KEYWORD false) at run time, can be used to provide
+        in CTFE, but $(D_KEYWORD false) otherwise, can be used to provide
         an alternative execution path to avoid operations which are forbidden
-        at compile time. Every usage of $(D __ctfe) is evaluated before
+        in CTFE. Every usage of $(D __ctfe) is evaluated before
         code generation and therefore has no run-time cost, even if no optimizer
         is used.
     )
 
 
-    $(P In order to be executed at compile time, the function
+    $(P In order to be executed in CTFE, the function
     must appear in a context where it must be so executed, for
     example:)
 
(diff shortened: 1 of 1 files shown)


spec/function.dd
```diff
@@ -2181,15 +2181,49 @@ void main()
 $(H2 $(LNAME2 interpretation, Compile Time Function Execution (CTFE)))
 
     $(P Functions which are both portable and free of global side-effects can be
-    executed at compile time. This is useful when constant folding
-    algorithms need to include recursion and looping. Compile time function
-    execution is subject to the following restrictions:
+    executed at compile time. In certain contexts, such compile time execution
```
ZombineDev: Execution -> Evaluation

spec/function.dd
```diff
@@ -2181,15 +2181,49 @@ void main()
...
+    is guaranteed. It is called Compile Time Function Execution (CTFE) then.
+    The contexts that trigger CTFE are:)
+
+    $(UL
+    $(LI initialization of a static variable or a manifest constant)
+    $(LI static initializers of struct/class members)
+    $(LI dimension of a static array)
+    $(LI argument for a template value parameter)
+    $(LI $(D_KEYWORD static if))
```
ZombineDev: I would recommend using `$(GLINK ...)` to link to the corresponding grammar, as that gives more information to the reader.

spec/function.dd
```diff
@@ -2181,15 +2181,49 @@ void main()
...
+    $(LI $(D_KEYWORD static assert))
+    $(LI $(D_KEYWORD mixin) statement)
     )
 
+$(SPEC_RUNNABLE_EXAMPLE_COMPILE
+---
+template eval( A... )
+{
+    const typeof(A[0]) eval = A[0];
+}
```
ZombineDev: Why not simply:
```D
enum eval(Args...) = A[0];
```
?

ZombineDev: Ah, I see you just moved the example up, but still I think we should show the simplest syntax that works to the readers.

spec/function.dd
```diff
@@ -2181,15 +2181,49 @@ void main()
...
+
+    $(UL
+    $(LI initialization of a static variable or a manifest constant)
+    $(LI static initializers of struct/class members)
+    $(LI dimension of a static array)
+    $(LI argument for a template value parameter)
+    $(LI $(D_KEYWORD static if))
+    $(LI $(D_KEYWORD static foreach))
+    $(LI $(D_KEYWORD static assert))
+    $(LI $(D_KEYWORD mixin) statement)
```
ZombineDev: `pragma(msg, ...)` too.

spec/function.dd
```diff
@@ -2354,11 +2355,11 @@ int foo(char[] s)
 const int x = foo("1");
 ---
 
-        $(P is illegal, because the runtime code for foo() cannot be
+        $(P is illegal, because the runtime code for `foo` cannot be
```
ZombineDev: Slightly off topic, but if possible, we should combine this paragraph with one above as currently they're rendered as two independent paragraphs, while they semantically form one single paragraph with an example in the middle.

spec/function.dd
aG0aep6G: I wouldn't mind changing it to "evaluation", but I'm not sure if this is the place to push it.

I had remembered it as "evaluation" before I started editing the section. [The tour](https://tour.dlang.io/tour/en/gems/compile-time-function-evaluation-ctfe) also calls it "evaluation". But [Programming in D](http://ddili.org/ders/d.en/functions_more.html#ix_functions_more.CTFE) follows the spec with "execution", and [Wikipedia](https://en.wikipedia.org/wiki/Compile_time_function_execution) lists "execution" first. So there doesn't seem to be a consensus.

MetaLang: I don't think there's much of a difference in this case and either is probably fine.

spec/function.dd
MetaLang: All `pragma` statements are compile-time only.

spec/function.dd
```diff
@@ -2181,15 +2181,50 @@ void main()
...
+    $(LI initialization of a static variable or a
+        $(DDSUBLINK spec/enum, manifest_constants, manifest constant))
+    $(LI static initializers of struct/class members)
+    $(LI dimension of a $(DDSUBLINK spec/arrays, static-arrays, static array))
+    $(LI argument for a $(DDSUBLINK spec/template, template_value_parameter,
+        template value parameter))
+    $(LI $(DDSUBLINK spec/version, staticif, $(D_KEYWORD static if)))
+    $(LI $(DDSUBLINK spec/version, staticforeach, $(D_KEYWORD static foreach)))
+    $(LI $(DDSUBLINK spec/version, static-assert, $(D_KEYWORD static assert)))
+    $(LI $(DDSUBLINK spec/statement, mixin-statement,
```
MetaLang: These two cause a weird colouration on the link. Probably because one word is a keyword and the other is not. Not sure if it's fixable.

image

spec/function.dd
ZombineDev: In general, I think the more the better, but of course in reasonable manner. E.g. search for `GLINK` in some of Walter's recent PRs to the spec (https://github.com/dlang/dlang.org/pull/2113/files, https://github.com/dlang/dlang.org/pull/2104/files).

spec/function.dd
ZombineDev: Yep, it seems like it would be better to remove "statement" from the `mixin` and `pragma` list items.

ZombineDev: Perhaps @andralex can lend his typographical expertise?

spec/function.dd
ZombineDev: It would be execution if the function was compiled to a shared object / dll via the standard compiler pipeline and then called through a function pointer. Execution implies that you're able to observe platform implementation-defined behavior and even perform timing attacks along the lines of [Meltdown and Spectre](https://meltdownattack.com/). This is far from the intention and reality as the CTFE interpreter shields the code from such platform-specific side-channel information. VMs like CLR, JVM and V8 expose much more information in comparison.

aG0aep6G: I suggest we leave it as "execution" for now. I'd prefer "evaluation" myself, but it's been "execution" in the spec for a long time. So this is also about continuity and about avoiding unnecessary friction. We can always go for "evaluation" in another, dedicated PR.

spec/function.dd
aG0aep6G: I've changed it to backticks instead of `$(D_KEYWORD ...)`. No more blue:

spectacle sm5509

ZombineDev:
Thanks for working on this @aG0aep6G!

PR #2113, opened 2018-01-21
module.dd: fix a number of minor issues

WalterBright:


WalterBright:
This time the CyberShadow failure is not an H3 problem, because there aren't any H3s. But it's another issue where the error message is as unhelpful as a seg fault.

PR #2117, opened 2018-01-23
Correct unittest point: with -> without

mathias-baumann-sociomantic:
I suspect that's what was intended?

### DIFF
--- a/spec/betterc.dd
+++ b/spec/betterc.dd
@@ -50,7 +50,7 @@ $(OL
     $(LI `synchronized` and $(MREF core, sync))
     $(LI Static module constructors or deconstructors)
     $(LI Struct deconstructors)
-    $(LI `unittest` (testing can be as usual with the `-betterC` flag))
+    $(LI `unittest` (testing can be done without the `-betterC` flag))
 )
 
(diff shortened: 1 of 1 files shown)


schveiguy:
The grammar doesn't make sense, even before the change. I'm not sure what the statement in parentheses is trying to say. Can you explain what it means?

schveiguy:
ping @wilzbach as the original author.

schveiguy:
OK, I think it came from this comment: https://github.com/dlang/dlang.org/pull/1796#discussion_r125882434

So yes, the intent is to say, you can unit test your betterC functions by running without -betterC.

wilzbach:
> I suspect that's what was intended?

Yes, thanks.

>  Can you explain what it means?

The point was to say that `unittest` doesn't work with `-betterC`, but for testing purposes the program can simply compiled and run _without_ `-betterC`.

schveiguy:
Suggestion (correct grammar, and removing superfluous "as usual"): "testing can be done without the `-betterC` flag"

wilzbach:
> Suggestion (correct grammar, and removing superfluous "as usual"): "testing can be done without the -betterC flag"

I changed it to your suggestion. Thanks!

PR #2121, opened 2018-01-24
property.dd: minor fixes

WalterBright:
Finally removed obsolete references to `typedef`, which has been gone from the compiler for what, 10 years?

### DIFF
--- a/spec/property.dd
+++ b/spec/property.dd
@@ -4,7 +4,7 @@ $(SPEC_S Properties,
 
 $(HEADERNAV_TOC)
 
-        $(P Every type and expression has properties that can be queried:)
+        $(P Every symbol,type and expression has properties that can be queried:)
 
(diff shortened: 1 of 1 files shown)


spec/property.dd
```diff
@@ -4,7 +4,7 @@ $(SPEC_S Properties,
 
 $(HEADERNAV_TOC)
 
-        $(P Every type and expression has properties that can be queried:)
+        $(P Every symbol,type and expression has properties that can be queried:)
```
JackStouffer: space after comma

andralex: @JackStouffer do you have rights to edit yet?

JackStouffer: @andralex I tried in the past to edit other people's PRs, but the problem I ran into is it seems to only work over `git://` and not `https://`, so while I'm at work the firewall stops me.

PR #2126, opened 2018-01-25
Parameter storage class in is an equivalent of const currently; see a…

carblue:
…lso discussion in issue #17928; may change later to const scope


MetaLang:
I'm gonna go out on a limb here and say that a decision is not needed from @andralex or @WalterBright. This change was deliberate on Walter's part and needs to be documented.

PR #2127, opened 2018-01-25
Ddoc: Auto-detected URLs should be wrapped in DDOC_LINK_AUTODETECT instead of LINK directly.

quickfur:
This is a followup to #1858


quickfur:
Compiler implementation: https://github.com/dlang/dmd/pull/7779

PR #2131, opened 2018-01-27
Update spec for multiple selective imports

wilzbach:
This is an attempt at updating the spec for the changes in https://github.com/dlang/dmd/pull/6589
I wasn't so sure about the grammar changes.
(`grammar.dd` would also need an update)

See also: https://dlang.org/changelog/pending.html#multi_module_selective_imports

### DIFF
--- a/spec/module.dd
+++ b/spec/module.dd
@@ -408,11 +408,12 @@ all `import`s are selective, or when the imported module has a qualified name:)
 
 $(SPEC_RUNNABLE_EXAMPLE_COMPILE
 ---
-import std.stdio : writeln, std.file : write, std.ascii;
+import std.stdio : writeln, writefln, std.file : write, std.ascii;
 
 void main()
 {
     writeln("hello!");           // ok, writeln from std.stdio
+    writefln("%s!", "hello");    // ok, writefln from std.stdio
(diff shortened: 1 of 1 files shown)


spec/module.dd
```diff
@@ -407,6 +403,22 @@ void main()
 
 $(P $(D static) cannot be used with selective imports.)
 
+$(P Selective `import`s can be made from multiple modules if
+all `import`s are selective, or when the imported module has a qualified name:)
+
+$(SPEC_RUNNABLE_EXAMPLE_COMPILE
+---
+import std.stdio : writeln, std.file : write, std.ascii;
```
quickfur: What about an example that has multiple symbols per module? E.g.,
````
import std.stdio : writeln, writefln, std.file : write;
````

quickfur: (TBH I don't really like this syntax, because it mixes module names and symbols and is harder to read. But since it's in the spec, it should be illustrated properly.)

wilzbach: Ok. done

MetaLang:
Andrei has approved these changes and it looks like Wilzbach has addressed all issues, so I'm gonna merge.

PR #2133, opened 2018-01-27
Change documentation style requirement to /++ +/.

jmdavis:
If we're going to require either `/++ +/` or `/** */`, `/++ +/` makes more
sense, because it allows for comments to be put into the documentation
if need be, whereas `/** */` does not, because it doesn't allow comments to
nest. And while comments would usually be in code examples inside a
documented unit test, sometimes we still need to put code examples
directly in the documentation.

I'm all for disallowing stars or pluses on the beginning of documentation lines (they're really annoying IMHO), but if we're going to require `/++ +/` or `/** */`, I very much would prefer `/++ +/`, and as the commit message points out, `/++ +/` is objectively better. In most cases, it doesn't matter, and it's personal preference, but in some cases, it really does matter, and in those cases, `/++ +/` works well, whereas `/** */` does not. I don't see any advantage to requiring `/** */`.

### DIFF
--- a/dstyle.dd
+++ b/dstyle.dd
@@ -404,7 +404,7 @@ enum myFancyConstant;
     $(LI Every public function should have a Ddoc description and documented
     `Params:` and `Returns:` sections (if applicable):)
 ---
-/**
+/++
 Checks whether a number is positive.
 `0` isn't considered as positive number.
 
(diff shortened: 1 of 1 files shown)


jmdavis:
Closing in favor of https://github.com/dlang/dlang.org/pull/2134.

PR #2138, opened 2018-01-29
arrays.dd: better wording

WalterBright:


spec/arrays.dd
```diff
@@ -24,11 +24,31 @@ int* p;
...
 
-        $(P These are simple pointers to data, analogous to C pointers.
+        $(P A pointer to type $(I T) has a value which is a reference (address) to another
+        object of type $(I T). It is commonly called a $(I pointer to T).
+        )
+
+        $(P If a pointer contains a $(I null) value, it is not pointing to a valid object.)
+
+        $(P A pointer to $(I T) must either contain a $(I null) value,
+        or point to a valid object of type $(I T).)
```
aG0aep6G: This can be interpreted to mean that even constructing an invalid pointer has undefined behavior, even if it's never dereferenced. I don't think that's the intended meaning.

Maybe: "When a pointer to $(I T) is being dereferenced, it must [...]."

spec/arrays.dd
```diff
@@ -24,11 +24,31 @@ int* p;
...
+
+        $(BEST_PRACTICE These are simple pointers to data.
         Pointers are provided for interfacing with C and for
         specialized systems work.
-        There
-        is no length associated with it, and so there is no way for the
+        There is no length associated with it, and so there is no way for the
         compiler or runtime to do bounds checking, etc., on it.
         Most conventional uses for pointers can be replaced with
         dynamic arrays, $(D out) and $(D ref) parameters,
```
quickfur: The elephant in the room is that `key in aa` for associative arrays returns a pointer. In that case, a pointer is not only necessary (indicates the key doesn't exist in the AA), but also perfectly safe (if it's not null, it always points to a valid value in the AA).  This text almost makes it sound like `key in aa` should be avoided, or is somehow "specialized systems work".

spec/arrays.dd
WalterBright: ok

spec/arrays.dd
WalterBright: Unfortunately, the `key in aa` is dangerous if not used carefully. For example, the value can be deleted from the aa and yet the pointer remains. It's a poor design.

quickfur: Fair enough.

spec/arrays.dd
```diff
@@ -43,26 +63,30 @@ int[3] s;
 ---------
 )
 
-        $(P These are analogous to C arrays. Static arrays are distinguished
-        by having a length fixed at compile time.
+        $(P Static arrays have a length fixed at compile time.
         )
 
         $(P The total size of a static array cannot exceed 16Mb.
```
JackStouffer: Isn't this a limitation imposed by the linker DMD uses and not necessarily a limitation of the language.

WalterBright: The maximum length set set by Optlink for Win32. It is still a fixed size, however, on all platforms.

PR #2143, opened 2018-01-29
struct.dd: improve wording

WalterBright:


spec/struct.dd
```diff
@@ -47,27 +38,64 @@ $(GNAME AggregateBody):
...
+    ))
+
+$(H2 $(LNAME2 POD, Plain Old Data))
+
+    $(P A struct or union is $(I Plain Old Data) (POD) if it meets the following criteria:)
+
+    $(OL
+    $(LI it is not nested)
+    $(LI it has no postblits, destructors, or assignment operators)
+    $(LI it has no `ref` fields or fields that are themselves non-POD)
```
ntrel, line 95: What's a `ref` field?

PR #2145, opened 2018-01-30
betterc.dd: expand and reword

WalterBright:


spec/betterc.dd
```diff
@@ -4,38 +4,88 @@ $(SPEC_S Better C,
...
-Additionally, this also makes embedding D libraries in larger projects easier by:
+    $(P D programs generally require:)
 
-$(OL
-    $(LI Simplifying the process of integration at the build-system level)
-    $(LI Removing the need to ensure that Druntime is properly initialized on calls to the library, when an initialization step is not performed before the library is used.)
-    $(LI Mixing memory management strategies (GC + manual memory management) can sometimes be tricky, hence removing D's GC from the equation may be a good solution)
-)
+    $(OL
+    $(LI The D runtime library (Phobos) to be linked in, because many features of
```
JinShil: I would prefer it if the runtime and Phobos were not conflated.

spec/betterc.dd
```diff
@@ -4,38 +4,88 @@ $(SPEC_S Better C,
...
+    the core language require runtime library support.)
+    $(LI The `main()` function to be written in D, to ensure that the required
+    runtime library support is properly initialized.)
+    )
```
wilzbach: A general remark: I always use indentation for OL/OL (and used it here), because as you have recently discovered yourself, it's quite tricky to keep track of all parentheses in Ddoc. Same motivation applies to indent the content in `$(P`.

spec/betterc.dd
WalterBright: druntime being a subset of Phobos is not normally visible to the user.

ZombineDev: Instead of potentially confusing the readers, we should have just added one more sentence to clarify the details. More information is generally better.

PR #2147, opened 2018-01-30
Add 'Run' + 'Explore' sections

wilzbach:
This content is important for people and currently very hard to reach
and know about.

Preview:

image

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -384,16 +384,16 @@ $(DIVC boxes,
             $(P Take the $(LINK2 https://tour.dlang.org,
                 Tour), explore
                 $(LINK2 $(ROOT_DIR)overview.html, major features) in D,
-                quick start with $(LINK2 $(ROOT_DIR)ctod.html, C) or
+                browse the $(LINK2 $(ROOT_DIR)comparison.html, quick overview),
+                start with $(LINK2 $(ROOT_DIR)ctod.html, C) or
                 $(LINK2 $(ROOT_DIR)cpptod.html, C++) background,
-                setup your $(LINK2 https://wiki.dlang.org/IDEs, IDE) or
-                $(LINK2 https://wiki.dlang.org/Editors, editor),
                 and ask questions in the
                 $(LINK2 https://forum.dlang.org/group/digitalmars.D.learn,
                     Learn forum).
             )
             $(P For a deeper dive into D
-                check out $(LINK2 https://wiki.dlang.org/Books, books) about D,
+                check out $(LINK2 https://wiki.dlang.org/Books, books) or
+                $(LINK2 https://wiki.dlang.org/Videos, videos)
                 such as Ali Çehreli's free book
                 $(LINK2 http://ddili.org/ders/d.en/index.html,
                 <cite>Programming in D</cite>).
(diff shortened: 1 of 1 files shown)


wilzbach:
CC @CyberShadow as this is a change to a non trivial change to the front page and I think an extra pairs of eyes won't hurt.

CyberShadow:
I think that if we are to link to DFmt/DScanner/DCD, then those pages should contain a list of editor/IDE integrations (or, if not the projects' READMEs, then a wiki page listing them). But DCD etc.'s READMEs really should contain or link to a setup guide.

wilzbach:
This came up today in [NG.learn](https://forum.dlang.org/post/vfxrjdxofwywdyyrbvxr@forum.dlang.org)

> It would be nice if there was a link to https://run.dlang.io/ on the dlang website - preferably under Resources drop-down menu or similar.
> Or is there somewhere and I am just missing it?
> I have bookmarked it for myself, but it would be better for newbies if it was also easily accessible from the website.

------------------

> I think that if we are to link to DFmt/DScanner/DCD, then those pages should contain a list of editor/IDE integrations (or, if not the projects' READMEs, then a wiki page listing them). But DCD etc.'s READMEs really should contain or link to a setup guide. 

Fair point.
Well, IDE and editor already include links to all existing plugins, so I went with the easy way out and removed the links to formatting, linting and completion:

image

andralex:
cc @mdparker @CyberShadow 

One further idea would be to add a third button beside "Edit" "Input" "Run" called e.g. "Open Online IDE" that opens run.dlang.io with the chosen code snippet.

wilzbach:
This came up in the NG again today: https://forum.dlang.org/post/czpwjosmmvnijzmkyexr@forum.dlang.org

CC @CyberShadow as this is blocked on your approval.

> I was wondering whether D also has a 'reliable' source code formatter.
(reliable being a key word there).

So a point could be made for adding the links back, but let's go step by step.

> One further idea would be to add a third button beside "Edit" "Input" "Run" called e.g. "Open Online IDE" that opens run.dlang.io with the chosen code snippet.

You mean like the button on the right for the snippets in Phobos?

image

Sure, this shouldn't be too difficult.

wilzbach:
> Sure, this shouldn't be too difficult.

https://github.com/dlang/dlang.org/pull/2239
https://github.com/dlang/dlang.org/pull/2240

wilzbach:
Preview:

image

wilzbach:
> One thing to keep in mind is that the more information is put on a web page, the harder it becomes to notice / find any information on it. Ideally, information would be placed there, where users are expected to look for it. The front page may not be that place.

Fair point. For comparison with other languages, I like how well structured https://www.python.org is.

> Perhaps promoting and expanding the "Tools" menu in the top navigation? While "Resources" does a sufficient job of vaguely yet encompassingly describing its menu contents, it doesn't exactly catch the eye.

Hmm, that's tricky because we can't arbitrarily add new menu items. I think we are already at the limit:

image
 
I would love something structured like the overview at the DWiki, but that's a lot more work than this PR and as everyone else seems to like the idea, I'm going ahead and merge this now.

wilzbach:
FYI DAutoTest gives a 500 for the preview:

image

http://dtest.dlang.io/artifact/website-6297ccb4876888b57018018760e9a202e3e5ed19-5813c0939931340a94d86a3d43f8df25/web/index.html

(the diff is working though)

(also I'm not sure why the std.exception changes pop up in the diff now, but I assume that's because an earlier build of master failed.)

CyberShadow:
> FYI DAutoTest gives a 500 for the preview:

Oops, fixed (stray `git cat-file` process from before a cache cleanup).

PR #2148, opened 2018-01-30
Mention sponsors + contributors on the front page

wilzbach:
image

CC @JackStouffer


wilzbach:
CC @andralex and @CyberShadow as this is a change to the front page

CyberShadow:
By the way, how does one pronounce "A huge ❤" aloud? "A huge heart"? Maybe "Lots of ❤" would be better? (I'm not very well-versed in Emoji, sorry!)

wilzbach:
> (I'm not very well-versed in Emoji, sorry!)

I'm neither - even though Emoji is [Oxford Dictionaries Word of the Year 2015](https://en.oxforddictionaries.com/word-of-the-year/word-of-the-year-2015).

> Maybe "Lots of ❤" would be better? 

Fair point. I like it too. Changed.

PR #2150, opened 2018-01-30
Disable building the man preview on OSX

wilzbach:
Apparentely the `man` tool behaves a bit differently on OSX.
As this is only used for previewing the man page changes at dmd (for now),
I think the best course of action is to simply disable this on OSX.

I don't use OSX, someone who does and wants to fix the underlying preview command (https://github.com/dlang/dmd/blob/master/docs/Makefile#L26), is of course always invited to do so.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -986,7 +986,7 @@ $W/$(MAN_PAGE): $(DMD_DIR)/generated/$(MAN_PAGE) | ${STABLE_DMD}
 	mkdir -p $(dir $@)
 	cp $< $@
 	# CircleCi + nightlies.dlang.org might not have `man` installed
-	if command -v man > /dev/null ; then \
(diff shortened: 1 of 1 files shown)


timotheecour:
@wilzbach does this help?

https://github.com/emacs-helm/helm/issues/1718 `man -l` bug keeps getting rediscovered #1718
```
If darwin, don't use -l
If man -l fails, retry with man
``

wilzbach:
Thanks a lot for the pointer! I wasn't aware of this before :/

Well, building the HTML of man page is only required for DAutoTest, so that changes to cli.d can be previewed nicely at DMD. It's absolutely not required nor even useful for people building dlang.org locally, so I think we are fine with this path for now.

Since, a couple of days, there's also a special DAutoTest target here, it could have been only be added to that one.

PR #2151, opened 2018-01-31
End the deprecation period of using the result of a comma expression

wilzbach:
See also: https://github.com/dlang/dmd/pull/7813


quickfur:
Finally! About time the comma operator died a long-overdue death!

MetaLang:
Hasn't this only been deprecated since mid-late 2016?

wilzbach:
> Hasn't this only been deprecated since mid-late 2016?

I thought that deprecations are for one generous year, but Martin recently reclarified this https://github.com/dlang/dmd/pull/6676#discussion_r164557024:

> We have a general schedule for deprecations (6 months), but deviate from that when it seems too disturbing.

JackStouffer:
We **really** need to make an official document standardizing the deprecation process. 

I think I'll write a short DIP; thoughts?

wilzbach:
> I think I'll write a short DIP; thoughts?

Yes, please :heart:

I assume you are aware of the recent discussion here: https://github.com/dlang/phobos/pull/6086#discussion_r164306440?

(As always: feel free to ping us on Slack if we can help.)

quickfur:
The DMD pull has merged.  This goes in now.

Yes!!!! Finally, the bad old comma operator is DEAD!

PR #2153, opened 2018-01-31
Group D Language Foundation pages

wilzbach:
@andralex has already approved the principal nature of these changes,
so just the execution/implementation needs to be reviewed.

Preview:

image


JackStouffer:
Shouldn't the navbar link take you to this new page then?

wilzbach:
> Shouldn't the navbar link take you to this new page then?

I was planning to do this in a follow-up PR as I'm planning, to move everything under a `foundation` folder, but sure I can also directly do it in this PR (and do the `mv` later).

Anything else blocking this?

CC @andralex

PR #2154, opened 2018-01-31
Always display the download icon

wilzbach:
Based on feedback from the NG:

>> https://dlang.org/install.html
> I've never seen that page. Would've helped me to see it earlier. The D download page should include a blurb with a link to that install page.
https://forum.dlang.org/post/zwizsobrbboibzqrussm@forum.dlang.org

From @quickfur:

> Wow.  I set out *deliberately* looking for that link, and couldn't find it until I looked at your screenshot. I definitely wouldn't have found it if I didn't even know it was there.
https://forum.dlang.org/post/mailman.2877.1517420581.9493.digitalmars-d@puremagic.com

The info buttons have been added in: https://github.com/dlang/dlang.org/pull/1697

And the install.sh page was added in: https://github.com/dlang/dlang.org/pull/1936

### DIFF
--- a/download.dd
+++ b/download.dd
@@ -269,12 +269,8 @@ Macros:
     )
     EXTRA_HEADERS=$(T style,
         .download_paragraph > span {
-            visibility: hidden;
             margin-left: 0.4em;
             font-size: 125%;
-        }
-        .download_paragraph:hover > span {
-            visibility: visible;
         }
         .download_paragraph a > i.fa {
             color: #999;


wilzbach:
Before:

image


After:

image

uvtc:
Oh my! All of those OS-specific icons have links associated with them, but only the little gray "(i)" ⓘ is clickable! I had no idea. I don't recognize that ⓘ as a link. Maybe it needs to be brighter (not gray) and/or enclosed in a rounded-corner box like the links to the os-specific DMD archives. It definitely should not only show up when hovering.

uvtc:
See also <https://issues.dlang.org/show_bug.cgi?id=18344>.

bachmeil:
I think it would be better to provide information about the script on the page, because someone new to D is unlikely to investigate the install script further (there's a lot of information on that page). I've been using D for years and didn't know the details of the install script until now. My recommendation is to put this text before the curl command:

> The install.sh script is the official D version manager for DMD, GDC and LDC on FreeBSD, Linux and macOS. It provides a convenient way to fetch and install a D compiler without requiring administrative privileges or external dependencies. For more information, see {link to https://dlang.org/install.html}.

PR #2156, opened 2018-02-01
const3.dd: start work on this one

WalterBright:
I moved the stuff about storage classes to the end, they really should be merged with https://dlang.org/spec/declaration.html#typequal_vs_storageclass


spec/const3.dd
```diff
@@ -4,209 +4,44 @@ $(D_S Type Qualifiers,
 
 $(HEADERNAV_TOC)
 
-        $(P Type qualifiers modify a type by applying a $(GLINK2 declaration, TypeCtor).
-        $(I TypeCtor)s are: $(D const), $(D immutable), $(D shared), and $(D inout).
-        Each applies transitively to all subtypes.
+        $(P Type qualifiers modify a type by applying a $(I type constructor).
+        $(GLINK2 declaration, TypeCtor)s are:
+        $(D const), $(D immutable), $(D shared), and $(D inout).
```
andralex: There are many type constructors in D (see definition at https://en.wikipedia.org/wiki/Type_constructor): the array syntax `T[]`, the pointer syntax `T*`, template instantiation `A!T` etc. are all type constructors. 

These are qualifiers (and indeed they are also type constructors), and the types resulting from the application are called qualified types.

WalterBright: you're right, I'll fix

andralex:
@WalterBright your changes overwrote my editorial changes, could you please check what your pushing process is? If you force push (with `-f`) into your branch, please change that to push without forcing. Then you'll get errors only if you need to pull first.

andralex:
For example I just created https://github.com/dlang/dlang.org/pull/2156/commits/5d7e5976dcf29ef0040fa3864c1ffbaed7d50569. Can you make sure it doesn't get overwritten. Thanks.

WalterBright:
@andralex How do I merge your changes into my git repository?

PR #2159, opened 2018-02-01
Build changelog menu dynamically (replace bash with D code)

wilzbach:
Now that we have the ddoc_preprocessor, we can slowly remove all these old things.
Advantage of auto-generating these files is that less unrelated changes will pop up on a PR.


CyberShadow:
What's "ZipException: no end record"? Another Dub failure mode?

wilzbach:
Yes, but we fixed that one quite recently: https://github.com/dlang/dub/pull/1339
That's also the reason for 2.078.2 ;-)

CyberShadow:
> Now that we have the ddoc_preprocessor, we can slowly remove all these old things.

Great!

Looks like it currently generates a broken link to 2.078.2_pre , is that intentional?

wilzbach:
> Looks like it currently generates a broken link to 2.078.2_pre , is that intentional?

Kinda - it mimicked the `update_nav.sh` script 1:1 and Martin typically hasn't run it for last beta releases. Though it has been done in the past and so I would put it in the "an additional step which has been accidentally forgotten but can't be forgotten anymore category".
Anyhow we fixed the prerelease pages to point to `changelog/<ver>.html` to have stable URLs a while ago.

While I updated the URL, I used it as an opportunity to improve the text for the preview item to:

image

Note that I also had to bump the DMD version on CircleCi - as interestingly the changelog is built by the release tool.

CyberShadow:
Thanks!

PR #2162, opened 2018-02-02
Checkout upstream/stable instead of LATEST

wilzbach:
From the private mail conversation:
>>    Hi all,
>>    It often happens that someone finds a typo in the docs or other Ddoc
    weirdness. At the moment, it takes about a month for this fix to
    reach dlang.org.
    We already have the version chooser with master/stable and the last
    releases at docarchives.dlang.io.
    So what do you think about building from upstream/stable instead of
    the latest release?
>> Any thoughts on this one? As mentioned it would be cool to see fixed typos going live immediately and not after one month.
> Sorry for the late answer. I think that's a fine idea. -- Andrei

This also brings it in more accordance to the version chooser:

image


posix.mak
```diff
@@ -388,10 +388,10 @@ ifeq (1,$(RELEASE))
 release : html dmd-release druntime-release phobos-release d-release.tag
 endif
 
-docs-latest: dmd-latest druntime-latest phobos-latest apidocs-latest
+docs-stable: dmd-stable druntime-stable phobos-stable apidocs-stable
```
wilzbach: AFAICT nothing outside of dlang.org is depending on this target.
All distros use the docs built in the release archive and the release archive uses it's own `release` build here.

andralex:
nice

CyberShadow:
Something has gone terribly wrong. Just a short excerpt from the VERY LARGE log file:

```
.generated/stable_dmd-2.077.1/dmd2/linux/bin64/dub(22666:4805)[error]: Declaration expected
```

wilzbach:
> Just a short excerpt from the VERY LARGE log file:

Ouch 143M log file??

Fixed & this is passing now, but it seems that the assert->rewrite pipeline isn't working correctly anymore :/

wilzbach:
>  but it seems that the assert->rewrite pipeline isn't working correctly anymore :/

Luckily I have been working on a separate PR that simplifies this pipeline: https://github.com/dlang/dlang.org/pull/2069 - this should help this PR too.

wilzbach:
Argh. I realized a big problem with this. We can't do this within the two weeks before release where `master` got merged into stable, but didn't get fully released yet.

We could check that with `https://raw.githubusercontent.com/dlang/dmd/stable/VERSION` == `git ls-remote --tags https://github.com/dlang/dmd | sed 's/.*refs\/tags\/\([a-z0-9.-]*\).*/\1/' | sort -r | head -n1` though ...

PR #2163, opened 2018-02-02
Issue 18351 - integrate dub changelog with changed.d tool

wilzbach:
This prefers dlang.org for cloning `dub` and adds it to `dautotest`, s.t. changes to DUB will be visible on the pending changelog.

See also: https://github.com/dlang/tools/pull/302

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -141,6 +141,7 @@ PHOBOS_DIR=../phobos
 DRUNTIME_DIR=../druntime
 TOOLS_DIR=../tools
 INSTALLER_DIR=../installer
+DUB_DIR=../dub
 
 include $(DMD_DIR)/src/osmodel.mak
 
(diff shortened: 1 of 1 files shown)


posix.mak
```diff
@@ -449,7 +450,7 @@ rsync : all kindle pdf
 rsync-only :
 	rsync -avzO --chmod=u=rwX,g=rwX,o=rX --delete $(RSYNC_FILTER) $W/ $(REMOTE_DIR)/
 
-dautotest: all verbatim pdf diffable-intermediaries
+dautotest: all verbatim pdf diffable-intermediaries ${DUB_DIR}
```
CyberShadow: Is this the right place to add the target? Wouldn't the target that builds the changelog be more suitable?

wilzbach: Ok

PR #2165, opened 2018-02-03
Fix Issue 16707 - [Templates] run variadic templates example failed

enckse:


spec/template.dd
```diff
@@ -708,6 +708,8 @@ $(GNAME TemplateSequenceParameter):
         another template, or as the list of parameters for a function.)
 
         ---
+        import std.stdio;
```
wilzbach: In my WIP conversion from the spec to a spec testsuite, I always add `std.stdio` automatically because it wasn't worth bothering and we do the same for Phobos too - it has the advantage that you can immediately of cluttering everything with `std.stdio` imports -> I'm not sure whether we really want to go this road here.

Also you want to wrap this in a block, s.t. the spec tester picks it up:

```
---

---
)
```

`make -f posix.mak test_dspec` (my progress on converting the docs has a bit stagnated, so not everything uses this macro)

PR #2166, opened 2018-02-03
Clean up line about leading stars in style guide.

jmdavis:
I'm not sure why it says something about no leading zeroes. It's true that there' shouldn't be any, but AFAIK, no one has been doing it, so it comes off as weird. It's stars that some folks like to do. IIRC, it's a javadoc thing.


wilzbach:
> I'm not sure why it says something about no leading zeroes.

Because you added it ;-)

https://github.com/dlang/dlang.org/pull/2134

jmdavis:
> Because you added it ;-)

Figures. I still have no idea why it's there. I was in a bit of a hurry when I created that PR and clearly wasn't careful enough.

PR #2170, opened 2018-02-05
Don't build the newest Phobos

wilzbach:
This was added with the dspec tester as still an old version of Phobos was used.
It has been bumped in the mean time, so let's see whether everything is still green.


wilzbach:
Hmm, looks like we can't avoid that. The spec tester needs to have the newest Phobos version.

PR #2172, opened 2018-02-05
Auto-generate a TOC for the compiler help pages

wilzbach:
We now can generate nice TOCs.

Before:

image

After:

image


wilzbach:
Argh - that's because the Ddoc preprocessor has zero understanding of the macros at the moment.
However, I doubt that adding the support for this is worth the pain. Sorry.

PR #2175, opened 2018-02-05
Check all files for undefined ddoc

wilzbach:
Extends the undefined ddoc check for all files and fixes the violations.

### DIFF
--- a/faq.dd
+++ b/faq.dd
@@ -225,7 +225,7 @@ $(ITEM q4, Why is printf in D?)
         is not part of D, it is part of C's standard
         runtime library which is accessible from D.
         D's standard runtime library has
-        $(LINK2 https://dlang.org/phobos/std_stdio.html#writefln, std.stdio.writefln),
+        $(REF writefln, std,stdio)
         which is as powerful as $(LINK2 http://www.digitalmars.com/rtl/stdio.html#printf, printf)
         but is much easier to use.
         )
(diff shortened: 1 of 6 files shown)


dlang.org.ddoc
```diff
@@ -155,6 +155,7 @@ DBLOG_LATEST_DATE2=
 DBLOG_LATEST_LINK2=https://dlang.org/blog
 DBLOG_LATEST_TITLE2=
 DMDSRC=$(HTTPS github.com/dlang/dmd/blob/master/src/dmd/$0, $0)
+DOLLAR=&dollar;
```
andralex: This should go in html.ddoc

dlang.org.ddoc
```diff
@@ -155,6 +155,7 @@ DBLOG_LATEST_DATE2=
 DBLOG_LATEST_LINK2=https://dlang.org/blog
```
andralex: Pedantically this should be `$(HTTPS dlang.org/blog)`. That way if the macro ever makes it in non-html documentation, there's no problem with it.

faq.dd
```diff
@@ -393,15 +393,15 @@ $(ITEM printf, How do I get printf() to work with strings?)
 $(CCODE
 char s[8];
 strcpy(s, "foo");
-printf("string = '$(B %s)'\n", s);
+printf("string = '%s'\n", s);
```
andralex: Custom escapes inside code are mightily difficult in latex, too. I'm very glad we're steering clear off them.

dlang.org.ddoc
wilzbach: Not related to this PR and it's only used for the front page. The reason it's defined here, is that DAutoTest doesn't error with undefined macros as these are dynamically injected.

dlang.org.ddoc
wilzbach: (done)

PR #2135, opened 2018-01-28
Automatically generate the grammar overview

wilzbach:
I guess that tells all?

>  61 additions and 1,851 deletions.

This isn't perfect yet, but already pretty cool.

TODO:
- [ ] https://github.com/dlang/dlang.org/pull/2137

There are 8 new modules which contained D grammar which have been forgotten.

PR #2177, opened 2018-02-05
Use the $(DDOC_BIN) preprocessor with --compiler"

wilzbach:
We always want to use the latest `$(DMD)` to build the documentation as
- this is an automatic test for the newest compiler binary
- we depend on same freshly fixed bugs (and probably we do so in the future) - one example is the fixed behavior of `-w` (https://github.com/dlang/dmd/pull/7440)

This makes it more explicit and also adds `$(DDOC_BIN)` to a couple of other instances as:
- this allows us to preview its output (verbatim)
- it avoids the problem in the future that we add more features to the preprocessor and they don't end up in the PDF specification build

(this might produce changes in the verbatim files)

### DIFF
--- a/ddoc_preprocessor.d
+++ b/ddoc_preprocessor.d
@@ -33,6 +33,7 @@ int main(string[] rootArgs)
     import std.getopt;
     auto helpInformation = getopt(
         rootArgs, std.getopt.config.passThrough,
+        std.getopt.config.required,
         "compiler", "Compiler to use", &config.dmdBinPath,
     );
     if (helpInformation.helpWanted)
--- a/posix.mak
+++ b/posix.mak
@@ -283,6 +283,7 @@ DDOC_VARS_PRERELEASE_VERBATIM=$(DDOC_VARS_PRERELEASE) \
 ################################################################################
 
 DDOC_BIN:=$G/ddoc
+DDOC_BIN_DMD:=$(DDOC_BIN) --compiler=$(DMD)
 
 ################################################################################
 # Resources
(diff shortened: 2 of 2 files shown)


ddoc_preprocessor.d
```diff
@@ -31,6 +31,7 @@ int main(string[] rootArgs)
     import std.getopt;
     auto helpInformation = getopt(
         rootArgs, std.getopt.config.passThrough,
+        std.getopt.config.required,
```
andralex: `config.required` suffices?

posix.mak
```diff
@@ -282,6 +282,7 @@ DDOC_VARS_PRERELEASE_VERBATIM=$(DDOC_VARS_PRERELEASE) \
 ################################################################################
 
 DDOC_BIN:=$G/ddoc
+DDOC_BIN_DMD:=$(DDOC_BIN) --compiler=$(DMD)
```
andralex: Not sure where `:=` became a thing in our makefiles. We've kept it for special cases so far.

wilzbach: Since we started to have ugly, hard-to-trace bugs due accidentally overwritten of variables, see e.g. https://github.com/dlang/dlang.org/pull/1953

wilzbach:
@ others. FYI: the diff is a bit larger because
- the verbatim files are now run through the preprocessor
- the dlangspec PDF and ebook files temporarily didn't run through the preprocessor (after the grammar section was automatically generated and this PR fixes this)

CyberShadow:
Something weird is happening in `dlangspec.html` now. Looks like the entire BNF grammar is getting dumped into the "Modules" section or something.

wilzbach:
> Looks like the entire BNF grammar is getting dumped into the "Modules" section or something.

It's the grammar file and expected (one of the reasons why I made this PR).
Background:

https://github.com/dlang/dlang.org/blob/master/spec/grammar.dd
https://github.com/dlang/dlang.org/pull/2135

(I accidentally forgot using the ddoc preprocessor for PDF + Ebook when I made the Grammar generating PR)

PR #2181, opened 2018-02-06
Use /bin/bash as SHELL for dlang.org to have more consistent behavior across platforms and distros

wilzbach:
To avoids things like https://github.com/dlang/dlang.org/pull/2180


wilzbach:
From:

https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html

> The program used as the shell is taken from the variable SHELL. If this variable is not set in your makefile, the program /bin/sh is used as the shell.
> Unlike most variables, the variable SHELL is never set from the environment.

PR #2182, opened 2018-02-06
Remove outdated rm blog.html in posix.mak

wilzbach:
This should hopefully fix our deployment issues.


wilzbach:
Re-pushed to get a higher priority.

PR #2183, opened 2018-02-06
Make Donate button responsive

wilzbach:
The first commit moves CSS styles affecting only the front page to `index.dd.`
This is done because:
- it keeps the main stylesheet thin and lean (easier for other sites to use it as base)
- it allows developers to know that all the changes they need to make are within the same file
- a step into only serving stylesheets needed for the current page to the user

We haven't done this before because Ddoc only allowed to do so with ugly workarounds (`padding: 0;` created a new section).
Though we have started following this pattern more and more over the last months and I used this opportunity to finally move a full chunk out of `style.css`

The second commit is the actual fix. Pretty simple.

CC @JackStouffer

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1353,196 +1353,6 @@ body.chm dt.d_decl
     width: 100%;
 }
 
-body#Home > .container
-{
-    max-width: none;
-    padding: 0;
-}
-
-body#Home #content #tools,
-body#Home #content > .intro
-{
-    background: #F5F5F5;
-}
-
-body#Home #content > h1
-{
-    margin: 0;
-}
-
-body#Home #content > .intro
-{
-    padding: 1em 0 1em;
-    border-bottom: 1px solid #E6E6E6;
-}
-
-body#Home #content > .intro > div > div
-{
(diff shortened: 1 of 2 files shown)


wilzbach:
Before:

image


After:

image

css/style.css
```diff
@@ -2229,47 +2005,6 @@ the home page, intro pitch and your-code-here layed out vertically */
...
-    .call_to_donate
-    {
-        padding: 40px 80px;
-    }
-}
-
-/* Narrow layout stage 3: no more two column boxes */
-@media only screen and (max-width: 40em)
-{
-    .boxes .row, .boxes .row .item
```
CyberShadow: These are also used by the foundation page.

wilzbach: Argh. Good catch! Thanks!

wilzbach:
@JackStouffer @CyberShadow (or others) - this is a serious UI regression on the mobile front page, so if you can give this some priority, this would be great!

JackStouffer:
Sorry, thought I approved this already.

PR #2184, opened 2018-02-06
Add contributor hints and explanation of pending.dd

wilzbach:
Follow-up to https://github.com/dlang/phobos/pull/6132#issuecomment-363557604

CC @schveiguy

### DIFF
--- a/changelog/pending.ddoc
+++ b/changelog/pending.ddoc
@@ -3,7 +3,33 @@ $(DIVC version,
 $(P
 $(B $(LARGE $(LINK2 http://nightlies.dlang.org, Download D nightlies)))$(BR)
 $(SMALL $1)
+
 )
+
+$(HR)
+
+$(P
+This changelog has been $(B automatically generated) from all commits in $(B $(CODE master))
+since the last release.
+)
+$(P
+    $(UL
+        $(LI The full-text messages are assembled from the $(CODE changelog/) directories of the respective repositories:
+            $(LINK2 https://github.com/dlang/dmd/tree/master/changelog, $(CODE dmd)),
+            $(LINK2 https://github.com/dlang/druntime/tree/master/changelog, $(CODE druntime)),
+            $(LINK2 https://github.com/dlang/phobos/tree/master/changelog, $(CODE phobos)),
+            $(LINK2 https://github.com/dlang/tools/tree/master/changelog, $(CODE tools)),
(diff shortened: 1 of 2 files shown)


CyberShadow:
Is this the only place with a description of the process? Could we not link to another place instead?

Seems like a lot of text that won't be interesting to D users...

wilzbach:
> Is this the only place with a description of the process? Could we not link to another place instead?

And that would be?

> Seems like a lot of text that won't be interesting to D users...

Agreed. OTOH it's the nightly changelog and only used by "hard-code" developers who will hopefully report regressions or typos.
I tried to reduce the text a bit, but maybe @schveiguy knows what exactly he would be looking for?

CyberShadow:
> And that would be? (I'm really open to ideas)

Maybe some place on the wiki? It has tons of pages of build instructions etc., does none of them mention the changelog?

schveiguy:
The issue I had was, when looking at this changelog (indeed only hard-code (is that a typo or is that a cool pun for hackers? 😆 ) users would find this), I clicked the "improve this page" link, and it was a 404.

I didn't know where the page came from.

I think it's reasonable to have this in here, especially since the improve link doesn't work.

However, I have a different suggestion: Make the "Compiler changes" etc. titles be a link to the github repository for that changelog. Or add a [(source)](https://github.com/dlang/dmd/tree/master/changelog) link right after the section title. You may even be able to do this on individual changelog entries.

wilzbach:
> Or add a (source) link right after the section title. You may even be able to do this on individual changelog entries.

Easy enough:

- https://github.com/dlang/tools/pull/314
- https://github.com/dlang/dlang.org/pull/2203

schveiguy:
I think we can close this, the buttons seem less intrusive.

wilzbach:
We could keep the first message?

> This changelog has been $(B automatically generated) from all commits in $(B $(CODE master))
since the last release.

schveiguy:
Hm.. maybe a link to the `changed.d` file that generates it would be useful.

wilzbach:
> Hm.. maybe a link to the changed.d file that generates it would be useful.

I think I had a better idea:

image

If someone really wants to dive into this, he can simply look at the the `pending_changelog` target which references `changed.d`, but imho just mentioning the Makefile target is still general-purpose.

Full preview:

image

PR #2185, opened 2018-02-06
DBlog now uses HTTPS by default

wilzbach:
Apparentely Jan changed the config today and now it redirects to HTTPS by default:

```
> curl http://blog.dlang.org
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>302 Found</title>
</head><body>
<h1>Found</h1>
<p>The document has moved <a href="https://blog.dlang.org/">here</a>.</p>
</body></html>
```

Our curl command here didn't use `-L`, hence this file was written and the following commands silently failed. I will work on making them fail louder, but for now we should pull this.
As an exception, I will merge this myself as it's a trivial change and we need to fix the front page ASAP.

edit: rebased to get the highest priority in the queue.


CyberShadow:
Thanks!

PR #2188, opened 2018-02-07
Escape commas in option flag arguments

marler8997:
I think this might address the issue with https://github.com/dlang/dmd/pull/7777 but I'm not 100% sure.


marler8997:
nice...was eerily similar to the new function I added :)

PR #2191, opened 2018-02-08
Fix REF1 to behave analogous to the other REF macros

wilzbach:
REF1 is currently only used once in entire DRuntime and Phobos, so maybe
a special OBJECTREF macro would have been better?
Anyhow, have a look at the only use of `REF1` to understand the
motivation for this PR:

https://dlang.org/phobos/std_file.html#.thisExePath

Also there are three other PRs being blocked on this:

- https://github.com/dlang/dmd/pull/7342
- https://github.com/dlang/druntime/pull/2082
- https://github.com/dlang/phobos/pull/6140


wilzbach:
Ping @CyberShadow or others - as the Phobos PR has already been merged, this leaves the current preview `std.process` in a suboptimal state, e.g.

https://dlang.org/phobos-prerelease/std_process.html#.ProcessPipes.stdin

(or check the diff)

PR #2193, opened 2018-02-08
Avoid duplicated SWNAMES in to auto-generated CLI output

wilzbach:
From https://github.com/dlang/dlang.org/pull/2068#issuecomment-363652740

Not so sure whether's really worth bother though.

### DIFF
--- a/ddoc_preprocessor.d
+++ b/ddoc_preprocessor.d
@@ -286,6 +286,9 @@ auto genSwitches(string fileText)
 {
     enum ddocKey = "$(CLI_SWITCHES";
     auto content = ddocKey ~ "\n";
+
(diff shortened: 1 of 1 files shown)


CyberShadow:
Thanks. Considering this is one step closer to generating pedantically correct HTML, I think it is.

wilzbach:
And it's also closer to the non-generated version...

CyberShadow:
(An alternative to not emitting them would be to make them unique)

wilzbach:
(Yes, but then I would have needed to modify the SWNAME macro - not impossible, but resulting in a larger diff. I was torn between both approaches and ultimately went with this one because of the smaller diff and being closer to the "original")

PR #2195, opened 2018-02-09
Remove PHOBOS macro in favor in REF

wilzbach:
Seems like the `PHOBOS` macro was a hack to get references to the standard library.
We should really use the universal `REF` here.

Note - this depends on https://github.com/dlang/dlang.org/pull/2194

edit: this PR has been merged -> this should be good to go.

### DIFF
--- a/dlang.org.ddoc
+++ b/dlang.org.ddoc
@@ -338,7 +338,6 @@ _=
 PC=$(TC p, $1, $+)
 _=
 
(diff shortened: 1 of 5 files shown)


wilzbach:
CC @CyberShadow or @ZombineDev - this has been pending for quite a while now (> one month). It's pretty straight-forward...

CyberShadow:
Thanks for the ping :)

PR #2198, opened 2018-02-09
Add DConf to the D Language Foundation menu

wilzbach:


dlang.org.ddoc
```diff
@@ -268,6 +268,7 @@ $(SUBMENU_MANUAL
     $(SUBMENU_LINK https://wiki.dlang.org/Get_involved, Get involved)
     $(SUBMENU_LINK_DIVIDER $(ROOT_DIR)foundation/contributors.html, Contributors)
     $(SUBMENU_LINK $(ROOT_DIR)foundation/index.html, Foundation)
+    $(SUBMENU_LINK $(ROOT_DIR)foundation/index.html, DConf)
```
JackStouffer: I would make more sense to have this go straight to the dconf homepage.

wilzbach: (I removed this entry)

PR #2199, opened 2018-02-10
Document __traits(isModule) and __traits(isPackage)

blm768:
This PR documents the language changes in https://github.com/dlang/dmd/pull/5290.

### DIFF
--- a/spec/traits.dd
+++ b/spec/traits.dd
@@ -44,6 +44,8 @@ $(GNAME TraitsKeyword):
     $(GLINK isLazy)
     $(GLINK isReturnOnStack)
     $(GLINK isZeroInit)
+    $(GLINK isModule)
+    $(GLINK isPackage)
     $(GLINK hasMember)
     $(GLINK identifier)
     $(GLINK getAliasThis)
@@ -593,6 +595,40 @@ $(SPEC_RUNNABLE_EXAMPLE_COMPILE
     $(LI Testing that the compiler does this correctly is normally hackish and awkward,
     this enables efficient, direct, and simple testing.)
     ))
+
+$(H2 $(GNAME isModule))
+
+        $(P Takes one argument. If that argument is a symbol that refers to a
+        $(DDLINK spec/module, module) then $(D true) is returned, otherwise $(D false).
+        $(DDSUBLINK spec/module, package-module, Package modules) are considered to be
+        modules even if they have not been directly imported as modules.
+        )
+
+---
+import std.algorithm.sorting;
+
+// A regular package (no package.d)
+static assert(!__traits(isModule, std));
+// A package module (has a package.d file)
+// Note that we haven't imported std.algorithm directly.
+// (In other words, we don't have an "import std.algorithm;" directive.)
+static assert(__traits(isModule, std.algorithm));
+// A regular module
+static assert(__traits(isModule, std.algorithm.sorting));
+---
+
+$(H2 $(GNAME isPackage))
+
+        $(P Takes one argument. If that argument is a symbol that refers to a
+        $(DDLINK spec/module, package) then $(D true) is returned, otherwise $(D false).
+        )
+
+---
+import std.algorithm.sorting;
+static assert(__traits(isPackage, std));
+static assert(__traits(isPackage, std.algorithm));
+static assert(!__traits(isPackage, std.algorithm.sorting));
+---
 
 $(H2 $(GNAME hasMember))
 
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -2048,6 +2048,8 @@ $(GNAME TypeSpecialization):
     $(D shared)
     $(D return)
     $(D __parameters)
+    $(D module)
+    $(D package)
 )
 
     $(P $(I IsExpression)s are evaluated at compile time and are
@@ -2132,11 +2134,15 @@ void foo()
                 $(D enum)
                 $(D function)
                 $(D delegate)
-           $(D const)
+                $(D const)
                 $(D immutable)
                 $(D shared)
+                $(D module)
+                $(D package)
 
-        then the condition is satisfied if $(I Type) is one of those.
+        then the condition is satisfied if $(I Type) is one of those. $(DDSUBLINK
+        spec/module, package-module, Package modules) are considered to be both
+        packages and modules.
 
 -------------
 alias Bar = short;


JinShil:
I noticed when review the DMD PR that if a module is not imported, that `__traits(isModule)` can still resolve it.  That's a pretty important condition, IMO, and I think it should be spelled out in the the specification.

wilzbach:
How about defining more explicitly what a module symbol and package symbol is?
The examples show it a bit, but I think the specification should be a bit more specific than just relying on examples.

blm768:
@wilzbach I've made the wording a bit less ambiguous, but let me know if you think we should expand more on what's there.

@JinShil Does that apply just to modules that have been imported during the current compiler run but not in the current module? I thought that the logic in `isPkgX` (`traits.d`, line 527) wouldn't return `true` for symbols that have never been imported, but it's been two years since I first wrote the PR, so my mental model could be out of date. Or are you referring to how `Package.resolvePKGUnknown` will check for a `package.d` even when the package has never been imported as a module?

spec/traits.dd
```diff
@@ -411,6 +413,33 @@ static assert(!__traits(isTemplate,foo!int()));
...
+import std.algorithm.sorting;
+static assert(!__traits(isModule, std));
+static assert(__traits(isModule, std.algorithm));
+static assert(__traits(isModule, std.algorithm.sorting));
+---
+
+$(H2 $(GNAME isPackage))
+
+        $(P Takes one argument. If that argument is a symbol that refers to a package
+        then $(D true) is returned, otherwise $(D false).
```
wilzbach: I would add another sentence here  - defining what a package is. Maybe sth. like this?

>  A package is a grouping of modules. `package.d` modules are considered as a package, however the existence of a `package.d` file for a package isn't required.

spec/traits.dd
```diff
@@ -411,6 +413,33 @@ static assert(!__traits(isTemplate,foo!int()));
...
+        )
+
+---
+import std.algorithm.sorting;
+static assert(__traits(isPackage, std));
+static assert(__traits(isPackage, std.algorithm));
+static assert(!__traits(isPackage, std.algorithm.sorting));
```
wilzbach: How about adding short comments for people not so familiar with the layout of `std.algorithm`?

```
static assert(__traits(isPackage, std)); // no package.d
static assert(__traits(isPackage, std.algorithm)); package.d
static assert(!__traits(isPackage, std.algorithm.sorting)); // module file
```

(similar suggestion for the example above)

JinShil:
> Does that apply just to modules that have been imported during the current compiler run but not in the current module? I thought that the logic in `isPkgX` (traits.d, line 527) wouldn't return `true` for symbols that have never been imported, but it's been two years since I first wrote the PR, so my mental model could be out of date. Or are you referring to how `Package.resolvePKGUnknown` will check for a package.d even when the package has never been imported as a module?

I'm referring to this: https://github.com/blm768/dmd/blob/381813d646c23cabbb043a0f1b722d707a232f21/test/runnable/traits.d#L629-L630

Just ensure the documentation clearly explains how the implementation works so the reader can accurately predict the behavior simply reading the documentation, without the need to write a test to figure it out.

spec/traits.dd
```diff
@@ -411,6 +413,40 @@ static assert(!__traits(isTemplate,foo!int()));
 static assert(!__traits(isTemplate,"string"));
 ---
 
+$(H2 $(GNAME isModule))
+
+        $(P Takes one argument. If that argument is a symbol that refers to a
+        $(DLINK spec/module, module) then $(D true) is returned, otherwise $(D false).
+        $(DDSUBLINK spec/module, package-module, Package modules) are considered to be
+        modules even if they have not been directly imported as modules.
+        )
```
wilzbach:
```
/dev/shm/dtest/work/repo/dlang.org/web/spec/traits.html:592: [UNDEFINED MACRO]         <!--UNDEFINED MACRO: "DLINK"--> then <span class="d_inlinecode donthyphenate notranslate">true</span> is returned, otherwise <span class="d_inlinecode donthyphenate notranslate">false</span>.
/dev/shm/dtest/work/repo/dlang.org/web/spec/traits.html:612: [UNDEFINED MACRO]         <!--UNDEFINED MACRO: "DLINK"--> then <span class="d_inlinecode donthyphenate notranslate">true</span> is returned, otherwise <span class="d_inlinecode donthyphenate notranslate">false</span>.
/dev/shm/dtest/work/repo/dlang.org/web/dlangspec.html:20849: [UNDEFINED MACRO]         <!--UNDEFINED MACRO: "DLINK"--> then <span class="d_inlinecode donthyphenate notranslate">true</span> is returned, otherwise <span class="d_inlinecode donthyphenate notranslate">false</span>.
/dev/shm/dtest/work/repo/dlang.org/web/dlangspec.html:20869: [UNDEFINED MACRO]         <!--UNDEFINED MACRO: "DLINK"--> then <span class="d_inlinecode donthyphenate notranslate">true</span> is returned, otherwise <span class="d_inlinecode donthyphenate notranslate">false</span>.
4 errors found. Exiting.
```

Things can use:

- `$(RELATIVE_LINK2 module, mytext)`
- `$(DDSUBLINK spec/module, module_declaration, mytext)`

blm768:
@wilzbach Do you think linking to the module documentation provides enough explanation of packages? It seems a little less redundant than fully explaining them in `traits.dd`, but they're not terribly prominent in `module.dd`, either.

@JinShil I'd almost forgotten that comment was there. I think the note I've added for `isModule` should cover that case now.

JinShil:
What about this?

```D
import std.algorithm;

void fun()
{
    // `std.stdio` was not imported into this file, but it is in the import path.
    // Does this return `true` or `false`?
    static assert(!__traits(isModule, std.stdio));
}
```

adamdruppe:
On Wed, Feb 14, 2018 at 11:47:51PM +0000, Mike Franklin wrote:
>     // `std.stdio` was not imported into this file, but it is in the import path.
>     // Does this return `true` or `false`?
>     static assert(!__traits(isModule, std.stdio));

so I haven't been following this discussion, but I think the answer is "neither"... it should be a compile error, undefined symbol, std.stdio.

I had an argument on IRC yesterday about is() (well, not so much argument, more like "violent agreement") and how it swallows errors and that causes a lot of problems for D's users because it returns false on typos when it should just say "T is undefined".

We might or might not change is(), but we should definitely not introduce more swallowed errors where we don't need to. The isModule trait should require an in-scope name already to even be called, otherwise it should be an error.

And since std.stdio is not imported in this module, the name is not in scope.

blm768:
The intended behavior is for undefined symbols to produce an error, and that's the result which a quick test on my machine gives. There's some special logic to check whether a package (which will be in scope but might not be directly imported) has a `package.d`, but otherwise the code should fall back on the normal symbol resolution mechanics.

FeepingCreature:
Ffuck.

Sure would have been nice to have this feature merged just now.

thewilsonator:
Ping me when this is good.

blm768:
@thewilsonator I think I've covered pretty much everything that's relevant now. Do you think it's worth having any cross-linking between the `__traits` and `is()` entries?

thewilsonator:
>  Do you think it's worth having any cross-linking between the __traits and is() entries?

Possibly? I haven't really thought about it. Regardless, it need't hold up merging this. Feel free to do it in another PR if you think it would be useful.

PR #2200, opened 2018-02-10
[unbreak nightlies] Run the full chmgen script to avoid regressions

wilzbach:
The nightlies are currently broken - see http://nightlies.dlang.org/dmd-master-2018-02-10/build.html

The problem seems to be that when we added the _NIGHTLY to the changelog navigation this broke the win32 chm navigation build, because of undefined macros in the pseudo-JSON.

The fix is trivial and I also think that it's a good idea to run the full CHMNavigation script on CircleCi to prevent any regressions.

CC @CyberShadow

@MartinNowak - any chance we can make the nightlies build failures more louder? Email? Simple simple `curl -X` to Slack?


wilzbach:
Quick ping on this. Anyone around for a quick merge? I think it would be great to get the nightlies working again ;-)

andralex:
Should we add an "URGENT" label?

PR #2201, opened 2018-02-10
Title font shrinks if too long

jercaianu:
Some modules from phobos, more precisely the allocators, had very long names.
All titles now have the separate class "d_title" and if the length exceeds a threshold, the font size shrinks.

### DIFF
--- a/css/html-skeleton.css
+++ b/css/html-skeleton.css
@@ -85,6 +85,16 @@ parameters will be rendered in this style everywhere.*/
 {
 }
 
+/* SPAN: D module title */
+.d_title
+{
+}
+
+/* SPAN: link to Phobos resource */
+.phobos_src
+{
+}
+
 /* DIV: Formats the "Date" section */
 .Date
 {
--- a/css/style.css
+++ b/css/style.css
@@ -1177,6 +1177,24 @@ td .d_code2, td .cppcode2, td .cppcode
     font-weight: bold;
 }
 
+.d_title
+{
+    font-weight: bold;
+    text-overflow: clip;
+    overflow: hidden;
+    display: block;
+    word-wrap: break-word;
+    font-family: Consolas, "Bitstream Vera Sans Mono", "Andale Mono", Monaco, "DejaVu Sans Mono", "Lucida Console", monospace;
+}
+
+.phobos_src
+{
+    text-overflow: clip;
+    overflow: hidden;
+    display: block;
+    word-wrap: break-word;
+}
+
 /* syntax highlighting
 
 d_* classes come from Ddoc.
--- a/html.ddoc
+++ b/html.ddoc
@@ -95,6 +95,7 @@ D_STRING  = $(SPANC d_string, $0)
 D_KEYWORD = $(SPANC d_keyword, $0)
 D_PSYMBOL = $(SPANC d_psymbol, $0)
 D_PARAM   = $(SPANC d_param, $0)
+D_TITLE   = $(SPANC d_title, $0)
 _=
 
 _=Main entry point
--- a/std.ddoc
+++ b/std.ddoc
@@ -8,7 +8,7 @@ LAYOUT_SUFFIX =
 $(SCRIPTLOAD ../js/listanchors.js)
 $(SCRIPTLOAD ../js/show_contributors.js)
 $(SCRIPT jQuery(document).ready(listanchors);)
-LAYOUT_TITLE=$(H1 $(D $(TITLE)))
+LAYOUT_TITLE=$(H1 $(D_TITLE $(TITLE)))
 _=
 
 LREF = <a href="#$1">$(D $1)</a>


wilzbach:
I think we also need to think about responsiveness into this:

image

How about simply cutting the module name for now?

```css
.d_title
{
...
    text-overflow: ellipsis;
    overflow: hidden;
    display: block;
}
```

jercaianu:
@wilzbach Thanks for taking a look!
Which browser did you use for the screenshot that you posted?

On shrinking, on safari for example it looks like this.
screen shot 2018-02-11 at 11 39 54 am

wilzbach:
Chrome

jercaianu:
@wilzbach 
Thanks, I tried on Chrome, but could not reproduce. The title does not go outside the window in my case, it just wraps around.
How did you perform the resizing?

wilzbach:
With the mobile device preview tool.

I just tested and it's the same for Chromium:

image

BTW what do you think of the CSS cropping I suggested? This would fix this issue ;-)

jercaianu:
@wilzbach 
I couldn't reproduce on my phone and machine with various browsers.
I changed the CSS as you suggested and it still works fine for me.

Since I can't reproduce the issue, can you please check again with my latest changes to see if it looks better?
Thanks!

wilzbach:
> Since I can't reproduce the issue, can you please check again with my latest changes to see if it looks better?

Well, I did test my proposed CSS changes for you (no worries), so it looks as expected:

image

BTW to reproduce what the behavior you want to have you can use `word-wrap: break-word;`

image

@CyberShadow @jercaianu (or others) what version do you like more? I am - by no means - a UX person.

CyberShadow:
Ideally the ellipsis would be on the left. Maybe something like this? http://jsfiddle.net/j08691/ZfbaD/2/

jercaianu:
@CyberShadow @wilzbach 
Thanks for the suggestion.
I tried the ellipsis on left, and it works well on Opera for example, but Safari does not support this feature. Also, I could not make right side ellipsis work with the <span> tag, only <p>, as in your example.

Personally, I think the wrap around is the best choice, because if we hide the title by means of ellipsis, it would be difficult for the user to see the full module path.

wilzbach:
Two unrelated comments

1) It looks like we need to do the same for the source link

image

2) It looks like we need to fix DDox too

https://dlang.org/library-prerelease/std/experimental/allocator/building_blocks/ascending_page_allocator.html

See e.g. https://github.com/dlang/dlang.org/blob/master/dpl-docs/views/layout.dt#L202

--------

> but Safari does not support this feature. 

Ugh, well we could also do something like this:

```css
_::-webkit-full-page-media, _:future, :root .safari_only {
 ... do safari hacks here...
}
```

jercaianu:
@wilzbach 
I added wrap around for both the title and phobos src link.
Does it look fine to you?
Thanks!

wilzbach:
Yes looks. Thanks a lot! Great work!

image

CyberShadow:
Why the switch from monospace to variable-width?

jercaianu:
@CyberShadow 
I added a new css class for the title which caused the title to change. Will fix

jercaianu:
@CyberShadow @wilzbach 
I readded the monospace font.
If the title is too long, I adjusted the font size to linearly decrease based on the length.

Can you please check if everything is alright?
Thanks

CyberShadow:
Allowing the line to break and fixing overlapping HTML DOM elements: :+1:

Using Javascript to change the font size depending on its length: not so sure. Some small issues:
- Using JavaScript means it's done after load, i.e. it causes a relayout. This is a little annoying (text moves around on the page after you started reading it) and a has a small negative performance impact.
- In extreme cases the font size is almost comically small - it's smaller than the body of the page.

There is definitely a lot we could improve here (i.e. avoid breaking in the middle of a package name either by adding word-wrap points or using a completely different way to present the hierarchy), depending on how much time we want to spend on this.

jercaianu:
@CyberShadow 
Another option would be to put the version-changer button below the title (for long titles) and let the title wrap around.

wilzbach:
> Using JavaScript means it's done after load, i.e. it causes a relayout. This is a little annoying (text moves around on the page after you started reading it) and a has a small negative performance impact.

The same could be said about the nice FontAwesome icons and we still use them.

Imho, this is a lot better than the current status quo and perfect shouldn't be the enemy of progress.
We can always redefine the approach if it really becomes a problem, but imho there are many other, more pressing issues to be dealt with.

> In extreme cases the font size is almost comically small - it's smaller than the body of the page.

The font-size is already clamped to a minimum of its current size:

```diff
+                if (fontSize < 0.5)
+                    fontSize = 0.5;
+                elem.css('font-size', fontSize + 'em');
```

CyberShadow:
> The same could be said about the nice FontAwesome icons and we still use them.

- They don't require JavaScript. There is a relayout when the font is loaded for the first time, but browsers cache it. That doesn't happen with JavaScript.
- The argument is fallacious. Just because X is bad but we already do X somewhere, doesn't make it OK to keep doing X everywhere.

CyberShadow:
I'd merge the CSS part. I think adding JavaScript needs a stronger justification.

wilzbach:
> They don't require JavaScript. There is a relayout when the font is loaded for the first time, but browsers cache it. That doesn't happen with JavaScript.

... but for example the version chooser is added via JavaScript too (and there are lots of other things we do with JS because Ddoc just doesn't cut it)

> I'd merge the CSS part. I think adding JavaScript needs a stronger justification.

Fair enough, but do you have a better suggestion?
FWIW there's a lot we could learn from Adam's dpldpocs:

http://dpldocs.info/experimental-docs/std.algorithm.searching.find.2.html

image

CyberShadow:
> Fair enough, but do you have a better suggestion?

Yes, see my post from 3 days ago.

I suggest to just let it wrap for now. Inconsistently changing presentation to fit arbitrary content-based metrics is a weird enough idea of its own.

wilzbach:
Okay I squash all commits into one and removed the JS function, s.t. we can move forward here.
For reference, the patch is:

```patch
diff --git a/js/dlang.js b/js/dlang.js
index 2ff4b721..320da614 100644
--- a/js/dlang.js
+++ b/js/dlang.js
@@ -74,5 +74,16 @@
                 elem.show();
             }
         });
+
+        $('.d_title').each(function() {
+            var elem = $(this);
+            var textLength = elem.html().length;
+            if (textLength > 30) {
+                var fontSize = -0.0135 * textLength + 1.375;
+                if (fontSize < 0.5)
+                    fontSize = 0.5;
+                elem.css('font-size', fontSize + 'em');
+            }
+        });
     });
 })(jQuery);
```

CyberShadow:
Thanks!

jercaianu:
@wilzbach 
Thanks!

PR #2203, opened 2018-02-11
[preparation] add an edit button to changelog entries

wilzbach:
Required for https://github.com/dlang/tools/pull/314

Pretty stupidly simple for now:

image

(if wanted the CSS can always be tweaked later, but for now these Ddoc classes are required for the tools PR)


schveiguy:
OK, so this should have no real effect (yet). The diff just shows an added css style. Seems fine, can be reversed if for some reason we find it's a problem.

wilzbach:
Yeah, the real effect comes into play in https://github.com/dlang/tools/pull/314

PR #2207, opened 2018-02-13
cent and ucent missing in the list BasicTypeX

schlupa:
Even if these types are not implemented, they are reserved and they should therefore be listed. They are listed on the page but not in the grammar.


MetaLang:
They're not rejected by the lexer so they should be included in the grammar spec.

PR #2209, opened 2018-02-15
Clarify behavior of debug statements

WalterBright:
See https://github.com/dlang/dmd/pull/7882 and https://issues.dlang.org/show_bug.cgi?id=18407 for discussion.


wilzbach:
Yes, https://github.com/dlang/dlang.org/pull/2209.

PR #2210, opened 2018-02-15
tips and tricks: document easy way to get compiler version and importPaths

timotheecour:
ppl may not be aware of such a way to get this on cmd line

### DIFF
--- a/dcompiler.dd
+++ b/dcompiler.dd
@@ -18,6 +18,7 @@ $(D_S dmd - $(WINDOWS Windows)$(LINUX Linux)$(OSX Mac OS X)$(FREEBSD FreeBSD) D 
     $(LI $(RELATIVE_LINK2 library, Building Libraries))
     $(LI $(RELATIVE_LINK2 compiling-dmd, Compiling dmd))
     $(LI $(RELATIVE_LINK2 compiling-phobos, Compiling Phobos))
+    $(LI $(RELATIVE_LINK2 tips-tricks, Tips and Tricks))
     )
 
 $(H2 $(LNAME2 requirements, Requirements and Downloads))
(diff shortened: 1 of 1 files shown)


dcompiler.dd
```diff
@@ -864,6 +865,23 @@ make -f posix.mak DMD=~/$(DMDDIR)/$(OS)/bin/dmd
 )
 
 
+$(H2 $(LEGACY_LNAME2 tips_tricks, tips-tricks, Tips and Tricks))
+
+    $(P Get compiler info on command line (posix):)
+
+
+$(UNIX
```
wilzbach: Please but the entire section and menu item in the `$(UNIX)` tag. Otherwise it will be empty and look weird for people on Windows.

dcompiler.dd
```diff
@@ -864,6 +865,23 @@ make -f posix.mak DMD=~/$(DMDDIR)/$(OS)/bin/dmd
 )
 
 
+$(H2 $(LEGACY_LNAME2 tips_tricks, tips-tricks, Tips and Tricks))
```
wilzbach: You don't need to start new things with legacy support ;-)
Just `$(H2 $(LNAME2 tips-tricks, Tips and Tricks))` should be fine

timotheecour:
> people might be unfamiliar with jq and then you would have to explain it which isn't really the job of the CLI documentation page.

that's exactly why i was mentioning it, to avoid ppl to resort to grep when the right tool for the job is something that understands json (like jq); allows for quick debugging on cmd line; but maybe i can also write smthg like: optionally, use a program like jq to quickly get field u need, eg : ...

i'll address ur other points if we want to proceed with this, otherwise happy to close

wilzbach:
Looks like no one is interested in adding this, so I guess we have to search for alternative places. Content on the wiki is pretty though to find, so maybe we could bootstrap a compiler-specific DTour like tour.dlang.io/tour/mir - this could potentially also include examples on how to get started with dmd development.
Has anyone a better idea?

PR #2087, opened 2018-01-15
Add d-prerelease.tag to DAutoTest

wilzbach:
I realized in https://github.com/dlang/dlang.org/pull/1982 that we don't run
`chmgen.d` on DAutoTest.

@CyberShadow: IIRC you were planning to use `chmgen` to check for dead, internal links?

PR #2212, opened 2018-02-17
remove chmgen (d.chm, dman)

MartinNowak:
- see https://github.com/dlang/installer/pull/298
- too much maintenance for too little gain


CyberShadow:
We need chmgen to lint the generated HTML, so we should not remove it. See https://github.com/dlang/installer/pull/298#issuecomment-366448221.

MartinNowak:
> We need chmgen to lint the generated HTML, so we should not remove it. See dlang/installer#298 (comment).

Mmh OK, could you please turn it into a linter only tool then?
Stuff like this is really not helpful http://nightlies.dlang.org/dmd-master-2018-02-17/build.html.

CyberShadow:
> Mmh OK, could you please turn it into a linter only tool then?

Well, from where I'm standing, that seems like an over-reaction at this point.

> Stuff like this is really not helpful http://nightlies.dlang.org/dmd-master-2018-02-17/build.html.

Feel free to contact me directly when that happens. But that really should be caught by our CI.

MartinNowak:
> Well, from where I'm standing, that seems like an over-reaction at this point.

It's not, I've asked for this several times, and it remains a common nuisance. Again 4 complex issues last year, and 39 nightly build failures (partly b/c I cannot jump onto every failure). I've prolly spent 10 hrs. debugging and trying to fix chmgen last year.

MartinNowak:
Having it as a linter in out CI path seems useful and works b/c issues are catched before being merged (and come with a html diff). Running into d.chm build issues in a release is way too late.

CyberShadow:
> Again 4 complex issues last year, and 39 nightly build failures (partly b/c I cannot jump onto every failure).

What are they?

Please email me when something like this happens, I'll try to address it quickly. I would like to keep maintaining chmgen.

> I've prolly spent 10 hrs. debugging and trying to fix chmgen last year.

I don't understand how that can be. Looking at the commit log, you have made 1 commit involving chmgen.d, and I made 18. Did I miss something?

CyberShadow:
> Running into d.chm build issues in a release is way too late.

We don't test actual chm generation currently, because that can only be done on Windows.

We can test dman on any platform, though.

Do you think you could make it a soft failure, so that it doesn't block a release? Then, let me know and I'll try to fix it.

MartinNowak:
> Do you think you could make it a soft failure, so that it doesn't block a release? Then, let me know and I'll try to fix it.

Either we ship a tool or we don't, a soft failure doesn't make sense for releases. It's not the nightlies, but more the delays this introduces in our release schedule.

MartinNowak:
> > I've prolly spent 10 hrs. debugging and trying to fix chmgen last year.
> 
> I don't understand how that can be. Looking at the commit log, you have made 1 commit involving chmgen.d, and I made 18. Did I miss something?

Yes bug reports and reproducing, hence [search chmgen](https://forum.dlang.org/search?q=chmgen&scope=forum).
https://issues.dlang.org/show_bug.cgi?id=15864
https://issues.dlang.org/show_bug.cgi?id=16167
https://issues.dlang.org/show_bug.cgi?id=17731

MartinNowak:
The current issue again is caused by broken json generation which isn't CI covered.
https://issues.dlang.org/show_bug.cgi?id=16167

MartinNowak:
Changed this PR to only remove d.chm and dman from the release target.

CyberShadow:
> https://issues.dlang.org/show_bug.cgi?id=15864

Not a bug in chmgen... looks like good dogfooding. But if it wasn't caught by CI, then that is a problem.

> https://issues.dlang.org/show_bug.cgi?id=16167

That looks like the same as http://nightlies.dlang.org/dmd-master-2018-02-17/build.html. Yes, it looks like a regression in something surrounding chmgen. It should have been caught by CI.

> https://issues.dlang.org/show_bug.cgi?id=17731

Not a bug in chmgen, but I fixed that on the same day it was filed. I would also file that under dogfooding.

Sorry, but right now this still seems like an overreaction.

CyberShadow:
https://github.com/dlang/dlang.org/pull/2213

MartinNowak:
> Sorry, but right now this still seems like an overreaction.

Those tools (we're only talking about the d.chm and dman, not HTML linting), provide very little value. It's unreasonably that you and I spent so much effort on this. This is time that could go into improving/replacing Jenkins, improving code.dlang.org, reviewing&merging the MinGW toolchain, working on `@safe @nogc`.
But instead we work on obscure macro generators to output json files, that are used to index our website and an offline copy of it :o. If you feel like maintaining that, that's fine for me, but I cannot afford the time and certainly no longer want this to be in our release pipeline.

wilzbach:
I'm totally with Martin here, I also had to waste time with chmgen and as Martin I don't enjoy it either:

https://github.com/dlang/dlang.org/pull/2200

FWIW I even tried to trackle the problem before. See e.g. the still open https://github.com/dlang/dlang.org/pull/2087 from January which shows that the interest in this is little.

> Those tools (we're only talking about the d.chm and dman, not HTML linting),

Yeah the linter would very useful, but FWIW it's not active at the moment because last time I checked too many URLs were broken. (Until my hotfix in https://github.com/dlang/dlang.org/pull/2200 it wasn't even run on the CI).
Maybe we can tackle this with a blacklist, s.t. no one has to do the hard work of fixing all URLs at once, but dump the blacklist file somewhere which can then be reduced in individual PRs step by step?
At least this would prevent regressions.

win32.mak
```diff
@@ -1,46 +0,0 @@
-# Building dlang.org on Windows is no longer supported.
```
wilzbach: :+1: (this file isn't tested or run on a CI anyways)

wilzbach:
FWIW dman and d.chm have now been removed from the 2.079 release - see https://issues.dlang.org/show_bug.cgi?id=18610

And there hasn't been a big outcry yet though there was a bit of discussion on the NG one month ago: https://forum.dlang.org/post/mailman.3726.1518902610.9493.digitalmars-d@puremagic.com

We should have just added the removal to the 2.079 changelog so that people are aware of this.

PR #2213, opened 2018-02-17
posix.mak: Add DMan tags to dautotest target

CyberShadow:
Pull in chmgen and related rules to ensure they work correctly.

See https://github.com/dlang/dlang.org/pull/2212#issuecomment-366451751


MartinNowak:
Should we also check json validity or can we drop json generation?

CyberShadow:
> Should we also check json validity or can we drop json generation?

The JSON that is getting broken is consumed by chmgen, so by pulling in chmgen, that will be tested.

CyberShadow:
Looks like this is exposing a bug right now. I need to leave, but I will definitely look into it tonight or tomorrow.

MartinNowak:
> Looks like this is exposing a bug right now. I need to leave, but I will definitely look into it tonight or tomorrow.

No hurries, I'll disable the d.chm build for now so I can build the beta and we can decide on readding or removing later.

wilzbach:
FWIW I tried that before too: https://github.com/dlang/dlang.org/pull/2087

But I hadn't had time to dig into the failures :/

CyberShadow:
> FWIW I tried that before too: #2087

Sorry, I hadn't realized at the time the implications of the CI failures for that PR.

wilzbach:
> Sorry, I hadn't realized at the time the implications of the CI failures for that PR.

And I'm sorry that I didn't post it there (or directly pinged you).
Thanks for taking this over!

CyberShadow:
> No hurries, I'll disable the d.chm build for now so I can build the beta and we can decide on readding or removing later.

Ah, I didn't realize that it was blocking a release. With that in mind, closing your PRs was pretty rude of me, so I apologize for that as well.

posix.mak
```diff
@@ -455,7 +455,7 @@ rsync : all kindle pdf
 rsync-only :
 	rsync -avzO --chmod=u=rwX,g=rwX,o=rX --delete $(RSYNC_FILTER) $W/ $(REMOTE_DIR)/
 
-dautotest: all verbatim pdf diffable-intermediaries
+dautotest: all verbatim pdf diffable-intermediaries d-latest.tag d-prerelease.tag
```
wilzbach: I will make a quick follow-up which removes `--only-tags` from `d-latest.tag`

CyberShadow: Whoops, right - with `--only-tags`, we don't actually pull in and test `chm-nav-*.json`. Well spotted.

MartinNowak:
> Ah, I didn't realize that it was blocking a release. With that in mind, closing your PRs was pretty rude of me, so I apologize for that as well.

No worries, I don't usually take things personally :).

MartinNowak:
What do you think of the appveyor -> d.chm, travis or X -> dman plan? I mean dman feels like maintaining Walters personal tool, so maybe we want to drop that altogether.

CyberShadow:
Well, the root of the problems so far has been that chmgen was not tested as part of CI. Other than that linker error, IIRC none of the problems were related to actually building dman or the CHM file once chmgen has successfully executed. So, I think such problems are now behind us, and there should be no trouble re-enabling CHM and dman for releases. We should probably add building dman to CI scripts if we are to do that, though.

The idea of moving them (and other rarely-needed tools and resources) outside of the main distribution is worth considering. Maybe a "D extras" package?

MartinNowak:
> Well, the root of the problems so far has been that chmgen was not tested as part of CI. Other than that linker error, IIRC none of the problems were related to actually building dman or the CHM file once chmgen has successfully executed. So, I think such problems are now behind us, and there should be no trouble re-enabling CHM and dman for releases. We should probably add building dman to CI scripts if we are to do that, though.

OK, let's try to readd them for the beta.3.

wilzbach:
> The idea of moving them (and other rarely-needed tools and resources) outside of the main distribution is worth considering. Maybe a "D extras" package?

Yeah, especially given that many other solutions already exist: e.g. http://devdocs.io/d, dash, zeal and what not, see also:

https://forum.dlang.org/post/ecyzbdsvuedwwgqpplsq@forum.dlang.org

CyberShadow:
> many other solutions already exist

None have (or can have) an accurate index, though. We generate our own.

wilzbach:
> None have (or can have) an accurate index, though. We generate our own.

Fair enough, would using GH releases on a new tag work?
It's just a few lines to setup and everyone interested could then grab the binary from the GH releases tab.
BTW can we build the chmdocs for Windows on Linux? Someone on the NG mentioned that it works with Wine, but the way I see it, for building the Json no Windows is needed?

CyberShadow:
> Fair enough, would using GH releases on a new tag work?

Potentially, though, someone may see the "Releases" tag and expect something else. Further fragmenting our infrastructure and increasing reliance on third-party providers is also not great.

> BTW can we build the chmdocs for Windows on Linux?

I don't see why not. Not sure about the availability of Wine on CI and build machines, though.

> Someone on the NG mentioned that it works with Wine, but the way I see it, for building the Json no Windows is needed?

Yes, the JSON file with the tags is used by dman which is currently built for Linux too.

wilzbach:
> So, I think such problems are now behind us, and there should be no trouble re-enabling CHM and dman for releases. 

I just tried running it with `wine` and there are still issues as `win32.mak` can't use any of the knowledge of `posix.mak` - see https://github.com/dlang/dlang.org/pull/2250

I suggest that if we really want to continue shipping the chm docs, we remove `win32.mak`, use Linux to build the json and just run the final step (i.e. the packaging with `hhc.exe` on Windows).
BTW using `wine` for the chm generation isn't that easy, because [the wine libraries don't support `SetControlData`](https://bugs.winehq.org/show_bug.cgi?id=7517#c17)), but even using the native libraries from the installer segfaults for me :/

> Potentially, though, someone may see the "Releases" tag and expect something else. Further fragmenting our infrastructure and increasing reliance on third-party providers is also not great.

Well it also doesn't seem like the best course of action to push the maintenance of Walter's favorite way to browse the docs (but rarely used) down to @MartinNowak.
We could also upload the chm releases into the S3 bucket (IIRC S3 access can be limited to only a part of a bucket and to write/upload-only) - what did you have in mind when you suggested moving rarely used tools outside of the main distribution?


> I don't see why not. Not sure about the availability of Wine on CI and build machines, though.

We have root access on CircleCi and also can use any Docker image as root.

CyberShadow:
> Well it also doesn't seem like the best course of action to push the maintenance of Walter's favorite way to browse the docs (but rarely used) down to @MartinNowak.

Agreed.

Maybe I should just move it back to http://thecybershadow.net/d/docs/ :)

> what did you have in mind when you suggested moving rarely used tools outside of the main distribution?

Lots of software packages have an `-extras` package that is still built/shipped by the package's maintainers, but contains things that most users aren't likely to see. However, that will just complicate the release process further.

Maybe we can think of some way that the community can plug in their own things into the D release process:
- Continuous integration in a way that doesn't require admin access or writing to the github.com/dlang repositories, and that the users are responsible of
- Hooks for new tags that automatically publish build artifacts

Then, we can just place links to the community-built artifacts on our download pages, alongside the official downloads.

Does something like that exists? I think some CI providers allow building repos you don't have access to, but not sure how the configuration happens in such cases. Maybe an automatically-updating fork? Or am I overthinking this?

CyberShadow:
@MartinNowak If we drop CHM and dman (but not chmgen) right now, will that significantly simplify the release process?

wilzbach:
Well, there are actually a few things out there that depend on the release process, e.g. all the D packages that are built on GitHub (brew, snap, ...)

> Maybe an automatically-updating fork

Yes, that would work. I wonder why no one has done so already...
My trick for the Docker images to stay up-to-date, is a daily Travis cron:

https://github.com/wilzbach/dlang-docker

> Maybe we can think of some way that the community can plug in their own things into the D release 
process:

We could create a repo at dlang-community for this, but I'm not so sure about it, because I already maintain too much stuff that regularly breaks on new releases. Though on the other hand, the community could ensure that it's more stable...

Anyhow, in my experience _anything_ which isn't constantly tested on every PR _will break_, so I guess this only makes sense if we are willing for people to plug in their own scripts etc. into our CI infrastructure.

> Or am I overthinking this?

It's a real problem.
For example, for wiki.dlang.org to stay up-to-date, I setup a simple daily cron job on one of my server that uses the mediawiki API and http://downloads.dlang.org/releases/LATEST to bump https://wiki.dlang.org/Template:Latest_DMD_Version_Raw every day, because I just got too tired of manually updating it.

CyberShadow:
> We could create a repo at dlang-community for this, but I'm not so sure about it, because I already maintain too much stuff that regularly breaks on new releases.

I think ideally it would be a tool, and the responsibility of keeping their artifacts building would fall to their maintainer. E.g. if I'm personally interested in building dman and the CHM, I want no one but myself burdened by it.

> Anyhow, in my experience anything which isn't constantly tested on every PR will break, so I guess this only makes sense if we are willing for people to plug in their own scripts etc. into our CI infrastructure.

What I had in mind definitely did not involve putting any amount of weight on PR authors and reviewers. At most, a non-blocking notification, and even so it's not hard to think of cases where it would be just a nuisance (e.g. if the test is flaky).

PR #2216, opened 2018-02-18
add __traits(isReturnOnStack, func)

WalterBright:
Implementation: https://github.com/dlang/dmd/pull/7908

### DIFF
--- a/spec/traits.dd
+++ b/spec/traits.dd
@@ -42,6 +42,7 @@ $(GNAME TraitsKeyword):
     $(GLINK isRef)
     $(GLINK isOut)
     $(GLINK isLazy)
+    $(GLINK isReturnOnStack)
     $(GLINK hasMember)
     $(GLINK identifier)
     $(GLINK getAliasThis)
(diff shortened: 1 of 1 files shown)


JinShil:
image

There' a stray list item #4 in there, but I can't seem to find it in the markup.

wilzbach:
Urgh this seems to be a common problem with `{OL,UL}` and `BEST_PRACTICE`, see e.g. https://dlang.org/spec/arrays.html

wilzbach:
So the problem is that 

```
BEST_PRACTICE= $(P best-practice, $(B Best Practices:) $0)
```

BUT `OL,UL` are individual HTML blocks:

```html
<p> 
foo
 <ol> ... </ol> // <-- "invalid" HTML, starts a new block
bar
</p>
```

The [spec](https://www.w3.org/TR/html5/grouping-content.html#the-p-element) says this pretty explicitly:

> List elements (in particular, ol and ul elements) __cannot__ be children of p elements. 


----

-> fix at https://github.com/dlang/dlang.org/pull/2252

PR #2217, opened 2018-02-18
Remove text justification in table cells

mdparker:
@andralex requested this.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1011,7 +1011,6 @@ table td
 {
     border: none;
     border-bottom: 1px solid #E6E6E6;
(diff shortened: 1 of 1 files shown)


wilzbach:
Before:

image


After:

image

CyberShadow:
Your screenshot is fine. There are places affected by this change where justification didn't hurt, but it doesn't look bad without justification either. I think this is fine.

(Full disclosure: this is also a step away from automatic hyphenation, which I subjectively dislike and think shouldn't be used on web pages favoring readability and utility over presentation, so I'm personally in favor of this change.)

PR #2218, opened 2018-02-18
Change C++ code background to white

mdparker:
@andralex requested this. The pink background made it look like we're trying to portray C++ in a bad light. IMO, we ought to syntax highlight it too.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1163,7 +1163,7 @@ body.dcompiler dt a.anchor:hover:before
 
 .ccode, .cppcode, .cppcode2 /* C/C++ code */
 {
-    background-color: #f6ecf0;
+    background-color: #fcfcfc;
 }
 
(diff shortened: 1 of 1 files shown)


wilzbach:
For reference, before:

image


After:

image

wilzbach:
> IMO, we ought to syntax highlight it too.

Yes, but that's more work, because Ddoc currently does the highlighting during compilation and no JS syntax highlighter is currently loaded (though there's CodeMirror).

adamdruppe:
I just saw this thread hit my email seconds after I tweet that adrdox now has Python syntax highlighting support (and it has had C++ support for a long time).

I suppose you could manually do it with like $(GREEN) macros and such. The C++ samples are already manually character-escaped... a little syntax stuff isn't horrid to toss in by hand for these short, rarely changed things.

PR #2219, opened 2018-02-19
Remove --only-tag from d-latest.tag

wilzbach:
See https://github.com/dlang/dlang.org/pull/2213#discussion_r168968067


wilzbach:
image


Probably API-rate limiting again: https://github.com/dlang-bots/dlang-bot/issues/166

PR #2221, opened 2018-02-19
Add a release schedule to the changelog pages

wilzbach:
For some reason the argument that D isn't stable and has no proper release model often comes up.

Also for all newcomers and not hard-core insiders its hard to know the release schedule (even many contributors on GitHub aren't aware of it).

Since @MartinNowak has revamped the release process, the releases follow a schedule, so imho we should advertise this and make it easy for people to learn about this.

If everyone is okay with the format (and it gets merged), I would do a quick follow-up that replaces that injects an automatically generated release schedule table, s.t. we don't have to update this file manually.

### DIFF
--- a/changelog/release-schedule.dd
+++ b/changelog/release-schedule.dd
@@ -5,8 +5,7 @@ $(D_S $(TITLE),
 $(UL
     $(LI New release are published every $(I two) months, on the first day of every uneven month.)
     $(LI Two weeks before a new release `master` is merged into `stable` and a first beta is released.)
-    $(LI Patch releases are usually released on the first day of every even month.)
-    $(LI In the unfortunate case of serious regressions being found, more point releases or betas may be released.)
+    $(LI Point releases are published unscheduled when important issues or regressions get fixed.)
 )
 
 $(P The release schedule for 2018 is as follows:)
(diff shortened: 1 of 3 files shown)


changelog/release-schedule.dd
```diff
@@ -0,0 +1,56 @@
+Ddoc
+
+$(D_S $(TITLE),
+
+$(UL
+    $(LI New release are published every $(I two) months, on the first day of every uneven month.)
+    $(LI Two weeks before a new release `master` is merged into `stable` and a first beta is released.)
+    $(LI Patch releases are usually released on the first day of every even month.)
```
MartinNowak: Patch releases are still unplanned. There is a good argument for releasing fixes asap. Sometimes we just don't have enough important fixes to warrant the effort of a patch release.

changelog/release-schedule.dd
```diff
@@ -0,0 +1,56 @@
...
+)
+
+$(P The release schedule for 2018 is as follows:)
+
+    $(DIVC release-schedule,
+        $(TABLE
+            $(MINOR_RELEASE 2018-03-01, 2.079.0)
+            $(PATCH_RELEASE 2018-04-01, 2.079.1)
+            $(BETA_RELEASE 2018-04-15, 2.080.0)
+            $(MINOR_RELEASE 2018-05-01, 2.080.0)
```
MartinNowak: 1st of every odd month and 15th of every even month is rather straighforward.
How about just giving a brief example with 4 dates for 2018 and linking to the calendar (#2222) instead?

wilzbach:
> How about just giving a brief example with 4 dates for 2018 and linking to the calendar (#2222) instead?

Well people love up-to-date things, so I was thinking of using a few lines of D to generate this release schedule and then stop worrying about this page.

changelog/release-schedule.dd
```diff
@@ -0,0 +1,46 @@
...
+            $(MINOR_RELEASE 2018-11-01, 2.083.0)
+            $(BETA_RELEASE 2018-12-15, 2.084.0)
+            $(MINOR_RELEASE 2019-01-01, 2.084.0)
+        )
+    )
+)
+
+Macros:
+    TITLE=D release schedule
+    MINOR_RELEASE=$(ROW minor, $1, $2, minor release)
```
ZombineDev: I wonder if "major" release would be more appropriate, especially given how many new features 2.079.0 comes with. Major is also closer to the SemVer terminology, since deprecations and removal of features is allowed only in major versions.

wilzbach: I wondered about that too, but major would mean that we bump from 2 to 3.
Anyhow the entire "releases aren't SemVer" is a long-standing issue though I have hopes that it will be finally fixed this year.
We could also use a non-semver word like "new feature release"?

ZombineDev: With the current version scheme, a bump from 2 to 3 would be mean a brand new language (like D1 -> D2), so I still believe that "major" release would be more true to the name.

PR #2222, opened 2018-02-19
embed dlang calendar on dlang.org/calendar

MartinNowak:


wilzbach:
I assume you this is your response to https://github.com/dlang/dlang.org/pull/2221 (adding a release schedule to the changelog pages)?

MartinNowak:
It was caused by https://forum.dlang.org/thread/opyiekgxrlbxitfkfbxx@forum.dlang.org, but indeed mentioning the release schedule is helpful as well.

MartinNowak:
In particular the calendar is also supposed to host meetups or D talks on other conferences, thus needs to become more visible to be useful.

wilzbach:
If we want something more customized, we could give FullCalendar a try.
A quick example:

http://files.wilzbach.me/dlang/calendar.html

Of course, it would be easy to make it look more "native".
Though this requires exposing a Google Calendar API token publicly:

https://fullcalendar.io/docs/google_calendar

(it's not super bad because it's restricted to a HTTP origins)

Also I'm wondering whether there's an easy way to put e.g. the next three events from the calendar onto the frontpage, but I guess that requires parsing and understanding the iCal format?

ZombineDev:
Let's merge now and improve later.

MartinNowak:
> Let's merge now and improve later.

Unless it's broken, but generally good approach.

ZombineDev:
> Unless it's broken, but generally good approach.

@MartinNowak before merging I checked DAutoTest and didn't notice anything obviously broken (except for the ugly default design). Did I miss something?

wilzbach:
No, I think it was just a general comment about the "Let's merge now and improve later" strategy.
Also it's already deployed and live:
https://dlang.org/calendar.html

PR #2227, opened 2018-02-19
Fix CSS for the contributors.html page

wilzbach:
Injecting ROOT_DIR is required :/


wilzbach:
CC @CyberShadow (or others) - could we give this some priority?  It's a simple fix and the live page is broken: https://dlang.org/foundation/contributors.html

PR #2228, opened 2018-02-19
add links for Integer Promotions and Usual Arithmetic Conversions

WalterBright:
Pretty straightforward.


JinShil:
The new links result in a 404 in DAutoTest.  I don't know if that's due to the fact that this isn't published or due to some problem in this PR.

wilzbach:
> I don't know if that's due to the fact that this isn't published or due to some problem in this PR.

I think it's a systematic typo. It should be `type` (not `types`).

PR #2231, opened 2018-02-21
Update examples for traits(isSame) to accommodate the use of function calls

RazvanN7:


spec/traits.dd
```diff
@@ -1040,9 +1047,10 @@ void main()
     int b;
     static assert(!__traits(isSame, a => a + b, a => a + b));
 
-    // lambdas calling a function inside their body are considered incomparable
+    // lambdas calling other functions are comparable
     int f() { return 3;}
-    static assert(!__traits(isSame, a => a + f(), a => a + f()));
+    static assert(__traits(isSame, a => a + f(), a => a + f()));
+    test!(a => a + f())();
```
wilzbach: Maybe you could inline `test` here, s.t. the reader has a consistent flow?

andralex: The idea here is to show the construct works across scopes.

wilzbach:
@RazvanN7 resolved the conflicts & rebased for your ;-)

RazvanN7:
Oh thanks, I was just about to do it.

wilzbach:
For reference, the respective PR was https://github.com/dlang/dmd/pull/7935

spec/traits.dd
```diff
@@ -1053,16 +1053,23 @@ void main()
...
         $(P If these constraints aren't fulfilled, the function is considered
         incomparable and `isSame` returns $(D false).)
 
 $(SPEC_RUNNABLE_EXAMPLE_COMPILE
 ---
+int f() { return 2; }
+void test(alias pred)()
+{
+    // f() from main is a different function from top-level f()
+    static assert(!__traits(isSame, a => a + f(), pred));
```
wilzbach: @RazvanN7 this fails at the moment:

```
__stdin.d(6): Error: static assert:  `!true` is false
__stdin.d(23):        instantiated from here: `test!((a) => a + f())`
```

RazvanN7: Yes, this has unveiled a bug which occurs on template lambdas. I temporarily made the template lambda into a non-template lambda so that this can be merged. I will make a PR tomorrow to fix the bug in the lambdacomp module.

wilzbach:
> Yes, this has unveiled a bug which occurs on template lambdas. I temporarily made the template lambda into a non-template lambda so that this can be merged. I will make a PR tomorrow to fix the bug in the lambdacomp module.

Nice. Automation for the win!
Ah this isn't critical and has been pending for three weeks, so I think waiting for another day is easier.
Thanks a lot!

RazvanN7:
Solving the bug might take a while so we better merge this. The problem is that the serialization is computed right after parsing, during semantic1, which is problematic since a lot of information is missing. It might not be possible to implement serialization for template instantiations and template lambdas using `traits(isSame)`

PR #2232, opened 2018-02-21
Remove unused articles

wilzbach:
I'm just checked whether this is really unused.
If so, we might want to move the content to the wiki, s.t. it's not entirely lost.


JackStouffer:
Or adapt them to blog posts. CC @mdparker

mdparker:
Blog posts would be nice, depending on the content.

JinShil:
I believe these articles came from:
https://digitalmars.com/d/2.0/cppstrings.html
https://digitalmars.com/d/2.0/cpp0x.html

wilzbach:
> Blog posts would be nice, depending on the content.

Well it's an old comparison between D and C++0x.
There's not so much we can revive - especially given the newer https://dlang.org/articles/cpptod.html exists

PR #2234, opened 2018-02-22
Show dummy content to the front page when run with DIFFABLE=1

wilzbach:
s.t. people aren't confused that something is broken


dlang.org.ddoc
```diff
@@ -148,7 +148,7 @@ DDSUBLINK=$(LINK2 $(ROOT_DIR)$1.html#$2, $3)
 _=
 
 DBLOG_LATEST_AUTHOR=A D enthusiast
-DBLOG_LATEST_DATE=Devember 17, $(YEAR)
```
wilzbach: That's not a typo. It was supposed to be a joke...

wilzbach: (though I have to admit a pretty bad one)

PR #2237, opened 2018-02-22
Revert "Update spec for multiple selective imports"

wilzbach:
Reverts dlang/dlang.org#2131 - see also: https://github.com/dlang/dmd/pull/7939


wilzbach:
I forgot that `stable` is still a thing for `dlang.org` -> https://github.com/dlang/dlang.org/pull/2245

Sorry for the noise :/

PR #2238, opened 2018-02-22
Use the FullCalendar widget instead of the embedded Google widget

wilzbach:
See also:
- https://github.com/dlang/dlang.org/pull/2222#issuecomment-367156497
- https://fullcalendar.io/

FullCalendar allows a lot of customization and also comes with a few themes:

https://fullcalendar.io/js/fullcalendar-3.8.2/demos/themes.html


calendar.dd
```diff
@@ -3,11 +3,48 @@ Ddoc
...
 )
 )
 
+
+
 Macros:
     TITLE=DLang Calendar
+    EXTRA_HEADERS=
+<script src="https://cdn.jsdelivr.net/momentjs/2.14.1/moment-with-locales.min.js"></script>
+<script src="https://code.jquery.com/jquery-3.3.1.js"></script>
```
JackStouffer: need to revive my jquery upgrade pr :/

wilzbach:
CC @CyberShadow - are you okay with this too?
The only downside here is that a Google calendar API key needs to be exposed, but I restricted the domain to localhost, dlang.org and dlang.io which should make abuse harder. And if this ever becomes a problem we can easily revert back to the old widget.

CyberShadow:
For some reason some of the CDNs used on this page and dlang.org are timing out for me. Is it just me?

CyberShadow:
When it does load, the calendar I see is completely blank. I see some events currently on https://dlang.org/calendar , but here there's nothing.

wilzbach:
>  Is it just me?

I think so. Does https://fullcalendar.io/js/fullcalendar-3.8.2/demos/themes.html work for you?

CyberShadow:
Yep, that works fine.

wilzbach:
Interesting it works fine in my Chrome and Firefox:

image

What error do you get in your Firefox?

CyberShadow:
No error, it's just blank.

PR #2239, opened 2018-02-22
Add an 'Open in IDE' button for the frontpage + spec examples

wilzbach:
From https://github.com/dlang/dlang.org/pull/2147#issuecomment-367089547

> One further idea would be to add a third button beside "Edit" "Input" "Run" called e.g. "Open Online IDE" that opens run.dlang.io with the chosen code snippet.

Note that we already have this feature for the Phobos documentation, that's why it was so easy to do here:

image

Preview:

image

image

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1582,6 +1582,10 @@ input.runButton, input.resetButton, input.argsButton, input.inputButton, input.e
     z-index: 2;
     cursor: pointer;
 }
+input.openInEditorButton {
+    float: right;
+    width: auto;
+}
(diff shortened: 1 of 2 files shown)


wilzbach:
> BTW, the -X / -D buttons could really use some hover text (title) (or otherwise accessible) explanations.

Good idea: https://github.com/dlang-tour/core/pull/683
Thanks!
It's always harder to judge how easy/usable a website is if you use it daily (or even have developed it), so feedback is very welcome :)

PR #2241, opened 2018-02-22
Add State of D 2018 banner

wilzbach:
Preparation for the "State of D" survey.

Survey: https://dlang.typeform.com/to/H1GTak
Results: https://dlang.typeform.com/report/H1GTak/PY9NhHkcBFG0t6ig

CC @mdparker

Preview:

image


JackStouffer:
IMO it would be more effective to have something like those cookie notifications on EU sites. I.E. a banner that appears on the bottom of the window for any page the user visits, and which links to the survey. But, once it's dismissed once, it doesn't appear again.

I can write up an implementation quickly if you like the idea.

wilzbach:
> I can write up an implementation quickly if you like the idea.

I like the idea, but it's just planned to be there for ~two weeks and I'm currently pretty busy with other things :/

You probably want to have a look at https://github.com/dlang-community/artwork

wilzbach:
Going to merge this for now as we want to display the banner on the front page and the DConf talk deadline has passed anyways.
Improvements are always welcome.

PR #2242, opened 2018-02-22
Auto-generate changelog navigation footer

wilzbach:
This finally gets rid of the last part of the ugly bash update script and unrelated changes when adding a new release.

CC @MartinNowak


wilzbach:
@CyberShadow - are you okay with replacing bash code with D?

MartinNowak:
OTOH, the bash script has worked since years wasn't even paid off ;).

PR #2243, opened 2018-02-22
Add a html-verbatim target analogous to html

wilzbach:
This allows just running `verbatim` for all HTML pages. Not really a neccessity, but sometimes handy and a trivial change.


wilzbach:
Seems like exactly at midnight the bot runs into API rate limits :/

PR #2246, opened 2018-02-23
Update the Objective-C documentation with support for class methods

jacob-carlborg:
The dlang.org part of https://github.com/dlang/dmd/pull/7762.


MetaLang:
Will wait for the DMD PR to be merged before merging this. Ping me if I forget @jacob-carlborg.

PR #2204, opened 2018-02-12
Improve Compile-time sequences docs

ntrel:
* Change `expression list` to `value sequence` (this matches https://dlang.org/spec/template.html#variadic-templates point 2).
* Change `list` to `sequence`.
* Describe value & symbol sequences.
* Add subheadings.
* Tweak examples.
* Mention `tupleof`.

PR #2248, opened 2018-02-23
Improve *Foreach over Tuples* spec

ntrel:
* Compile-time Tuple -> Sequence (see #2204)
* ExpressionTuple -> ValueSeq (this matches https://dlang.org/spec/template.html#variadic-templates point 2)
* Improve wording of *Foreach over Tuples* & example
* Add link to `static foreach`

### DIFF
--- a/spec/statement.dd
+++ b/spec/statement.dd
@@ -458,7 +458,7 @@ $(GNAME ForeachAggregate):
 $(P
         $(I ForeachAggregate) is evaluated. It must evaluate to an expression
         of type static array, dynamic array, associative array,
-        struct, class, delegate, or tuple.
+        struct, class, delegate, or sequence.
         The $(PS0) is executed, once for each element of the
         aggregate.
         At the start of each iteration, the variables declared by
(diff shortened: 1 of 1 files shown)


spec/statement.dd
```diff
@@ -835,52 +835,53 @@ void main()
...
+        When the sequence is a $(I ValueSeq), the element alias is a variable
+        and is set to each value in the sequence. If the type of the
+        variable is given, it must match the type of the sequence contents.
+        If no type is given, the type of the variable is set to the type
+        of the sequence element, which may change from iteration to iteration.
+)$(P
+        If there are
+        two symbols declared, the first is the $(I index variable)
+        and the second is the $(I element alias). The index
+        must be of $(D int) or $(D uint) type, it cannot be `ref`,
```
wilzbach: It can also be `long`: https://run.dlang.io/is/kxSAFP as the index is `size_t`

ntrel: @wilzbach I've added `long` and `ulong` here, and updated runtime foreach index types, which listed just `size_t` instead of these. BTW for runtime foreach only, dmd accepts `short` and `byte` - I've not documented this though.

PR #2250, opened 2018-02-24
Fix chmgen for Windows

wilzbach:
1) There are still things that are different with the win32.mak file (e.g. it doesn't use the ddoc preprocessor)
2) If we want to guarantee that it succeeds we probably need to use wine or other methods to verify that the build passers
3) An alternative method would be to build the JSON on Linux and just run the final step (the packaging with `hhc.exe` on Windows)

In any case as mentioned in https://github.com/dlang/dlang.org/pull/2213, I'm don't think that investing time into maintaining the `win32.mak` build is worthwhile..


win32.mak
```diff
@@ -18,7 +19,7 @@ chmgen.exe : chmgen.d
 	$(DMD) -g chmgen
 
 chm\d.hhp chm\d.hhc chm\d.hhk : chmgen.exe chm-nav-release.json $(TARGETS)
-	chmgen --target release
+	chmgen --root=$W --target release
```
CyberShadow: IIRC win32.mak used to build in-place, i.e. put the .html files alongside the .dd files. Has this changed?

win32.mak
```diff
@@ -30,11 +30,6 @@ d.chm : chm\d.chm
 chm-nav-release.json : $(DDOC) std.ddoc spec\spec.ddoc modlist-release.ddoc changelog\changelog.ddoc chm-nav.dd
 	$(DMD) -o- -c -Df$@ $**
 
-d.tag : chmgen.exe $(TARGETS)
-	chmgen --only-tags
-
-d-tags.json : d.tag
-
```
CyberShadow: Why? Is this obsoleted by posix.mak? Are we sure this is not used anywhere, incl. Martin's release-building scripts?

win32.mak
wilzbach: `win32.mak` doesn't generate the documentation anymore. This has long been deprecated, though it's `docs` for the installer: https://github.com/dlang/installer/blob/master/create_dmd_release/create_dmd_release.d#L514

win32.mak
wilzbach: Yes, `chmgen` already depends on a Linux host to build the HTML files, for which the `d.tags.json` is built too (in fact https://github.com/dlang/installer/pull/298 shows that this already has been done).
Apart from that building the `d-tags.json` comes with its problems (e.g. root of the HTML files etc.), so I thought it's best to keep only what's really run.

>  incl. Martin's release-building scripts?

Yes, as mentioned he copied the JSON file over from Linux.
See also:

https://github.com/dlang/installer/blob/master/create_dmd_release/create_dmd_release.d#L514
https://github.com/dlang/installer/pull/298

PR #2255, opened 2018-02-28
Include pending.dd in the changelog footer generation

wilzbach:
Required for https://github.com/dlang/tools/pull/327

### DIFF
--- a/ddoc/source/preprocessor.d
+++ b/ddoc/source/preprocessor.d
@@ -320,21 +320,34 @@ auto genChangelogVersion(string fileName, string fileText)
 
         // inject the changelog footer
         auto fileBaseName = fileName.baseName;
-        auto r = changelogFiles.enumerate.find!(a => a.value.baseName == fileBaseName);
+        auto r = changelogFiles.chain("pending.dd".only).enumerate.find!(a => a.value.baseName == fileBaseName);
         if (r.length != 0)
         {
             auto el = r.front;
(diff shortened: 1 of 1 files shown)


MartinNowak:
Apparently breaks detection of the latest changelog (http://dtest.dlang.io/diff/website-2fe0bc4bde757a4b78766769260b1ae5f2c4c57f-6eb46c6d29bfbbe2c48fdb8bd6f30f8d/website-e1e5ceba304d91721bee12217cd6580c6f3d45e8-5147cf4817a6e8eca361d19b8ee66c00/web/changelog/2.078.3.html)
2.078.3 links to 2.078.3 as previous

wilzbach:
Fixed. Should be good to go now.

wilzbach:
>  Documentation OK (no changes)

(though it will only have full effect with dlang/tools#327).

@CyberShadow what do you think? Anything blocking this?

CyberShadow:
We probably don't want to touch `pending.dd` for release builds. Is that accounted for?

wilzbach:
> We probably don't want to touch pending.dd for release builds. Is that accounted for?

Yes, the file isn't build for release builds, but at the moment it still appears in the menu. Though that's unrelated to this PR. We could check the build mode in the preprocessor, but I think that's better than in a different PR (and I don't think it's that important either as the offline docs are just for browsing the spec + library docs).

Next versions doesn't link to pending as pending isn't part of the `versions` array.

PR #2259, opened 2018-03-03
add detail about empty structs

WalterBright:
Empty structs are a minefield of more or less intractable issues.

### DIFF
--- a/spec/struct.dd
+++ b/spec/struct.dd
@@ -62,6 +62,8 @@ $(H2 $(LNAME2 struct_layout, Struct Layout))
     the first field and the start of the object.
     )
 
+    $(P Structs with no fields of non-zero size (aka $(I Empty Structs)) have a size of one byte.)
+
     $(P Non-static $(RELATIVE_LINK2 nested, function-nested D structs), which access the context of
     their enclosing scope, have an extra field.
     )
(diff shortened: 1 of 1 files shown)


jmdavis:
Wow. That is, umm... disgusting.

spec/struct.dd
```diff
@@ -70,11 +72,21 @@ $(H2 $(LNAME2 struct_layout, Struct Layout))
     $(OL
     $(LI The default layout of the fields of a struct is an exact
     match with the $(I associated C compiler).)
+    $(LI g++ and clang++ differ in how empty structs are handled. Both return `1` from `sizeof`,
+    however, clang++ does not push them onto the parameter stack while g++ does. This is a
+    binary incompatibility between g++ and clang++.
+    dmd follows clang++ behavior for OSX and FreeBSD, and g++ behavior for Linux and other
+    Posix platforms.
```
jmdavis: I suspect that in the long run, it'll make more sense to use clang's behavior on pretty much everything other than Linux or Linux-based systems such as Android (though Android may prefer clang too; I'm not sure; it's not supported by dmd at the moment though, so it doesn't really matter for dmd). Certainly, as I understand it, all of the BSDs are moving towards ditching gcc in favor of clang, but right now, dmd isn't officially supporting any of those platforms, so it isn't an issue at the moment, and this can be changed to take such systems into account if they ever do end up with dmd support. And given how clang supposedly tries to be compatible with gcc, I wonder if clang is ever going to change what it's doing to match gcc. The whole thing is ugly either way, and we can cross that bridge if it ever happens.

WalterBright: The plan is to match the behavior of the default compiler on the target platform, whether that is clang or gcc.

PR #2261, opened 2018-03-04
use UNDEFINED_BEHAVIOR macro

WalterBright:
Searched the spec for descriptions of undefined behavior, and delineated them with the UNDEFINED_BEHAVIOR macro.

### DIFF
--- a/spec/attribute.dd
+++ b/spec/attribute.dd
@@ -277,10 +277,12 @@ auto sz = S.sizeof;  // 12
         $(P Do not align references or pointers that were allocated
         using $(GLINK2 expression, NewExpression) on boundaries that are not
         a multiple of $(D size_t). The garbage collector assumes that pointers
-        and references to gc allocated objects will be on $(D size_t)
-        byte boundaries. If they are not, undefined behavior will
-        result.
+        and references to GC allocated objects will be on $(D size_t)
+        byte boundaries.
         )
+
+        $(UNDEFINED_BEHAVIOR If any pointers and references to GC
+        allocated objects are not aligned on `size_t` byte boundaries.)
 
         $(P The $(I AlignAttribute) is reset to the default when
         entering a function scope or a non-anonymous struct, union, class, and restored
--- a/spec/class.dd
+++ b/spec/class.dd
@@ -389,6 +389,20 @@ $(GNAME Constructor):
         $(P The following restrictions apply to class construction:)
 
     $(OL
+        $(LI It is illegal for constructors to mutually call each other.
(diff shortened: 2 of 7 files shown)


spec/attribute.dd
```diff
@@ -279,10 +279,12 @@ auto sz = S.sizeof;  // 12
         using $(GLINK2 expression, NewExpression) on boundaries that are not
         a multiple of $(D size_t). The garbage collector assumes that pointers
         and references to gc allocated objects will be on $(D size_t)
-        byte boundaries. If they are not, undefined behavior will
-        result.
+        byte boundaries.
         )
 
+        $(UNDEFINED_BEHAVIOR If any pointers and references to gc
```
wilzbach: Shouldn't GC be uppercase as it's an abbreviation?

spec/class.dd
```diff
@@ -392,14 +392,18 @@ $(GNAME Constructor):
...
 
         ------
         this() { this(1); }
         this(int i) { this(); } // illegal, cyclic constructor calls
         ------
+
+        $(IMPLEMENTATION_DEFINED The compiler is not required to detect
+        cyclic constructor calls.)
+
+        $(UNDEFINED_BEHAVIOR If the program executes with cyclic construct
```
wilzbach: constructor

spec/expression.dd
```diff
@@ -770,6 +766,12 @@ $(GNAME DeleteExpression):
         is called.
     )
 
+    $(UNDEFINED_BEHAVIOR
+    $(OL
+    $(LI Using `delete` to free memory allocated by the garbage collector.)
```
wilzbach: Isn't there a "not" missing?
Using `delete` with GC allocated objects is well-defined, e.g. https://github.com/dlang/druntime/blob/master/src/rt/lifetime.d#L177

WalterBright: Look closely at the link - it free's the data. There could still be other copies of the pointer.

wilzbach: 1) This is different to the existing statement that you removed:

> If the garbage collector was __not__ used to allocate the memory for the instance, undefined behavior will result.

2) Is `GC.free` on GC-allocated objects really undefined behavior?
AFAICT what it exactly does depends on the GC implementation, but currently it's `freeNoSync` which tries to free the memory if possible.

https://github.com/dlang/druntime/blob/master/src/gc/impl/conservative/gc.d#L818

Anyhow, `delete` is luckily finally deprecated, so we hopefully can remove it entirely soon.

PR #2262, opened 2018-03-04
Tweak betterC and front page

joakim-noah:
Saying TLS doesn't work with betterC has bothered me [ever since that page got added](https://github.com/dlang/dlang.org/pull/1796#discussion_r128740386), also took the opportunity to make other tweaks to that page and the front page.


ZombineDev:
Thanks @joakim-noah!

joakim-noah:
NP, good to see it fixed.

PR #2264, opened 2018-03-05
Improve function argument order spec.

JohanEngelen:
Related recent change: https://github.com/dlang/dlang.org/pull/2132

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -11,8 +11,9 @@ $(HEADERNAV_TOC)
 
 $(H2 $(LNAME2 order-of-evaluation, Order Of Evaluation))
 
-$(P Binary expressions and function arguments are evaluated in strictly
-left-to-right order.)
+$(P Binary expressions are evaluated in strictly left-to-right order.
+Function arguments are evaluated in strictly left-to-right order for
+functions with `extern (D)` $(DDSUBLINK spec/attribute, linkage, linkage).)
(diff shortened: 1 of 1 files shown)


spec/expression.dd
```diff
@@ -11,8 +11,9 @@ $(HEADERNAV_TOC)
 
 $(H2 $(LNAME2 order-of-evaluation, Order Of Evaluation))
 
-$(P Binary expressions and function arguments are evaluated in strictly
-left-to-right order.)
+$(P Binary expressions are evaluated in strictly left-to-right order.
```
joakim-noah: Don't think you need the dashes, `left to right`.

spec/expression.dd
```diff
@@ -25,10 +26,10 @@ left-to-right order.)
     $(IMPLEMENTATION_DEFINED
     $(OL
     $(LI The order of evaluation of the operands of $(GLINK AssignExpression).)
-    $(LI The order of evaluation of function arguments.)
+    $(LI The order of evaluation of function arguments (for functions with linkage other than `extern (D)`).)
```
joakim-noah: Remove the parantheses around "for functions," it's not an aside.

spec/expression.dd
JohanEngelen: I like the hyphens of the original, reads a little easier.

PR #2266, opened 2018-03-07
Update interfaceToC.dd

Dashster:
The string literal passed to printf appears to have a null character appended. The documentation currently says that it doesn't, and that it does. Change first sentence to say that it does.

### DIFF
--- a/spec/interfaceToC.dd
+++ b/spec/interfaceToC.dd
@@ -226,7 +226,7 @@ void foo(char[] string)
 }
 ------
 
-        While the $(P The $(CODE printf) format string literal
+        $(P While the $(CODE printf) format string literal
         in the example does not end with an explicit $(CODE '\0'), $(CODE printf) will still see one.
         This is because string literals,
         when they are not part of an initializer to a larger data structure,
--- a/spec/interfaceToC.dd
+++ b/spec/interfaceToC.dd
@@ -227,7 +227,7 @@ void foo(char[] string)
 ------
 
         $(P The $(CODE printf) format string literal
-        in the example doesn't end with $(CODE '\0').
+        in the example DOES end with $(CODE '\0').
         This is because string literals,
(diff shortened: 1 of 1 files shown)


spec/interfaceToC.dd
```diff
@@ -227,7 +227,7 @@ void foo(char[] string)
 ------
 
         $(P The $(CODE printf) format string literal
-        in the example doesn't end with $(CODE '\0').
+        in the example DOES end with $(CODE '\0').
```
wilzbach: Full capitalization isn't used in spec. How about `$(I ...)` for italic?

Dashster: Changed to $(I does).

aG0aep6G:
As far as I see, this change is wrong.

The string literal in the example does **not** end with `\0`. A string literal that ends with `\0` would be: `"foo\0"`. As the documentation says, the null character is stored **after** the end of the literal. The documentation is correct as it is. But maybe it could be improved to convey the message more clearly.

wilzbach:
@Dashster have a look at https://en.wikipedia.org/wiki/Printf_format_string#Precision_field

> For the string type, it limits the number of characters that should be output, after which the string is truncated.

tl;dr:  "printf will still see one" is incorrect, it's just a specific printf instruction.

aG0aep6G:
> @Dashster have a look at https://en.wikipedia.org/wiki/Printf_format_string#Precision_field
[...]
> tl;dr: "printf will still see one" is incorrect, it's just a specific printf instruction.

I think you're misunderstanding this. The text is not about the `string` variable (which should probably have a different name). It's about the format string `"my string is: %.*s\n"`. The `%.*s` in there doesn't matter. The code could just as well be `printf("foo");`.

Dashster:
I would like to withdraw this request but don't know how to do so. If anyone else can do so, I would appreciate it.

PR #2267, opened 2018-03-11
Remove 'State of D' survey banner

wilzbach:
To be merged on Wednesday, 14.3


wilzbach:
As it turns out, we already have a new item for the highlighted news area - the Amazon Smile promotion.
I already added this to this PR to avoid merge conflicts:

image

JackStouffer:
I'm not really sure about this. The turquoise really sticks out from the rest of the page.

Can we put this on the donate page or redesign it to be less obtrusive?

What do you think @CyberShadow?

wilzbach:
> redesign it to be less obtrusive?

How about something like the existing banner?

amazon-smile-banner

or:

open-collective-launchedpng

CyberShadow:
I think the banner is OK, considering it's temporary.

wilzbach:
@JackStouffer the plan is to swap this tonight, so if you have a better idea, please let me know.

JackStouffer:
If it's temporary I'm ok with it.

PR #2273, opened 2018-03-12
Add OpenCollective backers to foundation/sponsors

wilzbach:
Preview:

image

### DIFF
--- a/foundation/sponsors.dd
+++ b/foundation/sponsors.dd
@@ -41,8 +41,18 @@ $(UL
     $(LI C. Piker)
 )
 
+$(H2 Backers on OpenCollective)
+
+<script src='https://opencollective.com/dlang/banner.js?style={"a":{"color":"black"}}'></script>
 )
 
 Macros:
     TITLE=D Language Foundation Sponsors
     SUBNAV=$(SUBNAV_FOUNDATION)
+    EXTRA_HEADERS=$(T style,
+        #opencollective-banner h2 {
+          color: red;
+        }
+
+    )
+    _=


JackStouffer:
Can we add some CSS rules to bring the added elements in line with the rest of the site?

wilzbach:
Not sure as it's essentially embedding this IFrame (https://opencollective.com/dlang/banner.html?limit=10&id=banner-iframe-8970&style={}) and they don't seem to allow CORS :/

They allow to inject custom style though via the URL parameter, but that doesn't seem to work for the text as (1) the use inline styles for the colors and (2) they only seem to allow a fixed set of css classes and `a` is one of the few allowed.

See also:

https://opencollective.com/dlang/banner.html?style={%22a%22:{%22color%22:%22orange%22},%22body%22:{%22color%22:%22blue%22}}

https://opencollective.com/widgets

wilzbach:
Ping. I know that the design isn't ideal, but it's not on the front page and not hugely disturbing compared to the rest of the site, so imho we should merge this and whenever someone has time improve it.
We have done the same with the calendar..


CC @CyberShadow - would this be okay with you too?

wilzbach:
Do you have troubles when opening https://opencollective.com/dlang/banner.html directly? Then we could just use the iframe directly without their loading injection.

CyberShadow:
No, that works perfectly fine, even with JavaScript disabled.

JackStouffer:
Works fine on Firefox an Safari on macOS.

wilzbach:
> No, that works perfectly fine, even with JavaScript disabled.

Hmm, I would prefer that solution too, but the problem here is that the height of the iframe is dynamic and CORS restrictions prevent us from accessing it and checking for the height.

There are a few suggestions on SO, but they are all not really pretty and often require access to the source:

https://stackoverflow.com/questions/5589756/is-there-a-cross-domain-iframe-height-auto-resizer-that-works

> Works fine on Firefox an Safari on macOS.

And Chrome + Firefox on my Linux machine too.

PR #2275, opened 2018-03-13
Fix Issue 18603 - Update minimally required version for macOS

wilzbach:
See also: https://dlang.org/changelog/2.079.0.html#osx-10_9


wilzbach:
Ping. This is a really trivial documentation update ...

PR #1941, opened 2017-12-05
Fix Issue 9433 - Deprecate delete

JinShil:
Spec and documentation update to accompany https://github.com/dlang/dmd/pull/7342

PR #2277, opened 2018-03-16
Deprecate class allocators and deallocators

JinShil:
`delete` was deprecated in 2.079.  Before the `delete` keyword can be removed it will be necessary to also deprecate and remove class allocators and deallocators.  They have been planned for deprecation for years; this PR finally puts the deprecation process in motion.

See https://dlang.org/deprecate.html#Class%20allocators%20and%20deallocators

DMD PR:  https://github.com/dlang/dmd/pull/8042

### DIFF
--- a/deprecate.dd
+++ b/deprecate.dd
@@ -11,7 +11,7 @@ $(SPEC_S Deprecated Features,
     $(TABLE2 Deprecated Features,
         $(THEAD Feature,                                                          Spec,  Dep,    Error,  Gone)
         $(TROW $(DEPLINK Hexstring literals),                                     2.079, 2.079, &nbsp;, &nbsp;)
-        $(TROW $(DEPLINK Class allocators and deallocators),                      future, &nbsp;, &nbsp;, &nbsp;)
+        $(TROW $(DEPLINK Class allocators and deallocators),                      &nbsp;, 2.080, &nbsp;, &nbsp;)
         $(TROW $(DEPLINK Implicit comparison of different enums),                 2.075,  2.075, &nbsp;, &nbsp;)
         $(TROW $(DEPLINK Implicit string concatenation),                          2.072,  2.072, &nbsp;, &nbsp;)
         $(TROW $(DEPLINK Using the result of a comma expression),                 2.072,  2.072,  2.079, &nbsp;)


adamdruppe:
So we recently found `@disable new` is actually kinda cool....

JinShil:
> So we recently found @disable new is actually kinda cool....

Please elaborate.

adamdruppe:
```
struct S
{
  @disable new (size_t s) { return null; }
}

void main()
{
	S* s = new S; //  Error: allocator `e.S.new` is not callable because it is annotated with @disable
}
```

adamdruppe:
So the application there is you can force users to go through your custom factory functions, or in the case of structs, put them on the stack (or bypass with memcpy lol).

JackStouffer:
See https://github.com/dlang/phobos/pull/6003 for an example.

JinShil:
I've added support for the syntax `@disable new();` to the DMD PR.  You won't even need to add the useless `size_t` argument.  That should prevent any breaking changes.

adamdruppe:
That sounds good to me!

JinShil:
>  That sounds good to me!

Great!  If you don't mind, and truly approve of this PR, please take the explicit step of marking it "Approved".

adamdruppe:
oh lol i keep forgetting i actually have some power in this repo. I don't have such power over on the dmd side though.

JinShil:
> I don't have such power over on the dmd side though.

I still think it's helpful to mark PRs "approved" even in repos where you may not have merge rights.  It helps to know that it was reviewed and scrutinized by anyone with D language skills, especially a veteran like yourself.

wilzbach:
Shouldn't this PR remove or mark as deprecated the respective sections in the spec too?
e.g.
https://dlang.org/spec/class.html#allocators

JinShil:
> Shouldn't this PR remove or mark as deprecated the respective sections in the spec too?

The spec already states the following for allocators:

> Note: Class allocators are deprecated in D2. 

... and the following for class deallocators

> Note: Class deallocators and the delete operator are deprecated in D2

I remember a brief discussion we had with one of the other deprecations in 2.079 to leave the spec in tact until removal (See https://github.com/dlang/dlang.org/pull/1941#issuecomment-364106970) so users could still refer to it when updating their code.  Therefore, since the spec has already marked allocators and deallocators as deprecated, I think the spec should be left it as is until removal.

JinShil:
Ping!  DMD PR was merged.

PR #2278, opened 2018-03-17
Adjust guidelines for attributes on templated functions.

jmdavis:
If whether a particular attribute is applicable to a templated function
or not does not depend on the template arguments (e.g. it's a function
for operating on arrays of characters that's templated on character type
and @safe would always be inferred), then having the attributes be
explicit is better for the documentation. It also ensures that the
attribute is enforced so that it's caught if the code changes
accidentally conflict with the attribute. One of the major downsides to
attribute inference is that it's unclear which attributes apply, and
it's easier to accidentally make changes which change the attributes.

Attribute inference is critical when the applicability of an attribute
depends on the template arguments, but if the attribute is _always_
inferred regardless of the template arguments, then avoiding putting the
attributes on the function just give us the downsides of attribute
inference without the gains. So, ultimately, it makes the most sense to
be as explicit about attributes as we can be.

### DIFF
--- a/dstyle.dd
+++ b/dstyle.dd
@@ -355,8 +355,13 @@ $(LISTSECTION phobos_return_types, Return type,
 $(LISTSECTION phobos_attributes, Attributes,
     $(LI $(I Non-templated) functions should be annotated with
          matching attributes (`@nogc`, `@safe`, `pure`, `nothrow`).)
-    $(LI $(I Templated) functions should $(B not) be annotated with attributes
-         as the compiler can infer them.)
+    $(LI If the template arguments for a $(I templated) function affect whether
+         an attribute is appropriate, then the function should $(B not) be
+         annotated with that attribute so that the compiler can infer it.
+         However, if the attribute is not affected by the template arguments
+         (and thus would always be inferred), then the function should be
+         explicitly annotated with that attribute just like a non-templated
+         function would be.)
     $(LI Attributes should be listed in alphabetical ordering, e.g. `const @nogc nothrow pure @safe`
         (the ordering should ignore the leading `@`).
     )


jmdavis:
> I see the advantages of being explicit if possible, but what if a new attribute is ever added?

Then a new attribute is added. I don't understand what the question is.

wilzbach:
The new attribute would automatically be inferred if you don't explicitly add the attributes for templated functions.

jmdavis:
> The new attribute would automatically be inferred if you don't explicitly add the attributes for templated functions.

If a new attribute gets inferred, then a new attribute gets inferred. Ideally, we're as explicit as possible, but that doesn't mean that we're perfect about it. It's also potentially the case that some attributes would depend on the template arguments and some wouldn't, so only some would be explicit, though that's pretty unlikely.

The main point here is that we should be explicit about attributes where we can be. Leaving attributes to be inferred when they're guaranteed to be inferred just means that the documentation isn't clear when it could be. I created the PR, because some folks were taking what the style guide says to mean that you should never put attributes on templated functions, which I definitely think is the wrong approach. We should try to be as explicit as we can be even if we're not perfect about it.

jmdavis:
@wilzbach Can you look at this again, please?

I really think that it's bad policy to require that attribute inference be used when it's unnecessary. Attribute inference is fantastic when the attributes depend on the template arguments, but when they don't, using attribute inference just means making the documentation less clear. If an attribute gets added to the language later and thus ends up being inferred on old code, then ideally, the code would be updated later, but I don't see it as a big deal if occasionally, attributes end up being inferred instead of being explicit. The core thing here is that we try to be explicit about them when we reasonably can be, because that makes the documentation clearer, whereas right now, the style guide claims that we should never be explicit about attributes even when we can be, and that's just going to make the documentation worse for no benefit.

jmdavis:
Thanks!

PR #2281, opened 2018-03-19
Update struct.dd - postblit documentation

RazvanN7:


andralex:
@RazvanN7 - need to define what happens when the entire struct is qualified and then defines a postblit with no qualifier/same qualifier/different qualifier.

andralex:
@RazvanN7 please add examples after each major paragraphs. RUNNABLE_EXAMPLE should be great here. @wilzbach we'd really need a ERROR_EXAMPLE here, do you think you could add that to our build infra?

spec/struct.dd
```diff
@@ -566,6 +566,113 @@ $(GNAME Postblit):
...
+        }
+        ---
+
+        $(P Postblits are not overloadable.)
+        
+        $(P If a postblit is qualified with `immutable` or `shared` as in
+        $(D this(this) immutable;) or $(D this(this) shared;), then the code
+        is ill-formed. Historically, such definitions passed compilation but
+        could not be invoked. A deprecation diagnostic is issued by the front-end
+        of dmd 2.080 or later, see $(BUGZILLA 18417). Example:)
```
wilzbach: Don't add things that can get outdated into the spec, because they will.
Info this belongs into the changelog entry

spec/struct.dd
```diff
@@ -566,6 +566,113 @@ $(GNAME Postblit):
...
+        struct S2
+        {
+            // ill-formed/not callable, no diagnostic before dmd 2.080
+            this(this) shared {}
+        }
+        ---
+
+        $(P If a postblit is qualified with `const` as in $(D this(this) const;),
+        the code is ill-formed. A deprecation diagnostic is issued by the
+        front-end of dmd 2.080 or later, see $(BUGZILLA 18417). Prior versions
```
wilzbach: Again information that will get outdated, better not add it.

spec/struct.dd
```diff
@@ -566,6 +566,113 @@ $(GNAME Postblit):
...
+        void main()
+        {
+            immutable S a;
+            auto a2 = a;
+            import std.stdio: writeln;
+            writeln(a2.n);     // prints 1
+        }
+        ---
+        
+        $(P This allows incorrect code to compile, refer to $(BUGZILLA 18357) for examples.)
```
wilzbach: This doesn't work. The spec should be self-containing as Bugzilla might be offline or it simply might be print as ebook or PDF.

wilzbach:
Out of interest: does this try to address https://issues.dlang.org/show_bug.cgi?id=14246?
(see also the proposed changes + tests in https://github.com/dlang/dmd/pull/6816)

RazvanN7:
>Out of interest: does this try to address https://issues.dlang.org/show_bug.cgi?id=14246?

No, this PR documents solely the postblitting mechanism; it's a work in progress, though.

spec/struct.dd
```diff
@@ -566,6 +566,206 @@ $(GNAME Postblit):
...
+}
+---
+
+        $(P The following describes the behavior of the
+        qualified postblit definitions:)
+
+        $(OL
+        $(LI `const`. When a postblit is qualified with `const` as in
+        $(D this(this) const;) or $(D const this(this);) then the postblit
+        is succesfully called on mutable (unqualified), `const`,
```
andralex: I removed the backquotes around mutable because mutable is not a keyword.

spec/struct.dd
```diff
@@ -566,6 +566,206 @@ $(GNAME Postblit):
...
+    auto s4 = s3;
+    immutable S s5;
+    auto s6 = s5;
+}
+---
+)
+        $(LI `immutable`. When a postblit is qualified with `immutable`
+        as in $(D this(this) immutable) or $(D immutable this(this))
+        the code is ill-formed. The `immutable` postblit passes the
+        compilation phase but cannot be invoked. Example:)
```
andralex: Version when this will be diagnosed? I think we can leave this text as is and add a mention once we merge the appropriate PR.

spec/struct.dd
```diff
@@ -566,6 +566,206 @@ $(GNAME Postblit):
...
+    this(this) const {} // error: function X.__postblit conflicts with function X.__postblit
+}
+---
+
+        $(P The following describes the behavior of the
+        qualified postblit definitions:)
+
+        $(OL
+        $(LI `const`. When a postblit is qualified with `const` as in
+        $(D this(this) const;) or $(D const this(this);) then the postblit
```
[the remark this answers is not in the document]
RazvanN7: We will deprecate this with the next version, but @wilzbach argued that information that gets outdated should not be in the spec. I thought that saying it's ill-formed but not mentioning the deprecation is the best compromise since you can actually call the function but it's not doing what it is supposed to.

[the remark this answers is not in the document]
andralex: It's helpful to give that information so people who see invalid code getting compiled don't get confused.

wilzbach:
Sorry for being late to the party, but should the behavior for an array of structs be defined too?
At the moment DMD calls `__ArrayPostblit` which is automatically added to `object.d`, but that might change soon: https://github.com/dlang/dmd/pull/8067

In any case, the behavior for now is:

```d
foreach (ref e; cast(S*)this.v.ptr)[0 .. n])
      e.__xpostblit();
```
```

PR #2284, opened 2018-03-19
Use the dlang.org's XML stream at feedburner to convert the latest articles to ddoc macros

wilzbach:
This won't work on DAutoTest, but the generated Ddoc file will be e.g.

```
DBLOG_LATEST_TITLE_1=DConf 2018 Programme &amp; Open Registration
DBLOG_LATEST_LINK_1=https://dlang.org/blog/2018/03/19/dconf-2018-programme-open-registration/
DBLOG_LATEST_DATE_1=March 19, 2018
DBLOG_LATEST_AUTHOR_1=Michael Parker
DBLOG_LATEST_TITLE_2=User Stories: Funkwerk
DBLOG_LATEST_LINK_2=https://dlang.org/blog/2018/03/14/user-stories-funkwerk/
DBLOG_LATEST_DATE_2=March 14, 2018
DBLOG_LATEST_AUTHOR_2=Michael Parker
```

### DIFF
--- a/dblog_feed_example.xml
+++ b/dblog_feed_example.xml
@@ -1,0 +1,167 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?xml-stylesheet type="text/xsl" media="screen" href="/~d/styles/rss2full.xsl"?><?xml-stylesheet type="text/css" media="screen" href="http://feeds.feedburner.com/~d/styles/itemcontent.css"?><rss xmlns:content="http://purl.org/rss/1.0/modules/content/" xmlns:wfw="http://wellformedweb.org/CommentAPI/" xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:atom="http://www.w3.org/2005/Atom" xmlns:sy="http://purl.org/rss/1.0/modules/syndication/" xmlns:slash="http://purl.org/rss/1.0/modules/slash/" xmlns:feedburner="http://rssnamespace.org/feedburner/ext/1.0" version="2.0">
+
+<channel>
+	<title>The D Blog</title>
+	
+	<link>https://dlang.org/blog</link>
+	<description>The official blog for the D Programming Language.</description>
+	<lastBuildDate>Mon, 19 Mar 2018 13:27:32 +0000</lastBuildDate>
+	<language>en-US</language>
+	<sy:updatePeriod>hourly</sy:updatePeriod>
+	<sy:updateFrequency>1</sy:updateFrequency>
+	<generator>https://wordpress.org/?v=4.9.4</generator>
+<site xmlns="com-wordpress:feed-additions:1">112438176</site>	<atom10:link xmlns:atom10="http://www.w3.org/2005/Atom" rel="self" type="application/rss+xml" href="http://feeds.feedburner.com/OfficialDBlog" /><feedburner:info uri="officialdblog" /><atom10:link xmlns:atom10="http://www.w3.org/2005/Atom" rel="hub" href="http://pubsubhubbub.appspot.com/" /><feedburner:browserFriendly></feedburner:browserFriendly><item>
+		<title>DConf 2018 Programme &amp; Open Registration</title>
+		<link>https://dlang.org/blog/2018/03/19/dconf-2018-programme-open-registration/</link>
+		<comments>https://dlang.org/blog/2018/03/19/dconf-2018-programme-open-registration/#respond</comments>
+		<pubDate>Mon, 19 Mar 2018 13:27:32 +0000</pubDate>
+		<dc:creator><![CDATA[Michael Parker]]></dc:creator>
+				<category><![CDATA[D Foundation]]></category>
+		<category><![CDATA[DConf]]></category>
+
+		<guid isPermaLink="false">http://dlang.org/blog/?p=1493</guid>
+		<description><![CDATA[Registration for DConf 2018 Munich, May 2-5, is open! 21 hours of engaging presentations and panels, an all-day Hackathon, and intelligent conversation with members of the D programming language community! We have keynotes from language stewards Walter Bright and Andrei Alexandrescu, along with our invited guest from academia Martin Odersky, professor at EPFL and creator of the Scala language. Whether you know nothing of DLang or you're a seasoned pro, we invite you to spend 4 days with us to partake in the fun, the education, and the German beer. Check the programme and register today!]]></description>
+		<wfw:commentRss>https://dlang.org/blog/2018/03/19/dconf-2018-programme-open-registration/feed/</wfw:commentRss>
+		<slash:comments>0</slash:comments>
+	<post-id xmlns="com-wordpress:feed-additions:1">1493</post-id>	</item>
+		<item>
+		<title>User Stories: Funkwerk</title>
+		<link>https://dlang.org/blog/2018/03/14/user-stories-funkwerk/</link>
+		<comments>https://dlang.org/blog/2018/03/14/user-stories-funkwerk/#comments</comments>
+		<pubDate>Wed, 14 Mar 2018 14:06:32 +0000</pubDate>
+		<dc:creator><![CDATA[Michael Parker]]></dc:creator>
+				<category><![CDATA[Code]]></category>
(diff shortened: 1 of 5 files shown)


posix.mak
```diff
@@ -592,21 +592,12 @@ $G/dlangspec.verbatim.txt : $G/dlangspec-consolidated.d $(DDOC_BIN) verbatim.ddo
...
-	cat $< | grep -m2 'entry-title' | tail -n1 | \
-		sed -E 's/^.*<a href="(.+)" rel="bookmark">([^<]+)<\/a>.*<time.*datetime="[^"]+">([^<]*)<\/time><time class="updated".*Author *<\/span><a [^>]+>([^<]+)<\/a>.*/DBLOG_LATEST_TITLE2=\2|DBLOG_LATEST_LINK2=\1|DBLOG_LATEST_DATE2=\3|DBLOG_LATEST_AUTHOR2=\4/' | \
-		tr '|' '\n' > $(basename $@)2.ddoc
-	@if [ ! -s $(basename $@)2.ddoc ] ; then echo "$(basename $@)2.ddoc is empty. Please check the download"; rm -f $(basename $@)2.ddoc; exit 1 ; fi
-
-$G/dblog_latest2.ddoc: $G/dblog_latest.ddoc
+	curl -s --fail --retry 3 --retry-delay 5 -L http://feeds.feedburner.com/OfficialDBlog -o $@
+
+$G/dblog_latest.ddoc: $G/dblog_latest.xml $(STABLE_DMD) tools/ddoc_xml_extractor.d
+	$(STABLE_RDMD) tools/ddoc_xml_extractor.d -i $< -o $@
```
wilzbach: I used `--output` to avoid creating the file if an error happened

CyberShadow:
> This won't work on DAutoTest, but the generated Ddoc file will be e.g.

One thing we could do is in DIFFABLE mode, use the same code path, but process a dummy static file. That'll allow testing the conversion on CI.

tools/ddoc_xml_extractor.d
```diff
@@ -0,0 +1,86 @@
...
+
+auto toDdoc(const ref Article article, string suffix)
+{
+    string text;
+with(article)
+    text =
+`DBLOG_LATEST_TITLE%5$s=%1$s
+DBLOG_LATEST_LINK%5$s=%2$s
+DBLOG_LATEST_DATE%5$s=%3$s
+DBLOG_LATEST_AUTHOR%5$s=%4$s`.format(title, link, date, author, suffix);
```
CyberShadow: Any of these need HTML/XML escaping?

wilzbach: It looks to me like they are already escaped:

```
<title>DConf 2018 Programme &amp; Open Registration</title>
<link>https://dlang.org/blog/2018/03/19/dconf-2018-programme-open-registration/</link>
<comments>https://dlang.org/blog/2018/03/19/dconf-2018-programme-open-registration/#respond</comments>
<pubDate>Mon, 19 Mar 2018 13:27:32 +0000</pubDate>
<dc:creator><![CDATA[Michael Parker]]></dc:creator>
```

wilzbach:
> One thing we could do is in DIFFABLE mode, use the same code path, but process a dummy static file. That'll allow testing the conversion on CI.

Done.

tools/ddoc_xml_extractor.d
CyberShadow: > `<title>DConf 2018 Programme &amp; Open Registration</title>`

Won't std.xml parse that `&amp;` into an `&`, meaning we now have to escape it back into an `&amp;` ?

dblog_feed_example.xml
```diff
@@ -0,0 +1,167 @@
...
+		<pubDate>Mon, 12 Mar 2018 14:20:54 +0000</pubDate>
+		<dc:creator><![CDATA[Michael Parker]]></dc:creator>
+				<category><![CDATA[Community]]></category>
+		<category><![CDATA[D Foundation]]></category>
+
+		<guid isPermaLink="false">http://dlang.org/blog/?p=1441</guid>
+		<description><![CDATA[In its work guiding the development of D and promoting its adoption, the D Language Foundation is driven primarily by donations big and small. The money comes in from different sources, the most visible being those listed on the website’s donation page, and is put to use in different ways. Today, the D Language Foundation is opening a new chapter in the donation story.]]></description>
+		<wfw:commentRss>https://dlang.org/blog/2018/03/12/the-d-foundation-at-open-collective/feed/</wfw:commentRss>
+		<slash:comments>0</slash:comments>
+	<post-id xmlns="com-wordpress:feed-additions:1">1441</post-id>	</item>
```
CyberShadow: Cool, thanks! We only need at most two items, though.

CyberShadow:
So, how will this behave in case something fails?

dblog_feed_example.xml
wilzbach: I thought if we already test this, then we should do it with an actual real world example.

wilzbach:
> So, how will this behave in case something fails?

It will exit with a non-zero exit code before writing the file (hence `--output`) which means the entire deployment will fail.
This doesn't apply for HTML escaping errors though, but for everything else related to extracting the data from the XML.

tools/ddoc_xml_extractor.d
wilzbach: Yes you are right though I don't think we have HTML escaping out-of the box in Phobos, do we?
One could also make the point that dlang.org uses `<meta charset="utf-8">` and HTML escaping for non-ASCII characters isn't required.

CyberShadow: `std.xml.encode` ?

CyberShadow: >  One could also make the point that dlang.org uses <meta charset="utf-8"> and HTML escaping for non-ASCII characters isn't required.

Right, not unicode, just `&amp;` and `&gt;`

wilzbach: Ah nice! Didn't know about that one. Done!

CyberShadow:
> from Mon, 19 Mar 2018

Oof, I don't suppose we could easily use a more human-friendly date format there...

wilzbach:
> Oof, I don't suppose we could easily use a more human-friendly date format there...

Well, std.datetime doesn't support this :/
Am I allowed to use http://code.dlang.org/packages/datefmt?

jmdavis:
`std.datetime.systime.parseRFC822DateTime` would handle that particular monstrosity it if it had a time zone on it, but it doesn't, so it won't.

wilzbach:
It does have a timezone attached (see the XML), but I think @CyberShadow was asking about a more human-friendly output like "March" instead of "Mar" or "Monday" instead of "Mon".

CyberShadow:
> Am I allowed to use http://code.dlang.org/packages/datefmt?

Blah, that package has nothing on `ae.utils.time` :P

Formatting is actually much easier than parsing, could be done by hand. No library needed.

wilzbach:
> Formatting is actually much easier than parsing, could be done by hand. No library needed.
> Yeah, formatting is generally

I agree and `std.format` handles most of this, but for things like `weekday.toPretty` or `month.toPretty` I just feel like I'm not the first guy needing to do this conversion.

Anyhow, I now added a hard-coded month "prettification", s.t. we can emulate the current output on dlang.org.

Anything else required from this PR?

wilzbach:
(Updated)

wilzbach:
Going to merge this now as the RegExp seemed to have failed again:

image

tools/ddoc_xml_extractor.d
```diff
@@ -0,0 +1,101 @@
...
+}
+
+auto toDdoc(const ref Article article, string suffix)
+{
+    import std.xml : encode;
+    string text;
+with(article)
+{
+    auto prettyMonth = ["January", "February", "March", "April", "May", "June", "July",
+    "August", "September", "October", "November", "December"][date.month-1];
```
CyberShadow: Woah, what happened to the indentation here

wilzbach: -> https://github.com/dlang/dlang.org/pull/2305

PR #2286, opened 2018-03-20
Remove auto-cloning of Phobos on start

wilzbach:
Auto-cloning is useful, but not if you run `make -f posix.mak clean`.

This should now be possible due to https://github.com/dlang/dlang.org/pull/2069

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -155,6 +155,8 @@ TOOLS_DIR=../tools
 INSTALLER_DIR=../installer
 DUB_DIR=../dub
 
+# Auto-cloning missing directories
+$(shell [ ! -d $(DMD_DIR) ] && git clone --depth=1 ${GIT_HOME}/dmd $(DMD_DIR))
 include $(DMD_DIR)/src/osmodel.mak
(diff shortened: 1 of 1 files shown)


wilzbach:
CC @CyberShadow - I think you are going to like this ;-)

CyberShadow:
IIRC this (or a similar change?) was added there because of some difficulties with calculating dependencies. With that in mind do you think this could break anything?

wilzbach:
> IIRC this (or a similar change?) was added there because of some difficulties with calculating dependencies. With that in mind do you think this could break anything?

Yeah you are right. I even added it myself after someone complained about the build process failing due to non-existent repos: https://github.com/dlang/dlang.org/pull/1827

Anyhow - I just tested this with a build from scratch and it looks like everything is working fine and we only need to depend on DMD with the `shell` trick as `include osmodel.mak` doesn't work otherwise.
`PHOBOS_LATEST_FILES` is a relict from the times where we copied Phobos entirely for the writeln pipeline.

PR #2288, opened 2018-03-21
Fix unrelated DAutoTest diff changes

wilzbach:
The removal in https://github.com/dlang/dlang.org/pull/2268 was too eager.

See also: https://github.com/dlang/phobos/pull/6321


wilzbach:
@CyberShadow the respective Phobos PR is now merged and will be probably be part of 2.079.1
If you want the unrelated changes until 2.079.1 to gone, this PR would remove them. If this is too much churn, for little value, please close.
So either close or merge - really no hard feelings.

CyberShadow:
Thank you!

PR #2289, opened 2018-03-24
Add a pointer to dstep

wilzbach:
Imho htod shouldn't be documented on dlang.org, because

- it's not part of the release archives
- it's closed source
- it's Windows-only
- there are more modern tools like dstep

As removing the documentation might be a bit drastic, I went with adding a warning.

### DIFF
--- a/htod.dd
+++ b/htod.dd
@@ -2,9 +2,8 @@ Ddoc
 
 $(D_S htod,
 
-$(MESSAGE_BOX orange, $(B Warning) - $(D htod) is no longer considered a core tool and maintained
-    by the same people. More modern tools like $(LINK2 https://github.com/jacob-carlborg/dstep, dstep)
-    might work better for your use case.)
+$(MESSAGE_BOX orange, $(B Warning) - `htod` is not developed actively. Alternatives include 
+    $(HTTPS github.com/jacob-carlborg/dstep, dstep).)
 
 $(P     While D is binary compatible with C code, it cannot
         compile C code nor C header files. In order for
--- a/htod.dd
+++ b/htod.dd
@@ -2,6 +2,10 @@ Ddoc
 
 $(D_S htod,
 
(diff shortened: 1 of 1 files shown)


wilzbach:
Ping @CyberShadow and @andralex . So can we all agree that `htod` is super-outdated?
Maybe we can even remove it from the docs?

BTW funnily in the meantime Atila even posted his D <include> tool ...

andralex:
@atilaneves include a link to your tool as well?

CyberShadow:
Please link to the wiki instead: 
https://wiki.dlang.org/Bindings#Binding_generators

andralex:
@CyberShadow good idea feel free to amend

htod.dd
```diff
@@ -2,9 +2,8 @@ Ddoc
 
 $(D_S htod,
 
-$(MESSAGE_BOX orange, $(B Warning) - $(D htod) is no longer considered a core tool and maintained
-    by the same people. More modern tools like $(LINK2 https://github.com/jacob-carlborg/dstep, dstep)
-    might work better for your use case.)
+$(MESSAGE_BOX orange, $(B Warning) - `htod` is not developed actively. Alternatives include
```
CyberShadow: Please see $(LINK2 https://wiki.dlang.org/Bindings#Binding_generators, Binding generators on the D Wiki) for a list of alternatives.

atilaneves:
I added dpp to the wiki.

jacob-carlborg:
FYI htod is the first tool on that list in the wiki.

wilzbach:
@jacob-carlborg I know, but logging to the wiki currently doesn't work for me:

image

Maybe someone else can update the list (i.e. making sure that dstep + dpp are at the top)?

jacob-carlborg:
> Maybe someone else can update the list

Done.

PR #2291, opened 2018-03-24
Add D's docker image to the third-party downloads

wilzbach:
@lindt's docker images make using DMD on a platform without DMD really easy, so imho we should add the instructions to the Third-party downloads.

See also: https://hub.docker.com/r/dlanguage/dmd/


CyberShadow:
How far into the future do we expect these to keep working?

wilzbach:
> Does updating this require manual intervention for each major and point release? 

The docker builds from @lindt are almost automated and have been there for a long time: 
https://hub.docker.com/r/dlanguage/dmd/tags

AFAICT they only require pushing the new tag to GH.

>  Any contingency plans when the maintainer loses interest?

Hopefully someone else picks up the torch. Otherwise we can always easily remove them from the third-party downloads.
Though there is already a third-party downloads that has stopped to being updated (Chocolatey is at 2.073).

In this case, I think Funkwerk is actively using these Docker image within their D infrastructure/setup.

CyberShadow:
> Hopefully someone else picks up the torch.

Does the D foundation have access to the account in case that becomes necessary?


> Though there is already a third-party downloads that has stopped to being updated (Chocolatey is at 2.073).

We should remove it. Chocolatey is pretty problematic as a distribution platform anyway, they have some pretty shady practices.

wilzbach:
Regarding branding - see https://www.docker.com/brand-guidelines

>Do:
> Use the Docker logo to show your service is using Docker or is compatible with Docker
>Do not:
> Use the Docker logo without the letters 'docker'

Ugh - that just looks stupid with letters when it's so tiny :/
(updated)

wilzbach:
> Does the D foundation have access to the account in case that becomes necessary?

As far as I understand only write access to the GitHub repository (https://github.com/lindt/docker-dmd) is needed for tagging new releases.
If this ever becomes a problem, I think we could move it to dlang-community.

FWIW I'm currently experimenting with an fully automated setup that requires no interaction at all as it's based on the install script and uses Travis's daily crons: https://github.com/wilzbach/dlang-docker
(we currently use that one for DMD's CircleCI image, but apart from that it's rather experimental)

lindt:
we could either move/merge it together during the hackathon or if we find time after the D lightning talks.

wilzbach:
@CyberShadow rebased as #2292 is now in.

> we could either move/merge it together during the hackathon or if we find time after the D lightning talks.

Yeah let's talk on Thursday! My main use case for the docker images I built was using it for DMD's CircleCI. CircleCi doesn't offer Ubuntu 17.10 images (we wanted to have one CI with PIC hardening) and the way they build their images is a bit complicated because they patch their custom scripts and ssh code into the image which I needed to reproduce for the "Login via SSH" feature to work.

wilzbach:
Nevertheless I think we can already go ahead and merge this. Are you okay with this as well @CyberShadow? I think you said so on the other PR, but this is still lacking a GH approval.

Preview:

image

PR #2292, opened 2018-03-24
Add snap to the third-party downloads

wilzbach:
CC @WebDrake - maybe there's a better page than LaunchPad to link to?


WebDrake:
Will this do? https://snapcraft.io/dmd

Note that there is a known issue with the dmd snap: `rdmd` will fail if the D compiler it's invoking is itself snap-packaged (with the exception of `dmd`, where the fallback look-in-the-same-dir-first behaviour means that things Just Work).  This should only affect use-cases where someone uses the `--compiler` flag _and_ has the other compiler also installed as a snap.

The problem arises because rdmd passes a file descriptor to the compiler to serve as its "stdout", and apparmor blocks the snap-packaged D compiler from inheriting that file descriptor.

More details here:
https://forum.snapcraft.io/t/failed-to-flush-stdout-permission-denied/485
https://issues.dlang.org/show_bug.cgi?id=17838

WebDrake:
BTW, note the contact email address.  Any chance of getting a `snapcraft@dlang.org` email forwarder that I could use instead?

WebDrake:
These are all rather functional action item responses on my part.  I should add: thanks very much for wanting to add the snap package here!

Note that there is also a snap package for `dub`: https://snapcraft.io/dub

... and that both these packages should work effectively on (at least) Ubuntu, Debian, OpenSUSE, Arch, Solus, and others.  See: https://docs.snapcraft.io/core/install

wilzbach:
> Will this do? https://snapcraft.io/dmd

Yes. I was exactly looking for this page, but somehow couldn't find it.
I assume that `--classic` is no longer needed then?

> Any chance of getting a snapcraft@dlang.org email forwarder that I could use instead?

I don't control the dlang.org domain, Jan does. I do down dlang.io (and dlang.rocks + dlang.science), so I could easily make snapcraft@dlang.io to forward to you?

> Note that there is a known issue with the dmd snap: rdmd will fail if the D compiler it's invoking is itself snap-packaged (with the exception of dmd

Wow that's a long issue. Well, now that `-i` is part of `dmd` this isn't such a huge problem as before.

But to answer your question:

> So: can rdmd be either patched or invoked in order to avoid the D compiler it invokes needing to inherit a file descriptor from it? :-)

Yes it can. With the improvements in 2.079, DMD now supports JSON output via stdout (see e.g. https://run.dlang.io/is/DiSyCl).
That's what https://github.com/dlang/tools/pull/292 was supposed to do.

> Note that there is also a snap package for dub: https://snapcraft.io/dub

Hmm, is it possible to create meta-packages with snap? Because for ArchLinux and Gentoo we just recommend the `dlang` group - that might be an option. Otherwise, maybe "... install dmd dub" might work too (Ubuntu uses this too).
In any case I think you should make a PR to this README: https://github.com/dlang/dub

WebDrake:
>  I assume that `--classic` is no longer needed then?

If you type the instructions provided on the store page, you will get this response:

```
error: This revision of snap "dmd" was published using classic confinement and thus may perform
       arbitrary system changes outside of the security sandbox that snaps are usually confined to,
       which may put your system at risk.

       If you understand and want to proceed repeat the command including --classic.
```

I suspect it's quite deliberate on their part to not include the `--classic` flag in the store-side instructions, to ensure that people always see that warning.

> Wow that's a long issue. Well, now that -i is part of dmd this isn't such a huge problem as before.

I'm afraid that is contingent on the other compiler supporting that flag.  The snap-packaged `rdmd` should be able to work with _any_ snap-packaged D compiler, not just one that exposes `-i`.  (I forget: which frontend release was `-i` added?)

Note that this does NOT affect using `rdmd` with the `dmd` from the same snap package.  That's fine.

It really is an upstream issue (a bug in classic confinement), hence why I didn't push more strongly on it with rdmd itself; I couldn't think of an obvious and non-intrusive rdmd-side workaround, and no one else responded to the issue I raised, so it seemed best to leave it for the snapd devs to address.

> Hmm, is it possible to create meta-packages with snap?

No.  I did ask about this a while back and it looks like it's not a major concern for them right now.

> Otherwise, maybe "... install dmd dub" might work too (Ubuntu uses this too).

Unlike `apt`, `snap` expects a single package name at a time.

posix.mak
```diff
@@ -308,7 +308,7 @@ IMAGES=favicon.ico $(ORGS_USING_D) $(addprefix images/, \
 		d3 dconf_logo_2018 debian_logo dlogo fedora_logo freebsd_logo gentoo_logo homebrew_logo \
 		index-banner nix_logo \
 		opensuse_logo ubuntu_logo windows_logo pattern github-ribbon \
-		dlogo_opengraph \
+		dlogo_opengraph snapcraft_logo \
```
WebDrake: Have you checked/confirmed the obligations/permissions for using the snapcraft logo?

wilzbach:
> I suspect it's quite deliberate on their part to not include the --classic flag in the store-side instructions, to ensure that people always see that warning.

This is the default security policy applied to snaps. The snap has read and/or write rights only in its own install space and selected areas. It has access to libraries it bundles and/or provided by the core or ubuntu-core snap. Extended rights can be granted with interfaces, which are connected either at install time or by the user with the snap connect command. For example, the home interface will grant read rights in the user’s home.

(from: https://docs.snapcraft.io/reference/confinement)

Yeah looks like they want to promote their "interfaces"

>No. I did ask about this a while back and it looks like it's not a major concern for them right now.
> ...
> Unlike apt, snap expects a single package name at a time.

So the only way would be a separate line with `snap install --classic dub`?
Not sure that's really worth it.

CyberShadow:
As with #2291, how far into the future are these expected to keep working? Does updating this require manual intervention for each major and point release? Any contingency plans when the maintainer loses interest?

WebDrake:
>  how far into the future are these expected to keep working? Does updating this require manual intervention for each major and point release?

Snaps are released in channels (`stable`, `candidate`, `beta`, `edge`).  Every time the channel gets updated, all users should automatically receive the update (unless they've explicitly disabled it).  You can also run `snap refresh` to manually update, but this is rarely necessary.

For the compiler snaps (not dub), I've also made use of snap _tracks_: these are associated with successive minor releases of the compiler, and receive patch releases only.  So if you wanted to stay on (say) the 2.079.* release series, you could do:

```
sudo snap {install,refresh} --classic --channel=2.079/stable dmd
```

... and you would then receive only updates to 2.079.  In other words, you receive updates to the `stable` channel of the `2.079` track.

(Track names are arbitrary, but this track-per-minor-release-series is a reasonably common pattern for snap package maintainers to use, and the snap devs encourage it.)

Actual update of the package involves minor tweaks to the `snapcraft.yaml` file:
https://github.com/dlang-snaps/dmd.snap/pull/20/files

Right now things are set up so that each minor-release-series branch is tracked by Launchpad, and each time that branch is updated, a new package is generated and pushed to the `edge` channel of the corresponding track.

That does mean that right now it's necessary to set up a new minor-release-series build config each time a new minor release comes out, but that is easy:
https://code.launchpad.net/~dlang-snaps/+snap/dmd-2.077
https://code.launchpad.net/~dlang-snaps/+snap/dmd-2.078
https://code.launchpad.net/~dlang-snaps/+snap/dmd-2.079

> Any contingency plans when the maintainer loses interest?

@MartinNowak should have login/admin access to the snap-store account, the Launchpad group and the GitHub team.  I'm happy to extend that to others as is wished, and to mentor anyone who wants to know in how the snap packages are maintained.

CyberShadow:
Sounds good, thanks!

Speaking from experience, updating something on each new release can get tedious pretty quickly, so I suggest to automate as much as possible.

posix.mak
wilzbach: They don't provide any branding guidelines (I couldn't find any when I looked).
Anyhow, it's generally allowed to use trademarked logos if the customer isn't confused or a negative image is created towards the company.
Now that snapcraft itself is GPL licensed and we do "free advertisement" for them, I doubt that this is a problem. After all, they want that people use their logo like e.g. here: https://hexchat.github.io/downloads.html

wilzbach:
> So I'll let @wilzbach decide the merge order.

I would wait at #2291 for the feedback of Stefan (the maintainer of the Docker builds), so I would move ahead with this one first (though I really don't care. They merge conflict will be trivial).

Any last words from your side @WebDrake?

download.dd
```diff
@@ -190,6 +190,8 @@ sudo apt-get update && sudo apt-get install dmd-compiler dub)
 
 $(DOWNLOAD_OTHER $(OPENSUSE), OpenSUSE Tumbleweed, $(CONSOLE sudo zypper install dmd))
 
+$(DOWNLOAD_OTHER $(SNAP), $(LINK2 https://snapcraft.io/dmd, Snap), $(CONSOLE sudo snap install --classic dmd))
```
WebDrake: Given that you have explicit instructions for the APT repo above for separate dmd and dub install, why not either:

* a separate line for DMD snap and DUB snap, or

* `sudo snap install --classic dmd && sudo snap install --classic dub`

.... ?

WebDrake:
Other than a couple of suggestions for how to mention the separate DMD and DUB snaps, all good from my side.  Thanks!

PR #2294, opened 2018-03-24
Remove -fsS from the recommended flags to use for the install.sh script

wilzbach:
Motivation:

- these flags are only useful for scripts, not human usage
- I have seen actual human beings blindly typings these flags because we recommend them
- For server usage more flags would be needed anyways e.g. at least `--retry 3 --retry-delay 5`
- These flags aren't well-known and thus we somewhat require people to look them up

Imho we should only recommend flags that are useful in all cases.

As the usage of install.sh in CI scripts is more complicated, I added a dedicated section to its documentation.

So, what the heck does `-fsS` do?
-----------------------------

Exactly! One of the reasons why we shouldn't display it by default.
Anyhow, this saves you a trip to the man page:

```
-S, --show-error
	  When  used  with -s, --silent, it makes curl show an error mes‐
	  sage if it fails.

-s, --silent
	  Silent or quiet mode. Don't show progress meter or  error  mes‐
	  sages.   Makes Curl mute. It will still output the data you ask
	  for, potentially even to the terminal/stdout unless  you  redi‐
	  rect it.

	  Use  -S,  --show-error  in  addition  to this option to disable
	  progress meter but still show error messages.


-f, --fail
	  (HTTP) Fail silently (no output at all) on server errors.  This
	  is mostly done to better enable scripts etc to better deal with
	  failed attempts. In normal cases when an HTTP server  fails  to
	  deliver  a  document,  it  returns  an HTML document stating so
	  (which often also describes why and more). This flag will  pre‐
	  vent curl from outputting that and return error 22.

	  This method is not fail-safe and there are occasions where non-
	  successful response codes will slip  through,  especially  when
	  authentication is involved (response codes 401 and 407).
```

### DIFF
--- a/download.dd
+++ b/download.dd
@@ -88,7 +88,7 @@ $(DIVC download_channels,
          $(SBTN $(ARCH freebsd-32, tar.xz), i386) $(SBTN $(ARCH freebsd-64, tar.xz), x86_64)
     )
 
-    $(INSTALL_SCRIPT curl -fsS https://dlang.org/install.sh | bash -s dmd)
+    $(INSTALL_SCRIPT curl https://dlang.org/install.sh | bash -s dmd)
   )
 
   $(BETA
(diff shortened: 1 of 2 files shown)


wilzbach:
Ping @CyberShadow and @MartinNowak. This has been in here for three weeks now without a comment...

CyberShadow:
I don't think this is a good idea. Without these flags, curl will "succeed" if there is an error.

CyberShadow:
Particularly, without `-f`, if the website or an intermediary generates an error page, that will be piped into `bash`. That would be very bad.

wilzbach:
> Particularly, without -f, if the website or an intermediary generates an error page, that will be piped into bash. That would be very bad.

OK that's a fair point, but there's no need for the silent mode (`-s`) when people use `curl` locally, is there?
(and if no silent mode is used, then `-S` isn't needed either)

PR #2296, opened 2018-03-25
Add $(TAB) macro

wilzbach:
From:

https://github.com/dlang/tools/pull/336#discussion_r176952807

CC @CyberShadow

### DIFF
--- a/html.ddoc
+++ b/html.ddoc
@@ -435,4 +435,5 @@ HEARTS = &hearts;
 DIAMS = &diams;
 LANG = &lang;
 RANG = &rang;
+TAB = &#09;
 _=


html.ddoc
```diff
@@ -435,4 +435,5 @@ HEARTS = &hearts;
 DIAMS = &diams;
 LANG = &lang;
 RANG = &rang;
+TAB = &#09;
```
CyberShadow: This should probably not be at the end of this list though, which seems to track another one (https://dlang.org/entity.html).

wilzbach: -> https://github.com/dlang/dlang.org/pull/2297

PR #2298, opened 2018-03-26
Fix quotations in struct.dd

RazvanN7:


wilzbach:
@quickfur FYI: DAutoTest seems to be extremely busy these days, but usually it does get to build the PR ... eventually at least :P

PR #2299, opened 2018-03-26
Add postblit documentation for qualified structs

RazvanN7:


spec/struct.dd
```diff
@@ -831,6 +831,75 @@ void main()
...
 
+        $(P The postblits `__fieldPostblit` and `__aggrPostblit`
+        are generated without any implicit qualifiers and are not considered
+        struct members. This leads to the situation where qualifying an
+        entire struct declaration does not have any impact on the above
+        mentioned postblits. However, since `__xpostblit` is a member of
+        the struct and an alias of one of the other postblits, the qualifiers
+        applied to the struct will affect the aliased postblit:)
+
+---
```
wilzbach: `$(SPEC_RUNNABLE_EXAMPLE_FAIL`

spec/struct.dd
```diff
@@ -831,6 +831,75 @@ void main()
...
+    S a;        // error : mutable method B.__fieldPostblit is not callable using a const object
+    this(this)
+    { }
+}
+
+// `__xpostblit` aliases the field postblit; no error
+const struct B2
+{
+    S a;
+}
```
wilzbach: Maybe worth splitting up into multiple tests (?), s.t. you can use `SPEC_RUNNABLE_EXAMPLE_COMPILE`

spec/struct.dd
andralex: @RazvanN7 can you please add this when you get the chance?

PR #2304, opened 2018-03-27
DConf front page: registration is open

wilzbach:


wilzbach:
CC @CyberShadow this is a rather time-sensitive item for the front page.

CyberShadow:
It's still a bit hidden in a corner, should it be promoted to a banner higher at some point?

wilzbach:
Yeah why not :)
Do you want to use something similar to the "State of D" banner?
We could even make the "D" move a bit (but from the feedback on the proposed [dconf18](https://wilzbach.github.io/dconf18/) website, I'm guessing this wouldn't be popular)

CC @mdparker

CyberShadow:
> We could even make the "D" move a bit

**NO**. Annoying as hell when you're trying to read something else on the same page.

PR #2290, opened 2018-03-24
Modernize dcompiler documentation

wilzbach:
A few small improvements to the respective dmd.html files.

PR #2306, opened 2018-03-27
Fix Issue 2473, 10941 - Linux system install instructions incorrect

wilzbach:
The dmd.html page needs serious work and apparentely it always has been like this since 2008!
I wasn't sure on the best way to fix this because the shipped `dmd.conf` looks like this:

```
[Environment32]
DFLAGS=-I%@P%/../../src/phobos -I%@P%/../../src/druntime/import -L-L%@P%/../lib32 -L--export-dynamic

[Environment64]
DFLAGS=-I%@P%/../../src/phobos -I%@P%/../../src/druntime/import -L-L%@P%/../lib64 -L--export-dynamic -fPIC
```

In other words, the shipped `dmd.conf` can't be simply copied to `/etc/`. This means we have the following solutions

a) just tell the user to add it to his `PATH`
b) advertise to use e.g. a symlink to `/usr/bin/dmd`
c) provide an alternative `dmd.conf` for copy/pasting to `/etc/dmd.conf`
d) provide a command to patch the shipped `dmd.conf`
e) suggest to move the entire folder to `/usr/local` and symlink `dmd` then
...

(the FreeBSD + MacOS sections of this need similar adjustments, but I left them for now until I know how to proceed here)

See also: https://github.com/dlang/dlang.org/pull/2290

### DIFF
--- a/dcompiler.dd
+++ b/dcompiler.dd
@@ -85,28 +85,16 @@ C:\&gt;unzip dmc.zip
 
     $(INSTALLATION_SCRIPT_HINT)
 
-    $(LI Unzip the archive into your home directory.
     It will create
     a $(D ~/dmd) directory with all the files in it.
+
     All the tools are command line tools, which means
     they are run from a console window.)
-
-    $(LI Copy $(D $(DMD_CONF)) to $(D /etc):
-
+    $(LI All tools can be run directly from the archive:
 $(CONSOLE
-cp $(DMDDIR)/linux/bin/$(DMD_CONF) /etc
+~/dmd/linux/bin64/dmd
 )
-    )
-
-    $(LI Put $(D $(DMDDIR)/linux/bin) on your $(B PATH),
-    or copy the linux executables
-    to $(D /usr/local/bin))
-
-    $(LI Copy the library to $(D /usr/lib):
-
-$(CONSOLE
-cp $(DMDDIR)/linux/lib/$(LIB) /usr/lib
-)
+    To install $(D dmd) globally, add this bin folder to your $(B PATH).
     )
     )
     $(FREEBSD


quickfur:
Using relative paths in the default dmd.conf makes no sense.  We should just decide on a global path like `/usr/bin`, `/usr/local/bin`, or `/opt/bin`, and just hardcode that into dmd.conf.  Either that, or provide an easy way for users to say "please install to directory xyz" and have the right version of dmd.conf automatically created and installed in the right place for them.

wilzbach:
@quickfur relative paths are necessary - otherwise you can't use the dmd binary without "installing" it, which is a major use case, e.g. the install script which is also used by Travis depends on this.

adamdruppe:
I wouldn't be a D user today if there was insistence on global paths. Relative paths in the config were (and still are, though nowadays I'd just fork it and fix it myself), believe it or not, THE #1 reason I chose Digital Mars back in the day. These compilers *just worked* while the competition wouldn't even start up.

quickfur:
You're right, I said that too brashly.  Actually, even I myself rely on the relative paths. :-D Otherwise my current setup based on git master wouldn't even work.  And you're right, too much reliance on global paths is bad (I tried installing two different versions of gcc side-by-side once... it scarred me for life. It's *possible*, no doubt, but involves black magic that mere mortals have no business dabbling in).

quickfur:
Perhaps a script for installing the correctly-fixed dmd.conf to /etc would be best.

Or just suggest all of the alternatives and let the user choose.

adamdruppe:
There's really no reason to put it in /etc at all. My preference is to just delete these insane instructions and replace them with the real options:

* use the bash script. if it uses /etc, document it there so people know but they don't have to do it themselves.
* use the distro packages. ditto
* use the zip, explaining that it just works in place, no moving/copying necessary.

quickfur:
Though I'd at least put a more detailed description of how installing to /etc works somewhere (not necessarily on this page), should the user prefer to do that. Well, it's more of a documentation of how it would work, rather than something the user would actually do. So that in case somebody screws up his installation, there's at least a google-able document describing how things actually work that he could use to troubleshoot whatever it is he did wrong.

wilzbach:
> So anything I can still do here?

Ping.

PR #2307, opened 2018-03-28
Start to opt-out of Ddoc's automatic keyword highlighting

wilzbach:
Let's whether DAutoTest likes this. In theory
- no changes should show up in the diff
- existing suppression of the auto-highlighting can be gradually removed (e.g. with the SUPPRESS helper or `grep`)
- ddoc should no longer do automatic highlighting (not sure whether we can verify this here, but @JackStouffer just run into this issue again - https://github.com/dlang/phobos/pull/6356)

See also:

- https://dlang.org/changelog/2.079.0.html#fix18361
- https://github.com/dlang/dmd/pull/7834

CC @quickfur

### DIFF
--- a/dlang.org.ddoc
+++ b/dlang.org.ddoc
@@ -552,3 +552,10 @@ PROJECT_SOURCE_DIR=
 
 FA_ICON=<i class="fa fa-$1" aria-hidden="true"></i>
 _=
+
+_ = Opt-out of automatic keyword highlighting - see https://dlang.org/changelog/2.079.0.html#fix18361
+DDOC_AUTO_PSYMBOL =
+DDOC_AUTO_KEYWORD =
+DDOC_AUTO_PARAM =
+_ = DDOC_AUTO_PSYMBOL_SUPPRESS = FIXME_UNDERSCORE_PREFIX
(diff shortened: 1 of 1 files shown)


dlang.org.ddoc
```diff
@@ -552,3 +552,10 @@ PROJECT_SOURCE_DIR=
 
 FA_ICON=<i class="fa fa-$1" aria-hidden="true"></i>
 _=
+
+_ = Opt-out of automatic keyword highlighting - see https://dlang.org/changelog/2.079.0.html#fix18361
+DDOC_AUTO_PSYMBOL =
+DDOC_AUTO_KEYWORD =
+DDOC_AUTO_PARAM =
+_ = DDOC_AUTO_PSYMBOL_SUPPRESS = FIXME_UNDERSCORE_PREFIX
```
andralex: why?

wilzbach: Because DDoc doesn't support comments and `DDOC_AUTO_PSYMBOL_SUPPRESS` is within reach. After all, there are still tons of underscores used to suppress auto highlighting like `_range` references in Phobos.

PR #2310, opened 2018-03-28
posix.mak: Use a wildcard for including images

wilzbach:
We already use a `wilcard` for the organization images and it's getting really annoying to edit the list by hand and resolve the merge conflicts.
Automation for the win!


wilzbach:
How ironic:

image

CyberShadow:
Wildcards in build description files are evil! We should be using them less, not more.

wilzbach:
> Wildcards in build description files are evil! 

And merge conflicts aren't evil too? What's the big problem with this?
I doubt that someone "accidentally" adds an image to `image` that shouldn't be add to `web/images` and we haven't had any problem with the ORG_IMAGES whereas I run into a conflict with this manual list almost weekly.

CyberShadow:
> And merge conflicts aren't evil too?

No, they're just annoying.

> What's the big problem with this?

Explicit file lists in build description files create the sort of redundancy that you want to have. It is too easy to mess up the file list by accidentally deleting or creating files.

An alternative to both that can be used in build systems is dependency discovery, but for HTML it would mean parsing the .html and .css files and looking at which files are referenced in `` tags and `url(...)` CSS properties.

> whereas I run into a conflict with this manual list almost weekly.

Another way to handle this is to have a flat file list, plus a git merge driver that automatically fixes conflicts by sorting the list alphabetically.

If you feel strongly about this, I wouldn't be against letting go of best practices in favor of pragmatism, but it is not going to be an endorsement of the general practice.

wilzbach:
> It would still mean that all you have to do is rebase locally, and git should automatically resolve the conflict.

```
Documentation OK (no changes)
```

-> looks good :)

wilzbach:
Yes good catch.
They were intended to be used for [the orgs page](https://dlang.org/orgs-using-d.html) (e.g. Retina displays), but this never got to my priority list. I will remove them in a quick follow-up.

PR #2315, opened 2018-03-31
[discuss] Use full width for the documentation pages

wilzbach:
So I don't expect that everyone likes this idea, but I just accidentally saw this and imho it looks more structured. This is entirely subjective, so just "up" or "down" vote to show your personal opinion.

Having the menu aligned to the right isn't too uncommon:

- readthedocs (e.g. http://docs.readthedocs.io/en/latest/getting_started.html)
- Python (e.g. https://docs.python.org/3/library/shutil.html)
- scod (http://martinnowak.github.io/bloom/bloom.html)
- Git Book (https://doc.rust-lang.org/book/second-edition/)
- devdocs.io (http://devdocs.io/d/
...

This is just a quick CSS trick to roughly show how it would look. If there's interest, it would probably be done, s.t. there's no padding between the navigation menu and the browsers' right side and s.t. there's a fixed width (e.g. like currently) for the main content.

Current 76em width:

image

Full-width:

image

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -83,6 +83,12 @@ body > .container
     position: relative;
 }
 
+/*full width for the documentation pages*/
+body.std > .container
+{
+    max-width: 100%;
+}
+
 body.doc input,
 body.doc textarea,
 body.std input,


CyberShadow:
We definitely do not want any text long enough to wrap to be full-width. It's bad usability.

PR #2319, opened 2018-03-31
Use full page for search results

wilzbach:
A start and experiment with using the full site for showing the search results.

Preview:

image

This is the initial "proof of concept". When I find more time, I will try to tweak it and make it more "visually appealling". Ideas or examples of other good searches are of course always welcome.

See also:
- https://github.com/dlang/dlang.org/pull/2314 (for using the same search with Ddoc)
- http://www.scala-lang.org/api/current/scala/collection/immutable/Vector.html?search=fun (for a nice search)

BTW in case someone wants to test this locally, the most pleasant way is with:

```
make -f posix.mak apidocs-serve
```

(as then the JS/CSS is always "up-to-date" / needs to manual copy-overs)

### DIFF
--- a/css/ddox.css
+++ b/css/ddox.css
@@ -53,25 +53,36 @@ a.inherited:after { content: url(../images/ddox/inherited.png); padding-left: 3p
 
 #symbolSearch { width: 112pt; }
 
+#symbolSearchResultsContainer {
+	width: 80%;
+	margin: 0 auto;
+	padding: 0.3em;
+	display: none;
+}
+
 #symbolSearchResults {
-	background: #F5F5F5;
-	border: 1px solid #CCC;
 	font-size: small;
 	list-style: none;
-	margin: 0;
 	margin-top: 5px;
-	padding: 0.3em;
-	position: absolute;
-	right: -1px;
-	top: 100%;
-	width: 100%;
-	z-index: 1000;
+	-moz-column-gap: 20px;
+	-webkit-column-gap: 20px;
+	column-gap: 20px;
+	-moz-column-width: 17.5em;
+	-webkit-column-width: 17.5em;
+	column-width: 17.5em;
+	list-style-type: none;
+	font-size: 1.2em;
+}
+
+#symbolSearchCloseButton {
+	float: right;
+	font-size: 1.5em;
 }
 
 #symbolSearchResults li {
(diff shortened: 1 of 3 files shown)


wilzbach:
Okay. I finally got back to this and while this isn't perfect, imho it's a lot better than the dropdown.

Old:

image

New:

image

What do you think?

CyberShadow:
Not sure it's an improvement... it just throws more information in your face, most of which you don't care about, without any organization, and blocks the entire page while doing it. The screenshot in your example isn't of a typical search, I think they will typically be of more than two letters.

wilzbach:
What do other people think about this?

CC @thewilsonator maybe

thewilsonator:
The dropdown is dynamically repopulated, correct? Is the new one?

wilzbach:
> The dropdown is dynamically repopulated, correct? Is the new one?

Yes and yes. Just test it out yourself via the DAutoTest preview. Select one page on the diff and enter things in the search bar.

wilzbach:
New:

image


Old:

image

https://dlang.org/library/core/atomic.html

It only works for the Ddox doc pages.
See https://github.com/dlang/dlang.org/pull/2314 for bringing these auto-complete feature to Ddoc.

thewilsonator:
I think I like the older one more, possibly because I'm used to dropdown menus and the new ones is a bit jarring if you're not expecting that to happen, but not by a whole lot. 

All the new one does for me is having a slightly bigger font size. If it made use of that space e.g:
* had the first sentence of the documentation beneath the symbol link, providing a description of the symbol rather than just its name (useful for unfamiliar symbols (`std.mathspecial` anyone?)), 
* increased the size/usfulness of the icon next to the symbol e.g. I know some of those are e.g classes, but its very hard to see and I'm not sure what some of those icons are supposed to be.

it might have some use over the dropdown. 

I note that pressing enter take you to the page for the top result for the partial search, given the apparently random ordering of the results, I'd be all for making it "expand" to the new one with the extra info mentioned above instead.

wilzbach:
We should probably make the search bar more prominent and similar to e.g. like Rust docs does it:

image

image


https://doc.rust-lang.org/std/env/index.html

RazvanN7:
@wilzbach Seems like people are reluctant to this. Is it ok to close?

PR #2320, opened 2018-03-31
Remove legacy DMD excludes for building the DDox documentation

wilzbach:
2.079 has been released and it looks like we can remove this code :)
Let's see if DAutoTest agrees.


wilzbach:
The div is so big because e.g. `dmd/mscoeff.d` is now part of the menu

PR #2321, opened 2018-04-01
rewrite struct initialization text

WalterBright:
It was incomplete, inaccurate, and needed links to the grammar.

### DIFF
--- a/spec/struct.dd
+++ b/spec/struct.dd
@@ -132,68 +132,94 @@ S s;      // error, cannot initialize unknown contents
 S* p;     // ok, knowledge of members is not necessary
 ---
 
-        $(P They can be used to implement the
+        $(BEST_PRACTICE They can be used to implement the
         $(LINK2 https://en.wikipedia.org/wiki/Opaque_pointer, PIMPL idiom).)
 
 
+$(H2 $(LNAME2 default_struct_init, Default Initialization of Structs))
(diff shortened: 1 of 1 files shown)


wilzbach:
>  @WalterBright  WalterBright requested review from andralex and removed request for andralex 9 hours ago

FYI: We can adapt the CODEOWNERS file if this isn't helpful.

spec/struct.dd
```diff
@@ -132,68 +132,96 @@ S s;      // error, cannot initialize unknown contents
...
-------
-struct S { int a, b, c, d = 7; }
-static S x = { a:1, b:2 };           // c is set to 0, d to 7
-static S z = { c:4, b:5, a:2, d:5 }; // z.a = 2, z.b = 5, z.c = 4, z.d = 5
-------
+        $(P Fields not specified in the $(I StructInitializer) are default initialized.)
 
-$(P C-style initialization, based on the order of the members in the struct
-declaration, is also supported:)
+        $(P Initializers are evaluated at compile time.)
```
wilzbach: They can also be run-time, can't they?

```d
void bar(int b)
{
    struct F {
        int k;
    }
    F f = {k: b};
}
```

WalterBright: Yes, you're right. Fixing.

PR #2328, opened 2018-04-02
[WIP] Ddoc: in-site (Ddox-like) search (chmgen version)

wilzbach:
This is an initial experiment with using the symbols from chmgen for an in-site search.

Cons:
- JSON file is rather large (1.7M, 171K gzipped) - though the JSON file can be loaded lazily, s.t. this shouldn't have an impact
- no information about the symbol (class vs. struct, enum?, member?, field?)

See also: https://github.com/dlang/dlang.org/pull/2314

Todo:
- [ ] filter for duplicates?
- [ ] sort the results in a meaningful way
- [ ] generate a shortname for long names
- [ ] expand the name, e.g. `std_regex#.splitter` should be displayed as such not a just `splitter`

Example:

Chmgen:

image

Ddox's symbols:

image

### DIFF
--- a/css/ddox.css
+++ b/css/ddox.css
@@ -53,52 +53,6 @@ a.inherited:after { content: url(../images/ddox/inherited.png); padding-left: 3p
 
 #symbolSearch { width: 112pt; }
 
-#symbolSearchResults {
-	background: #F5F5F5;
-	border: 1px solid #CCC;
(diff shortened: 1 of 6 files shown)


wilzbach:
Closing in favor of #2314 (using DDox's symbols)

PR #2329, opened 2018-04-03
improve class construction and field initialization verbiage

WalterBright:
Improved wording and added examples.

### DIFF
--- a/spec/class.dd
+++ b/spec/class.dd
@@ -299,16 +299,12 @@ $(GNAME Constructor):
     $(GLINK2 template, ConstructorTemplate)
 )
 
-        $(P Members are always initialized to the
+        $(P Fields are by default initialized to the
         $(LNAME2 class-default-initializer, default initializer)
-        for their type, which is usually 0 for integer types and
-        NAN for floating point types.
-        This eliminates an entire
-        class of obscure problems that come from
-        neglecting to initialize a member in one of the constructors.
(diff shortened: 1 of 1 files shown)


spec/class.dd
```diff
@@ -439,27 +425,30 @@ $(GNAME Constructor):
     $(OL
         $(LI Storage is allocated for the object.
         If this fails, rather than return $(D null), an
-        $(D OutOfMemoryError) is thrown.
+        $(LINK2 ../library/core/exception/out_of_memory_error.html, OutOfMemoryError) is thrown.
```
andralex: @wilzbach do we have a canonical way for such links?

wilzbach: Yup - `$(REF OutOfMemoryError, core, exception)` or `$(REF_ALTTEXT OutOfMemoryError, OutOfMemoryError, core, exception)`  (though that will link to Ddoc, not Ddox which the link currently uses)

andralex: Aha, build failed with:

Error while processing file: /dev/shm/dtest/work/repo/dlang.org/web/dlangspec.html
object.Exception@tools/chmgen.d(588): Attempting to escape site root (dereferencing ../library/core/exception/out_of_memory_error.html from dlangspec.html)

andralex: so... `$(ROOT_DIR)`?

PR #2331, opened 2018-04-05
rewrite struct constructor section

WalterBright:
It was woefully incomplete. Copied a bunch from the recent class construction spec.

### DIFF
--- a/spec/struct.dd
+++ b/spec/struct.dd
@@ -154,6 +154,8 @@ $(H2 $(LNAME2 default_struct_init, Default Initialization of Structs))
 
         $(P The default initializers are evaluated at compile time.)
 
+        $(P The default initializers may not contain references to mutable data.)
+
 $(H2 $(LNAME2 static_struct_init, Static Initialization of Structs))
 
         $(P If a $(GLINK2 declaration, StructInitializer) is supplied, the
@@ -365,7 +367,6 @@ $(H2 $(LEGACY_LNAME2 ConstStruct, const-struct, Const, Immutable and Shared Stru
         $(SPEC_RUNNABLE_EXAMPLE_FAIL
         ----
         const struct S { int a; int b = 2; }
-
         void main()
         {
             S s = S(3); // initializes s.a to 3
@@ -378,23 +379,27 @@ $(H2 $(LEGACY_LNAME2 ConstStruct, const-struct, Const, Immutable and Shared Stru
 
 $(H2 $(LEGACY_LNAME2 Struct-Constructor, struct-constructor, Struct Constructors))
 
-        $(P Struct constructors are used to initialize an instance
-        of a struct.
-        The $(I ParameterList) may not be empty.
-        Struct instances that are not instantiated with a constructor
-        are default initialized to their $(CODE .init) value.
+        $(P Struct constructors are used to initialize an instance of a struct when a more
+        complex construction is needed than is allowed by
+        $(LINK2 #static_struct_init, static initialization) or a
+        $(LINK2 #struct-literal, struct literal).
+        )
+
+        $(P Constructors are defined with a function name of `this` and having no return value.
+        The grammar is the same as for the class $(GLINK2 class, Constructor).
+        )
+
+        $(P A struct constructor is called by the name of the struct followed by
+        $(GLINK2 class, Parameters).
+        $(P If the $(GLINK2 class, ParameterList) is empty,
+        the struct instance is default initialized.)
         )
 
         $(SPEC_RUNNABLE_EXAMPLE_FAIL
-        ------
+        ---
         struct S
         {
-            int x, y;
-
-            this()  // error, cannot implement default ctor for structs
-            {
-            }
-
+            int x, y = 4, z = 6;
             this(int a, int b)
             {
                 x = a;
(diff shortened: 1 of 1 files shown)


spec/struct.dd
```diff
@@ -559,6 +609,124 @@ $(H2 $(LNAME2 disable_default_construction, Disabling Default Struct Constructio
         $(BEST_PRACTICE Disabling default construction is useful when the default value,
         such as `null`, is not acceptable.)
 
+
+$(H2 $(LNAME2 UnionConstructor, Union Constructors))
+
+        $(P Unions are constructed in the same way as structs.)
```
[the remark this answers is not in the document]
WalterBright: Try it and see :-) Redefining what it should do is a topic for another PR.

spec/struct.dd
```diff
@@ -559,6 +609,124 @@ $(H2 $(LNAME2 disable_default_construction, Disabling Default Struct Constructio
...
+            void $(CODE_HIGHLIGHT opAssign)(A rhs) {}
+        }
+        struct S
+        {
+            A val;
+            this(int i)
+            {
+                val = A(i);  // val is initialized to the value of A(i)
+                val = A(2);  // rewritten to val.opAssign(A(2))
+            }
```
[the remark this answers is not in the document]
WalterBright: Specifying it that way would be redundant. All fields are statically initialized before the constructor gets control, so any value the field has is therefore readable and legal.

WalterBright:
@andralex all comments addressed.

spec/struct.dd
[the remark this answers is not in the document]
andralex: Oi.

spec/struct.dd
```diff
@@ -372,108 +378,122 @@ $(H2 $(LEGACY_LNAME2 Struct-Constructor, struct-constructor, Struct Constructors
...
+        }
+        ---
+        )
+        )
 
-$(P Also, member assignments inside loops are never considered constructors,
-even if it can be determined statically that the loop executes at most once.
-Example:)
+        $(LI It is illegal to refer to $(D this) implicitly or explicitly
+        prior to making a delegate constructor call.)
```
RazvanN7: This needs an example. I'm not sure what you are referring to.

spec/struct.dd
```diff
@@ -372,108 +378,122 @@ $(H2 $(LEGACY_LNAME2 Struct-Constructor, struct-constructor, Struct Constructors
...
-{
-    A a;
-    this(int x)
-    {
-        foreach (i; 0 .. x ? 0 : 1) a = i; // error, cannot assign
-    }
-}
-------
+        $(LI Once the delegate constructor returns, all fields are considered
+        constructed.)
```
RazvanN7: I would point out the fact that the compiler does not enforce that the delegate constructor *actually* initializes the field.

spec/struct.dd
```diff
@@ -493,24 +513,70 @@ constructor will have their `.init` values upon destruction.)
...
+            // Based on the definition, this creates a mutable object. But the
+            // created object cannot contain any mutable global data.
+            // Therefore the created object is unique.
+
+            this(int[] arr) immutable pure;
+            // Based on the definition, this creates an immutable object. But
+            // the argument int[] never appears in the created object so it
+            // isn't implicitly convertible to immutable. Also, it cannot store
+            // any immutable global data.
+            // Therefore the created object is unique.
```
RazvanN7: I think this should be reconsidered : https://issues.dlang.org/show_bug.cgi?id=10368 . Consider : 

```d
immutable int i;

struct C
{
    int* p;

    this(int) immutable pure
    { p = &i; }

}
void main()
{
    ++*C(9).p; // changes `i`
}
```
One cannot reliably construct a mutable object with an `immutable pure` constructor.

spec/struct.dd
```diff
@@ -559,6 +609,124 @@ $(H2 $(LNAME2 disable_default_construction, Disabling Default Struct Constructio
         $(BEST_PRACTICE Disabling default construction is useful when the default value,
         such as `null`, is not acceptable.)
 
+
+$(H2 $(LNAME2 UnionConstructor, Union Constructors))
+
+        $(P Unions are constructed in the same way as structs.)
+
+
+$(H2 $(LNAME2 field-init, Field initialization inside constructor))
```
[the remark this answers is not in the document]
RazvanN7: I suggest that the delegate constructor paragraph from above (lines 401-488) be added at this section as a `Restrictions` section.

spec/struct.dd
```diff
@@ -555,10 +622,161 @@ $(H2 $(LNAME2 disable_default_construction, Disabling Default Struct Constructio
...
+
+        $(P In a constructor body, if a delegate constructor is called,
+        all field assignments are considered assignments.
+        Otherwise, the first instance of field assignment is
+        its initialization, and assignments of the form `field = expression`
+        is treated as equivalent to `typeof(field)(expression)`.
```
RazvanN7: "and assignments of the form `field = expression` **are** treated..."

spec/struct.dd
```diff
@@ -555,10 +622,161 @@ $(H2 $(LNAME2 disable_default_construction, Disabling Default Struct Constructio
...
+        The values of fields may be read before initialization or construction
+        with a delegate constructor.
```
RazvanN7: Except if the field is an aggregate declaration with default construction disabled. In that case, the compiler will (or should) issue an error.

spec/struct.dd
```diff
@@ -555,10 +622,161 @@ $(H2 $(LNAME2 disable_default_construction, Disabling Default Struct Constructio
...
+                    default: break;
+                }
+                num = j;        // Error: `case` and `default` are also labels
+            }
+        }
+        ---
+        )
+
+        $(P If a field's type has disabled default construction, then it must be initialized
+        in the constructor.)
```
RazvanN7: Please mention that using the field before initialization results in an error (or if it doesn't then it is a bug)

RazvanN7: After this is merged : https://github.com/dlang/dmd/pull/8201 , the code in the original bug report provides a nice example which, in my opinion, should be present here.

spec/struct.dd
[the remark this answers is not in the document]
RazvanN7: Not if `A` has disabled default construction

spec/struct.dd
RazvanN7: What I mean is that this is valid code:
```d
struct A
{   
    int a, b, c;
    this(int a)
    {
        this.a = a;
    }
 
    this(int a, int b)
    {
        this(a);
    }
}
```

spec/struct.dd
WalterBright: It's a great point, but I am trying to document what is, not what should be. This should be addressed when dealing with 10368.

spec/struct.dd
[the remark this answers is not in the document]
WalterBright: Copy/pasting 87 lines of spec is not the best way to do things. Things should only be specified once.

spec/struct.dd
WalterBright: ok

spec/struct.dd
WalterBright: I'm not sure that change is right. In any case, again, here I am documenting what is, not what it should be.

spec/struct.dd
WalterBright: Again, I am documenting what is, not what should be. Just like we should not mix refactoring code with redesigning it.

PR #2332, opened 2018-04-06

MartinNowak:


download.dd
```diff
@@ -209,10 +209,10 @@ Macros:
...
     DMDV2=$(LATEST)
 
-    BETA=$(COMMENT $0)
-    _=BETA=$0
-    B_DMDV2=2.079.0
-    B_SUFFIX=rc.1
+    _=BETA=$(COMMENT $0)
+    BETA=$0
+    B_DMDV2=2.079.1
+    B_SUFFIX=beta.1
```
wilzbach: Should we auto-detect this too?

MartinNowak: What do you mean by auto-detect?

As this page change is part of a git tag, it has to be commited before the tag is published.
Updating those links is already scripted, so not much need for improvements.
https://gist.github.com/MartinNowak/a471fe7ddbfeef205cdf04c93a94c6d0/8694c361cc255ec2e82114df1a2d8ab219c10ecd#file-release-sh-L67-L76

PR #2335, opened 2018-04-09
Bump Ddox to 0.16.10

wilzbach:
To incorporate e.g. https://github.com/rejectedsoftware/ddox/pull/203

(I just pushed the tag, so this might need a restart if DAutoStart builds this too soon)

### DIFF
--- a/dpl-docs/dub.selections.json
+++ b/dpl-docs/dub.selections.json
@@ -4,18 +4,19 @@
 		"botan": "1.12.9",
 		"botan-math": "1.0.3",
 		"ddoc_preprocessor": {"path":"../ddoc"},
-		"ddox": "0.16.7",
-		"diet-ng": "1.4.3",
+		"ddox": "0.16.10",
+		"diet-ng": "1.4.5",
 		"dmd": {"path":"../../dmd"},
-		"eventcore": "0.8.17",
+		"eventcore": "0.8.34",
 		"hyphenate": "1.1.1",
 		"libasync": "0.8.3",
-		"libdparse": "0.7.1",
+		"libdparse": "0.7.2-alpha.4",
 		"libevent": "2.0.2+2.0.16",
-		"memutils": "0.4.9",
-		"openssl": "1.1.5+1.0.1g",
-		"taggedalgebraic": "0.10.7",
-		"vibe-core": "1.2.0",
-		"vibe-d": "0.8.1"
+		"memutils": "0.4.10",
+		"openssl": "1.1.6+1.0.1g",
+		"stdx-allocator": "2.77.1",
+		"taggedalgebraic": "0.10.11",
+		"vibe-core": "1.4.0",
+		"vibe-d": "0.8.3"
 	}
 }


wilzbach:
Ehm no changes... looking at the logs reveals:

```
.generated/stable_dmd-2.078.2/dmd2/linux/bin64/dub upgrade --missing-only --root=dpl-docs
Upgrading project in /dev/shm/dtest/work/repo/dlang.org/dpl-docs
```

This is done by DAutoTest automatically (not in the Makefile) - see: https://github.com/CyberShadow/ae/blob/58a4e2b26caffd1a4da569c3b17a992483659306/sys/d/manager.d#L1714

Apparently according to this it was a workaround for a dub bug that has been fixed three years ago, so we could remove the manual Makefile overwrite.
OTOH bumping the version manually isn't fun and I wouldn't mind if it's done automatically.

CC @CyberShadow

ZombineDev:
> OTOH bumping the version manually isn't fun and I wouldn't mind if it's done automatically.

I believe upgrading dependencies needs to be a conscious decision (though opening a PR automatically would certainly be helpful), because you can never know if patch release breaks your application.

wilzbach:
Yep I sadly agree and I think the [current behavior of DAutoTest to run `dub upgrade`](https://github.com/CyberShadow/ae/blob/58a4e2b26caffd1a4da569c3b17a992483659306/sys/d/manager.d#L1714) is just a relict of an old workaround that was long forgotten.

CyberShadow:
> Yep I sadly agree and I think the current behavior of DAutoTest to run dub upgrade is just a relict of an old workaround that was long forgotten.

I don't think that's right. To quote [Sonke](https://github.com/dlang/dlang.org/pull/1017#issuecomment-110487529):

> The `--missing-only` switch will cause it to download only what is already present in dub.selections.json, without looking for actual upgrade possibilities. Together with `dub build --nodeps` it basically just resembles the default behavior of `dub build` without `--nodeps`.

So, the switch should be a no-op, unless dub.selections.json is missing some packages.

We went over this recently when a dependency appeared in the dependency tree, but dub.selections.json hadn't been updated, which broke the build.

CyberShadow:
Additionally, Digger actually checks if something in the build modified versioned files, and fails the build when that happens. This includes updates to `dub.selections.json`, and is what caused the build to be broken in the above-mentioned incident. So, whatever you think happened (or didn't happen) was certainly not because `dub upgrade` updated Dub files.

CyberShadow:
@wilzbach @ZombineDev Here is the actual problem:

> ```
> Selected package ddox 0.16.10 doesn't exist. Using latest matching version instead.
> Fetching ddox 0.16.9 (getting selected version)...
> ```

It looks like code.dlang.org flaked out when building this PR, so it used the previous version (0.16.9)?

@s-ludwig Is this the expected behavior of `dub upgrade --missing-only`? This seems contrary to the switch's description.

> Additionally, Digger actually checks if something in the build modified versioned files, and fails the build when that happens.

Actually that doesn't happen in DAutoTest, but only during deploy. code.dlang.org did not flake out during deploy, which is why no error was raised.

wilzbach:
This could also be due to https://github.com/dlang/ci/issues/190 (tl;dr: i.e. a mirror being out of date and thus not being aware of the newest dependencies)

PR #2339, opened 2018-04-10
update contracts.dd to reflect DIP1009

zachthemystic:
I edited this on github, but then was unable to compile the docs it to check its final appearance in html (See here for why: https://forum.dlang.org/post/azdfcrfqegubikcyntfj@forum.dlang.org ). At any rate, I'm hoping this can be looked at, in particular by @tgehr for accuracy. 

Also, the grammar spec is in a different file and still needs to be changed.

### DIFF
--- a/spec/contracts.dd
+++ b/spec/contracts.dd
@@ -66,11 +66,9 @@ $(H2 $(LNAME2 pre_post_contracts, Pre and Post Contracts))
         In D, pre contracts begin with `in`, and post contracts begin with `out`. They
         come at the end of the function signature and before the opening brace of the function body.)
 
-        $(P As of the acceptance of
-        $(LINK2 https://github.com/dlang/DIPs/blob/master/DIPs/accepted/DIP1009.md, DIP1009),
-        pre and post contracts can be written either in expression form, with
-        syntaxes similar to $(B assert), or as block statements containing arbitrary
-        code.)
+        $(P Pre and post contracts can be written either in expression form (feature introduced in DMD
+        2.081.0), with a syntax similar to $(B assert), or as block statements
+        containing arbitrary code.)
 
         $(P The expression form is:)
 ------
@@ -112,16 +110,16 @@ do
         an `assert` statement within the corresponding `in` or `out` block will throw an
         `AssertError`.)
 
-        $(P Notice that the keyword `do` can be used to announce the function body. While any number
+        $(P The keyword `do` can be used to announce the function body. Although any number
         of pre or post contracts of any form may follow each other, `do` is
         required only when the last contract before the body is a block statement.
         (Before the acceptance of
         $(LINK2 https://github.com/dlang/DIPs/blob/master/DIPs/accepted/DIP1003.md, DIP1003),
-        the keyword $(D body) was required instead of `do`, and may still be encountered in
-        old code bases. In the long term, $(D body) may be deprecated, but for now it's allowed both
-        as a keyword in this context and as an identifier elsewhere, although $(D do) is preferred.) )
+        the keyword `body` was required instead of `do`, and may still be encountered in
+        old code bases. In the long term, `body` may be deprecated, but for now it's allowed both
+        as a keyword in this context and as an identifier elsewhere, although `do` is preferred.) )
 
-        $(P While any arbitrary D code is allowed in the `in` and `out` contract blocks,
+        $(P Though any arbitrary D code is allowed in the `in` and `out` contract blocks,
         their only function should be to verify incoming and outgoing data. It is important
         to ensure that the
         code has no side effects, and that the release version of the code
@@ -158,7 +156,7 @@ do
 }
 ------
 
-        $(P The two functions are almost semantically identical. The expressions in the
+        $(P The two functions are almost identical semantically. The expressions in the
         first are lowered to contract blocks that look almost exactly like the second,
         except that a separate block is created for each expression in the first, thus
         avoiding shadowing variable names.)
@@ -216,8 +214,7 @@ class Date
 ------
 
         $(P Invariant blocks should contain `assert` expressions, and should throw
-        $(D AssertError)s when they fail. As of
-        $(LINK2 https://github.com/dlang/DIPs/blob/master/DIPs/accepted/DIP1009.md, DIP1009),
+        `AssertError`s when they fail. Since DMD version 2.081.0,
         invariants can also be written as expression statements, with `assert` implied:
         )
 
--- a/spec/contracts.dd
+++ b/spec/contracts.dd
@@ -33,9 +33,11 @@ $(H2 $(LNAME2 assert_contracts, Assert Contract))
 
         $(P The most basic contract is the
         $(GLINK2 expression, AssertExpression).
-        An $(B assert) declares an expression that must evaluate to true:)
+        An $(B assert) declares an expression that must evaluate to true, with an
+        optional failure string as a second argument:)
 ------
 assert(expression);
+assert(expression, "message to display on failure");
 ------
         $(P As a contract, an $(D assert) represents a guarantee that the code
         $(I must) uphold. Any failure of this expression represents a logic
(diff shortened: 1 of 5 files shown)


zachthemystic:
> But it failed to build due to trailing whitespace. I fixed that for you.
Much appreciated.

spec/contracts.dd
```diff
@@ -59,82 +61,111 @@ $(H2 $(LNAME2 pre_post_contracts, Pre and Post Contracts))
...
-    In the past, $(D body) was used, and could still be encountered in old code bases.
-    In the long term, $(D body) may be deprecated, but for now it's allowed
-    as a keyword in this context and as an identifier in the function body,
-    although $(D do) is preferred.)
+        $(P The optional identifier in either type of `out` contract is set to the return value
+        of the function, and should be tested immediately in the following expression,
+        or at some point in the block statement. An `out` contract of the
+        expression form should omit the identifier when testing for anything other than
+        the return value. In the block statement form, the identifier (along with its
+        accompanying parentheses) should likewise be omitted if no test requires it.
```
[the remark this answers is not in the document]
zachthemystic: I'll take your word for it. My objection would be that the documentation can also be intended to help beginners understand the intended use of the features. The whole document is written somewhat like that, as opposed to pure technical documentation. But as I said, I'll take your word for it.

zachthemystic:
Okay, I added the grammar changes as well, but I'll need to wait for the DAutoTest to finish to check the html appearance.

zachthemystic:
Uh, the DAutoTest failed for the commits after adding the grammar files, but I don't know why.

spec/contracts.dd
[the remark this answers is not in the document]
schveiguy: There are lots of places in the spec, where the syntax is not described in english. It's pretty clear to me in the syntax that you can omit the identifier if you don't intend to test it.

schveiguy:
Wow, the build log is quite an explosion of errors. I'm not sure how to diagnose!

I'm wondering if it's somehow actually *testing* the grammar to see if it's valid within dmd.

schveiguy:
ping @CyberShadow

zachthemystic:
I got as far as realizing that the successful build also spews the errors. It spews about twice as much as the failed build. The failed build ended seemingly arbitrarily when it built LaTex, about halfway through the process.

CyberShadow:
Uh oh, you have angered the PDF generator :)

It seems to complain about this line somewhere:
```
            main(string[] args) { ... }
```

It's not part of this diff, so it's likely that some bit of syntax went through unescaped in the Latex intermediary file, which in turn caused some tag to not be properly closed (or prematurely closed).

The next step would be to hunt which part of this diff causes the problem, and add escapes as necessary.

> I got as far as realizing that the successful build also spews the errors.

Nope, it doesn't - go to the DAutoTest page, click on the link by "Base result", then see the log linked there - there are lots of PDF warnings, but it ultimately succeeds.

spec/contracts.dd
```diff
@@ -59,82 +61,106 @@ $(H2 $(LNAME2 pre_post_contracts, Pre and Post Contracts))
...
+        `AssertError`.)
+
+        $(P Notice that the keyword `do` can be used to announce the function body. While any number
+        of pre or post contracts of any form may follow each other, `do` is
+        required only when the last contract before the body is a block statement.
+        (Before the acceptance of
+        $(LINK2 https://github.com/dlang/DIPs/blob/master/DIPs/accepted/DIP1003.md, DIP1003),
+        the keyword $(D body) was required instead of `do`, and may still be encountered in
+        old code bases. In the long term, $(D body) may be deprecated, but for now it's allowed both
+        as a keyword in this context and as an identifier elsewhere (although $(D do) is preferred).) )
```
CyberShadow: Is it just me, or there is one more `)` here than there should be?

zachthemystic: I put the parens around "(Before the acceptance of... elsewhere (although `do` is preferred).) )"
The last parens are to close the paragraph $(P ...). I'll remove the "although `do`"  set of parens.
This file did pass DAutoTest before the grammar commits, so it's not the cause of the errors.

schveiguy: vim says the parens are legit.

zachthemystic:
> go to the DAutoTest page, click on the link by "Base result", then see the log linked there - there are lots of PDF warnings, but it ultimately succeeds.

Yes, it succeeds. The failed build fails about halfway through. Search for "LaTeX Warning: Hyper reference `MAIN' on page" on the successful build, and you'll see it appears halfway through, whereas that's the line that caused the failed build to fail. Baffling! Anyway, I made two tiny edits and now have to wait again on the results.

wilzbach:
Why don't you build it locally and see where the error is coming from?
Might be a bit faster then waiting..

zachthemystic:
> Why don't you build it locally and see where the error is coming from?
> Might be a bit faster then waiting..

I couldn't build it. See the first comment in this PR: https://github.com/dlang/dlang.org/pull/2339#issue-180730844

zachthemystic:
> The next step would be to hunt which part of this diff causes the problem, and add escapes as necessary.

This is hard for me, because I couldn't get it to build on my machine, so I was forced to use DAutoTest. The instructions in CONTRIBUTING.md failed when "make -f posix.mak html" errored with:
"...
make[1]: Nothing to be done for `all'.
.generated/stable_dmd-2.078.2/dmd2/osx/bin/dub build --compiler=.generated/stable_dmd-2.078.2/dmd2/osx/bin/dmd --root=ddoc && \
		mv ddoc/ddoc_preprocessor .generated/ddoc_preprocessor
Invalid source/import path: /Users/zach/dlang/dmd/src/dmd
Invalid source/import path: /Users/zach/dlang/dmd/generated/dub
Running pre-generate commands for dmd:frontend...
/bin/sh: ./config.sh: No such file or directory
Command failed with exit code 127
make: *** [.generated/ddoc_preprocessor] Error 2"

zachthemystic:
I remembered seeing something odd, i.e. naked '( )' in the existing class and struct invariant grammars, instead of $(LPAREN) and $(RPAREN). Let's see if that fixes it.

zachthemystic:
> It seems to complain about this line somewhere: `main(string[] args)`

That line appears in struct.dd as part of a big commented-out section in a $(COMMENT ...). Line 406. Maybe the $(COMMENT) macro has something to do with it?

zachthemystic:
> Should be fixed by #2340

I could be doing something wrong, but with that pull it still isn't reading the dub.sdl file for me: 
"...
make[1]: Nothing to be done for `all'.
.generated/stable_dmd-2.078.2/dmd2/osx/bin/dub build --compiler=.generated/stable_dmd-2.078.2/dmd2/osx/bin/dmd --root=ddoc && \
		mv ddoc/ddoc_preprocessor .generated/ddoc_preprocessor
No package file found in /Users/zach/dlang/dmd/, expected one of dub.json/dub.sdl/package.json
make: *** [.generated/ddoc_preprocessor] Error 2"

spec/function.dd
```diff
@@ -77,28 +77,56 @@ $(GNAME MemberFunctionAttribute):
...
 $(GRAMMAR
 $(GNAME FunctionBody):
-    $(GLINK2 statement, BlockStatement)
-    $(GLINK FunctionContracts)$(OPT) $(GLINK BodyStatement)
-    $(GLINK FunctionContracts)
+    $(GLINK SpecifiedFunctionBody)
+    $(GLINK MissingFunctionBody)
+
+$(GNAME FunctionLiteralBody):
+    $(GLINK SpecifiedFunctionBody)
```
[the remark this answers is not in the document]
zachthemystic: Thanks for testing. I'm out of my depth on this, I'm afraid.

CyberShadow:
If it's another crazy problem with paging or whatever that causes it to break on a completely innocent-looking spot, then I don't think there's much I can do rather than suggest to drop PDF generation, or switch to a different method to generate it, like rendering+printing HTML in a headless browser.

wilzbach:
I think the failure is related to this https://github.com/dlang/dlang.org/pull/2363#issuecomment-401639993

>    Your main error here is using \lstinline in the argument of another command, like most verbatim constructs that does not work. You almost got away with it except in the last case, the page break happens inside a listings block so you end up calling listings recursively inside itself and things go wrong.

https://tex.stackexchange.com/questions/229113/another-extra-or-forgotten-endgroup

However, I don't know how to fix it either. Probably breaking the Grammar sections into individual pieces could help...

wilzbach:
> I think the failure is related to this #2363 (comment)

So this makes it compile:

```diff
diff --git a/spec/function.dd b/spec/function.dd
index 685d597e..1b7b10e8 100644
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -52,7 +52,11 @@ $(GNAME InOutX):
     $(D ref)
     $(RELATIVE_LINK2 return-ref-parameters, $(D return ref))
     $(D scope)
+)
+
+$(H4 $(LNAME2 grammar, foo))
 
+$(GRAMMAR
 $(GNAME FunctionAttributes):
     $(GLINK FunctionAttribute)
     $(GLINK FunctionAttribute) $(I FunctionAttributes)
```

So as mentioned just splitting up the grammar section in multiple pieces...
I pushed a more specific commit that splits the grammar up in multiple parts to your PR. Feel free to modify it, it's just a suggestion.

zachthemystic:
It passes. That is all that matters at this point. Thank you.

spec/function.dd
```diff
@@ -75,30 +83,66 @@ $(GNAME MemberFunctionAttribute):
...
+
+$(GNAME SpecifiedFunctionBody):
+    $(D do)$(OPT) $(GLINK2 statement, BlockStatement)
+    $(GLINK FunctionContracts)$(OPT) $(GLINK InOutContractExpression) $(D do)$(OPT) $(GLINK2 statement, BlockStatement)
+    $(GLINK FunctionContracts)$(OPT) $(GLINK InOutStatement) $(D do)$(OPT) $(GLINK2 statement, BlockStatement)
+
+$(GNAME MissingFunctionBody):
+    $(D ;)
+    $(GLINK FunctionContracts)$(OPT) $(GLINK InOutContractExpression) $(D ;)
+    $(GLINK FunctionContracts)$(OPT) $(GLINK InOutStatement)
```
wilzbach: Isn't there a semi-colon missing?

spec/function.dd
```diff
@@ -75,30 +83,66 @@ $(GNAME MemberFunctionAttribute):
...
+$(GRAMMAR
 $(GNAME FunctionContracts):
-    $(GLINK InStatement) $(GLINK OutStatement)$(OPT)
-    $(GLINK OutStatement) $(GLINK InStatement)$(OPT)
+    $(GLINK FunctionContract)
+    $(GLINK FunctionContract) $(I FunctionContracts)
+
+$(GNAME FunctionContract):
+    $(GLINK InOutContractExpression)
+    $(GLINK InOutStatement)
```
wilzbach: Couldn't you group all of them under FunctionContract:

FunctionContract:
- InContractExpression
- OutContractExpression
- InStatement
- OutStatement

This would also save you two lines above.

spec/contracts.dd
```diff
@@ -61,82 +63,106 @@ $(H2 $(LNAME2 pre_post_contracts, Pre and Post Contracts))
         typical use of this would be in validating the parameters to a function. The post
         contracts validate the result of the statement. The most typical use of this
         would be in validating the return value of a function and of any side effects it has.
-        The syntax is:)
+        In D, pre contracts begin with `in`, and post contracts begin with `out`. They
```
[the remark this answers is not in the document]
zachthemystic: I just copied the style of the previous documentation. It seems subjective, and not a source of confusion.

spec/contracts.dd
```diff
@@ -61,82 +63,106 @@ $(H2 $(LNAME2 pre_post_contracts, Pre and Post Contracts))
...
+        come at the end of the function signature and before the opening brace of the function body.)
 
+        $(P As of the acceptance of
+        $(LINK2 https://github.com/dlang/DIPs/blob/master/DIPs/accepted/DIP1009.md, DIP1009),
+        pre and post contracts can be written either in expression form, with
```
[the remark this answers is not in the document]
zachthemystic: It should probably just indicate the first compiler version that incorporates the feature then. ( 2.081.0 )

spec/function.dd
zachthemystic: No. The existing grammar is peculiar, and possibly bad design, but it is meant to prevent a semicolon coming after the end brace, ` }; `, as in:

int func() in { ... }; // <-- bad, already an error

The way you indicate the body here would be to now say `do`. Anything else (other than another `in` or `out`) means the function body is missing, and the declaration is over, and the parser moves on.

Arguably the no-semicolon-after-brace design is questionable. It hasn't ever mattered much, because currently a contact without a body is only legal in interface declarations, and rarely appears. It could however be expanded to ordinary function declarations eventually, at which point we might see more complaints about it. (e.g. "Why can't I just add the semicolon?")

Anyway, after a parenthesized contract with no body, the expectation would be for the semicolon to be there:

int func() in(...); // <-- legal and expected

spec/function.dd
zachthemystic: There might be a more concise way to write it. The reason they are separate is the need to write `do` between an In/OutStatement and the function body, whereas there is no need for In/OutExpressions. Timon Gehr wrote it, and he is generally very meticulous, so I left it as is, although I sympathize with the desire to be more concise.

spec/function.dd
wilzbach: Thanks a lot for the extensive explanation.
For the record, I do think this will be confusing for users if contracts are ever allowed for declaration-only functions/structs/classes:

```
int func() in(...);  // required
int func() in(...); // error
```

But yeah that's DMD's current behaviour: https://run.dlang.io/is/UwxsE0

zachthemystic:
One last concern. Maybe we should squash into two commits, with the split being at where the grammar finally passed the pdf generation, for future reference?

wilzbach:
Good idea.

zachthemystic:
Erm, I think I botched my pull. My git skills are clouded...

zachthemystic:
I tried to merge the existing dlang.org in my local machine before pushing it. But now there are 66 files changed, all in the last commit... hmmm...

zachthemystic:
For reference, the only files this pull is supposed to change are:
spec/contracts.dd
spec/class.dd
spec/expression.dd
spec/function.dd
spec/struct.dd

zachthemystic:
Okay, making progress. I got it down to 2 unwanted files in the pull...

zachthemystic:
Alright, I don't want my poor git skills to delay it. If you can reduce it to 2 commits do it. It currently pulls the master branch into it as the 2nd to last commit. Sorry about my git skills.

wilzbach:
No worries. Fixed it for you ;-)

zachthemystic:
Thanks :)

PR #2340, opened 2018-04-13
Fix building dlang.org from scratch

wilzbach:
See: https://forum.dlang.org/post/azdfcrfqegubikcyntfj@forum.dlang.org

The problem was that `GIT_HOME` was undefined and thus the git clone failed.
I think we should really have a CI for this. @CyberShadow what do you think about re-enabling Travis?

I also added a hint about `DIFFABLE=1` to the docs while I was at it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,4 +6,6 @@ os:
  - linux
 
 script:
-  - make -f posix.mak test
+  - make -f posix.mak html
+  - make -f posix.mak clean && rm -rf ../{dmd,druntime,phobos,tools,installer,dub}
+  - make -f posix.mak all
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,6 +39,16 @@ website pages. You may see warnings while the compiler is built. After `make`
 ended with error code 0, directory `web` contains the produced HTML
 files. Take a moment to open `web/index.html` in a browser.
 
+### Building with a slow internet connection
+
+The dlang.org build process depends on files which are downloaded during
+the build process. If you happen to have a slow internet connection, you can skip
+these downloads with `DIFFABLE=1`:
+
+```
+make -f posix.mak html DIFFABLE=1
+```
+
 ## Building the standard library documentation
 
 Now that the main site is in place, the standard library documentation would be
@@ -99,3 +109,7 @@ The output is in `web/phobos` and `web/library`.
 ### Learning more about DDoc
 
 Please see the [Ddoc fundamentals](https://wiki.dlang.org/Contributing_to_dlang.org).
+
+### Learn about more Makefile targets and options
+
+The full list of available targets and options is documented in the header of `posix.mak`.
--- a/posix.mak
+++ b/posix.mak
@@ -154,6 +154,7 @@ DRUNTIME_DIR=../druntime
 TOOLS_DIR=../tools
 INSTALLER_DIR=../installer
 DUB_DIR=../dub
+GIT_HOME=https://github.com/dlang
 
 # Auto-cloning missing directories
 $(shell [ ! -d $(DMD_DIR) ] && git clone --depth=1 ${GIT_HOME}/dmd $(DMD_DIR))
@@ -166,7 +167,6 @@ PHOBOS_LIB=$(PHOBOS_DIR)/generated/$(OS)/release/$(MODEL)/dmd/libphobos2.a
 # External directories
 DOC_OUTPUT_DIR:=$(PWD)/web
 W:=$(DOC_OUTPUT_DIR)
-GIT_HOME=https://github.com/dlang
 DPL_DOCS_PATH=dpl-docs
 DPL_DOCS=$(DPL_DOCS_PATH)/dpl-docs
 REMOTE_DIR=d-programming@digitalmars.com:data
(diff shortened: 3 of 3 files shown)


.travis.yml
```diff
@@ -6,4 +6,6 @@ os:
  - linux
 
 script:
-  - make -f posix.mak test
+  - make -f posix.mak html
+  - make -f posix.mak clean && rm -rf ../{dmd,druntime,phobos,tools,installer,dub}
+  - make -f posix.mak all
```
wilzbach: I thought we should test the two most important targets: `html` and `all`

posix.mak
```diff
@@ -154,6 +154,7 @@ DRUNTIME_DIR=../druntime
 TOOLS_DIR=../tools
 INSTALLER_DIR=../installer
 DUB_DIR=../dub
+GIT_HOME=https://github.com/dlang
 
 # Auto-cloning missing directories
 $(shell [ ! -d $(DMD_DIR) ] && git clone --depth=1 ${GIT_HOME}/dmd $(DMD_DIR))
```
wilzbach: Now `GIT_HOME` is no longer undefined and checking out `/dmd` doesn't fail.

posix.mak
```diff
@@ -585,7 +585,7 @@ $G/%-${LATEST} :
 
 # Clone all main D repositories
 ${DMD_DIR} ${DRUNTIME_DIR} ${PHOBOS_DIR} ${TOOLS_DIR} ${INSTALLER_DIR} ${DUB_DIR}:
-	git clone ${GIT_HOME}/$(notdir $(@F)) $@
+	git clone --depth=1 ${GIT_HOME}/$(notdir $(@F)) $@
```
wilzbach: It takes quite a long time to do a full checkout of `dmd` and it isn't really required.

posix.mak
```diff
@@ -877,13 +877,13 @@ $G/changelog/next-version: ${DMD_DIR}/VERSION
 	@echo $(NEXT_VERSION) > $@
 
 changelog/prerelease.dd: $G/changelog/next-version $(LOOSE_CHANGELOG_FILES) | \
-							${STABLE_DMD} $(TOOLS_DIR) $(INSTALLER_DIR) $(DUB_DIR)
+							${STABLE_DMD} $(DMD_DIR) $(DRUNTIME_DIR) $(PHOBOS_DIR) $(TOOLS_DIR) $(INSTALLER_DIR) $(DUB_DIR)
```
wilzbach: `druntime` and `phobos` weren't cloned by default for the changelog script anymore.

posix.mak
CyberShadow: Git history is needed for `git describe` to work properly, so this might not be a good idea.

CONTRIBUTING.md
```diff
@@ -39,6 +39,16 @@ website pages. You may see warnings while the compiler is built. After `make`
 ended with error code 0, directory `web` contains the produced HTML
 files. Take a moment to open `web/index.html` in a browser.
 
+### Building without changing files
+
+The dlang.org build process depends on many files which are downloaded during
+the build process. If you happen to have a slow internet connection, you can skip
+these downloads with `DIFFABLE=`:
```
CyberShadow: How true is this really?

CONTRIBUTING.md
```diff
@@ -39,6 +39,16 @@ website pages. You may see warnings while the compiler is built. After `make`
 ended with error code 0, directory `web` contains the produced HTML
 files. Take a moment to open `web/index.html` in a browser.
 
+### Building without changing files
```
CyberShadow: "without changing files"  is kind of ambiguous here

CONTRIBUTING.md
```diff
@@ -99,3 +109,7 @@ The output is in `web/phobos` and `web/library`.
 ### Learning more about DDoc
 
 Please see the [Ddoc fundamentals](https://wiki.dlang.org/Contributing_to_dlang.org).
+
+### Learn about the more Makefile targets and options
```
CyberShadow: This doesn't sound right. s/the // ?

CyberShadow:
> @CyberShadow what do you think about re-enabling Travis?

Why was it disabled?

I don't know if it's going to bring enough benefit vs. the overhead of managing another CI. I guess we could try and see if it doesn't get in the way too much?

zachthemystic:
I just tried "make -f posix.mak all" again. It still fails for me, after:"
CC="c++" /tmp/.host_dmd-2.072.2/dmd2/osx/bin/dmd -conf=/tmp/.host_dmd-2.072.2/dmd2/osx/bin/dmd.conf -of../generated/osx/release/64/dmd -m64 -vtls -J../generated/osx/release/64 -J../res -L-lstdc++ -version=MARS -fPIC -w -de dmd/access.d .....
.......................(Long list of files).................
sed -i -e "s|../druntime/import |../druntime-2.079.0/import |" -e "s|../phobos |../phobos-2.079.0 |" .generated/dmd-2.079.0/generated/osx/release/64/dmd.conf
sed: -e: No such file or directory
make: *** [.generated/dmd-2.079.0/generated/osx/release/64/dmd] Error 1
make: *** Deleting file `.generated/dmd-2.079.0/generated/osx/release/64/dmd'
"

CyberShadow:
Looks like your `sed` isn't GNU sed.

zachthemystic:
> Looks like your sed isn't GNU sed.

This link might help:
https://stackoverflow.com/questions/2320564/i-need-my-sed-i-command-for-in-place-editing-to-work-with-both-gnu-sed-and-bsd

wilzbach:
> Why was it disabled?

I think because we moved the testsuite to DAutoTest to keep the CI overhead low here.

> I don't know if it's going to bring enough benefit vs. the overhead of managing another CI. I guess we could try and see if it doesn't get in the way too much?

Ok enabled. Ideally it should just work "out of the box" (as for any fresh checkout).

> Looks like your sed isn't GNU sed.

Ok. I added a "Posix-compatible" way with using a temporary file. Not sure why we need to build this compiler anyhow, Maybe we can even remove the dependence on this entirely, but for now "... && mv" should work.

zachthemystic:
Build failed. But at least we know that it ran for 15 minutes and 50 seconds before doing so!

angelkennedy492 (approved):

PR #2311, opened 2018-03-30
Fix Issue 15449 - CSS and JS files should be minified

wilzbach:
A start anyhow:

```
> ll js
total 304K
drwxr-xr-x  2 seb users 4,0K Mar 20 12:22 ./
drwxr-xr-x 21 seb users 4,0K Mar 30 06:56 ../
-rw-r--r--  1 seb users 156K Feb  5 12:26 codemirror-compressed.js
-rw-r--r--  1 seb users 3,6K Feb  5 18:54 ddox.js
-rw-r--r--  1 seb users 3,0K Feb 27 23:16 dlang.js
-rw-r--r--  1 seb users  93K Feb  5 12:26 jquery-1.7.2.min.js
-rw-r--r--  1 seb users 5,8K Mar 20 12:22 listanchors.js
-rw-r--r--  1 seb users 1,6K Feb  5 12:26 platform-downloads.js
-rw-r--r--  1 seb users 3,8K Mar 13 11:56 run_examples.js
-rw-r--r--  1 seb users  16K Mar 13 11:56 run.js
-rw-r--r--  1 seb users 2,1K Feb  5 12:26 show_contributors.js

> ll web/js
drwxr-xr-x  2 seb users 4,0K Mar 30 06:56 ./
drwxr-xr-x 10 seb users 4,0K Mar 30 06:56 ../
-rw-r--r--  1 seb users 156K Mar 30 06:54 codemirror-compressed.js
-rw-r--r--  1 seb users 3,6K Mar 30 06:54 ddox.js
-rw-r--r--  1 seb users 3,0K Mar 30 06:54 dlang.js
-rw-r--r--  1 seb users  93K Mar 30 06:54 jquery-1.7.2.min.js
-rw-r--r--  1 seb users 5,8K Mar 30 06:54 listanchors.js
-rw-r--r--  1 seb users 1,6K Mar 30 06:54 platform-downloads.js
-rw-r--r--  1 seb users 3,8K Mar 30 06:54 run_examples.js
-rw-r--r--  1 seb users 7,7K Mar 30 06:56 run.js
-rw-r--r--  1 seb users 2,1K Mar 30 06:54 show_contributors.js
```

(the biggest chunks are jQuery + CodeMirror, but the other scripts are reduced by half.)

```
> ll css
total 64K
drwxr-xr-x  2 seb users 4,0K Mar 30 01:29 ./
drwxr-xr-x 21 seb users 4,0K Mar 30 06:56 ../
[...]

PR #2341, opened 2018-04-13
Compress JS and CSS with MINIFY=1

wilzbach:
Split of https://github.com/dlang/dlang.org/pull/2311 that just compresses JS + CSS.

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -74,7 +74,7 @@
 #
 #       DIFFABLE=1          Removes inclusion of all dynamic content and timestamps
 #       RELEASE=1           Release build (needs to be set for the `release` target)
-#       CSS_MINIFY=1        Minify the CSS via an online service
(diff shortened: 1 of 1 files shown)


posix.mak
```diff
@@ -230,6 +230,22 @@ else
...
 
+################################################################################
+# Build compression variables
+################################################################################
+# Set to 1 in the command line to minify css and js files
+MINIFY=0
+
+YUICOMPRESSOR_VERSION=2.4.8
+YUICOMPRESSOR=yuicompressor-$(YUICOMPRESSOR_VERSION).jar
+JS_COMPRESSOR_BIN=$G/$(YUICOMPRESSOR)
```
CyberShadow: That's just confusing, why not call it `YUI_COMPRESSOR_BIN`? It's not clear why there's a rule to download only the JS compression but not the CSS one.

PR #2342, opened 2018-04-13
Fixed crashing example in slicing section

schlupa:
The first example in the slicing section fails to build as it tries to link the external symbol foo which is not provided.

I also enhanced the other example to be a little bit more interesting for casual visitors. Examples which display only 'Succeed without output.' are completely useless.

### DIFF
--- a/spec/arrays.dd
+++ b/spec/arrays.dd
@@ -165,7 +165,10 @@ $(H2 $(LNAME2 slicing, Slicing))
 
 $(SPEC_RUNNABLE_EXAMPLE_COMPILE
 ---------
-void foo(int);
+void foo(int value)
+{
+  writeln("value=", value);
+}
 
 int[10] a;   // declare array of 10 ints
 int[] b;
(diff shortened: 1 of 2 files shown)


spec/arrays.dd
```diff
@@ -165,7 +165,10 @@ $(H2 $(LNAME2 slicing, Slicing))
 
 $(SPEC_RUNNABLE_EXAMPLE_COMPILE
```
wilzbach: Let's better use `SPEC_RUNNABLE_EXAMPLE_RUN` here, s.t. the spec tester actually runs the example?

spec/lex.dd
```diff
@@ -600,7 +600,7 @@ $(GNAME DecimalInteger):
     $(GLINK NonZeroDigit) $(GLINK DecimalDigitsUS)
 
 $(GNAME BinaryInteger):
-    $(GLINK BinPrefix) $(GLINK BinaryDigitsUS)
+    $(GLINK BinPrefix) $(GLINK BinaryDigitsNoSingleUS)
```
ZombineDev: @thewilsonator this change looks completely unrelated.

thewilsonator: Ohh, good catch.

thewilsonator: Hmm i think I must have `git reset --soft HEAD~N` for `N` equals one more than I thought. I was fighting both `git` and the stupid whitespace checker a lot with this, that that happened is not really all that surprising.

ZombineDev: When you do changes to people's own branches be sure to make backups and then compare your changes with the backed up branch before force pushing.

PR #2343, opened 2018-04-14
Add documentation for getOrAdd function to associative arrays

GilesBathgate:
Documentation for https://github.com/dlang/druntime/pull/2162


JinShil:
I recommend that the documentation for both of these functions mentions the fact that they avoid the need for multiple lookups.  IMO, that will often be the deciding factor when a user chooses to use these functions.

PR #2345, opened 2018-04-14
Only compile the failing spec tester example

wilzbach:
This currently fails on some PRs with:

```
Test the D Language specification
../dmd/generated/linux/release/64/dmd -run tools/dspec_tester.d --compiler=../dmd/generated/linux/release/64/dmd
---
void main() {
import std.stdio;
import std.algorithm;
int[] s = [1, 2, 3, 4];

copy(s[1..3], s[0..2]);
assert(s == [2, 3, 3, 4]);

}
---
gcc: error: __stdin.o: No such file or directory
Error: linker exited with status 1
```

(e.g. https://github.com/dlang/phobos/pull/6443, https://github.com/dlang/phobos/pull/6453 or https://github.com/dlang/dmd/pull/8168)

Hence, I think for now the best strategy is to just compile the respective test.

### DIFF
--- a/spec/arrays.dd
+++ b/spec/arrays.dd
@@ -255,7 +255,7 @@ s[1..3] = s[0..2]; // error, overlapping copy
         $(REF copy, std,algorithm,mutation):
         )
 
-$(SPEC_RUNNABLE_EXAMPLE_RUN
+$(SPEC_RUNNABLE_EXAMPLE_COMPILE
 ---------
 import std.algorithm;
 int[] s = [1, 2, 3, 4];
--- a/spec/struct.dd
+++ b/spec/struct.dd
@@ -304,7 +304,7 @@ $(H2 $(LEGACY_LNAME2 StructLiteral, struct-literal, Struct Literals))
         $(P Struct literals consist of the name of the struct followed
         by a parenthesized argument list:)
 
-        $(SPEC_RUNNABLE_EXAMPLE_RUN
+        $(SPEC_RUNNABLE_EXAMPLE_COMPILE
         ---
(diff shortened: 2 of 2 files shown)


wilzbach:
I also opened a Bugzilla issue, s.t. this doesn't get lost and can be tracked when there's time: https://issues.dlang.org/show_bug.cgi?id=18762
(it's probably related to the way the spec tester runs the test in parallel as this would explain why it happens only spuriously)

The TravisCi error is unrelated as we are currently trying to re-add Travis to dlang.org (see e.g. https://github.com/dlang/dlang.org/pull/2340)

CyberShadow:
> Observed another probably related failure:

This doesn't look like a scalable approach. Won't failures like these just keep cropping up until the root  problem is fixed?

> (it's probably related to the way the spec tester runs the test in parallel as this would explain why it happens only spuriously)

From a glance at the relevant code in `dspec_tester` and DMD, it seems fine, but there may be other parts like the linker affected by it.

wilzbach:
(wrong button)

Fully agreed and I would love to fix the actual problem, but I have honestly no idea why we sometimes get these linker errors...

CyberShadow:
What do you think about instead of using `-run` and making DMD read from stdin, to save it to a `.d` file to a guaranteed-unique directory (probably just use an integer counter)?

CyberShadow:
I can reproduce this on my machine, looking into it.

CyberShadow:
The problem is in DMD's handling of programs from stdin:
https://issues.dlang.org/show_bug.cgi?id=18762#c1

PR #2349, opened 2018-04-17
Remove '@' from attributes

GilesBathgate:
It seems the attributes in the grammar are all prefixed with @ but this is implied by the rule for properties i.e:
    
    Property:
        @ PropertyIdentifier
    
Furthermore, disable is referenced twice, they are not in alphabetical order, and `Property` and `PropertyIdentifier` are not technically correct terms anymore (or so I believe). I've ultimately just done a bit of a re-org, please see changes.

### DIFF
--- a/spec/attribute.dd
+++ b/spec/attribute.dd
@@ -28,24 +28,23 @@ $(GNAME Attribute):
     $(RELATIVE_LINK2 inout, $(D inout))
     $(RELATIVE_LINK2 shared, $(D shared))
     $(RELATIVE_LINK2 gshared, $(D __gshared))
-    $(GLINK Property)
+    $(GLINK AtAttribute)
     $(RELATIVE_LINK2 nothrow, $(D nothrow))
     $(RELATIVE_LINK2 pure, $(D pure))
     $(RELATIVE_LINK2 ref, $(D ref))
     $(RELATIVE_LINK2 return, $(D return))
 
-$(GNAME Property):
-    $(D @) $(GLINK PropertyIdentifier)
+$(GNAME AtAttribute):
+    $(D @) $(RELATIVE_LINK2 disable, $(D disable))
+    $(D @) $(RELATIVE_LINK2 nogc, $(D nogc))
+    $(GLINK Property)
+    $(D @) $(RELATIVE_LINK2 safe, $(D safe))
+    $(D @) $(RELATIVE_LINK2 safe, $(D system))
+    $(D @) $(RELATIVE_LINK2 safe, $(D trusted))
     $(GLINK UserDefinedAttribute)
 
-$(GNAME PropertyIdentifier):
-    $(RELATIVE_LINK2 property, `property`)
-    $(RELATIVE_LINK2 safe, `@safe`)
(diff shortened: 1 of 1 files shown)


GilesBathgate:
also @property on the same page, hmm If I follow the grammar correctly it would mean I have to use @@safe and @@system etc. @wilzbach thoughts?

rainers:
I'd recommend removing all `@`s from the PropertyIdentifiers. `@safe` is not a single token, you can also write
```
@
safe
int foo();
```
I can't say I like that...

quickfur:
@rainers  Wow, I had no idea `@` was treated separately like that.  That's... scary.  I hope to never encounter code that has whitespace between `@` and the identifier!

GilesBathgate:
Hmm, `PropertyIdentifier` isn't even the right name for that grammar rule anymore is it?

spec/attribute.dd
```diff
@@ -28,25 +28,21 @@ $(GNAME Attribute):
     $(RELATIVE_LINK2 inout, $(D inout))
     $(RELATIVE_LINK2 shared, $(D shared))
     $(RELATIVE_LINK2 gshared, $(D __gshared))
-    $(GLINK Property)
+    $(GLINK PrefixedAttribute)
```
BBasile: suggestion, use `AtAttribute`, like in dparse.

GilesBathgate:
@rainers It didn't seem to help. @CyberShadow Any ideas?

CyberShadow:
The error is:

```
<recently read> \egroup 
                        
l.6014  
         align (4):
```

Here is the surrounding generated tex source:

```tex
\begin{lstlisting}
struct S
{
  align (1):
    byte a;   // placed at offset 0
    int b;    // placed at offset 1
    long c;   // placed at offset 5
}
auto sz = S.sizeof;  // 16
\end{lstlisting}
\par
        The alignment for the fields of an aggregate does not affect the alignment
        of the aggregate itself - that is affected by the alignment setting outside
        of the aggregate. \par
\par
\begin{lstlisting}
align (2) struct S
{
  align (1):
    byte a;   // placed at offset 0
    int b;    // placed at offset 1
    long c;   // placed at offset 5
}
auto sz = S.sizeof;  // 14
\end{lstlisting}
\par
        Setting the alignment of a field aligns it to that power of 2, regardless
        of the size of the field. \par
\par
\begin{lstlisting}
struct S
{
  align (4):           // <---------------------------- ERROR HERE
    byte a;   // placed at offset 0
    byte b;   // placed at offset 4
    short c;  // placed at offset 8
}
auto sz = S.sizeof;  // 12
\end{lstlisting}
```

I see absolutely nothing wrong with it.

@andralex I recall you recently ran into a similar problem, and successfully worked around it. Any idea?

GilesBathgate:
@rainers I have found a workaround, and the documentation now builds correctly.

rainers:
@GilesBathgate Thank you for your perseverence. I guess the problem was that there are links to `Property` from elsewhere, too.

Looking at it again, there are still quite some inconsistencies and omissions regarding attributes, storage classes and function attributes. But that's for another PR.

PR #2350, opened 2018-04-20
 fix issue 18782 - Documentation error: ProtectionAttributes should say Visibility Attribute

aG0aep6G:
Also removing an outdated sentence in the vicinity.


aG0aep6G:
Added a commit to fix similar wrong links in lex.html.

aG0aep6G:
Ping. This has been open for nearly two months without any reaction.

PR #2351, opened 2018-04-21
Fix Issue 16206 - traits getOverloads fails when one of the overload is a templatized function

Biotronic:
Spec change for https://github.com/dlang/dmd/pull/8195

### DIFF
--- a/spec/traits.dd
+++ b/spec/traits.dd
@@ -658,8 +658,10 @@ $(H2 $(GNAME getOverloads))
 
         $(P The first argument is an aggregate (e.g. struct/class/module).
         The second argument is a string that matches the name of
-        one of the functions in that aggregate.
+        one of the members in that aggregate.
+        The third argument is a bool, and is optional.
         The result is a tuple of all the overloads of that function.
+        If the third member evaluates to true, the result will include all members that matches the supplied name.
         )
 
 ---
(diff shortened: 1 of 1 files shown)


spec/traits.dd
```diff
@@ -659,6 +659,7 @@ $(H2 $(GNAME getOverloads))
         $(P The first argument is an aggregate (e.g. struct/class/module).
         The second argument is a string that matches the name of
         one of the functions in that aggregate.
+        The third argument is a bool, and is optional.
```
JinShil: Please add an explanation of what the 3rd argument does.

spec/traits.dd
```diff
@@ -658,8 +658,10 @@ $(H2 $(GNAME getOverloads))
 
         $(P The first argument is an aggregate (e.g. struct/class/module).
         The second argument is a string that matches the name of
-        one of the functions in that aggregate.
+        one of the members in that aggregate.
+        The third argument is a bool, and is optional.
         The result is a tuple of all the overloads of that function.
+        If the third member evaluates to true, the result will include all members that matches the supplied name.
```
JinShil: I recommend swapping this line with the one above it so the "The result..." is the last in the paragraph, and the explanation of the 3rd argument is not fragmented.

RazvanN7: Also, "the result will include all members that **match** the supplied name"

wilzbach:
CircleCi failure is unrelated and sorry for the unexpected problems.
The fix and explanation is here -> https://github.com/dlang/dlang.org/pull/2371

PR #2352, opened 2018-04-22
Add missing SpecialToken as Primary, TemplateSingleArg and TemplateDefArg

BBasile:


BBasile:
Actually i see now that there's an issue in the grammar. Everywhere that a SpecialKeyword is allowed, a SpecialToken is too.

PR #2358, opened 2018-05-02
Deploy install.sh from dlang.org

wilzbach:
The installer script is updated very rarely (less than once a month),
so imho hosting a copy directly in this repo makes deployment more transparent and avoids the bus factor permission problem.

See also: https://github.com/dlang/installer/blob/master/script/deploy.sh

CC @MartinNowak @CyberShadow

(I'm aware that this has the downside that updates get a bit more annoying - the alternative would be to make sure more than one person can deploy the script.)


wilzbach:
It was decided to avoid this additional step.

PR #2359, opened 2018-05-06
Error for implicit string concatenation

JinShil:
PR to accompany https://github.com/dlang/dmd/pull/8220

See https://dlang.org/deprecate.html#Implicit%20string%20concatenation

This PR moves a deprecation forward, further narrowing the gap between specification and implementation.


wilzbach:
(rebased)

PR #2363, opened 2018-05-08
[spec] Add UDAs for parameters

wilzbach:
A first start. I wasn't sure how much into detail of UDAs we should go in the functions page as they are already defined in depth in https://dlang.org/spec/attribute.html

DMD PR: https://github.com/dlang/dmd/pull/7576

### DIFF
--- a/dlang.org.ddoc
+++ b/dlang.org.ddoc
@@ -175,6 +175,7 @@ _=
 GEN_DATETIME=$(DATETIME)
 GLINK=$(RELATIVE_LINK2 $0, $(I $0))
 GLINK2=$(DDSUBLINK spec/$1,$2,$(I $2))
+GLINK2_ALTTEXT=$(DDSUBLINK spec/$1,$2,$(I $3))
 GLOSSARY = $(HTTP dlang.org/glossary.html#$0, $0)
 GLOSSARY2 = $(HTTP dlang.org/glossary.html#$1, $2)
 GNAME=<a id="$0">$(SPANC gname, $0)</a>
--- a/ebook.ddoc
+++ b/ebook.ddoc
@@ -49,6 +49,7 @@ MODDEFFILE=<pre class="moddeffile">$0</pre><p></p>
 _=
 
 GLINK2=$(GLINK $2)
+GLINK2_ALTTEXT=$(RELATIVE_LINK2 $2, $3)
 DDLINK=$(RELATIVE_LINK2 $2, $3)
 DDSUBLINK=$(RELATIVE_LINK2 $2, $3)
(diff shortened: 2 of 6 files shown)


jacob-carlborg:
> A first start. I wasn't sure how much into detail of UDAs we should go in the functions page as they are already defined in depth in https://dlang.org/spec/attribute.html

I think it's almost a bit too much detail since, as you mentioned, it's already specified elsewhere.

wilzbach:
So I merged `InOut` and `UserDefinedAttribute` into `ParameterAttributes` as `const @(2)` and `@(2) const` are both legal (only `int @(2)` is forbidden as UDAs can only be prefix).
For the lazy who are wondering if this works for `AliasDeclaration`s too:

```
AliasDeclaration:
    alias StorageClassesopt BasicType Declarators ;
    alias StorageClassesopt BasicType FuncDeclarator ;
    alias AliasDeclarationX ;

AliasDeclarationX:
    AliasDeclarationY
    AliasDeclarationX , AliasDeclarationY

AliasDeclarationY:
   Identifier TemplateParametersopt = FunctionLiteral
```

```
FunctionLiteral:
    function Typeopt ParameterAttributes opt FunctionLiteralBody
    delegate Typeopt ParameterMemberAttributes opt FunctionLiteralBody
    ParameterMemberAttributes FunctionLiteralBody
    FunctionLiteralBody
    Lambda
```

```
Lambda:
    function Typeopt ParameterAttributes => AssignExpression
    delegate Typeopt ParameterMemberAttributes => AssignExpression
    ParameterMemberAttributes => AssignExpression
    Identifier => AssignExpression
```

```
ParameterAttributes:
    Parameters FunctionAttributesopt

ParameterMemberAttributes:
    Parameters MemberFunctionAttributesopt
```

```
Parameters:
    ( ParameterListopt )

ParameterList:
    Parameter
    Parameter , ParameterList
```

wilzbach:
I'm not sure why this fails when building the PDF, but as removing two lines anywhere in the GRAMMAR section makes the error go away I presume it has something to do with this:

> Your main error here is using \lstinline in the argument of another command, like most verbatim constructs that does not work. You almost got away with it except in the last case, the page break happens inside a listings block so you end up calling listings recursively inside itself and things go wrong.

https://tex.stackexchange.com/questions/229113/another-extra-or-forgotten-endgroup

However, I tried replacing `\lstinline` with `\texttt` or `\verb`, but without success.
Has anyone more experience with this or an idea how to workaround it?

wilzbach:
Rebased. Should be passing DAutoTest now.

ZombineDev:
Why close?

wilzbach:
I think something went wrong when I purged my GitHub branches. Thanks for noticing!

PR #2364, opened 2018-05-08
Remove old & invalid msie < 9 support code

wilzbach:
Turns out that DDox's macro processor doesn't insert a newline when newlines occur in macros.
The workaround is to use good, old HTML.

The old MSIE < 9 support code lead to the final "}" being removed.

See also: https://github.com/rejectedsoftware/ddox/issues/204

### DIFF
--- a/dlang.org.ddoc
+++ b/dlang.org.ddoc
@@ -386,7 +386,9 @@ _=
 RELATIVE_LINK2=$(ALOCAL $1, $+)
 _=
 
-RUNNABLE_EXAMPLE=$(DIVC runnable-examples, $1)
+RUNNABLE_EXAMPLE=<div class="runnable-examples">
+$1
+</div>
 RUNNABLE_EXAMPLE_STDIN=<code class="runnable-examples-stdin">$0</code>
 RUNNABLE_EXAMPLE_ARGS=<code class="runnable-examples-args">$0</code>
 
--- a/js/run.js
+++ b/js/run.js
@@ -282,18 +282,13 @@ function setupTextarea(el, opts)
 
     var prepareForMain = function()
     {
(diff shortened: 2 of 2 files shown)


wilzbach:
Before:

```
 $(RUNNABLE_EXAMPLE
 ----
 string letters()
 {
   char[] result = new char['z' - 'a' + 1];
   foreach (i, ref e; result)
   {
     e = cast(char)('a' + i);
   }
   return assumeUnique(result);
 }
 ----
 )
```

And after [rendering the macros](https://github.com/rejectedsoftware/ddox/blob/040983fffadc4a224f0eadf3ea0ccd5647496700/source/ddox/ddoc.d#L349):

```
<div class="runnable-examples"> ----
 string letters()
 {
   char[] result = new char['z' - 'a' + 1];
   foreach (i, ref e; result)
   {
     e = cast(char)('a' + i);
   }
   return assumeUnique(result);
 }
 ----
 </div>
```

----

After:

```
$(RUNNABLE_EXAMPLE
 ----
 string letters()
 {
   char[] result = new char['z' - 'a' + 1];
   foreach (i, ref e; result)
   {
     e = cast(char)('a' + i);
   }
   return assumeUnique(result);
 }
 ----
 )
```

And after [rendering the macros](https://github.com/rejectedsoftware/ddox/blob/040983fffadc4a224f0eadf3ea0ccd5647496700/source/ddox/ddoc.d#L349):

```
<div class="runnable-examples">
 ----
 string letters()
 {
   char[] result = new char['z' - 'a' + 1];
   foreach (i, ref e; result)
   {
     e = cast(char)('a' + i);
   }
   return assumeUnique(result);
 }
 ----
```

PR #2366, opened 2018-05-12
Issue 18743 - ConditionalExpression with AssignExpression requires parens

ntrel:
Depends on https://github.com/dlang/dmd/pull/8237.

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -165,6 +165,28 @@ $(GNAME ConditionalExpression):
         the result type of the conditional expression.
     )
 
+    $(P $(B Note:)
+        When a conditional expression is the left operand of
+        an $(RELATIVE_LINK2 assign_expressions, assign expression), they would be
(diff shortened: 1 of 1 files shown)


spec/expression.dd
```diff
@@ -165,6 +165,28 @@ $(GNAME ConditionalExpression):
         the result type of the conditional expression.
     )
 
+    $(P $(B Note:)
+        When a conditional expression is the left operand of
+        an $(RELATIVE_LINK2 assign_expressions, assign expression), they would be
+        parsed as an assignment to $(I the result) of the conditional. This
+        has now been deprecated. To preserve the existing behaviour, add
```
wilzbach: "now" doesn't make sense in a spec that might be read by someone in a year.

ntrel:
@wilzbach Thanks, you're right. Should be fixed now.

ntrel:
Tests pass.

PR #2367, opened 2018-05-15
fix wrong link and absolute links

JohnCoconut:
Issuse reference: https://issues.dlang.org/show_bug.cgi?id=18861

A question, how to build dlang.org website without building dmd compiler?

### DIFF
--- a/articles/cpptod.dd
+++ b/articles/cpptod.dd
@@ -560,7 +560,7 @@ $(H4 The D Way)
 
         The D version is analogous, though a little simpler, taking
         advantage of
-        $(HTTPS dlang.org/spec/template.html#implicit_template_properties,
+        $(DDSUBLINK spec/template, implicit_template_properties,
         Eponymous Templates) - promotion of single template members to the
         enclosing name space:
 ------
@@ -580,7 +580,7 @@ void test()
 }
 ------
         The template blocks can be made shorter using
-        $(HTTPS dlang.org/spec/template.html#variable-template,
+        $(DDSUBLINK spec/template, variable-template,
         Enum Template) syntax:
 ------
 enum factorial(int n) = n * .factorial!(n-1);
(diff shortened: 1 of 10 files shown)


wilzbach:
> A question, how to build dlang.org website without building dmd compiler?

You might be able to set the DMD Makefile variable to your host compiler, but DMD still needs to be partially build because it's used as a library for the ddoc preprocessor.
Are you running into any specific troubles?

articles/cpptod.dd
```diff
@@ -560,7 +560,7 @@ $(H4 The D Way)
 
         The D version is analogous, though a little simpler, taking
         advantage of
-        $(HTTPS dlang.org/spec/template.html#implicit_template_properties,
+        $(DDSUBLINK spec/template, implicit_template_properties,
```
wilzbach: Any specific reason why you sometimes use LINK2 and sometimes DDSUBLINK?

mywtfmp3: When it referes to a sublink in a page, I use `DDSUBLINK`. When it refers to the whole page, I use `LINK2`. I saw this commit, [use DDSUBLINK instead of LINK2 ](https://github.com/dlang/druntime/commit/ea4785d189ca7303e034e6c392f9b8fde31faef7) in druntime repo in the past, I thought it would be prefered.

mywtfmp3:
I  have no problem building it. Just wondering if I can use the compiler from my distro. I will build dmd then.

wilzbach:
Btw all you need to do should be running `make -f posix.mak html -j8`?

For overavoiding to download DMD stable you might be able to do `STABLE_DMD=dmd`, but I haven't tried this yet and I am currently on my phone.

mywtfmp3:
It uses distro dmd after specifying `STABLE_DMD` and `STABLE_DMD_CONF`,

`make -f posix.mak html -j4 STABLE_DMD=dmd STABLE_DMD_CONF=/etc/dmd.conf`

PR #2368, opened 2018-05-15
fix Issue 18712 by documenting the behavior

WalterBright:


spec/statement.dd
```diff
@@ -1163,14 +1163,31 @@ switch (name)
         less error prone. char, wchar and dchar strings are allowed.
         )
 
-        $(P $(D Implementation Note:) The compiler's code generator may
+        $(P If a $(GLINK CaseStatement), $(GLINK CaseRangeStatement), or $(GLINK DefaultStatement)
+        appears in the top level of the $(GLINK ScopeStatement), then a new scope for local
```
rainers: According to -vcg-ast output, this happens for `case` in nested scopes, too, not only on top level.

spec/statement.dd
```diff
@@ -1163,14 +1163,31 @@ switch (name)
...
+        symbols is introduced immediately following it, extending to the first appearance of one of
+        $(GLINK CaseStatement), $(GLINK CaseRangeStatement), $(GLINK DefaultStatement), `}`,
+        the end of a $(GLINK MixinStatement), or the end of the file.
```
rainers: The end of the file cannot be reached anyway as the switch statement is inside part of a function body (or similar).

spec/statement.dd
```diff
@@ -1163,14 +1163,31 @@ switch (name)
...
 
+        $(BEST_PRACTICE
+        $(OL
+        $(LI Use a $(GLINK ScopeBlockStatement) rather than a $(GLINK ScopeStatement)
+        for the $(B switch) body. (I.e. use `{ }`.))
+        $(LI $(GLINK CaseStatement), $(GLINK CaseRangeStatement), and $(GLINK DefaultStatement)
+        should appear directly in the `{ }` of the $(GLINK ScopeStatement), rather than
+        in nested `{ }`.)
+        )
+        )
```
rainers: Is there an example where allowing CaseStatement and the others in nested scopes is actually useful?

WalterBright: I wouldn't allow it in our code, but people do use it in C. In any case, I'm trying to document what it does, not what we'd like it to do.

JinShil:
I don't understand why we don't do the due diligence to fix this.  It can be taken through the deprecation process, emitting a warning for a year or two informing users that the behavior will change in a future release, and then submitting the fix after that time has passed.  Let's not so easily give in to mistakes of the past.

WalterBright:
What I'm doing here is documenting the way it behaves now. Redoing the semantics is another matter entirely and beyond the scope of this PR.

Fixing this is not just a tweak, it's substantial and could be fairly disruptive, and probably rises to the level of a DIP.

JinShil:
It makes sense to document existing behavior for experimental features that are expected to grow and change with time, e.g. -betterC.  This PR, however, is accepting defeat and endorsing a mistake in the implementation.   In this case we should be acknowledging that the current behavior is incorrect and keep the bug in bugzilla.  Then devise a solution to correct the behavior.  This PR, or any PR that clarifies current behavior, should only be accepted if the current behavior is exactly what we want, or the feature is experimental.

rainers:
I think it is ok to document the current behavior if it is unlikely to change in the near future. It should be marked as a current limitation, though, as e.g. for multiple `alias this`.

thewilsonator:
#2512

PR #2369, opened 2018-05-16
Spec update for the inline assembler (DMD PR 8250)

leitimmel:
This is the companion pull request for [this one](https://github.com/dlang/dmd/pull/8250), which proposes adding the `IRETQ` instruction to DMD's inline assembler backend.


leitimmel:
I have inserted the instruction behind `IRET` and `IRETD` so it's alphabetically ordered, but that caused all elements to move by one column. Many of them had to be moved to a new row because the row in which I inserted had 6 elements.

PR #2371, opened 2018-05-29
Don't export RELEASE=1 to the DMD master build

wilzbach:
For the `RELEASE` build of dlang.org (the release build is the built that generates the docs which are run during the tarball generation by the installer script), the `RELEASE` variable is set to distinguish for this and use a few different Makefile rules.

As until recently it was possible to use `RELEASE=1` for dmd too and this option finally got turned into a hard error, this yieleded a conflict.
This PR simply unsets `RELEASE` for the dmd makefile invocations which cause this conflict.
In theory `unexport` or `undefine` should prevent a variable to be passed to a recursive Make call, but apparently those only work for Makefile environment variables and not for Makefile options.

See also:
- https://github.com/dlang/dmd/pull/8231
- https://circleci.com/gh/dlang/dlang.org/1464?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -593,20 +593,20 @@ ${DMD_DIR}/VERSION : ${DMD_DIR}
 ################################################################################
 
 $(DMD) : ${DMD_DIR}
-	${MAKE} --directory=${DMD_DIR}/src -f posix.mak AUTO_BOOTSTRAP=1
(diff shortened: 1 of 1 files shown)


wilzbach:
Ping @CyberShadow as this unbreaks the currently broken CircleCi at dlang.org

PR #2372, opened 2018-05-29
[trivial] [deprecate.dd] Fix *Hiding base class functions* wording

ntrel:


ntrel:
Ping

JinShil:
What was the problem with the original.  It appears it's just restating the same thing in a different way.

ntrel:
@JinShil 
> Declaration functions

`Declaration functions` is not a known term. I expect `Declaring functions` or `Function declarations` was meant.

> in a derived class that can be called with the same arguments as a function in a base class,
> but do not override that function causes the base class function to be hidden

This whole sentence is awkward to read and does not flow well. It's tortuous to have all those clauses before the meaning/structure of the sentence is clear. In written English this is uncommon (perhaps in other languages (German?) it's more common). Prefixing the sentence with `This occurs when` helps let the reader know a bit more about the structure of the complicated sentence.

`but do not override` -> `without overriding` - *but do not* is less clear (the reader may initially wonder if it applies to *arguments* rather than *functions* (before they see the word *override*), my wording helps to avoid this.

`causes the base class function to be hidden` -> `The base class function gets hidden` - starting a new sentence helps break up the complicated form of the first sentence.

PR #2374, opened 2018-06-02
Fix Issue 18874 - Add thatneedle.com to organizations using D

wilzbach:
Preview:

image


wilzbach:
From Bugzilla:

> Looks good. Thanks for looking into this.

So anyone around for merging this?

PR #2380, opened 2018-06-06
fix release doc build with ENABLE_RELEASE=1

MartinNowak:
- set ENABLE_RELEASE=1 in release target and rerun makefile
- use same ENABLE_RELEASE variable as dmd and create_dmd_release
- Need both a target and a variable to be set for the build to work
  with dlang.org's current makefile architecture (deps in variables).
  The workaround is messy and separate targets should be used instead
  of variables that affect file lists.

The approach in #2371 didn't work as it seems that variables in prerequisites (e.g. of the html target) get expanded before target-specific variables are applied. So even though `release : CHANGELOG_FILES:=$(filter-out changelog/pending,$(CHANGELOG_FILES))` does modify the list it has no effect on `html : $(ALL_FILES)` at the time the html target is run.

### DIFF
--- a/.circleci/run.sh
+++ b/.circleci/run.sh
@@ -109,7 +109,7 @@ run_make()
     make -v
 
     # -j1 is used for a better error log
-    make -f posix.mak -j1 RELEASE=1 DIFFABLE=1 release
+    make -f posix.mak -j1 DIFFABLE=1 release
 }
 
 case $1 in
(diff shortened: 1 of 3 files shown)


.circleci/run.sh
```diff
@@ -109,7 +109,7 @@ run_make()
     make -v
 
     # -j1 is used for a better error log
-    make -f posix.mak -j1 RELEASE=1 DIFFABLE=1 release
+    make -f posix.mak -j1 DIFFABLE=1 release
```
wilzbach: Shouldn't we set `ENABLE_RELEASE=1` here if we're about to renable this?

edit: saw the Makefile re-invocation below. Fair enough then.

MartinNowak: Yes, wanted to test that re-invocation ;).

PR #2382, opened 2018-06-06
Fix Issue 18558 - Template alias spec incomplete

ntrel:
* Document that template alias parameters can evaluate their argument expression at compile-time.
* Document template parameterized by a lambda.

### DIFF
--- a/spec/template.dd
+++ b/spec/template.dd
@@ -644,7 +644,7 @@ $(GNAME TemplateAliasParameterDefault):
         )
 
         $(LI Compile-time values
-
+        $(SPEC_RUNNABLE_EXAMPLE_RUN
         ------
         template Foo(alias x)
         {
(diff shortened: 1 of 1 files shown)


spec/template.dd
```diff
@@ -637,6 +642,44 @@ $(GNAME TemplateAliasParameterDefault):
...
+
+        $(LI Lambdas
+
+        ------
+        template Foo(alias fun)
+        {
+            enum val = fun(2);
+        }
+
+        alias foo = Foo!(x => x * x);
```
aG0aep6G: `x => x * x` is not a value. It has no type. It's a template. `(int x) => x * x` would be a value.

spec/template.dd
```diff
@@ -637,6 +642,44 @@ $(GNAME TemplateAliasParameterDefault):
         }
         ------
         )
+
+        $(LI Compile-time values
+
+        ------
```
wilzbach: You could wrap the block with this

```
$(SPEC_RUNNABLE_EXAMPLE_RUN
---
...
---
)
```

then the spec tester would verify the (static) asserts.
There's also SPEC_RUNNABLE_EXAMPLE_COMPILE

PR #2384, opened 2018-06-06
Update __traits(getLinkage) to include struct

TurkeyMan:


TurkeyMan:
I see @RazvanN7 already made one!

PR #2386, opened 2018-06-08
Fix Issue 18959 - [Change Log: 2.079.0] substitute was added in std.algorithm.iteration but the changelog points to std.algorithm.searching

wilzbach:


changelog/2.079.0.dd
```diff
@@ -43,7 +43,7 @@ $(BUGSTITLE_TEXT_HEADER Library changes,
 $(LI $(RELATIVE_LINK2 compile-time-format-optimized,`std.format` with strings passed during compile-time has been optimized))
 $(LI $(RELATIVE_LINK2 fix18397,Changed `std.conv.hexString` to return an immutable string literal))
 $(LI $(RELATIVE_LINK2 nullable-class,`Nullable!C.nullify` no longer calls .destroy when `C` is a class or interface))
-$(LI $(RELATIVE_LINK2 std-algorithm-iteration-substitute,`std.algorithm.searching.substitute` was added))
+$(LI $(RELATIVE_LINK2 std-algorithm-ite-substitute,`std.algorithm.iteration.substitute` was added))
```
tirkarthi: Thanks for this. Shouldn't this be "std-algorithm-iteration-substitute,`std.algorithm.iteration.substitute` was added" since the below content has "$(LI $(LNAME2 std-algorithm-iteration-substitute,`std.algorithm.iteration.substitute` was added)" . I think `iteration` was cut off to `ite` here.

schveiguy: I fixed it, definitely seems like it should be that way.

schveiguy:
Feel free to merge when the docs are done building and work correctly.

changelog/2.079.0.dd
wilzbach: Thanks!

PR #2392, opened 2018-06-20
Discourage usage of property functions

JinShil:
There is significant debate over what's happening with the `@property` feature demonstrated in the review of https://github.com/dlang/dmd/pull/8320

The uncertainty and lack of direction on this feature is causing too much confusion and conflict accompanied by speculative claims.

If there is any truth to the claims that `@property` was a mistake, then we need to make that known instead of setting traps for users by remaining silent.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -592,6 +592,10 @@ $(H2 $(LNAME2 optional-parenthesis, Optional Parentheses))
 
 $(H2 $(LNAME2 property-functions, Property Functions))
 
+    $(P WARNING: The definition and usefulness of property functions is being reviewed, and the implementation
+    is currently incomplete.  Using property functions is not recommended until the definition is
+    more certain and implementation more mature.)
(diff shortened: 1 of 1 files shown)


spec/function.dd
```diff
@@ -592,6 +592,9 @@ $(H2 $(LNAME2 optional-parenthesis, Optional Parentheses))
 
 $(H2 $(LNAME2 property-functions, Property Functions))
 
+    $(P WARNING: The future of property functions is uncertain.  Using property functions is
+    discouraged until the feature can be properly designed and implemented.)
```
Geod24: I would be a bit less dramatic:

"The definition and usefulness of `@property` function is being reviewed, and their implementation is currently incomplete w.r.t the current specifications. Using property functions is not recommended until the feature [...]"

JinShil: Done.

PR #2397, opened 2018-06-27
Deprecate scope as a type constraint

JinShil:
Spec update to accompany https://github.com/dlang/dmd/pull/8410


JinShil:
This is blocked waiting on the outcome of @WalterBright use of `scope struct` in https://github.com/dlang/dmd/pull/8369#discussion_r198675586

PR #2400, opened 2018-07-01
Fix local Latex PDF build

wilzbach:
At least on my system the PDF build doesn't default to UTF-8 anymore.
Also, there seems to be an issue with the cyrillic letters, so for the sake of simplicity I used one of the already predefined variables which are correctly encoded for HTML, ebook and TEX.


wilzbach:
Ping @CyberShadow - are you okay with this?

CyberShadow:
Thanks :)

PR #2402, opened 2018-07-02
Change deprecated C array syntax to D array syntax in the spec example

wilzbach:
Required for https://github.com/dlang/dmd/pull/8438


JinShil:
Thanks!

PR #2403, opened 2018-07-02
hasMember doc: add import module effect to example

JohanEngelen:
It was very surprising to me that all imported functions because "members" of the struct according to "hasMember" trait. Let's add that behavior to the documentation.


wilzbach:
Removed the trailing whitespace.

wilzbach:
Definitely a good idea to document the current behaviour. Thanks!

PR #2410, opened 2018-07-06
Specify that the struct invariant may be violated when the destructor is called

FeepingCreature:
Specify that the struct invariant may be violated when the destructor is called, because the struct may be in its T.init state.
See also Issue #19065, PR #19037.
Note that `moveEmplace` already implicitly assumes that this is the case! Otherwise, there'd be little sense in it resetting its source to T.init if it has an elaborate destructor.

The underlying problem is that struct destructors are the one struct lifetime method that is always, unavoidably called by the compiler and cannot be bypassed. (this() can be bypassed with T.init, this(this) can be bypassed with moveEmplace.)

### DIFF
--- a/spec/struct.dd
+++ b/spec/struct.dd
@@ -1223,6 +1223,11 @@ $(H2 $(LEGACY_LNAME2 StructDestructor, struct-destructor, Struct Destructors))
 
         $(P Unions may not have fields that have destructors.)
 
+        $(P It is important to keep in mind that even in $(D @safe) code,
+        the object will not necessarily have been constructed when the destructor is called.
+        However, it will either be constructed or in its $(D .init) state, which the destructor
+        must be able to handle gracefully.)
+
 $(H2 $(LNAME2 StructInvariant, Struct Invariants))
 
 $(GRAMMAR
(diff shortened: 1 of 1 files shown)


FeepingCreature:
The deeper reasoning behind this is that since we allow `t = T.init` in @safe code, this is equivalent to me to admitting that struct variables may at any time either be an initialized struct value or `T.init`. Since struct destructors are called on variables regardless of whether they're properly initialized or not, struct destructors must be able to handle both `T.init` and `T(...)`.

dukc:
I think this sould be done EVEN if there wasn't the case with `.init` values: The last public call checks the same value anyway. thus a check on destructor is waste of computational power.

FeepingCreature:
Are you sure this commit is still correct! The diff looks weird.

thewilsonator:
It looks fine to me. Done. Ideally the compiler should insert a unittest to make sure that the destructor hanles the case in `(T.init).~this()`.

FeepingCreature:
To copypaste the reasoning from the `Nullable` debate: the problem is that this code becomes impossible:

    struct S {
      Object o;
      invariant (o !is null);
      @disable this();
    }

thewilsonator:
... so ah, 3 days later ...

thewilsonator:
Should this be merged or closed?

FeepingCreature:
Yeah, this can be removed. The change in question was never implemented, and the `union` hack has proven a viable - if exceedingly dirty - workaround.

PR #2411, opened 2018-07-06
fix various broken links in articles

aG0aep6G:
fixes issue 19067 - [Mixins] Dead Link to template-mixin


aG0aep6G:
Rebased to get DAutoTest on board. I've also fixed an oversight by amending. I think this is ready to go now.

PR #2413, opened 2018-07-07
upgrade dependencies to fix build

MartinNowak:
- apparently resolves conflict between libdparse 0.7.2-alpha.4 and ~>0.8


MartinNowak:
Let's check the html outcome before merging this.

wilzbach:
image

I manually browsed the HTML and it doesn't seem to be broken. The changes between ddox 0.16.10 and 0.16.11 was just the libdparse upgrade anyhow: https://github.com/rejectedsoftware/ddox/compare/v0.16.10...master

PR #2415, opened 2018-07-10
-betterC spec update: struct destructors now work

wilzbach:


JinShil:
I believe this was fixed with https://github.com/dlang/druntime/pull/2204 and https://github.com/dlang/druntime/pull/2178, but when I made those PRs, I was just focused on remove a specific limitation and didn't intend to holistically enable struct destruction in the language.

@WalterBright What are your thoughts?  Is there anything preventing struct destruction from working in `betterC?

WalterBright:
@JinShil nope!

PR #2418, opened 2018-07-11
Change regarding invariant-noncompilant .init values

dukc:
See [https://forum.dlang.org/post/ohohfaisyaregskmkzme@forum.dlang.org](https://forum.dlang.org/post/ohohfaisyaregskmkzme@forum.dlang.org)

### DIFF
--- a/spec/contracts.dd
+++ b/spec/contracts.dd
@@ -203,8 +203,13 @@ class Date
         $(LI postconditions)
         )
 
-        $(P The invariant is not checked if the class or struct is implicitly constructed using
-        the default $(CODE .init) value.
+        $(P The invariant is not checked if the class is implicitly constructed using
+        the default $(CODE .init) value, since it is $(CODE null), meaning no instantiation.
+        However, the invariant is checked when a struct is implicitly constructed via
+        $(CODE .init) value. $(CODE .init) value of a struct does not need to satisfy
+        the invariant, but in that case it must not be instantiated. The $(CODE .init) value
+        may always be used without instantiating it, such as inside $(CODE is) or $(CODE typeof)
+        expressions.
         )
 
         $(P The code in the invariant may not call any public non-static members


FeepingCreature:
Note that the invariant is not checked on access to `.init` at the moment, which would be both hard to do (since `.init` is usually a constant) and pointless (since `.init` usually violates invariants). My impression was that `.init` would in fact silently violate invariants, but that this was acceptable because the generated values were still "basically safe", ie. valid or null.

jmdavis:
Strictly speaking, the invariant is called after a constructor is completed, before and after public functions are called and before the destructor is called. `T.init` itself never triggers the invariant, because no functions are called. IMHO, what you're saying here confuses matters considerably.

It would be _far_ better to say something like "Default initialization does not trigger the invariant as no functions are called during default initialization."

FeepingCreature:
The problem is the asymmetry: the invariant is not called on `T foo = T.init;`, but it _is_ called on scope exit of `foo`.

dukc:
As FeepingCreature said, non-invariantcompliant only make sense when you are not going to change the value at all, because a change would trigger the invariant first. But even that does not make sense when there is the destructor, which inevitably also triggers the invariant. Either making `init` values always checked or disabling checks before the destructor would solve the problem.

But I just realised this option would mean code breakage. The disadvantage of the other option is that it closes the future possibility of removing the check before a function call, unless we allow doing work on .init value which does not satisfy the invariant. But I believe preventing code breaking is more important, so I'll close this in few days unless someone objects.

jmdavis:
> The problem is the asymmetry: the invariant is not called on T foo = T.init;, but it is called on scope exit of foo.

Honestly, I don't see any problem with that. If you understand how an invariant works, then you can code accordingly. The key issue as far as updating the spec goes is ensuring that it's clear. Also, checking the invariant after default initialization is kind of silly given that it's always going to be the same value, and if there were any problem with its invariant, then it would fail as soon as you called any functions on it. And honestly, from a strictly pedantic point of view, having an invariant that fails for `T.init` is in violation of the concept of an invariant anyway, since it's always supposed to be true for a value of type `T`. That can be annoying in some cases just like it can be annoying sometimes that we don't have default constructors in D, but having `T.init` comes with its own set of advantages, so it's a matter of pros and cons.

Regardless, if you want to argue about why the current behavior should be changed and write a DIP about it, then that's fine, and we may very well end up with something better as a result, but that's a completely separate issue from updating the spec. If the spec isn't clear enough about how the language currently works, then we can update it to clarify it, but a DIP is required if you actually want to change the semantics of the language.

FeepingCreature:
I disagree **fundamentally** with both your points, but this is probably not the place, yeah.

Are you on IRC or Discord? This might be easier to work out in conversation.

That said, I'll probably just work around it with union.

dukc:
I said: deprecating **instantiating** invalid `.init` values, not deprecating structs that have an invalid `.init` value. So the final behaviour after deprecation period would be exactly as described in this PR.

dukc:
Closed as I promised.

PR #2424, opened 2018-07-19
betterC section title: add 'features'

wilzbach:
Just "Not Available" sounds pretty strange.
BTW should we also add an example here on how to use D with -betterC without any runtime?

### DIFF
--- a/spec/betterc.dd
+++ b/spec/betterc.dd
@@ -83,7 +83,7 @@ $(H2 $(LNAME2 retained, Retained Features))
     $(LI `assert` failures are directed to the C runtime library)
     )
 
-$(H2 $(LNAME2 consequences, Not Available))
+$(H2 $(LNAME2 consequences, Unavailable Features))
 
     $(P D features not available with $(B BetterC):)
 


spec/betterc.dd
```diff
@@ -83,7 +83,7 @@ $(H2 $(LNAME2 retained, Retained Features))
     $(LI `assert` failures are directed to the C runtime library)
     )
 
-$(H2 $(LNAME2 consequences, Not Available))
+$(H2 $(LNAME2 consequences, Not Available Features))
```
jacob-carlborg: Doesn’t it sound better with “Features Not Available”?

wilzbach: The section about is titled: "Retained Features"
https://dlang.org/spec/betterc.html#retained

Hence, I went with "Not Available Features", though I don't have a strong feeling about this as a non-native speaker.

jmdavis: "Unavailable Features" would be much better than "Not Available Features," though "Features Not Available" works fine too.

wilzbach:
> "Unavailable Features" would be much better than "Not Available Features," though "Features Not Available" works fine too.

Changed. Thanks!

PR #2428, opened 2018-07-23
Add SAoC banner

wilzbach:
Disclaimer: I don't like this banner (and I didn't create it).
There's a thing called Corporate Identity (and imho that applies to Open Source organizations too).

Our Corporate Identity has been created by an actual designer -> https://github.com/dlang-community/artwork

In short
--------

- Font: Montserrat
- the _new_ logo (it has been decided over a year ago that this is the new logo for new projects)
- Background color: #532360 (or the nice stars)
- Main Font color: white

image

CC @aldacron


jacob-carlborg:
Unfortunately this logo has the same problem as the old one. It's not a recognizable shape, it doesn't work in a single color. Which means the logo is very inflexible. We would be stuck with a single color scheme. There will be cases where the logo will not work for these reasons.

wilzbach:
@jacob-carlborg I'm aware of this problem (and that's the unfortunately main reason why we haven't swapped the icon on the menu and the favicon).
Anyhow, I'm more criticizing the fact that the pink font on blue background doesn't fit at all with the existing dlang.org design.

jacob-carlborg:
Oh, I didn't see that image before. That Symmetry Autumn of Code image looks absolutely horrible. Did the designer came up with that as well?

wilzbach:
> Did the designer came up with that as well?

No. He is absolutely innocent (actually I think since 2017's DConf at Sociomantic he hasn't done any D-related work and he doesn't even work at Sociomantic anymore).

jacob-carlborg:
Come on, we cannot put this on our web site. The more I look at the image the more upset I become.

wilzbach:
The new banner Mike send me today:

image

(PR updated with it)

jacob-carlborg:
Looks better but it still looks a bit out of place.

wilzbach:
Updated again (thanks @mdparker)

image

I still think it could be more beautiful, but the application deadline is very soon, so I think we should move forward with this and we can always improve it over the next days.

Are you okay with this as well @CyberShadow?

CyberShadow:
It's not bad, but if I'm allowed to nitpick:

- Should it take as much vertical space? Maybe something in the shape of the 2nd banner but aesthetics of the 3rd?
- Font rendering looks a bit weird and pixelated
- Maybe use one of the fonts we are already using on dlang.org?
- What is the license of the laptop-on-autumn-leaves photo, are we free to use it?

mdparker:
That version used Montserrat Black as the font

. I paid for a royalty-free license for the photo via Depositphotos.

Here's a leaderboard version using Roboto Slab. 

saoc_comp3

wilzbach:
Ping @mdparker @CyberShadow 
I think we should try to get this to the frontpage of dlang.org ASAP (it's only two weeks to the application deadline) and I'm inclined to merge this as we don't seem to be capable to come with something better :/

> Here's a leaderboard version using Roboto Slab.

Hmm, the font shadowing is imho way too big. Also the earlier mentioned problem that no one knows what the Symmetry Autumn of Code is applies here.

wilzbach:
Oh I am far from happy, but as mentioned I think it's important to get something up on the front page for increased visibility of this campaign.

PR #2429, opened 2018-07-24
Document pragma(mangle) special case on OSX

Geod24:


spec/pragma.dd
```diff
@@ -144,6 +144,11 @@ $(H3 $(LNAME2 mangle, $(D pragma mangle)))
     $(P There must be one $(ASSIGNEXPRESSION) and it must evaluate at compile time to a string literal.
     )
 
+    $(IMPLEMENTATION_DEFINED On Mac OSX, an extra underscore (`_`) is prepended to the string since 2.079,
```
jacob-carlborg: Not sure what the rest of the documentation says but the official name by Apple is `macOS`.

Geod24:
Looks like this would also happen on Win32 ( https://github.com/dlang/dmd/pull/7620 ) ?
CC @ibuclaw

spec/pragma.dd
```diff
@@ -144,6 +144,11 @@ $(H3 $(LNAME2 mangle, $(D pragma mangle)))
...
+        as is done by the C/C++ toolchain. This allows to use the same `pragma(mangle)` for all POSIX
```
JinShil: "allows to use" --> "allows using"

ibuclaw:
@Geod24 yes that's correct IIRC.

Geod24:
Fixed:
- Added win32 and rephrased the last sentence
- `Mac OSX` => `macOS`
- `allows to use` => `allows using`

PR #2432, opened 2018-07-30
Bump stable DUB/DMD

wilzbach:
Not sure if this helps anything, but trying to dig to root of https://github.com/dlang/dmd/pull/8519#issuecomment-408674418.


CyberShadow:
Shouldn't this target the stable branch, considering that's the branch that's failing?

wilzbach:
Yes, sorry!

Though it seems like `stable` doesn't build at all anymore :/
You don't by chance see which the first commit was that didn't build anymore?

wilzbach:
Lol. So with the recent master to stable merge, the build issues of the stable branch seemed to have disappeared. So we probably don't need this anymore, but OTOH having a more recent dub binary doesn't never hurt either.

PR #2436, opened 2018-08-02
Document how to run unittests in -betterC

wilzbach:
I think this isn't obvious to people, especially as it was listed in the "Not Available" features section.

### DIFF
--- a/spec/betterc.dd
+++ b/spec/betterc.dd
@@ -82,7 +82,33 @@ $(H2 $(LNAME2 retained, Retained Features))
     $(LI COM classes and C++ classes)
     $(LI `assert` failures are directed to the C runtime library)
     $(LI `switch` with strings)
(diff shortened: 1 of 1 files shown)


spec/betterc.dd
```diff
@@ -81,8 +81,34 @@ $(H2 $(LNAME2 retained, Retained Features))
...
+        u();
+}
+---
+
+$(CONSOLE
+> dmd -betterC -unittest -run test.d
+dmd_runpezoXK: foo.d:3: Assertion `0' failed.
+)
+
+However, in `-betterC` `assert` don't use Druntime's assert and are directed to `assert` of the C runtime library instead.
```
JinShil: don't --> doesn't, are --> is

Unless you wan to change `assert` to `assert`s.

PR #2438, opened 2018-08-03
Un-document D_NoBoundsChecks and replace it with D_BoundsChecks

Geod24:
As explained in dlang/dmd#8532 the prefered style is positive version identifiers.
The DMD PR adds the new version identifier to be used for at least a release before D_NoBoundsChecks is deprecated.
In the meantime, it can already be undocumented.


JinShil:
I don't know what the policy is exactly, but I would prefer if `D_NoBoundsCheck` was just labeled as deprecated in the next release.  I think it's valuable for users to be able to refer to the deprecated feature, so they can understand what it is that they need to transition away from.  Once the feature is removed, *then* it can be removed from the documentation.  Just my opinion, though.  I'll go along with whatever the established policy or convention is.

Geod24:
Pausing this for the time being.

PR #2439, opened 2018-08-05
[function.dd] Improve Lazy Parameters section

ntrel:
* Add two sub-headings for better Table of Contents.
* Be more precise about lazy parameters.
* Add link from lazy params to lazy variadics.


jacob-carlborg:
Please squash the commits.

PR #2442, opened 2018-08-10
Fix DDoc DDLINK macro usage

wilzbach:
I don't know why, but `DDLINK` is defined as:

```
DDLINK=$(LINK2 $(ROOT_DIR)$1.html, $3)
```

### DIFF
--- a/gpg_keys.dd
+++ b/gpg_keys.dd
@@ -36,7 +36,7 @@ $(P You can also download them as $(LINK2 d-keyring.gpg,keyring) file.)
 
 $(H2 D Security Team)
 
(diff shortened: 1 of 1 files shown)


aG0aep6G:
> I don't know why, but DDLINK is defined as:
> 
> DDLINK=$(LINK2 $(ROOT_DIR)$1.html, $3)

In ebook.ddoc it's defined as `DDLINK=$(RELATIVE_LINK2 $2, $3)`. As far as I know, the ebook only uses the spec pages. So the second argument matters on spec pages, but not on other pages.

PR #2448, opened 2018-08-19
dstyle: Avoid newlines in DDoc sections

JinShil:
This is to accommodate @WalterBright's preference state at https://github.com/dlang/dmd/pull/8572#discussion_r211086434 ...

> All these blank lines detract a bit from readability, as one has less code that fits on the screen.

... so that we can enforce a consistent coding style in DMD.

EDIT: I'm not really in favor of this, but it's impossible to be productive and consistent with contributors if @WalterBright is not on board with our coding conventions and style.  Either he has to bend to our preferences (unlikely in my experience), or we have to bend to his.  I'm getting frustrated enforcing these style conventions only to have @WalterBright recommend the opposite later.  It makes us look inept, and contributors, understandably, get annoyed by the mixed messages.

### DIFF
--- a/dstyle.dd
+++ b/dstyle.dd
@@ -412,12 +412,9 @@ enum myFancyConstant;
 /**
 Checks whether a number is positive.
 `0` isn't considered as positive number.
-
 Params:
     number = number to be checked
-
 Returns: `true` if the number is positive, `0` otherwise.
-
 See_Also: $(LREF isNegative)
 */
 bool isPositive(int number)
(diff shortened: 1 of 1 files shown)


jmdavis:
I don't believe that this is what is done in much of Phobos right now and IMHO would seriously harm readability in many cases. Also, as far as Phobos goes, he doesn't work on it much, so forcing it for Phobos on Walter's whim really doesn't make much sense IMHO.

JinShil:
@jmdavis, You're right, this would be too disruptive for Phobos, and looking through the DDoc comments in Phobos it is apparent that without the newline separation, the documentation just becomes blog of text, difficult to visual parse and glance through.  This is not going to fly.  I guess I'll continue recommendations in conflict with @WalterBright's preferences.

PR #2450, opened 2018-08-20
Update spec for Issue 19180 - Expose... __traits(isZeroInit, T)

n8sh:
This is the companion of https://github.com/dlang/dmd/pull/8583

### DIFF
--- a/spec/traits.dd
+++ b/spec/traits.dd
@@ -43,6 +43,7 @@ $(GNAME TraitsKeyword):
     $(GLINK isOut)
     $(GLINK isLazy)
     $(GLINK isReturnOnStack)
+    $(GLINK isZeroInit)
     $(GLINK hasMember)
     $(GLINK identifier)
     $(GLINK getAliasThis)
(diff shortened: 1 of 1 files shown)


spec/traits.dd
```diff
@@ -400,6 +401,36 @@ static assert(!__traits(isTemplate,foo!int()));
...
 
+$(H2 $(GNAME isZeroInit))
+
+        $(P Takes one argument which must be a type. If the type's
+        $(LINK2 property.html#init, default initializer) has no non-zero bits
+        then $(D true) is returned, otherwise $(D false). This always will
+        return $(D true) for a class $(D C) because $(D C.init) is a null
+        reference.)
+
+---
```
wilzbach: Bonus points for:

```
---
...
---
)
```

because then the spec tester will ensure that this code continues to work. I know that we have unit tests, but code that doesn't get tested slowly breaks over time :/

spec/traits.dd
```diff
@@ -400,6 +401,36 @@ static assert(!__traits(isTemplate,foo!int()));
 static assert(!__traits(isTemplate,"string"));
 ---
 
+$(H2 $(GNAME isZeroInit))
+
+        $(P Takes one argument which must be a type. If the type's
+        $(LINK2 property.html#init, default initializer) has no non-zero bits
```
wilzbach: Please don't hard-code HTML links as they won't work in the ebook or PDF.
Use sth. like `$(DDSUBLINK spec/property, init, default initializer)`.
(I now that this macros are poorly documented.)

spec/traits.dd
```diff
@@ -400,6 +401,36 @@ static assert(!__traits(isTemplate,foo!int()));
...
+        then $(D true) is returned, otherwise $(D false). This always will
```
wilzbach: s/This/`isZeroInit`/

spec/traits.dd
```diff
@@ -400,6 +401,36 @@ static assert(!__traits(isTemplate,foo!int()));
...
+        return $(D true) for a class $(D C) because $(D C.init) is a null
```
wilzbach: FYI: Backticks can be used instead of `$(D ..)`. It results in the same rendered HTML

n8sh:
Changed the style and the formatting as suggested.

I'm going to give a little pushback on removing specific mention of how this works with classes. I found a similar remark helpful reading the documentation of `std.traits.hasElaborateDestructor`.  How about if I move it out of the definition and instead add it to the example code?

<details><summary>Example Code</summary>

```d
struct S1 { int x; }
struct S2 { int x = -1; }

static assert(__traits(isZeroInit, S1));
static assert(!__traits(isZeroInit, S2));

void test()
{
    int x = 3;
    static assert(__traits(isZeroInit, typeof(x)));
}

// `isZeroInit` will always return true for a class C
// because `C.init` is null reference.

class C { int x = -1; }

static assert(__traits(isZeroInit, C));
```
</details>
<p></p>

`TypeInfo.initializer()` treating classes specially is further reason to emphasize that "initializer" here doesn't mean anything other than `T.init`.

ZombineDev:
Yes, it's a good thing to note and I think the example is the right place for this, though interfaces, function pointers, delegates, associative arrays and array slices have the same initializer.

n8sh:
OK. Is it good as it now stands?

PR #2451, opened 2018-08-21
Fix a few visual flaws in the DMD help page

wilzbach:
- Fix overwritten LIB macro (the macro was never used, but overwrote the actual LIB macro)
- Use individual install steps for Linux (all other pages have them too) + fix list indentation
- Fix links to the packages and install script

This page is quite complicated, so it's better to check the resulting diff of DAutoTest here.

### DIFF
--- a/dcompiler.dd
+++ b/dcompiler.dd
@@ -914,7 +914,6 @@ Macros:
     DMD_CONF=$(RELATIVE_LINK2 dmd-conf, dmd.conf)
     SC_INI=$(RELATIVE_LINK2 sc-ini, $(DMDDIR)\windows\bin\sc.ini)
     DMC=$(LINK2 http://www.digitalmars.com/ctg/sc.html, dmc)
-    LIB=$(LINK2 http://www.digitalmars.com/ctg/lib.html, lib)
     OPTLINK=$(LINK2 http://www.digitalmars.com/ctg/optlink.html, link.exe)
     DUB=$(LINK2 https://github.com/dlang/dub, dub$(BINEXT))
     RDMD=$(LINK2 $(ROOT_DIR)rdmd.html, rdmd$(BINEXT))
--- a/dcompiler.dd
+++ b/dcompiler.dd
@@ -891,10 +891,10 @@ Macros:
         SUBNAV=$(SUBNAV_CLI_REFERENCE)
     INSTALLATION_SCRIPT_HINT=
 $(P
-$(MESSAGE_BOX gray, $(B Hint) - The official $(LINK2 $(ROOT_DIR)download.sh, packages)
+$(MESSAGE_BOX gray, $(B Hint) - The official $(LINK2 $(ROOT_DIR)download.html, packages)
 performs these steps automatically.
 Alternatively, you can install DMD in your user directory with
-the $(LINK2 $(ROOT_DIR)install.sh, install script).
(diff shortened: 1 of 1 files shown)


dcompiler.dd
```diff
@@ -85,22 +85,53 @@ C:\&gt;unzip dmc.zip
...
+
+$(CONSOLE
+sudo cp $(DMDDIR)/linux/bin64/dmd.conf /etc/$(DMD_CONF)
+)
+    )
+
+    $(LI Copy binaries to $(D /usr/local/bin):
+
+$(CONSOLE
+sudo cp $(DMDDIR)/linux/bin64/{dmd,$(DUMPOBJ),$(OBJ2ASM),$(SHELL),rdmd} /usr/local/bin
```
ZombineDev: I think `$(OBJ2ASM)` and `$(SHELL)` are windows only.

ZombineDev:
```
ls dmd2/linux/bin64/
ddemangle*  dmd.conf  dumpobj*   obj2asm*  README.TXT
dmd*        dub*      dustmite*  rdmd*

File: dcompiler.dd [hunk 2]
```diff
@@ -85,22 +85,53 @@ C:\&gt;unzip dmc.zip
...
+)
+    )
+
+    $(LI Copy the library to $(D /usr/local/lib):
```
ZombineDev: The user will also need to copy the "headers":
* `dmd2/src/druntime/import/`
* `dmd2/src/phobos/std/`
* `dmd2/src/phobos/etc/`

dcompiler.dd
```diff
@@ -85,22 +85,53 @@ C:\&gt;unzip dmc.zip
...
+export PATH="$HOME/dmd/linux/bin64:$PATH"
+)
+
+    Alternatively, you can install dmd locally yourself by:
+    $(P $(UL
+
+    $(LI Copy $(D $(DMD_CONF)) to $(D /etc):
+
+$(CONSOLE
+sudo cp $(DMDDIR)/linux/bin64/dmd.conf /etc/$(DMD_CONF)
```
ZombineDev: Copying `dmd.conf` as it is won't work as paths to druntime and phobos inside are relative to the location of the dmd binary.

dcompiler.dd
```diff
@@ -85,50 +85,109 @@ C:\&gt;unzip dmc.zip
...
+[Environment64]
+DFLAGS=-I/usr/local/include/d/dmd -L-L/usr/local/lib -L--export-dynamic -fPIC
+---
+
+    )
+
+    $(LI Copy binaries to $(D /usr/local/bin):
+
+$(CONSOLE
+sudo cp $(DMDDIR)/linux/bin64/{dmd,$(DUMPOBJ),$(OBJ2ASM),rdmd,ddemangle,dub,dustmite} /usr/local/bin
```
wilzbach: I guess this is the only controversial part as it uses `bin64`, but as the existing example above (L95) already just shows 64-bit, I went for simplicity with 64-bit manual installation instructions only.
32-bit users are rare these days (some distros like mine don't support 32-bit anymore) and if they want to install dmd 32-bit globally, they know how to do it anyhow.
This is for the casual user, who for some reason, doesn't want neither the local folder nor the distro package globally installed.

wilzbach:
Ping. Anything blocking this?

PR #2452, opened 2018-08-21
[spec] Document ref and out parameters

ntrel:
* Split the `ref` example from the `out` examples, and put it first.
* Briefly describe each.
* Tweak the `scope` description.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -1265,13 +1265,13 @@ $(H3 $(LNAME2 ref-params, Ref and Out Parameters))
 
 $(SPEC_RUNNABLE_EXAMPLE_RUN
 ---
-void def(ref int x)
+void inc(ref int x)
 {
     x += 1;
 }
 
 int z = 3;
-def(z);
+inc(z);
 assert(z == 4);
 ------------
 )
@@ -1290,7 +1290,7 @@ void foo(out int x)
 
 int a = 3;
 foo(a);
-assert(x == 0);
+assert(a == 0);
 
 void abc(out int x)
 {
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -1263,7 +1263,7 @@ $(H3 $(LNAME2 ref-params, Ref and Out Parameters))
     A Ref parameter takes an lvalue argument, so changes to its value will operate
     on the caller's argument.)
 
-$(SPEC_RUNNABLE_EXAMPLE_COMPILE
+$(SPEC_RUNNABLE_EXAMPLE_RUN
 ---
 void def(ref int x)
 {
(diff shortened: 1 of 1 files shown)


spec/function.dd
```diff
@@ -1234,20 +1234,42 @@ int foo(in int x, out int y, ref int z, int q);
...
     $(TROW $(D inout), argument is implicitly converted to an inout type)
     )
 
+$(H3 $(LNAME2 ref-params, Ref and Out Parameters))
+
+    $(P By default, parameters take rvalue arguments.
+    A Ref parameter takes an lvalue argument, so changes to its value will operate
+    on the caller's argument.)
+    
+$(SPEC_RUNNABLE_EXAMPLE_COMPILE
```
wilzbach: Why not run + assert?

ntrel: I just re-ordered existing code. Assert would be better, but this file doesn't seem to use it.

spec/function.dd
```diff
@@ -1234,20 +1234,42 @@ int foo(in int x, out int y, ref int z, int q);
...
     to avoid using $(D in) until it is properly defined and implemented.  Use $(D scope const)
     or $(D const) explicitly instead.)
-    $(TROW $(D out), parameter is initialized upon function entry with the default value
+    $(TROW $(D ref), parameter is passed by reference)
+    $(TROW $(D out), parameter is passed by reference and initialized upon function entry with the default value
     for its type)
 
-    $(TROW $(D ref),   parameter is passed by reference)
-    $(TROW $(D scope), references in the parameter
+    $(TROW $(D scope), references to the parameter
```
ntrel: I presume the original was just a mistake, as `scope` is not transitive anyway.

spec/function.dd
ntrel: @wilzbach Changed to use assert.

spec/function.dd
JinShil: It doesn't seem to make sense either way.  I recommend not even including it in this PR.  Leave it for another PR as it might derail the progress on this PR.

spec/function.dd
```diff
@@ -1230,35 +1230,61 @@ int foo(in int x, out int y, ref int z, int q);
...
     for its type)
 
-    $(TROW $(D ref),   parameter is passed by reference)
-    $(TROW $(D scope), references in the parameter
+    $(TROW $(D scope), References to the parameter
     cannot be escaped (e.g. assigned to a global variable).
-    Ignored for parameters with no references)
+    Ignored for parameters with no references.
+    Note: The `scope` attribute is currently only enforced by `dmd`
+    when the `-dip1000` switch is passed.
```
JinShil: Since DIP 25 and DIP 1000 are experimental features (at least I think they are). I'd prefer if we had separate a separate document in the spec describing their semantics.  Then when they are official, their documentation can be folded into the spec.

ntrel: `-dip25` is already in the docs (I wrote it): 
https://dlang.org/spec/function.html#return-ref-parameters

JinShil: Thanks. We're lucky to have it.  But I think it should be moved until DIP 25 and 1000 are official.

spec/function.dd
ntrel: @JinShil OK, moved to #2454.

ntrel:
Probably this just needs someone to do the `git commit --amend` trick. (I don't have my terminal available ATM).

ntrel:
`--amend` retriggered tests, but now there's a different error. AFIACT this is the problem (which I also get for #2454 but with different input line numbers):
```
<recently read> \egroup 
                        
l.19214 }
         
? 
! Emergency stop.
<recently read> \egroup 
                        
l.19214 }
         
!  ==> Fatal error occurred, no output PDF file produced!
```

ntrel:
Same test error for #2459 - it seems anything that touches `spec/function.dd` gets this error. @wilzbach Any idea?

ntrel:
I forgot about this, but the test error blocking this has long since gone away - which then revealed an actual error, now fixed.

PR #2453, opened 2018-08-23
Document recent -dip1000 changes

thewilsonator:
Document recent dip1000 related changes in:
https://github.com/dlang/dmd/pull/8346
https://github.com/dlang/dmd/pull/8408
https://github.com/dlang/dmd/pull/8504

@WalterBright is this an accurate description?

cc  @JinShil

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -1241,8 +1241,13 @@ int foo(in int x, out int y, ref int z, int q);
     $(TROW $(D scope), references in the parameter
     cannot be escaped (e.g. assigned to a global variable).
     Ignored for parameters with no references)
+    $(TROW $(D return), $(ARGS Parameter may be returned or copied to the first parameter,
+    but otherwise does not escape from the function.
+    Such copies are required not to outlive the argument(s) they were derived from.
+    Ignored for parameters with no references.
+    See $(DDSUBLINK spec/memory-safe-d, scope-return-params, Scope Parameters).))
     $(TROW $(D lazy), argument is evaluated by the called function and not by the caller)
-        $(TROW $(D const), argument is implicitly converted to a const type)
+    $(TROW $(D const), argument is implicitly converted to a const type)
     $(TROW $(D immutable), argument is implicitly converted to an immutable type)
     $(TROW $(D shared), argument is implicitly converted to a shared type)
     $(TROW $(D inout), argument is implicitly converted to an inout type)
--- a/spec/memory-safe-d.dd
+++ b/spec/memory-safe-d.dd
@@ -24,18 +24,51 @@ $(HEADERNAV_TOC)
 
 $(H2 $(LNAME2 usage, Usage))
 
-        $(P Memory safety can be enabled on a per-function basis using
-        the $(DDSUBLINK spec/function, safe-functions, $(D @safe) attribute).
-        This can be inferred when the compiler has the function body
-        available. The $(DDSUBLINK spec/function, trusted-functions, `@trusted` attribute) can be used when a function has a safe
-        interface, but uses unsafe code internally. These functions can
-        be called from $(D @safe) code.
+        $(P There are three categories of functions from the perspective of memory safety:)
+        $(UL
+                $(LI $(DDSUBLINK spec/function, safe-functions, `@safe`) functions)
+                $(LI $(DDSUBLINK spec/function, trusted-functions, `@trusted`) functions)
+                $(LI $(DDSUBLINK spec/function, system-functions, `@system`) functions)
         )
 
+        $(P `@system` functions may perform any operation legal from the perspective of the language including inherently
+        memory unsafe operations like returning pointers to expired stackframes. These functions may not be called directly from
+        `@safe` functions.)
+
+        $(P `@trusted` functions have all the capabilities of `@system` functions but may be called from
+        `@safe` functions. For this reason they should be very limited in the scope of their use. Typical uses of
+        `@trusted` functions include wrapping system calls that take buffer pointer and length arguments separately so that
+        @safe` functions may call them with arrays.)
+
+        $(P `@safe` functions have a number of restrictions on what they may do and are intended to disallow operations that
+        may cause memory corruption. See $(DDSUBLINK spec/function, safe-functions, `@safe` functions).)
+
+        $(P These attributes may be inferred when the compiler has the function body available, such as with templates.)
+
         $(P Array bounds checks are necessary to enforce memory safety, so
-        these are enabled (by default) for $(D @safe) code even in $(B
-        -release) mode.
+        these are enabled (by default) for `@safe` code even in $(B -release) mode.
         )
+
+    $(H3 $(LNAME2 scope-return-params, Scope and Return Parameters))
+
+         $(P The function parameter attributes `return` and `scope` are used to track what happens to low-level pointers
+         passed to functions. Such pointers include: raw pointers, arrays, `this`, classes, `ref` parameters, delegate/lazy parameters,
+         and aggregates containing a pointer.)
+
+         $(P $(D scope) ensures that no references to the pointed-to object are retained, in global variables or pointers passed to the
+         function (and recursively to other functions called in the function), as a result of calling the function.
+         Variables in the function body and parameter list that are `scope` may have their allocations elided as a result.)
+
+         $(P $(D return) indicates that either the return value of the function or the first parameter is a pointer derived from the
+         `return` parameter or any other parameters also marked `return`.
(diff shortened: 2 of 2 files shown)


JinShil:
This is great, but I think it needs to be more formally described in the spec:

https://github.com/dlang/dmd/pull/8346
--------------
That PR causes the compiler to silently remove the `return` storage class under certain conditions.  I'd like that documented at https://dlang.org/spec/function.html#return-ref-parameters

Furthermore the `return` storage class is not yet documented in the table at https://dlang.org/spec/function.html#param-storage

https://github.com/dlang/dmd/pull/8408
----------------
First of all, I think that should be reverted, and we should come up with another way for the programmer to meta-programatically infer the storage classes.

But, since that's unlikely, I'd like for the inference rules to documented in the spec.  I'd also like for the spec to make it clear when inference takes place -- I think it only takes place for templates, not for ordinary functions or methods.  That needs to clarified.

Also, what is `maybeScope` and how does that work.  Maybe you can distill the following into something comprehendible. 
https://github.com/dlang/dmd/blob/74f89cc8f876279093c3dd6ef9a3918db329db73/src/dmd/escape.d#L1601-L1620

https://github.com/dlang/dmd/pull/8504
--------------
I think this requires mention in https://dlang.org/spec/function.html#return-ref-parameters.  1) The first parameter is special, iff it's `ref` or `out` (I think).  2) special mention for constructors that the first parameter is essentially a `ref this`.  But I'm still not sure how this works for member functions as those too have an implicit `this`. 3) The relationship between `return` parameters and the special first `ref` parameter. 

You're not going to buy me off with a summary and a few paragraphs 😛.  I think there's more to it than that.  We also have to consider what this does to the DIP 1000 document.  Personally, I think before any of these PR were/are merged there should be a PR to DIP1000 approving the design first.

JinShil:
There's also a lot in https://wiki.dlang.org/DIP25 and https://github.com/dlang/DIPs/blob/master/DIPs/DIP1000.md.  The spec should include multiple examples illustrating the concepts.

thewilsonator:
>You're not going to buy me off with a summary and a few paragraphs

I wasn't expecting to. This is also for me to understand the state of the implementation and verify that it is what I think it is, which is probably not true in the general case.

>1) The first parameter is special, iff it's ref or out

With dlang/dmd#8346 it's ignored if its not a pointer (incl. ref/out), so I think not?

>Also, what is maybeScope and how does that work. Maybe you can distill the following into something comprehendible.

No idea either.

ntrel:
@JinShil:
> a lot of these features are feature-gated behind -dip25 and -dip1000 command-line switches

Return ref is already documented with a note about needing -dip25. I've added a note about -dip1000 for scope in #2452.

JinShil:
> Return ref is already documented with a note about needing -dip25. I've added a note about -dip1000 for scope in #2452.

Thank you.  But I still recommend moving anything related to DIP 25 and 1000 to a separate page until the design and implementation are finalized.

ntrel:
@thewilsonator I've made some changes [here](https://github.com/ntrel/dlang.org/commits/dip1000) to improve wording (and formatting). (Unfortunately github won't let me set your repo as the base for a PR for some reason).

thewilsonator:
Thanks will merge them tomorrow.

thewilsonator:
Pinging @WalterBright again.

ZombineDev:
@andralex can you review this?

spec/memory-safe-d.dd
```diff
@@ -24,19 +24,53 @@ $(HEADERNAV_TOC)
...
         $(P Array bounds checks are necessary to enforce memory safety, so
         these are enabled (by default) for $(D @safe) code even in $(B
         -release) mode.
         )
 
+    $(H3 $(LNAME2 scope-return-params, Scope & Return Parameters))
+
+         $(P The function parameter attributes $(D return) and $(D scope) are used to track what happens to low-level pointers 
+         passed to functions. Such pointers include: pointers, $(D this), classes, `ref` parameters, delegate/lazy parameters, 
+         and aggregates containing a pointer.)
```
jacob-carlborg: Arrays? Or do you count those as aggregates?

ntrel: Suggest we use `raw pointers, this, variables of reference type, ref/lazy parameters, ref return values and aggregates containing such pointers`.

thewilsonator:
Pinging @WalterBright again: if you want me to be a functional PR manager, you need to review this.

12345swordy:
@thewilsonator Try emailing him directly if he is not responding.

spec/memory-safe-d.dd
```diff
@@ -24,19 +24,53 @@ $(HEADERNAV_TOC)
...
+
+         $(P $(D return) indicates that either the return value of the function or the first parameter is a pointer derived from the 
+         `return` parameter or any other parameters also marked $(D return). 
+         For constructors, `return` applies to the (implicitly returned) $(D this) reference. 
+         For void functions, `return` applies to the first parameter $(I iff) it is `ref`; this is to support UFCS,
+         property setters and non-member functions (e.g. `put` used like `put(dest, source)`).)
+
+         $(P These attributes may appear after the formal parameter list, in which case they apply either to a method's $(D this) parameter, or to
+         a free function's first parameter $(I iff) it is `ref`.
+         $(D return) or $(D scope) is ignored when applied to a type that is not a pointer.)
```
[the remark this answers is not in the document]
thewilsonator: Hmm,  I should reword that to `low level pointer` using the definition a few paragraphs further up.

thewilsonator:
Woohoooo! Now I just need to figure out how to fix the LaTeX errors...

spec/memory-safe-d.dd
```diff
@@ -24,19 +24,53 @@ $(HEADERNAV_TOC)
...
+        may cause memory corruption. See $(DDSUBLINK spec/function, safe-functions, $(D @safe) functions).)
+
+        $(P These attributes may be inferred when the compiler has the function body available, such as with templates.)
+
         $(P Array bounds checks are necessary to enforce memory safety, so
         these are enabled (by default) for $(D @safe) code even in $(B
         -release) mode.
         )
 
+    $(H3 $(LNAME2 scope-return-params, Scope & Return Parameters))
```
wilzbach: I'm pretty sure LaTeX complains about the raw &. Use the Ddoc macro for ampersand or just "and".

spec/memory-safe-d.dd
```diff
@@ -24,19 +24,53 @@ $(HEADERNAV_TOC)
...
+                $(LI $(DDSUBLINK spec/function, safe-functions, $(D @safe)) functions)
+                $(LI $(DDSUBLINK spec/function, trusted-functions, $(D `@trusted`)) functions)
+                $(LI $(DDSUBLINK spec/function, system-functions, $(D `@system`)) functions)
         )
 
+        $(P $(D `@system`) functions may perform any operation legal from the perspective of the language including inherently
+        memory unsafe operations like returning pointers to expired stackframes. These functions may not be called directly from
+        $(D `@safe`) functions.)
+
+        $(P $(D `@trusted`) functions have all the capabilities of $(D `@system`) functions but may be called from
```
wilzbach: Why do you double escape things here? Markdown backticks do the same as the D Ddoc macro.

spec/memory-safe-d.dd
thewilsonator: Thanks for the suggestion

thewilsonator:
Hmm, can't tell if that actually improved things. I'm going to split this PR.

wilzbach:
Why don't you build it locally?

spec/function.dd
```diff
@@ -1241,8 +1241,13 @@ int foo(in int x, out int y, ref int z, int q);
     $(TROW $(D scope), references in the parameter
     cannot be escaped (e.g. assigned to a global variable).
     Ignored for parameters with no references)
+    $(TROW $(D return), Parameter may be returned or copied to the first parameter,
```
andralex: The LaTeX errors are caused by the last comma in this line. The comma is interpreted by ddoc as a parameter separator, which in LaTeX is translated into an column separator (ampersand) because you're in a table. To fix, you may want to use `$(TROW $(D return),  $(ARGS Parameter may be returned...))`

andralex:
Made a commit directly in the interest of time.

thewilsonator:
Thanks, I fixed some whitespace you left. Hopefully this works.

thewilsonator:
IT WORKS!!! Thanks everyone!

WalterBright:
@tgehr I cannot see a rationale here for this change.

dkorpel:
Probably just an oversight.

PR #2454, opened 2018-08-23
[spec/function.dd] Add scope parameter section

ntrel:
* Tweak existing wording
* Mention `@safe` & -dip1000 needed for enforcement

*Update:*
* Split existing `return ref` example.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -1242,9 +1242,18 @@ int foo(in int x, out int y, ref int z, int q);
     $(TROW $(D out), parameter is passed by reference and initialized upon function entry with the default value
     for its type)
 
-    $(TROW $(D scope), references in the parameter
-    cannot be escaped (e.g. assigned to a global variable).
-    Ignored for parameters with no references)
+    $(TROW $(D scope), $(ARGS
+    The parameter does not escape the function call
+    (e.g. by being assigned to a global variable).
+    Ignored for any parameter that is not a reference type.
+    Escape analysis is only done for `@safe` functions, for other functions `scope` semantics must
+    be manually enforced.
+    As the parameter must not escape, the compiler can avoid heap-allocation of a unique argument
+    to a `scope` parameter. Due to this, passing e.g. a `new T` argument may not violate `@nogc`.
+    A typical example of this is passing a delegate literal to a `scope` parameter.
+    $(B Note): `scope` escape analysis is currently only done by `dmd`
+    when the `-dip1000` switch is passed.
+    ))
     $(TROW $(D return), $(ARGS Parameter may be returned or copied to the first parameter,
     but otherwise does not escape from the function.
     Such copies are required not to outlive the argument(s) they were derived from.
@@ -1421,10 +1430,6 @@ pure void f()
         )
 
 $(H3 $(LNAME2 return-ref-parameters, Return Ref Parameters))
-
-        $(P Note: The `return` attribute is currently only enforced by `dmd`
-        when the `-dip25` switch is passed.
-        )
 
         $(P Return ref parameters are used with
         $(RELATIVE_LINK2 ref-functions, ref functions) to ensure that the
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -1246,14 +1246,6 @@ int foo(in int x, out int y, ref int z, int q);
     The parameter must not escape the function call
     (e.g. by being assigned to a global variable).
     Ignored for any parameter that is not a reference type.
-    `scope` escape analysis is only done for `@safe` functions. For other functions `scope`
-    semantics must be manually enforced.)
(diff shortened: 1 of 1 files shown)


spec/function.dd
```diff
@@ -1238,9 +1238,13 @@ int foo(in int x, out int y, ref int z, int q);
...
 
     $(TROW $(D ref),   parameter is passed by reference)
-    $(TROW $(D scope), references in the parameter
+    $(TROW $(D scope), References to the parameter
     cannot be escaped (e.g. assigned to a global variable).
-    Ignored for parameters with no references)
+    Enforced only for `@safe` functions.
+    Ignored for parameters with no references.
+    Note: The `scope` attribute is currently only enforced by `dmd`
+    when the `-dip1000` switch is passed.
```
jacob-carlborg: Doesn't `scope` on delegate has some affect even without `-dip1000`?

thewilsonator: Delegates are reference types though. But the elision of GC allocation of the closure should be mentioned.

ntrel: @jacob-carlborg I've now documented this, in a more general way. Walter has said that heap allocation for other reference types besides delegates can be elided too, e.g. for array concatenation:
https://forum.dlang.org/post/ofo9f5$2bt7$1@digitalmars.com

ntrel:
@thewilsonator Sorry I forgot about this. Rebased! (At least I get CircleCI notifications now when my builds fail :-) ).

spec/function.dd
```diff
@@ -1242,9 +1242,18 @@ int foo(in int x, out int y, ref int z, int q);
...
+    The parameter does not escape the function call
+    (e.g. by being assigned to a global variable).
+    Ignored for any parameter that is not of reference type.
+    Escape analysis is done for `@safe` functions, for other functions `scope` semantics must
+    be manually enforced.
+    As the parameter does not escape, the compiler can avoid heap-allocation of a unique argument
+    to a `scope` parameter. Due to this, such a function call can become `@nogc`.
+    This is typically done when passing a delegate literal to a `scope` parameter.
+    $(B Note): Escape analysis is currently only done by `dmd`
+    when the `-dip1000` switch is passed.
```
MoonlightSentinel, line 1255: This statement isn't exactly true, there is some scope analysis without `-dip1000` (esp. given that [`-dip25` is now partially enabled by default](https://dlang.org/changelog/2.092.0.html#dip25)). DIP1000 enables some idioms that are otherwise forbidden, e.g. pointers to local variables.

spec/function.dd
```diff
@@ -1408,7 +1417,7 @@ pure void f()
 
 $(H3 $(LNAME2 return-ref-parameters, Return Ref Parameters))
 
-        $(P Note: The `return` attribute is currently only enforced by `dmd`
+        $(P $(B Note): The `return` attribute is currently only enforced by `dmd`
```
MoonlightSentinel, line 1420: This could mention the deprecations issued since v2.092 ([which partially enables DIP25 by default](https://dlang.org/changelog/2.092.0.html#dip25)

spec/function.dd
ntrel: -dip25 only analyses escaping refs, not scope parameters. But I changed it to 'scope escape analysis...' for clarity.

spec/function.dd
ntrel: Thanks. I decided to just remove the note, as the deprecation warnings will become errors at some point.

spec/function.dd
```diff
@@ -1242,9 +1242,18 @@ int foo(in int x, out int y, ref int z, int q);
...
-    cannot be escaped (e.g. assigned to a global variable).
-    Ignored for parameters with no references)
+    $(TROW $(D scope), $(ARGS
+    The parameter does not escape the function call
+    (e.g. by being assigned to a global variable).
+    Ignored for any parameter that is not of reference type.
+    Escape analysis is done for `@safe` functions, for other functions `scope` semantics must
+    be manually enforced.
+    As the parameter does not escape, the compiler can avoid heap-allocation of a unique argument
+    to a `scope` parameter. Due to this, such a function call can become `@nogc`.
```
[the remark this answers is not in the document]
ntrel: Yes, please see 5843070 for the changed wording.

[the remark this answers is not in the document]
MoonlightSentinel: That's definitly better but I would change the example because passing `new T` to a scope parameter isn't `@nogc` (yet), You could also mention array literals (which were implemented recently)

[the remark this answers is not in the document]
ntrel: OK, I've now listed several things including array literals that can potentially avoid allocation.

ntrel:
The scope parameter changes were getting a bit big for the function parameter table, so I have moved them to their own section and split out an example from the `return scope` docs.

PR #2457, opened 2018-08-26
Fix Issue 18104: improve example showing that aliases cannot be used …

ghost91-:
…for expressions

### DIFF
--- a/spec/declaration.dd
+++ b/spec/declaration.dd
@@ -473,8 +473,9 @@ alias abc = foo.bar; // is it a type or a symbol?
 struct S { static int i; }
 S s;
 
-alias a = s.i; // illegal, s.i is an expression
-alias b = S.i; // ok
+alias a = s.i; // ok, s.i is a symbol
+alias b = S.i; // ok, S.i also is a symbol
+alias c = a + b; // illegal, a + b is an expression
 b = 4;         // sets S.i to 4
 -----------
 


spec/declaration.dd
```diff
@@ -473,8 +473,9 @@ alias abc = foo.bar; // is it a type or a symbol?
 struct S { static int i; }
 S s;
 
-alias a = s.i; // illegal, s.i is an expression
-alias b = S.i; // ok
+alias a = s.i; // ok, s.i is a symbol
+alias b = S.i; // ok, S.i also is a symbol
```
JinShil: `alias b = s.i` seems redundant and doesn't illustrate anything new.  I suggest maybe adding a member `static int j;` to `S` and use that instead to illustrate the idea, or something like that.

aG0aep6G: >     a = 2; // sets s.i to 2

Doesn't work like that. An alias of `s.i` is the same as one of `S.i`. It doesn't carry `this` with it. It can't be used to access `s.i`.

ghost91-: Oh yeah, of course 🤦‍♂️ I suppose I was bit too quick here.

ghost91-: Or should we simplify the example even more? After all, it is about alias not working on expressions. Structs are not necessary to show this:
```D
int a;

alias b = a; // ok, a is a symbol
alias c = a + a; // illegal, a + a is an expression
```

JinShil:
I recommend this:

```D
struct S 
{
    static int i;
    static int j;
}

alias a = S.i; // OK, `S.i` is a symbol
alias b = S.j; // OK. `S.j` is also a symbol
alias c = a + b; // illegal, `a + b` is an expression
a = 2;         // sets `S.i` to `2`
b = 4;         // sets `S.j` to `4`
```

This purpose of this example is simply to illustrate the one can't alias an expression.  Trying to show that `S.i` and `s.i` are equivalent is irrelevant, and actually [that feature is currently being debated](https://github.com/dlang/dmd/pull/8242#issuecomment-388594197).

PR #2459, opened 2018-08-31
Improve spec/function#lazy-params example

ZombineDev:
To better showcase the D's expressiveness.


ZombineDev:
@jacob-carlborg @thewilsonator can you guys help me debug the ddoc failure. I tried various things locally ~2 months ago, but it always seemed to fail for no reason.

thewilsonator:
this is using `SPEC_RUNNABLE_EXAMPLE_COMPILE` which I have just discovered in #2507 wraps a `void main(){ ... }` around the example. so that code becomes

```d
void main()
{
    import std.stdio : writeln;

    void main()
    {
        int x;
        3.times(writeln(x++)); // <<< fails here
        writeln("-");
        writeln(x);
    }

    void times(int n, lazy void exp)
    {
        while (n--)
            exp();
    }
}
```

which for me fails with `Error: no property times for type int`, presumably because times is nested. Whats with all the LaTeX failures, I have no idea. Maybe it confuses the macros. Either way it doesn't look like you will be able to show off `3.times(writeln(x++));` with  `SPEC_RUNNABLE_EXAMPLE_COMPILE `, you'd have to do a no main wrapping version.

ZombineDev:
Thanks for looking into this @thewilsonator. Actually I had gotten only so far as the LaTeX problems and I haven't even considered the the problem with `SPEC_RUNNABLE_EXAMPLE_COMPILE`. I propose we add `SPEC_RUNNABLE_EXAMPLE_COMPILE_CUSTOM_MAIN` version which would not try to wrap the code in another `main()`. I'll see if that's easy to do when I have some free time on my hands.

wilzbach:
>  I propose we add SPEC_RUNNABLE_EXAMPLE_COMPILE_CUSTOM_MAIN version which would not try to wrap the code in another main(). I'll see if that's easy to do when I have some free time on my hands.

Please don't. The script already checks for `void main`. Better to fix that, but looking on this from my phone I don't think that's the root cause. We have had tons of problems with the function page getting too big and thus LaTeX throwing weird errors.

thewilsonator:
> The script already checks for void main.

Good to know

>  Better to fix that, but looking on this from my phone I don't think that's the root cause. We have had tons of problems with the function page getting too big and thus LaTeX throwing weird errors.

Ahh, right.

wilzbach:
It could be related to the `-` and LaTeX not liking it (though I can't build the PDF at the moment either), but I tested this commit here.

PR #2464, opened 2018-09-06
Change download button text to be less confusing.

mdparker:


jacob-carlborg:
Please squash commits.

mdparker:
Doesn't that happen automatically?

jacob-carlborg:
No.

mdparker:
I made the edits online and will be away fom my computers for a while. So is the 'Squash & Merge' button a config option?

ZombineDev:
> So is the 'Squash & Merge' button a config option?

Yep, I just squashed your commits with the help of it ;)

PR #2465, opened 2018-09-06
Undocument Pascal calling convention

Geod24:
As @ibuclaw said:
```
- D dropped Windows 3.x and 9.x support back in 2.058 - http://dlang.org/deprecate.html#Windows%203.x%20and%20Windows%209x%20support
- It causes problems with ambiguity between `extern(Pascal)` functions and template value arguments - https://issues.dlang.org/show_bug.cgi?id=14591
- Is not supported by LLVM (http://llvm.org/docs/LangRef.html#calling-conventions) or GCC (https://gcc.gnu.org/onlinedocs/gcc/x86-Function-Attributes.html)
```

Then we can deprecate it in DMD. @WalterBright ?


ibuclaw:
FYI @rainers

rainers:
As noted in the dmd PR: `extern(Pascal)` is still used in a couple of sys.windows modules. Running the corresponding Windows SDK header files through the preprocessor shows the `PASCAL` annotation gets translated to `__stdcall` for both x86 and x64. So the D declarations are actually wrong.

Geod24:
Ping @WalterBright

thewilsonator:
@WalterBright you've got 72h to make up your mind on this.
@Geod24 please rebase

Geod24:
Rebased. Also fixed the druntime header.

Geod24:
Time's up

PR #2466, opened 2018-09-13
Add D_DoubleFloat version identifier

ibuclaw:
This is for when the target may have a floating point coprocessor (`D_HardFloat` is true), but only supports single precision operations.  So double precision is still handled using software emulation.

`D_DoubleFloat` implies that `D_HardFloat` is also true.  Conversely, `D_SoftFloat` implies that `D_DoubleFloat` is false.

The combination of `D_HardFloat=true` and `D_DoubleFloat=false` can occur on MIPS and RISC-V targets.


ibuclaw:
As the version is quite involved across all targets, I'm open to making it `MIPS_DoubleFloat` and `RISCV_DoubleFloat` respectively.  Though I'd rather avoid adding more target-specific version conditions if we can instead use a platform agnostic identifier.

ibuclaw:
Or maybe `HardFloatDouble` ?

ibuclaw:
Any more discussions about the colour of the shed here?

ibuclaw:
Have been blocked on adding RISC-V ports for over a month now.

PR #2467, opened 2018-09-13
Deprecated Target_SoftFloat and Target_HardFloat version identifiers

ibuclaw:
These don't have any benefit over `D_HardFloat` and `D_SoftFloat` respectively.


spec/version.dd
```diff
@@ -262,15 +262,10 @@ $(H3 $(LEGACY_LNAME2 PredefinedVersions, predefined-versions, Predefined Version
         $(TROW $(ARGS $(D X86_64)) , $(ARGS Intel and AMD 64-bit processors))
         $(TROW $(ARGS $(D ARM)) , $(ARGS The ARM architecture (32-bit) (AArch32 et al)))
         $(TROW $(ARGS $(D ARM_Thumb)) , $(ARGS ARM in any Thumb mode))
-        $(TROW $(ARGS $(D ARM_SoftFloat)) , $(ARGS The ARM $(D soft) floating point ABI))
-        $(TROW $(ARGS $(D ARM_SoftFP)) , $(ARGS The ARM $(D softfp) floating point ABI))
```
klickverbot: Doesn't this lose information? Granted, `softfp`/`soft` isn't an ABI change, but it's still potentially interesting information.

ibuclaw: There's no benefit that I could find to have it.  It's a bit of a misnomer, as `ARM_SoftFP` means `D_HardFloat` is available.

klickverbot: Say you are writing naked inline asm on ARM. How do you decide on ABI vs. instruction set when D_SoftFloat/D_HardFloat only describes the latter?

ibuclaw: I would just let the compiler do that for you.

```
@attribute("naked")
double foo(double a, double b)
{
  asm { /* do something. */; }
  bar(b, a);
  asm { /* do something else. */; }
}
```

klickverbot: That's… not what I meant. Let me rephrase my question: With the versions gone, how do you detect (`version`) the floating point ABI? And how the instruction set available? (Naked inline asm was only an example where both matter.)

klickverbot: > Calling a function from asm instruction is not something you'd do yourself […]

What are you basing this statement on?

By the way, your above example is invalid as per the GCC documentation:

> Only basic asm statements can safely be included in naked functions (see Basic Asm). While using extended asm or a mixture of basic asm and C code may appear to work, they cannot be depended upon to work reliably and are not supported.

ibuclaw: Arguably made redundant by `__traits(floatAbi)`, which gives the information much more accurately.

ZombineDev: If I understand correctly the discussion, with this PR we won't be able to differentiate between the FP ABI (how to link with external code) and whether native FP instructions are available. It seems that the distinction is only necessary when you you're writing code that needs to link with code using the soft ABI, but you can and want to use native FP instructions internally, @klickverbot correct? Is this a scenario important enough, that we need to support it?

klickverbot: > You could restrict yourself to basic asm, but the compiler could still change the abi of the function you are calling on a whim - for instance LTO - or even better remove the function as dead code.

If a compiler does this on code that might be referenced from inline ASM without a way to get around it (e.g. by specifying a list of "used" symbols), then I'd consider it a bug. This doens't only concern inline assembly, by the way, but also symbols exported for other reasons (e.g. to be referenced from a non-LTO library you link against). You can know the ABI for public (as in ELF global, exported, non-internalized) functions – LTO doesn't touch them, and that's the entire point of having an ABI.

> It seems that the distinction is only necessary when you you're writing code that needs to link with code using the soft ABI, but you can and want to use native FP instructions internally

It is quite common in the ARM world (`softfp`). Debian used to ship with it by default (still does?).

PR #2468, opened 2018-09-15
Add a small paragraph about implicit conversion of strings

BBasile:
Unless i missed it elsewhere the case is never mentioned. It can be confusing because implicit conversion works for string literals and not for other arrays of char so now it's explained that this case works because the conversion is done at compile time.

### DIFF
--- a/spec/arrays.dd
+++ b/spec/arrays.dd
@@ -1064,13 +1064,16 @@ $(H2 $(LNAME2 implicit-conversions, Implicit Conversions))
         )
 
         $(P A dynamic array $(D T[]) can be implicitly converted to one of the
-        following ($(D U) is a base class of $(D T)):)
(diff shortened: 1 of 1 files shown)


spec/arrays.dd
```diff
@@ -1064,14 +1064,17 @@ $(H2 $(LNAME2 implicit-conversions, Implicit Conversions))
...
-        following ($(D U) is a base class of $(D T)):)
+        following ($(D U) is a base class of $(D T)) and neither U nor T are character types:)
 
         $(UL
         $(LI $(D const(U)[]))
         $(LI $(D const(U[])))
         $(LI $(D void[]))
         )
 
+        $(P Arrays of character are never implicitly converted,
```
jacob-carlborg: This compiles:

```d
char[] a;
void[] b = a;
```

PR #2471, opened 2018-09-21
Spec for __traits(getTargetInfo)

TurkeyMan:
For https://github.com/dlang/dmd/pull/8724


TurkeyMan:
It's mentioned in the spec PR.
I think some subset would be mentioned in the spec, and the rest implementation defined.

TurkeyMan:
DMD PR was merged...

PR #2475, opened 2018-09-25
Update spec to disallow invalid binary literals

wolframw:
This is the spec update for https://github.com/dlang/dmd/pull/8738 / https://issues.dlang.org/show_bug.cgi?id=19246
Binary literals like `0b` and `0b_` are now deprecated. A valid binary literal must now have at least one digit after the prefix.


wolframw:
I managed to screw up my local repository big time and now I can't reopen this PR, so I made a new proper one: https://github.com/dlang/dlang.org/pull/2476
I apologize for the mess.

PR #2477, opened 2018-09-29
All target infos

TurkeyMan:
For https://github.com/dlang/dmd/pull/8763

### DIFF
--- a/spec/traits.dd
+++ b/spec/traits.dd
@@ -64,6 +64,7 @@ $(GNAME TraitsKeyword):
     $(GLINK classInstanceSize)
     $(GLINK getVirtualIndex)
     $(GLINK allMembers)
+    $(GLINK allTargetInfos)
     $(GLINK derivedMembers)
     $(GLINK isSame)
     $(GLINK compiles)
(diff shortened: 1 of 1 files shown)


spec/traits.dd
```diff
@@ -842,6 +844,24 @@ export
...
+version (CppRuntime_Microsoft)
+    static assert(__traits(getTargetInfo, "cppRuntimeLibrary") == "libcmt");
+---
+
+        $(P Keys are implementation defined, allowing relevant data for exotic targets.
+        A reliable subset exists which are always available:
+        )
+
+        $(UL
+        $(LI $(D "cppRuntimeLibrary"))
```
jacob-carlborg: Can we document what it means as well?

TurkeyMan: I thought about it briefly, but I wouldn't know how to create a table with 2 columns in ddoc.
It's also astoundingly self-explanatory :)

jacob-carlborg: > I thought about it briefly, but I wouldn't know how to create a table with 2 columns in ddoc.

For simplicity I would go with a list (as it currently is) with the name in bold, a dash (or a colon) and then the description, like this:

* **cppRuntimeLibrary** - description

Or if you want to go with a table, you can have a look at how it's done in the algorithm modules in Phobos, they all have a table at the top.

> It's also astoundingly self-explanatory :)

In this case perhaps. BTW, is it the C++ runtime library or C++ standard library. I know Clang has a separate C++ ABI and standard library.

TurkeyMan:
Walter rejected the DMD PR

PR #2478, opened 2018-10-03
Fix mistake in C++ code

Pharap:
`int` is not guaranteed to be at least 32 bits,
it can be as small as 16 bits in some implementations (e.g. AVR chips).
`long` on the other hand is guaranteed to be at least 32 bits.

### DIFF
--- a/articles/cpptod.dd
+++ b/articles/cpptod.dd
@@ -638,7 +638,7 @@ struct Integer&lt; 16 &gt;
 
 struct Integer&lt; 32 &gt;
 {
-    typedef int int_type ;
+    typedef long int_type ;
 };
 
 struct Integer&lt; 64 &gt;


jmdavis:
Given that C++11 finally does have portable integer types of specific sizes, that whole section should probably be replaced, but I don't expect that anyone really wants to take the time to come up with a better example.

Pharap:
@jmdavis 
I had assumed C++98/C++03 style was being used on purpose for some reason.
If pre-C++11 style isn't being used on purpose then most of the other examples are also quite outdated.

For example, a C++11 example for "Recursive Templates" would be:
```cpp
template<int n> class factorial
{
  public:
    constexpr int result = n * factorial<n - 1>::result;
};

template<> class factorial<1>
{
  public:
    constexpr int result = 1;
};

void test()
{
    std::printf("%d\n", factorial<4>::result); // prints 24
}
```

It's worth noting that as of the addition of `contexpr`, the following is also common:
```cpp
constexpr int factorial(int n)
{
	return (n <= 1) ? 1 : (n * factorial(n - 1));
}
```

And a C++14 replacement for the "Meta Templates" section would be something like this:
```cpp
#include <type_traits>

template<std::size_t nbits> struct Integer
{
	using int_type =
		std::conditional_t<(nbits <= 8), std::int_least8_t,
		std::conditional_t<(nbits <= 16), std::int_least16_t,
		std::conditional_t<(nbits <= 32), std::int_least32_t,
		std::int_least64_t>>>;
	
	static_assert(nbits <= 64, "No suitable integer type");
};

#include <cstdio>

int main()
{
    Integer< 8 > i ;
    Integer< 16 > j ;
    Integer< 29 > k ;
    Integer< 64 > l ;
    std::printf("%d %d %d %d\n", sizeof(i), sizeof(j), sizeof(k), sizeof(l));
    return 0 ;
}
```

Which renders the boost example somewhat obsolete.

And the "Type Traits" example would be something like:
```cpp
#include <type_traits>
#include <cassert>

template<typename T> class IsFunctionT
{
  private:
    template<typename U> static std::false_type test(...);
    template<typename U> static std::true_type test(U (*)[1]);
  public:
    constexpr bool value = IsFunctionT<T>::test<T>(0);
};

void test()
{
    using fp = int (*)(int);

    std::assert(IsFunctionT<fp>::value);
}
```
(Unless you assume `std::true_type` and `std::false_type` aren't available, in which case it's more similar to its original version.)

jmdavis:
I'm pretty sure that the entire page predates C++11 (possibly with some tweaks), but I'd have to go digging through the git history to know for sure. Sometimes, someone takes the time to rework pages like this, but often they were written fairly early on in D2's history and have stayed largely the same for years.

jmdavis:
I don't know what Walter would want to do with pages like this with regards to the version of C++.

Pharap:
Indeed, [the oldest commit](https://github.com/dlang/dlang.org/commit/4542ef3a1def67d4bc16cd959ccb7a93baac813a#diff-1f83f940097c886d33e6a52dea25af2d) I can find for it is from 2008.

There's been a lot of updates since then though, so I'd be surprised if nobody else has brought it up before.

---

However, this PR is fixing an actual mistake.
The mistake holds for all versions of C++, from C++98 through to C++17.
As such, it wouldn't make sense to try to extend this PR into a modernisation project.

Perhaps "Update cpptod.dd to use more modern C++ techniques" should be raised as a separate issue?

PR #2481, opened 2018-10-07
Remove GNEX ltd.

potato4d:
## about

I am an engineer of GNEX ltd. (See screenshot).
GNEX has finished using DLang. Please delete from orgs-using-d page.

## refs

screen shot 2018-10-07 at 17 37 34

https://github.com/orgs/gnexltd/people

### DIFF
--- a/orgs-using-d.dd
+++ b/orgs-using-d.dd
@@ -120,13 +120,6 @@ $(DIVC orgs-using-d center,
             $(FA_HIRING www.funkwerk.com/karriere/stellenangebote)
         )
     )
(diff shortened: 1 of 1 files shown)


potato4d:
@wilzbach 
Can you keep the logo?
There are no developers using D language from our organization, because it may come time to use D language.
I am looking forward to the opportunity to use it again!

CyberShadow:
Removing a file from a git repository's tree will not remove it from its history; one only needs to know that the file was there at some point, and given that, it's trivial to resurrect. So, don't worry about deleting files that aren't used right now.

potato4d:
@CyberShadow 
delete it.
thanks.

PR #2485, opened 2018-10-12
fix Issue 19292 - Mixin expressions should take an argument list

WalterBright:


WalterBright:
Implementation:
https://github.com/dlang/dmd/pull/8813
https://github.com/dlang/dmd/pull/8809
https://github.com/dlang/dmd/pull/8808

PR #2486, opened 2018-10-14
Add guideline for contracts

belka-ew:
- Pre and post contracts should have the same indentation level as their
declaration. Put no space between `in`/`out` and an open parenthesis:
```d
int foo(int r)
in(r == 0)
out(result; result == 0)
do { … }
```
```d
int foo(int r)
in { … }
out(result) { … }
do { … }
```

### DIFF
--- a/dstyle.dd
+++ b/dstyle.dd
@@ -300,6 +300,7 @@ $(LISTSECTION phobos_whitespace, Whitespace,
 -------------------------------
 for (…) { … }
 foreach (…) { … }
+static foreach (…) { … }
 if (x) { … }
 static if (x) { … }
 while (…) { … }
@@ -381,6 +382,19 @@ $(LISTSECTION phobos_declarations, Declarations,
 -------------------------------
 void foo(R)(R r)
 if (R == 1)
+-------------------------------
+    $(LI Pre and post contracts should have the same indentation level as their
+         declaration. Put no space between `in`/`out` and an open parenthesis:)
+-------------------------------
+int foo(int r)
+in(r == 0)
+out(result; result == 0)
+do { … }
+
+int foo(int r)
+in { … }
+out(result) { … }
+do { … }
 -------------------------------
 )
 $(LISTSECTION phobos_class_struct_field_declaration, Class/Struct Field Declarations,


ZombineDev:
At first I also preferred indenting constraints, so I would have agreed with @FeepingCreature that indenting contracts as well would look better, but having followed this style guide for so long, I now prefer not indenting either.

ZombineDev:
CC @andralex

andralex:
I think a paren should be there. Consider:

* In the constructs `func(argument)` the spacing rules around the parens are informed by mathematical notation. So no space before (or after) the open paren.
* In the constructs `if (expr)`, `while (expr)`,  `for (decl; expr; expr)`, `foreach (decl; expr)`, `switch (expr)` etc, the spacing rules around the parens are NOT informed by math and they are more open to taste. Different authors/groups use different conventions, such as `if(expr)`, `if( expr )` etc. In our codebase we go with `if (expr) ` i.e. "space before the open paren and after the close paren, no space after the open paren or before the close paren". That rule is informed by the use of parens as punctuation in written natural language.

Adding the new `in` and `out` notations naturally poses the question of aligning them with the existing rules we have. They are not function calls so they'd naturally align with `if` and others.

andralex:
@ZombineDev thanks for bringing this to my attention

ZombineDev:
@andralex thanks for getting back to me on this - I'll open a PR to correct the style accordingly.

andralex:
Without looking, I think our codebase uses `extern(C)` preponderantly, with the implied justification that it needs no space drawing from syntagma like "401(k)", "501(c)(3)", etc.

ZombineDev:
The predominant style is `extern (...)` by a long margin in dmd and druntime:

| "*.d" files in: | dmd/ | druntime/ | phobos/ |
|-----------------|------|-----------|---------|
| `extern(.*)`    | 436  | 570       | 57      |
| `extern (.*)`    | 2242 | 2017      | 64      |

See for yourselves:
```
grep --color=always -Hnr --include="*.d" 'extern(.*)' dmd/ | wc -l
grep --color=always -Hnr --include="*.d" 'extern (.*)' dmd/ | wc -l
grep --color=always -Hnr --include="*.d" 'extern(.*)' druntime/ | wc -l
grep --color=always -Hnr --include="*.d" 'extern (.*)' druntime/ | wc -l
grep --color=always -Hnr --include="*.d" 'extern(.*)' phobos/ | wc -l
grep --color=always -Hnr --include="*.d" 'extern (.*)' phobos/ | wc -l
```

ZombineDev:
@andralex If there was rationale, I believe it would be that it is consistent with the rest of the language (`for (..)`, `if (..)`, `version (..)`, etc.). It's also much simpler to teach "put space before open paren, unless it is a function call", than a list of exceptions to remember. Are you ok with putting `extern (C)` in the style guide as well?

andralex:
Cool, yeah, if the space after `extern` is prevalent let's. It simplifies everybody's life!

PR #2487, opened 2018-10-15
fix Issue 10692 - Deprecation isn't checked using alias this

WalterBright:


spec/class.dd
```diff
@@ -1179,6 +1179,8 @@ void test()
     int i = s; // i == 4
 }
 ---
+        $(P $(GLINK2 attribute, Attribute)s are ignored for $(AliasThis).
```
RazvanN7: The documentation tester complains that $(AliasThis) is an nonexistent macro. I guess you meant `$(D AliasThis)`

PR #2488, opened 2018-10-16
DStyle: Adjust guidance for contracts

ZombineDev:


ZombineDev:
@andralex please review as discussed in https://github.com/dlang/dlang.org/pull/2486.

PR #2493, opened 2018-10-23
Add changelog clarifications from forum

CyberShadow:
https://forum.dlang.org/post/osxjfludeknpcphxjvzi@forum.dlang.org
https://forum.dlang.org/post/shmmwojvtvuiinnoqjye@forum.dlang.org

### DIFF
--- a/changelog/2.083.0_pre.dd
+++ b/changelog/2.083.0_pre.dd
@@ -57,6 +57,11 @@ $(UL
  $(LI `CppRuntime_Sun`)
 )
 )
+
+$(P
+The C++ runtime in use generally depends on the C++ compiler used to build the C++ code that will be linked against.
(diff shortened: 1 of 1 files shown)


wilzbach:
This will be overwritten by the changelog generator when 2.083 is released is tht changelog generator is still not capable of incremental updates.

CyberShadow:
> The tool is here fwiw

Does it have a way of knowing when the file it's generating is going to be eventually overwritten, or if it's the final version and thus the primary source of truth henceforth (its creation coincides with deleting the individual changelog/ files)?

wilzbach:
No it doesn't have this at the moment, though it could probably just detect that if the output file exists already, that an incremental update is desired (or use a cli flag for this).

Btw a small step forward:

https://github.com/dlang/dlang.org/pull/2494

PR #2499, opened 2018-10-30
Specify effects of parameterized package attrib

mdparker:
The existing doc gave the impression that any package could be used to parameterize the `package` attribute and was not specific about which packages are granted access.


jacob-carlborg:
Please squash commits.

PR #2502, opened 2018-11-05
use newer host compiler to workaround dub regression

MartinNowak:
- also see https://github.com/dlang/dlang.org/pull/2497


MartinNowak:
Mmh, CircleCI still cannot handle stable PRs :/.

PR #2470, opened 2018-09-19
Add new C++ mangle syntax.

look-at-me:
See https://github.com/dlang/dmd/pull/8667

PR #2511, opened 2018-11-18
Don't pluralize with apostrophe s, plus a few other apostrophe fixes

n8sh:


articles/d-array-article.dd
```diff
@@ -148,21 +148,21 @@ $(H2 $(LNAME2 determinism, Determinism))
 
 $(P There is one caveat with slice appending that can bite inexperienced, and even experienced D coders: the apparent non-deterministic behavior of appending.)
 
-$(P Let's say we have a function which is passed a buffer, and writes some number of A's to the buffer (appending if necessary), returning the filled buffer:)
+$(P Let's say we have a function which is passed a buffer, and writes some number of 'A's to the buffer (appending if necessary), returning the filled buffer:)
```
n8sh: Here we can add a leading apostrophe to make it correct because we're talking about the character `'A'`.

thewilsonator:
Please remove the `function.dd` change into a separate PR because that will most probably cause this to fail.

n8sh:
That's strange. Why will it (probably) do that?

thewilsonator:
Because its too big and causes the doc generator to spew LaTeX errors. e.g. #2470 http://dtest.dlang.io/results/7d7cc2e5e4c8abac0eb5e76920efb05821fbe7e9/af558942e752f686d819f2a7eb1e54ee6b603e81/build.log

spec/function.dd
```diff
@@ -1231,7 +1231,7 @@ int foo(in int x, out int y, ref int z, int q);
     $(TROW $(I none), parameter becomes a mutable copy of its argument)
 
     $(TROW $(D in), defined as $(D scope const).  However $(D in) has not yet been properly
-    implemented so it's current implementation is equivalent to $(D const).  It is recommended
+    implemented so its current implementation is equivalent to $(D const).  It is recommended
```
n8sh: Removing this from the PR as @thewilsonator suggested.

PR #2512, opened 2018-11-19
Document the behaviour of switches w.r.t 18712

thewilsonator:


spec/statement.dd
```diff
@@ -1163,14 +1163,31 @@ switch (name)
         less error prone. char, wchar and dchar strings are allowed.
         )
 
-        $(P $(D Implementation Note:) The compiler's code generator may
+        $(P A new scope for local symbols is introduced immediately following a
+        $(GLINK CaseStatement), $(GLINK CaseRangeStatement), or $(GLINK DefaultStatement),
+        extending to the first appearance of one of
+        $(GLINK CaseStatement), $(GLINK CaseRangeStatement), $(GLINK DefaultStatement), `}`
+        or the end of a $(GLINK MixinStatement). One effect of this is 
+        )
```
n8sh: Sentence cuts off.

thewilsonator: Fixed

thewilsonator:
@n8sh can you approve this?

n8sh:
@thewilsonator I'll look it over.

spec/statement.dd
```diff
@@ -1163,14 +1163,46 @@ switch (name)
         less error prone. char, wchar and dchar strings are allowed.
         )
 
-        $(P $(D Implementation Note:) The compiler's code generator may
+        $(P A new scope for local symbols is introduced immediately following a
+        $(GLINK CaseStatement), $(GLINK CaseRangeStatement), or $(GLINK DefaultStatement),
+        extending to the first appearance of one of
+        $(GLINK CaseStatement), $(GLINK CaseRangeStatement), $(GLINK DefaultStatement), `}`
+        or the end of a $(GLINK MixinStatement). One current limitation of this is that
```
n8sh: What do you mean about the scope extending to the end of a mixin statement? It sounds to me like you're saying this code shouldn't work because `x` is out of scope at `writeln(x)`, but it works:

https://run.dlang.io/is/cSGXa6
```d
void main(string[] args)
{
    import std.stdio;

    switch (args.length)
    {
    case 1:
        int x;
        mixin("x += 2;");
        writeln(x);
        break;
    default:
        writeln("default");
    }
}
```

thewilsonator: Perhaps that should be `case labels`, as 
```d
void main(string[] args)
{
    import std.stdio;

    switch (args.length)
    {
    mixin("case 1:");
        int x;
        x += 2;
        writeln(x);
        break;
    default:
        writeln("default");
    }
}
```
doesn't work.

n8sh: This also fails with the same error message:
```d
void main(string[] args)
{
    import std.stdio;

    switch (args.length)
    {
    case 1:
        int x;
        x += 2;
        mixin("writeln(x);");
    mixin("default:");
        writeln("default");
    }
}
```
So the end of a mixin statement definitely doesn't close a case label's implicit scope.

thewilsonator: Huh. So the advice should be "don't mixin a case label".

n8sh: I think the problem is that a mixed-in case label neither starts a new scope nor closes the previous case label's scope. This works:
```d
void main(string[] args)
{
    import std.stdio;

    switch (args.length)
    {
    case 1:
        int x;
        x += 2;
        writeln(x);
        break;
    case 2:
        writeln("2");
        break;
    mixin("default:");
        writeln("default");
    }
}
```

thewilsonator: Wow. I'll have to think a bit more about how to word this. Thanks!

PR #2513, opened 2018-11-19
Fix __traits(getFunctionAttributes) example

n8sh:
The example code runner wraps the code in a function if no `main` is found, which makes the functions nested and changes attribute inference. Also the message printed by `__pragma(msg)` is formatted `tuple("a", "b", "c")` instead of `("a", "b", "c")` as the comment claimed.

### DIFF
--- a/spec/traits.dd
+++ b/spec/traits.dd
@@ -735,7 +735,6 @@ $(SPEC_RUNNABLE_EXAMPLE_COMPILE
 ---
 int sum(int x, int y) pure nothrow { return x + y; }
 
-// prints ("pure", "nothrow", "@system")
 pragma(msg, __traits(getFunctionAttributes, sum));
 
 struct S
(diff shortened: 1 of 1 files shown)


spec/traits.dd
```diff
@@ -735,26 +735,41 @@ $(SPEC_RUNNABLE_EXAMPLE_COMPILE
...
 )
 
     $(P Note that some attributes can be inferred. For example:)
 
 $(SPEC_RUNNABLE_EXAMPLE_COMPILE
 ---
-// prints ("pure", "nothrow", "@nogc", "@trusted")
 pragma(msg, __traits(getFunctionAttributes, (int x) @trusted { return x * 2; }));
+
+void main(){}
```
n8sh: The `main()` for this example is unnecessary because function literals have the same function attribute inference rules as nested functions, but I'm including it so if someone clicks "Edit" and puts something else in the output will not become misleading.

PR #2514, opened 2018-11-20
fix issue 19374 - TypeVector undefined in grammar

BBasile:


BBasile:
Done, renamed the rule "Vector".

PR #2517, opened 2018-11-27
Update getAliasThis documentation per issue 19439

n8sh:
Companion PR to https://github.com/dlang/dmd/pull/9016

### DIFF
--- a/spec/traits.dd
+++ b/spec/traits.dd
@@ -630,11 +630,34 @@ writeln(__traits(identifier, var));             // "var"
 
 $(H2 $(GNAME getAliasThis))
 
-    $(P Takes one argument, a symbol of aggregate type.
-        If the given aggregate type has $(D alias this), returns a list of
-        $(D alias this) names, by a tuple of $(D string)s.
-        Otherwise returns an empty tuple.
+    $(P Takes one argument, a type. If the type has `alias this` declarations,
+        returns a sequence of the names (as `string`s) of the members used in
+        those declarations. Otherwise returns an empty sequence.
     )
+
+$(SPEC_RUNNABLE_EXAMPLE_COMPILE
+---
+alias AliasSeq(T...) = T;
+
+struct S1
+{
+    string var;
+    alias var this;
+}
+static assert(__traits(getAliasThis, S1) == AliasSeq!("var"));
+static assert(__traits(getAliasThis, int).length == 0);
+
+pragma(msg, __traits(getAliasThis, S1));
+pragma(msg, __traits(getAliasThis, int));
+---
+)
+
+        Prints:
+
+$(CONSOLE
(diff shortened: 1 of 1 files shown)


spec/traits.dd
```diff
@@ -630,12 +630,35 @@ writeln(__traits(identifier, var));             // "var"
 
 $(H2 $(GNAME getAliasThis))
 
-    $(P Takes one argument, a symbol of aggregate type.
-        If the given aggregate type has $(D alias this), returns a list of
-        $(D alias this) names, by a tuple of $(D string)s.
-        Otherwise returns an empty tuple.
+    $(P Takes one argument, a symbol. If the symbol is of a type which has
+        $(D alias this), returns a list of $(D alias this) names, by a tuple
```
mdparker: names, by a tuple -> names as a tuple

ZombineDev: The name we use to call what `AliasSeq` represents was debated at length in the newsgroup a couple years ago and we agreed that we should call them "sequences" to distinguish them from tuple-like objects like the one defined in `std.typecons.Tuple`.

The difference is the following:
* Tuples are value-types holding other run-time specified values, whose types are fixed at compile-time, i.e. at run-time you can change value of the i-th element, but not its type. Nesting is respected, i.e. `Tuple!(int, Tuple!(Tuple!double)) != Tuple!(int, double)`.

* Sequences are lists of compile-time entities (which can include literal values, types, and symbols). They are immutable and they don't nest, i.e. `AliasSeq!(int, AliasSeq!(AliasSeq!double)) == AliasSeq!(int, double)`

So, in summary I recommend the following wording:

```suggestion
        Takes one argument, a type. If the type has `alias this` declaration(s),
        returns the sequence members (represented as `string`s) which were
        part of those declarationd. Otherwise returns an empty sequence.
```

spec/traits.dd
```diff
@@ -630,12 +630,35 @@ writeln(__traits(identifier, var));             // "var"
...
+        of $(D string)s. Otherwise returns an empty tuple.
     )
 
+$(SPEC_RUNNABLE_EXAMPLE_COMPILE
+---
+alias TypeTuple(T...) = T;
```
ZombineDev: As mentioned above, `TypeTuple` is the wrong name, as it's neither a tuple, nor in the example below holds any types.

```suggestion
alias AliasSeq(Args...) = Args;
```

n8sh: Implemented.

spec/traits.dd
n8sh: I have not adopted the suggested wording exactly, because to me "returns the sequence members which [...]" sounds like we are selecting some members of an existing sequence. Please review whether the new wording is acceptable.

PR #2518, opened 2018-12-04
Update sponsors page

mdparker:


mdparker:
We try to keep it sorted by total amount donated.

PR #2519, opened 2018-12-04
Update donate.dd

EmTee70:
I tried to add the link to Mikes blog post and a second Iframe for the campaign. Not sure howto see if ist ok.

### DIFF
--- a/foundation/donate.dd
+++ b/foundation/donate.dd
@@ -34,10 +34,12 @@ $(DONATE_ITEM Donate through Flipcause, users,
 
     Flipcause helps non-profit organizations like the D Language Foundation manage their
     donations and provide more value to donors. Donating through Flipcause can enable us
-    to more efficiently put your donation to use.
+    to more efficiently put your donation to use. As described in this $(LINK2 https://dlang.org/blog/2018/11/10/the-new-fundraising-campaign/, D Blog post).
     $(BR)$(BR)
+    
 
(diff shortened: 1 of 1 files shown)


mdparker:
We don't need a second iframe. I want to add the integration code for their menu. It's going to require modifying more than just the donate.dd file, I believe, so I'll need to experiment with a local build of the site. I intend to set aside some time in the coming weekend to get it done.

EmTee70:
I have seen, that you might have to integrate some javascript code to get there into the head.

With some div's around it might be possible to get the iframes placed side by side, Not brilliant, but a fast solution, that would need an update of the page, after every campaign, but you will  have to update the donate page anyway, if you start a new campaign and want to get your blog post linked?

https://stackoverflow.com/questions/23737427/how-to-put-two-iframes-side-by-side

mdparker:
> you will have to update the donate page anyway, if you start a new campaign and want to get your blog post linked?

No, that's the whole point of using the Flipcause menu. I can add campaigns to the menu via their site. As long as I link to the menu, I'll never have to update the link. The Donate button is the sidebar of the blog does what I'm talking about.

And I don't need to link to the blog post from the donate page. The menu is sufficient.

EmTee70:
I understand. But I am not convinced :-)

mdparker:
This one can be closed since #2523 was merged.

PR #2523, opened 2018-12-10
Replace Flipcause General Fund  card with campaign menu.

mdparker:


mdparker:
@thewilsonator @EmTee70 This is in preference to #2519. 

It turned out to be easier than I expected. I struggled with integrating their massive CSS and JS in a way that kept donate.dd clean, only to find it had trouble on slow connections. Then I discovered they had a lightweight version that's just an iframe -- no CSS, no JS. Easy.

EmTee70:
So your search for a simpler solution was with more success, than my own.
I just looked at the output of 
https://www.flipcause.com/widget/widget_home/NDMzMzI=
Which is  what you embedded. Unfortunately, it is not linking to the detail page(es) and not displaying the status of the campaign.

This is not your fault, it is just as it is. 

(Should link to:

https://www.flipcause.com/secure/cause_pdetails/NDUwNTY=

and not 

https://www.flipcause.com/widget/give_now/NDUwNTY=

as it is.)

mdparker:
It's not a matter of just changing the link. The link I have is for the campaign menu, not the individual campaign. That comes directly from the Integration Center as in our Flipcause account. When you click a campaign in the menu, you get an integrated campaigns that has different configuration options than the popup window you get from the individual cards like the one I embedded in the blog post. 

If you click on the campaign card in the campaign menu, you'll see a menu in the top right corner of the campaign page. There's an 'About' page there. I have no control over that layout.

I did find an option to display the goal, end date, and contributors for the integrated campaign. I've configured it to show, so you can see it now.

wilzbach:
I still don't like that we have to pay 100$/month to Flipcause. That's a lot of money in consideration of the amount of donations we receive each month.

wilzbach:
> What is happening in the background after the approval

Nothing. What should be happening?

> and when will the change be visible on dlang.org?

Auto-deployment takes 10-20 minutes (depending on how busy DAutotest is).

PR #2524, opened 2018-12-15
convert 'char[]' to 'string'

Flying-Toast:


Flying-Toast:
I purposely left `char[]`s in some places where it made more sense for the example (mostly in array.dd).
Run `grep -R "char\[\]"` to see all unchanged instances.

articles/exception-safe.dd
```diff
@@ -342,7 +342,7 @@ class Mailer
     void Send(Message msg)
     {
         {
-            char[] origTitle = msg.Title();
+            string origTitle = msg.Title();
```
wilzbach: Why not `auto` or `const` here.

articles/lazy-evaluation.dd
```diff
@@ -34,7 +34,7 @@ value:
 )
 
 ---
-void log(char[] message)
+void log(string message)
```
wilzbach: `const(char)[]` would be better for this entire file.

spec/ddoc.dd
```diff
@@ -335,7 +335,7 @@ $(DD Explains the return value of the function.
  * Returns: The contents of the file.
  */
 
-void[] readFile(char[] filename) { ... }
+void[] readFile(string filename) { ... }
```
wilzbach: `const(char)[]` is what the function in Phobos accepts.

spec/statement.dd
```diff
@@ -1889,11 +1889,11 @@ void main()
             writeln(x + i, ++j);
         ");    // ok
 
-    const char[] s = "int y;";
+    string s = "int y;";
```
wilzbach: `const`

spec/statement.dd
```diff
@@ -1889,11 +1889,11 @@ void main()
...
     mixin(s);  // ok
     y = 4;     // ok, mixin declared y
 
-    char[] t = "y = 3;";
+    string t = "y = 3;";
```
wilzbach: `const`

wc.dd
```diff
@@ -31,7 +31,7 @@ int main(string[] args)
         foreach(line; file.byLine(KeepTerminator.yes))
         {
             lCharCount += line.length;
-            foreach(char[] word; splitter(line))
+            foreach(string word; splitter(line))
```
wilzbach: `byLine` uses a mutable buffer, so this shouldn't compile better just remove the type completely.

PR #2527, opened 2018-12-21
add d-man to 404 page

Flying-Toast:


wilzbach:
That's how it looks:

image

Maybe a good idea to center it and add a `max-width` for mobile devices?

Flying-Toast:
I centered and added `max-width`. Here is what it looks like now:


And mobile:

404.dd
```diff
@@ -1,12 +1,20 @@
 Ddoc
 
-$(D_S 404 Not Found,
+<div class="center-children">
```
ZombineDev: Replace HTML tags with macros like `$(DIV ..`.

404.dd
```diff
@@ -1,12 +1,20 @@
...
 
-$(P
-Page not found!
+$(D_S $(TITLE),
+
+<img src="$(STATIC images/dman-scared.jpg)">
```
ZombineDev: here too.

404.dd
wilzbach: Ah this is page won't be part of the spec which is rendered to LaTeX, so it's fine to use plain old HTML here.

PR #2534, opened 2018-12-29
add missing information to return-ref documentation

WalterBright:


spec/function.dd
```diff
@@ -1362,8 +1362,8 @@ ref int gun(return ref int x) {
 }
 ---
 
-        $(P Ref methods marked with the `return` attribute ensure the returned
-        reference will not outlive the respective aggregate instance.
+        $(P Struct non-static methods marked with the `return` attribute ensure the returned
+        reference will not outlive the struct instance.
```
thewilsonator: Classes?

jacob-carlborg: Interfaces?

WalterBright: Classes are part of the `return scope` thing.

thewilsonator: If its going to confuse us, chances are it'll confuse someone else.

PR #2536, opened 2018-12-30
add Return Scope Parameters section

WalterBright:


spec/function.dd
```diff
@@ -1455,6 +1455,74 @@ inout(int)* neptune(inout ref int i)
...
+{
+     gp = p; // error, p escapes to global gp
+     gp = q; // error, q escapes to global gp
+     gp = r; // ok
+
+     thorin(p); // ok, p does not escape thorin()
+     thorin(q); // ok
+     thorin(r); // ok
+
+     global(p); // error, gloin() escapes p
```
JinShil: `gloin` instead of `global`?

spec/function.dd
```diff
@@ -1455,6 +1455,74 @@ inout(int)* neptune(inout ref int i)
...
```
JinShil: `gloin` instead of `global`?

thewilsonator:
I'd rather the LotR references were dropped and just call then `f`,`g` and `h`.

WalterBright:
I got bored with the planets :-)

Anyhow, others have complained about the boringness of f(), func(), foo(), etc., and they do get tiresome.

thewilsonator:
https://github.com/Droogans/unmaintainable-code/blob/master/README.md#bedazzling-names

wilzbach:
@thewilsonator: this doesn't apply here as it's the documentation with no real code where fun names make the example more interesting.

WalterBright:
While the advice is good, the trouble is, those functions have no purpose and do nothing. But they *are* all from the same group, and are all in one group in one section. I find one letter function names a little off-putting.

jacob-carlborg:
> No idea where the names come, but they sound fun.

They are names of the dwarfs in the The Hobbit.

PR #2539, opened 2019-01-02
Add Discord join link to the community section

WebFreak001:
Adds the [Discord](https://discordapp.com/) chat link for "D Language Code Club" (previously known as "D Code Club" and "Wilds Code Club") to the Community section. (in the following referred to as "Server") Discord is a lot like Slack.

The server currently has 378 members, where we usually have around 100 online users. (comparable to IRC) It is the official chat of [dplug](https://code.dlang.org/packages/dplug), [code-d](https://github.com/Pure-D/code-d) and [PowerNex](https://github.com/PowerNex/PowerNex). We have topic channels around general Programming, web development and OS development, a user-submitted D resources collection, a user-submitted D projects collection and a News channel about D news from the Announce forum and announcing new dub projects. There are also some off-topic channels. We are currently also working on a imaging library inside a channel on the server there too, which is supposed to eventually also become a GUI library. With that workgroup we are trying to improve certain areas in D.

We help a lot of users with their programming questions on the server. Currently the server is administered by @Vild, @M4GNV5 and me. @0xEAB, @rikkimax and @p0nce are additional moderators.

I think adding the Discord server to the list is a valuable addition because so far it seems that a lot of "younger" people visit our Discord instead of the IRC, which is considered a bit dated.

### DIFF
--- a/dlang.org.ddoc
+++ b/dlang.org.ddoc
@@ -260,6 +260,7 @@ $(SUBMENU_MANUAL
     $(SUBMENU_LINK $(ROOT_DIR)calendar.html, Calendar)
     $(SUBMENU_LINK_DIVIDER https://forum.dlang.org, Forums)
     $(SUBMENU_LINK irc://irc.freenode.net/d, IRC)
+    $(SUBMENU_LINK https://discord.gg/fva44g2, Discord)
     $(SUBMENU_LINK https://wiki.dlang.org, Wiki)
     $(SUBMENU_LINK_DIVIDER https://github.com/dlang, GitHub)
     $(SUBMENU_LINK $(ROOT_DIR)bugstats.html, Issues)


0xEAB:
I've collected a few quotes from the `#general` channel on the Discord server:

>> So that's kinda D Discord Server? Was wondering why anything like that isn't promoted on dlang.org. Way better than IRC.
>
> Yes. the majority here, to my knowledge, are D enthusiasts.
> It's not heavily promoted as it is not a official discord and it started out as a discord for my livestreaming.
> I wouldn't really call it better than IRC. Discord and IRC are both chatservices, but they are used in different ways. I would say Discord is more for communities and IRC is more for projects or small groups where no history is needed.

> you can also use this discord as faster stackoverflow :p

> had no idea D had some sorta Discord

> Cool to see there's actually a D discord server

> found the invite link on code-d vscode plugin github page

> Twitter brought me here :smile:
> I thought there was no server for dlang

> the link on the extension page on vscode
> jk, didn't know there was a discord server before

wilzbach:
@thewilsonator please please don't merge changes to the front page without @CyberShadow's, @andralex's or my approval and in any case not after seven hours.
Also, (1) a change to the menu would need to be propagated to a lot of things (ddox, dub, forum, tour, visualD,...) and (2) there's the UX rule of not using more than eight items and we already overstretched that.
Then there's (3) the point that it's not official and we just promoted it to be official.

I don't object to this in general, but I think it deserves more than seven hours of review time.

PR #2540, opened 2019-01-02
Add documentation for Objective-C classes

jacob-carlborg:
This is the spec update for https://github.com/dlang/dmd/pull/9179.

### DIFF
--- a/spec/class.dd
+++ b/spec/class.dd
@@ -52,6 +52,7 @@ $(P Classes consist of:)
 $(UL
         $(LI a super class)
         $(LI interfaces)
+        $(LI a nested anonymous metaclass for classes declared with `extern (Objective-C)`)
         $(LI dynamic fields)
         $(LI static fields)
         $(LI types)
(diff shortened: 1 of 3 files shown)


thewilsonator:
You need to escape all your `,` in `$(P some, text)` with `$(P $(ARGS some, text))` see https://github.com/dlang/dlang.org/pull/2453#discussion_r245378324

adamdruppe:
ddoc is so unbelievably bad. you guys really should ditch it so writers can spend less time on syntax and more time on content. just fyi.

jacob-carlborg:
> You need to escape all your , in $(P some, text) with $(P $(ARGS some, text)) see #2453 (comment)

You're kidding? The existing text doesn't do that.

thewilsonator:
Oh I was dealing with lists, hmm. IDK what the problem is then.

thewilsonator:
>  you guys really should ditch it

Would love to.

jacob-carlborg:
@thewilsonator it's green now. I had an extra comma where there shouldn't be one.

PR #2541, opened 2019-01-03
Revert "Add Discord join link to the community section"

wilzbach:
Reverts dlang/dlang.org#2539 as it was merged a bit too hastily and without any proper discussion (and as dlang.org her's auto-deployed this change is now live).

### DIFF
--- a/dlang.org.ddoc
+++ b/dlang.org.ddoc
@@ -260,7 +260,6 @@ $(SUBMENU_MANUAL
     $(SUBMENU_LINK $(ROOT_DIR)calendar.html, Calendar)
     $(SUBMENU_LINK_DIVIDER https://forum.dlang.org, Forums)
     $(SUBMENU_LINK irc://irc.freenode.net/d, IRC)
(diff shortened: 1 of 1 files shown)


wilzbach:
CC @thewilsonator

WebFreak001:
> The last time I was there, the vibe I got is that it was the personal server of some people who also happened to be D programmers,

it used to be that personal server, but as more users have joined it became pretty much just a D server with very little personal stuff (we still have a streaming notifications channel where anyone can share their livestreams and also automated livestream messages for Wild still)

> and the same server included channels for non-D stuff like Anime and some semi-NSFW pictures, though I might be mis-remembering

one could argue those pictures were more SFW than semi-NSFW, but I have put the channel on an invite-only basis anyway as I realized normal people probably wouldn't want to receive messages for that. :p

I agree that the Community section on the Wiki is pretty good, but it has been there for a while and basically all people who joined (and responded to our question) said they didn't know it was there or didn't even look on there. Eventually they joined when they discovered it for example in forum posts, somewhere in a tweet retweeted by the D foundation or in the READMEs of the projects which are on the Discord.

PR #2543, opened 2019-01-04
document return-ref-scope parameters

WalterBright:


ZombineDev:
There's a pretty big overlap between the new section "Return Ref Scope Parameters" and the previous "Ref Return Scope Parameters", especially given that the previous one specifies that the lexical order doesn't have any semantic significance.
I suggest moving the example in the previous section in order to reduce confusion.

RazvanN7:
ping @WalterBright

dkorpel:
This is superseded by https://issues.dlang.org/show_bug.cgi?id=22790

PR #2546, opened 2019-01-05
specify Unsigned, Integral, Floating, Arithmetic and Scalar Types

WalterBright:
and some links to them.


spec/type.dd
```diff
@@ -73,6 +75,71 @@ $(H2 $(LEGACY_LNAME2 Base Types, base-types, Base Types))
...
+        $(LI char) $(LI wchar) $(LI dchar)
+    )
+
+    $(P If the $(GLINK2 enum, EnumBaseType) of an enum is an $(I Unsigned Type), the enum
+    is an $(I Unsigned Type).)
+
+    $(P If the element type of a vector is an $(I Unsigned Type), the vector
+    is an $(I Unsigned Type).)
+
+$(H2 $(LNAME2 Integral Type, Integral Type))
```
jacob-carlborg: Should this say "Integral Type, Signed Type"?

jacob-carlborg: Perhaps not.

thewilsonator:
```
$(P If the arguments are all either types that are $(DDSUBLINK type, Arithmetic Type, $(I Arithmetic Type)),
```

You need to escape your commas , see https://github.com/dlang/dlang.org/pull/2453#discussion_r245378324

spec/type.dd
WalterBright: Integral includes signed and unsigned types.

jacob-carlborg: I got confused when one section said "Unsigned Type" and one said "Integral Type".

PR #2547, opened 2019-01-06
spec/vector.dd: make more formal and complete

WalterBright:
Chatty exposition doesn't really belong in a formal spec.

Fixed error in grammar for `Vector`.

### DIFF
--- a/spec/declaration.dd
+++ b/spec/declaration.dd
@@ -106,8 +106,10 @@ $(GNAME BasicType):
     $(GLINK2 grammar, Traits)
 
 $(GNAME Vector):
-    $(D __vector)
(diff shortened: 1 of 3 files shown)


WalterBright:
That usage is special to the `is expression` and should appear there in the grammar. It cannot be where it is now because it would allow these declarations:
```
___vector v;
```
which are invalid and are rejected by the parser.

WalterBright:
Added it to the https://dlang.org/spec/expression.html#is_expression grammar.

WalterBright:
By the way, when I redo these spec descriptions I read the code the implements it. It's the only way to pick up missing nuggets like the vector element type qualifiers are ignored.

PR #2556, opened 2019-01-16
fix issue 19592 - Rule for IdentifierList is wrong

Basile-z:
The problem is revealed for example when you try to follow the rules for a BaseClass made of an element of a TList.


BBasile:
There was another rule affected, the "old" alias declaration.

BBasile:
[Direct link](http://dtest.dlang.io/artifact/website-efc66aed933d4cc7be3e567c05cd8a50cd715e6b-77f27fdc2808fda6daa6c1d6198e1703/web/spec/declaration.html#IdentifierList) to the new version of the rule.

PR #2557, opened 2019-01-20
Bump the stable version of Dub and DMD used to build the DPL docs

jacob-carlborg:
It's required to have a later version of Dub that supports the `DUB_EXE` environment variable when building the DMD Dub package.


jacob-carlborg:
@thewilsonator everything's green again.

PR #2558, opened 2019-01-21
TLS access is slower than classic global vars

JohanEngelen:
Fix the text discussing performance considerations of TLS globals (the difference is not 3 instructions, and TLS is not faster with or without PIC enabled).

### DIFF
--- a/articles/migrate-to-shared.dd
+++ b/articles/migrate-to-shared.dd
@@ -17,8 +17,8 @@ $(HEADERNAV_TOC)
 $(H2 $(LNAME2 performance, Performance of TLS variables))
 
         $(P Reading or writing TLS variables is slower than for classic
-        global variables. The exact difference depends, a.o., on compiler
-        code generation settings and the target architecture.
+        global variables. The exact difference depends, among others,
+        on compiler code generation settings and the target architecture.
         For PIC (position independent code), which is the default setting
         on Linux, TLS access is slower because a call to $(D __tls_get_addr)
         is required. However for non-PIC, the performance difference is
--- a/articles/migrate-to-shared.dd
+++ b/articles/migrate-to-shared.dd
@@ -16,14 +16,15 @@ $(HEADERNAV_TOC)
 
 $(H2 $(LNAME2 performance, Performance of TLS variables))
 
-        $(P Yes, reading or writing TLS variables will be slower
-        than for classic global variables. It'll be about 3
-        instructions rather than one.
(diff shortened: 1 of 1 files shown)


articles/migrate-to-shared.dd
```diff
@@ -16,14 +16,15 @@ $(HEADERNAV_TOC)
...
-        $(P Yes, reading or writing TLS variables will be slower
-        than for classic global variables. It'll be about 3
-        instructions rather than one.
-        But on Linux, at least, TLS will be slightly faster
-        than accessing classic globals using PIC (position
-        independent code) compiler code generation settings.
-        So it's hard to see that this would be an unacceptable
-        problem. But let's presume it is. What can we do about it?
+        $(P Reading or writing TLS variables is slower than for classic
+        global variables. The exact difference depends, a.o., on compiler
```
ZombineDev: What does `a.o.` mean? "Among other"? Can you expand the abbreviation not to confuse readers?

JohanEngelen: yes

jacob-carlborg:
Please squash commits when done.

JohanEngelen:
> Please squash commits when done.

Can you do that using github's squash+merge button please? (don't know another way with web interface) Thanks!

jacob-carlborg:
> Can you do that using github's squash+merge button please? (don't know another way with web interface) Thanks!

I don't think that will create a merge commit. Why not using git?

JohanEngelen:
> I don't think that will create a merge commit.

True. It will create a commit on top of the target branch, with a title including the number of the PR. I like it and we use it a lot with LDC: it prevents a merge bubble and it's also much easier for improvements like this one created using the web interface.
Anyway, if you guys don't like that, I'll have to get a checkout of the repo, do the squash, force push, ... don't feel like doing that now. Please be patient ;)

wilzbach:
> I don't think that will create a merge commit. 

We had the argument plenty of times and no tool requires merge commits, so it's perfectly safe and fine to use GitHub's squash and merge or even dlang-bot's squash and merge...

ZombineDev:
@jacob-carlborg no need to waste contributors' time with something reviewers can easily do themselves. The goal of every reviewer should be to ensure that contributors' time is used effectively as much ensuring the code quality.

ZombineDev:
@JohanEngelen no need, that's for the contribution ;)

JohanEngelen:
thanks :)

PR #2559, opened 2019-01-23
Improve TLS-related perf discussion

JohanEngelen:
The previous change was not fair to Windows code. With these changes, the discussion is longer, but fairer for all platforms.
I also added a mitigation idea of grouping related variables into a struct such that TLS overhead is only paid once.

@jacob-carlborg @ZombineDev

### DIFF
--- a/articles/migrate-to-shared.dd
+++ b/articles/migrate-to-shared.dd
@@ -19,11 +19,10 @@ $(H2 $(LNAME2 performance, Performance of TLS variables))
         $(P Reading or writing TLS variables is potentially slower than for
         classic global variables. The exact difference depends, among others,
         on compiler code generation settings and the target architecture.
-        On Linux with PIC (position independent code), which is the default
-        setting, TLS access is slower because a call to $(D __tls_get_addr)
-        is required (similar on macOS). However for non-PIC, the performance
-        difference is smaller or even negligible.
-        On Windows, the performance difference is negligible.
+        The performance difference ranges from negligible (Windows) to a call
+        to a special function (PIC on Linux, macOS, *BSD. Search for example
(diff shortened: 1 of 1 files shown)


jacob-carlborg:
How come three platforms are mentioned? Can the text be more generic to cover all platforms or mention all platforms?

JohanEngelen:
It turns out it's different on all platforms, I guess. So I picked Windows, Linux, macOs, as 3 common ones.
I guess the generic statement would be something like: "depending on the target OS, the overhead ranges from zero to a call to a special function".
What do you mean with "all" platforms? (there are many ;-)

jacob-carlborg:
> What do you mean with "all" platforms? (there are many ;-)

Yes. But the spec is usually geared towards DMD, which means one additional platform. Anyway it's not that important. But if you want to cover more platforms it's better to be less specific.

JohanEngelen:
Talking about performance details in relation to DMD doesn't make sense. 
The text was wrong before, then I tried to fix that in my previous PR, but I introduced another error. Here I'm trying to fix that, by giving some details about example platforms. But we can remove all of it and just say that the performance difference ranges from zero to a call to a special function.

thewilsonator:
@JohanEngelen If you can find some general consensus for the BSD's that would be great, if not ping me.

JohanEngelen:
Changed some things. Better now?

thewilsonator:
Yep. adding 72h ->h merge.

thewilsonator:
Hmm, apparently DlangBot doesn't check 72h no objection -> merge here.

wilzbach:
> Hmm, apparently DlangBot doesn't check 72h no objection -> merge here.

It never does. That feature isn't implemented (though often manual judgment is required anyhow).

PR #2561, opened 2019-01-28
spec: close a loophole in definition of `@trusted`

aG0aep6G:
This would solidify my preferred interpretation of `@trusted` in the spec. Others have disagreed with me on what `@trusted` should mean exactly, so this might be controversial.

### DIFF
--- a/spec/const3.dd
+++ b/spec/const3.dd
@@ -262,14 +262,18 @@ char[] p = "world"; // error, cannot implicitly convert immutable
 
         $(P
         The second way is to cast data to immutable.
-        When doing so, it is up to the programmer to ensure that no
-        other mutable references to the same data exist.
+        When doing so, it is up to the programmer to ensure that any mutable
+        references to the same data are not used to modify the data after the
+        cast.
         )
 
 ---
-char[] s = ...;
-immutable(char)[] p = cast(immutable)s;     // undefined behavior
-immutable(char)[] p = cast(immutable)s.dup; // ok, unique reference
+char[] s = ['a'];
+s[0] = 'b'; // ok
+immutable(char)[] p = cast(immutable)s; // ok, if data is not mutated
+                                        // through s anymore
+s[0] = 'c'; // undefined behavior
+immutable(char)[] q = cast(immutable)s.dup; // always ok, unique reference
 ---
 
         $(P
--- a/spec/const3.dd
+++ b/spec/const3.dd
@@ -274,6 +274,13 @@ immutable(char)[] p = cast(immutable)s; // ok, if data is not mutated
                                         // through s anymore
 s[0] = 'c'; // undefined behavior
 immutable(char)[] q = cast(immutable)s.dup; // always ok, unique reference
+
+char[][] s2 = [['a', 'b'], ['c', 'd']];
+immutable(char[][]) p2 = cast(immutable)s2.dup; // dangerous, only the first
+                                                // level of elements is unique
+s2[0] = ['x', 'y']; // ok, doesn't affect p2
+s2[1][0] = 'z'; // undefined behavior
+immutable(char[][]) q2 = [s2[0].dup, s2[1].dup]; // always ok, unique references
 ---
 
         $(P
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -2756,6 +2756,40 @@ $(H3 $(LNAME2 trusted-functions, Trusted Functions))
         ensure that these guarantees are upheld.
         )
 
+        $(P Example:)
+
+        ---
+        immutable(int)* f(int* p) @trusted
+        {
+            version (none) p[2] = 13;
+            // Invalid. p[2] is out of bounds. This line would exhibit undefined
+            // behavior.
+
+            version (none) p[1] = 13;
+            // Invalid. In this program, p[1] happens to be in-bounds, so the
+            // line would not exhibit undefined behavior, but a trusted function
+            // is not allowed to rely on this.
+
+            version (none) return cast(immutable) p;
+            // Invalid. @safe code still has mutable access and could trigger
+            // undefined behavior by overwriting the value later on.
+
+            int* p2 = new int;
+            *p2 = 42;
+            return cast(immutable) p2;
+            // Valid. No @safe code can have knowledge of p2.
+        }
+
+        void main() @safe
+        {
+            int[2] a = [10, 20];
+            int* mp = &a[0];
+            immutable(int)* ip = f(mp);
+            assert(a[1] == 20); // Guaranteed. f cannot access a[1].
+            assert(ip !is mp); // Guaranteed. f cannot introduce unsafe aliasing.
+        }
+        ---
+
         $(P Trusted functions may call safe, trusted, or system functions.
         )
 
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -2749,8 +2749,11 @@ $(H3 $(LNAME2 trusted-functions, Trusted Functions))
 
         $(P Trusted functions are marked with the $(CODE @trusted) attribute.)
 
-        $(P Trusted functions are guaranteed by the programmer to not exhibit
-        any undefined behavior if called by a safe function.
+        $(P Trusted functions are guaranteed to not exhibit any undefined
+        behavior if called by a safe function. Furthermore, calls to trusted
+        functions cannot lead to undefined behavior in @safe code that is
+        executed afterwards. It is the responsibility of the programmer to
+        ensure that these guarantees are upheld.
         )
 
         $(P Trusted functions may call safe, trusted, or system functions.
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -2751,7 +2751,7 @@ $(H3 $(LNAME2 trusted-functions, Trusted Functions))
 
         $(P Trusted functions are guaranteed to not exhibit any undefined
         behavior if called by a safe function. Furthermore, calls to trusted
-        functions cannot lead to undefined behavior in @safe code that is
+        functions cannot lead to undefined behavior in `@safe` code that is
         executed afterwards. It is the responsibility of the programmer to
         ensure that these guarantees are upheld.
         )
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -2795,7 +2795,7 @@ $(H3 $(LNAME2 trusted-functions, Trusted Functions))
 
         $(P Trusted functions are covariant with safe or system functions.)
 
-        $(BEST_PRACTICE Generally, trusted functions should be kept small so
+        $(BEST_PRACTICE Trusted functions should be kept small so
         that they are easier to manually verify.
         )
 
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -2751,14 +2751,16 @@ $(H3 $(LNAME2 trusted-functions, Trusted Functions))
 
         $(P Trusted functions are guaranteed by the programmer to not exhibit
         any undefined behavior if called by a safe function.
-        Generally, trusted functions should be kept small so that they are
-        easier to manually verify.
         )
 
         $(P Trusted functions may call safe, trusted, or system functions.
         )
 
         $(P Trusted functions are covariant with safe or system functions.)
+
+        $(BEST_PRACTICE Generally, trusted functions should be kept small so
+        that they are easier to manually verify.
+        )
 
 $(H3 $(LNAME2 system-functions, System Functions))
 
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -2777,7 +2777,7 @@ $(H3 $(LNAME2 trusted-functions, Trusted Functions))
             int* p2 = new int;
             *p2 = 42;
             return cast(immutable) p2;
-            // Valid. No @safe code can have knowledge of p2.
+            // Valid. After f returns, no mutable aliases of p2 can exist.
         }
 
         void main() @safe


thewilsonator:
cc @andralex @WalterBright

spec/function.dd
```diff
@@ -2749,17 +2749,56 @@ $(H3 $(LNAME2 trusted-functions, Trusted Functions))
...
+            // is not allowed to rely on this.
+
+            version (none) return cast(immutable) p;
+            // Invalid. @safe code still has mutable access and could trigger
+            // undefined behavior by overwriting the value later on.
+
+            int* p2 = new int;
+            *p2 = 42;
+            return cast(immutable) p2;
+            // Valid. No @safe code can have knowledge of p2.
```
ZombineDev: I'd say:
```
Valid. After `f` returns, no mutable references to the value pointed by `p2` can exist.
```

Or:
```
Valid. After `f` returns, the value pointed by `p2` can't be mutated, hence for the callers of `f` the result is immutable, for all intents and purposes.
```

aG0aep6G:
> Just because it has not been spelt out explicitly, this would preclude type-based alias analysis based on mutability (that is, infering that a mutable and an immutable pointer don't alias). Whether that's a large price to pay or not is another question, though.

I think that's already not possible. As discussed in issue [issue 19621](https://issues.dlang.org/show_bug.cgi?id=19621), the spec allows to remove `immutable` with a cast. So this is valid, according to the spec:

```d
void main()
{
    immutable int x = 42;
    f(&x, cast(int*) &x);
}
void f(immutable int* i, int* m)
{
    /* Can't mutate through `m` here, but also can't
    assume that `m` is different from `i`. */
}
```

With the additional restriction to `@trusted`, it's at least guaranteed that `@safe` and `@trusted` functions don't introduce such aliasing.

> "Exist" is also problematic, though – maybe "use" would be a bit better defined?

I don't follow. I am replacing one occurrence of "exist" with "used to modify" here.

aG0aep6G:
> I'd argue that continuing to use `immutable` references after converting one to mutable is an error, just like it is the other way round.

So if I understand you correctly, your alternative solution to issue 19621 would have this effect:

```d
int* m = ...;
foo(cast(immutable int*) m); /* Allowed. */
bar(m); /* Not allowed; m has been invalidated. */

immutable(int)* i = ...;
baz(cast(int*) i); /* Allowed (as long as baz doesn't actually mutate). */
qux(i); /* Not allowed; i has been invalidated. */
```

(We'd have to define what it means to "use a reference". I'm passing them in calls here, because that probably qualifies.)

Doesn't look too bad. We can assume no aliasing, yet still have functions like `assumeUnique`.

Problems:

1) It would be a very subtle breaking change. The spec currently allows the `qux(i)` call.
2) The rule itself is also subtle, easy to miss. Might make for hard-to-find bugs.

aG0aep6G:
This has been idle for almost two weeks. Can someone take a look?

ZombineDev:
Ping @andralex @WalterBright this is an important specification clarification / change. Can you take a look?

aG0aep6G:
> This has been idle for almost two weeks. Can someone take a look?

Four weeks.

thewilsonator:
I've emailed W&A

spec/function.dd
```diff
@@ -2749,17 +2749,56 @@ $(H3 $(LNAME2 trusted-functions, Trusted Functions))
 
         $(P Trusted functions are marked with the $(CODE @trusted) attribute.)
 
-        $(P Trusted functions are guaranteed by the programmer to not exhibit
-        any undefined behavior if called by a safe function.
-        Generally, trusted functions should be kept small so that they are
-        easier to manually verify.
+        $(P Trusted functions are guaranteed to not exhibit any undefined
```
andralex: The old text "by the programmer" was adequate, no? Put another way: "Trusted functions are assumed without checking that ..."

spec/const3.dd
```diff
@@ -262,14 +262,18 @@ char[] p = "world"; // error, cannot implicitly convert immutable
 
         $(P
         The second way is to cast data to immutable.
-        When doing so, it is up to the programmer to ensure that no
-        other mutable references to the same data exist.
+        When doing so, it is up to the programmer to ensure that any mutable
+        references to the same data are not used to modify the data after the
+        cast.
```
andralex: This is meant as a relaxation, right? Data cast to immutable can still be read legally through mutable references, and as long as that's done things should still be fine. I think this works (famous last words...). cc @tgehr @WalterBright

spec/const3.dd
```diff
@@ -262,14 +262,18 @@ char[] p = "world"; // error, cannot implicitly convert immutable
...
 ---
-char[] s = ...;
-immutable(char)[] p = cast(immutable)s;     // undefined behavior
-immutable(char)[] p = cast(immutable)s.dup; // ok, unique reference
+char[] s = ['a'];
+s[0] = 'b'; // ok
+immutable(char)[] p = cast(immutable)s; // ok, if data is not mutated
+                                        // through s anymore
+s[0] = 'c'; // undefined behavior
+immutable(char)[] q = cast(immutable)s.dup; // always ok, unique reference
```
andralex: This example is good, but a more illustrative one would also include two levels of indirection. Then a simple `.dup` would not suffice because each member must be `dup`e as well. Would you want to insert a second example?

spec/function.dd
```diff
@@ -2749,17 +2749,56 @@ $(H3 $(LNAME2 trusted-functions, Trusted Functions))
...
+        behavior if called by a safe function. Furthermore, calls to trusted
+        functions cannot lead to undefined behavior in @safe code that is
```
andralex: code format for `@safe`

spec/function.dd
```diff
@@ -2749,17 +2749,56 @@ $(H3 $(LNAME2 trusted-functions, Trusted Functions))
...
+            // Invalid. @safe code still has mutable access and could trigger
+            // undefined behavior by overwriting the value later on.
+
+            int* p2 = new int;
+            *p2 = 42;
+            return cast(immutable) p2;
+            // Valid. After f returns, no mutable aliases of p2 can exist.
+        }
+
+        void main() @safe
```
andralex: I think `main` should go at the top, followed by `f`. This is because the comments explaining what works and what doesn't in `f` require the reader to have read the call site first.

spec/function.dd
```diff
@@ -2749,17 +2749,56 @@ $(H3 $(LNAME2 trusted-functions, Trusted Functions))
...
+        immutable(int)* f(int* p) @trusted
+        {
+            version (none) p[2] = 13;
+            // Invalid. p[2] is out of bounds. This line would exhibit undefined
+            // behavior.
+
+            version (none) p[1] = 13;
+            // Invalid. In this program, p[1] happens to be in-bounds, so the
+            // line would not exhibit undefined behavior, but a trusted function
+            // is not allowed to rely on this.
```
andralex: I'm not sure I understand the utility of this example. What if the user passes the length of `p` as a separate integral parameter? That would be good, right?

So for this example it would be "valid if and only if the actual extent of `p` is two or more elements".

If I understand correctly, do you want to limit `@trusted` functions taking pointers to accessing at most ONE element pointed to by those pointers? That would be a large restriction.

spec/function.dd
```diff
@@ -2749,17 +2749,56 @@ $(H3 $(LNAME2 trusted-functions, Trusted Functions))
...
+            return cast(immutable) p2;
+            // Valid. After f returns, no mutable aliases of p2 can exist.
+        }
+
+        void main() @safe
+        {
+            int[2] a = [10, 20];
+            int* mp = &a[0];
+            immutable(int)* ip = f(mp);
+            assert(a[1] == 20); // Guaranteed. f cannot access a[1].
```
andralex: I don't think we can ask that much of trusted functions. What do we do with the many functions that manipulate C-style strings (or any similar sentinel-terminated structures)?

spec/const3.dd
tgehr: Yes, but maybe it should be clarified what "ensure" means. In particular, those mutable references should still not even be accessible from @safe code. (Even if the current version of the @safe code does not use them to write.)

spec/function.dd
aG0aep6G: done

spec/const3.dd
aG0aep6G: > Would you want to insert a second example?

added one

spec/const3.dd
aG0aep6G: > maybe it should be clarified what "ensure" means. In particular, those mutable references should still not even be accessible from @safe code. (Even if the current version of the @safe code does not use them to write.)

I tried covering that in the definition of `@trusted` ("cannot lead to undefined behavior in @safe code that is executed afterwards"). I'm not sure if I managed to express what's in your parentheses, though. I've got it in an example, but in the prose it might not be clear enough.

spec/function.dd
aG0aep6G: The old text was fine. I just moved the "by the programmer" part to a new sentence, because I'm adding another requirement, and repeating "by the programmer" didn't look nice to me.

If you think "assumed without checking" is better than "guaranteed", I can change that. I don't see what difference it makes, though.

spec/function.dd
aG0aep6G: > I'm not sure I understand the utility of this example. What if the user passes the length of `p` as a separate integral parameter? That would be good, right?

No. That's not allowed either. Taking a D array is where it's at. Then it's guaranteed that the length is valid.

> So for this example it would be "valid if and only if the actual extent of `p` is two or more elements".

Exactly not that. In your version, it's possible to call an `@trusted` function in an unsafe way  from `@safe` code. In my opinion, that's what we want to prevent.

> If I understand correctly, do you want to limit `@trusted` functions taking pointers to accessing at most ONE element pointed to by those pointers?

Yes, and I don't think it's a new limitation. It's the reason why many C functions, e.g. `memcpy`, aren't marked as `@trusted` in core.stdc. But we can write an `@trusted` wrapper that takes D arrays and calls `memcpy` with a known-good length.

spec/function.dd
aG0aep6G: > I think `main` should go at the top

I put `f` at the top because its validity is independent of whatever `main` does. But either way is fine with me. If you want `main` at the top, just say the word and I'll move it.

spec/function.dd
aG0aep6G: >  What do we do with the many functions that manipulate C-style strings (or any similar sentinel-terminated structures)?

We don't mark them as `@trusted`. Note that we're already not trusting those functions in core.stdc.

andralex:
I think I see the merit of this idea. Very interesting. The one difficult issue is that until now `@trusted` was for all purposes "like `@system` but the compiler trusts you". With this, we'll have additional checks on `@trusted` function that make that view invalid.
I'll bring it up to @WalterBright.

dnadlinger:
@andralex: Regarding

> If this attempts to restrict trusted functions to only use one pointed-to element for pointers, that seems excessive.

and

> If I understand correctly, do you want to limit `@trusted` functions taking pointers to accessing at most ONE element pointed to by those pointers? That would be a large restriction.

: From the way I read your and @aG0aep6G comments, there is likely a misunderstanding here. Both @aG0aep6G and me are discussing a change/clarification of the outside semantics of a `@trusted` function, whereas your comments seem to concern what the function can do internally.

Incidentally, this is one of the two issues I pointed out in my ["`@trusted` considered harmful"](https://forum.dlang.org/thread/blrglebkzhrilxkbprgh@forum.dlang.org?page=1) post from almost seven years ago. As far as the API contract is concerned, there is no such thing as a `@trusted` function. There is only (callable from) `@safe` and (requiring) `@system`. The only difference of `@trusted` over `@safe` is the implementation detail that unsafe code is internally used.

Conflating API contracts and implementation details only creates confusion such as this one. (The linguistic ambiguity in "guaranteed" not specifying who provides the guarantee also doesn't help, e.g. regarding https://github.com/dlang/dlang.org/pull/2561/files/870e57149ad320eb253aed83a87599750e54eeb1#diff-9b28a4621c56f48c86e29641ee8fa10b).

spec/const3.dd
```diff
@@ -262,14 +262,25 @@ char[] p = "world"; // error, cannot implicitly convert immutable
...
+        cast.
         )
 
 ---
-char[] s = ...;
-immutable(char)[] p = cast(immutable)s;     // undefined behavior
-immutable(char)[] p = cast(immutable)s.dup; // ok, unique reference
+char[] s = ['a'];
+s[0] = 'b'; // ok
+immutable(char)[] p = cast(immutable)s; // ok, if data is not mutated
```
dnadlinger: I'd still prefer to make this `accessed through s` – is there any tangible benefit to relaxing this restriction?

dnadlinger: The reason for my preference (although it has admittedly been a while since I thought about this) is conceptual simplicity: If casting invalidates all old references, then any object is always in one of two states that are straightforwardly defined: mutable or immutable.

Immutable objects can be assumed not to be modified, mutable objects can be assumed not to be shared (in theory, anyway). If there can be both mutable and immutable references at the same time, there is a third state where neither guarantee applies.

(Whether it was wise to make immutable implicitly sharable is a question for another day, of course.)

aG0aep6G: > I'd still prefer to make this `accessed through s` – is there any tangible benefit to relaxing this restriction?

The goal is to fix issue [19621](https://issues.dlang.org/show_bug.cgi?id=19621). That can be done by either relaxing the rules for mutable → immutable (as done here), or by tightening the rules for immutable → mutable.

If I keep the rules tight here, we have to tighten them in the other direction as well, because both directions are equally dangerous. I.e., if I write "accessed" here, then the following kind of access should also be illegal, but it's currently [allowed by the spec](https://dlang.org/spec/const3.html#removing_with_cast):

```d
int main()
{
    immutable int* i = new int(42);
    return f(i, cast(int*) i); /* immutable → mutable */
}
int f(immutable int* i, int* m)
{
    return *i + *m; /* the spec currently says (at least implies) that this is ok */
}
```

I guess we could make that illegal, but it would be a silent, subtle breaking change, as far as I see. The benefit of relaxing the rules here is to avoid that breaking change. I don't know if it outweighs the benefits of forbidding the aliasing.

aG0aep6G:
> The one difficult issue is that until now `@trusted` was for all purposes "like `@system` but the compiler trusts you". With this, we'll have additional checks on `@trusted` function that make that view invalid.

To me it always meant "like `@safe` but the compiler trusts you (with the implementation)". And I believe that makes much more sense.

If `@trusted` functions can be unsafe, then we're back to manually checking the calls to those functions in `@safe` code. But that's what the `@safe` attribute is supposed to eliminate.

aG0aep6G:
> I'll bring it up to @WalterBright.

That was another two weeks ago. Ping @WalterBright @andralex

andralex:
@WalterBright has been busy with family matters, will reach out today.

spec/const3.dd
WalterBright: The point of @safe code is to offer guarantees, and allowing @safe code to continue to have mutable references to immutable data means that the compiler has no way to guarantee the @safe code isn't attempting mutation. I don't see a purpose to this relaxation, either, i.e. what important use case would it enable?

spec/const3.dd
WalterBright: I don't understand your f() example. What part of the code is @trusted?

spec/function.dd
WalterBright: "guaranteed" by itself implies guaranteed by the compiler. Guaranteed by something else must be explicitly stated.

spec/function.dd
```diff
@@ -2749,17 +2749,56 @@ $(H3 $(LNAME2 trusted-functions, Trusted Functions))
...
 
-        $(P Trusted functions are guaranteed by the programmer to not exhibit
-        any undefined behavior if called by a safe function.
-        Generally, trusted functions should be kept small so that they are
-        easier to manually verify.
+        $(P Trusted functions are guaranteed to not exhibit any undefined
+        behavior if called by a safe function. Furthermore, calls to trusted
+        functions cannot lead to undefined behavior in `@safe` code that is
+        executed afterwards. It is the responsibility of the programmer to
+        ensure that these guarantees are upheld.
```
WalterBright: ok

spec/function.dd
WalterBright: The working rule here is that an @trusted function must present a safe interface. Therefore, any pointer parameter that will perform arithmetic on the parameter and then dereference it cannot be @trusted, it must be @system. Your example and rationale with `memcpy` is correct.

spec/function.dd
WalterBright: I agree with @aG0aep6G

spec/function.dd
```diff
@@ -2749,17 +2749,56 @@ $(H3 $(LNAME2 trusted-functions, Trusted Functions))
...
+            assert(ip !is mp); // Guaranteed. f cannot introduce unsafe aliasing.
+        }
+        ---
+
         $(P Trusted functions may call safe, trusted, or system functions.
         )
 
         $(P Trusted functions are covariant with safe or system functions.)
 
+        $(BEST_PRACTICE Generally, trusted functions should be kept small so
```
WalterBright: Omit `Generally` as it's redundant.

spec/function.dd
aG0aep6G: > "guaranteed" by itself implies guaranteed by the compiler. Guaranteed by something else must be explicitly stated.

It's stated later in the paragraph. Good enough?

spec/const3.dd
aG0aep6G: > The point of @safe code is to offer guarantees, and allowing @safe code to continue to have mutable references to immutable data means that the compiler has no way to guarantee the @safe code isn't attempting mutation.

Allowing that is not my goal. I'm trying to:

1. allow creating such references in `@system`/`@trusted` code (not `@safe` code), and
2. disallow leaking dangerous stuff like that from `@trusted` code to `@safe` code.

Point 1 is a relaxation of the rules, done to fix an inconsistency in the spec ([issue 19621](https://issues.dlang.org/show_bug.cgi?id=19621)).
Point 2 is a tightening of the rules, done to fix a safety hole. Note that the hole is not being opened by point 1; it's there regardless.

If it makes for easier review, I could split this PR in two, as the two points are mostly independent of each other. Let me know if that would help.

> I don't see a purpose to this relaxation, either, i.e. what important use case would it enable?

The spec currently allows [casting `immutable` away](https://dlang.org/spec/const3.html#removing_with_cast). The motivating example given for that is calling a function where a parameter is de-facto `const`, but it's not annotated as such.

Similarly, you might want to call a function where a parameter is de-facto `const`, but it's marked as `immutable`. As far I can see, GtkD, a popular library, is full of such functions. Here's one example: [Window.setTitle](https://github.com/gtkd-developers/GtkD/blob/1349b9e8bd78f3bbb50b8cb1fddc48d8e4414226/generated/gtkd/gtk/Window.d#L2300-L2303)

spec/const3.dd
aG0aep6G: > I don't understand your f() example. What part of the code is @trusted?

There is no `@trusted` code. This discussion is about where exactly undefined behavior is triggered. That applies to `@system` code, too.

If a compiler looks at `f` in isolation, can it assume that the parameters `i` and `m` won't both be used to access the same data? @dnadlinger would like the answer to be "yes". My example is supposed to show that the answer must be "no" with the current spec, regardless of the changes in this PR.

spec/function.dd
aG0aep6G: I just moved this part. [You wrote "generally".](https://github.com/dlang/dlang.org/commit/beea2678611529d32bd9e93597d2d4333a879499) Anyway, I've removed it.

aG0aep6G:
@WalterBright, @andralex, you've both had a look at this now. Where do we stand?

andralex:
LGTM. @WalterBright ?

thewilsonator:
Ping @WalterBright

PR #2562, opened 2019-01-31
final switch can be used in -betterC

wilzbach:
See also: https://github.com/dlang/druntime/pull/2274


ZombineDev:
For example: https://run.dlang.io/is/0nuCGl

PR #2565, opened 2019-02-05
foreach index type must be size_t

thewilsonator:
Spec changes for  issue 16976 / https://github.com/dlang/dmd/pull/8941


JohanEngelen:
This should be clearly stated in the release notes.

PR #2568, opened 2019-02-06
Fix Issue 19648 - OSX install page has no correspondence with reality

wilzbach:
See also https://github.com/dlang/dlang.org/pull/2451 for more fixes.

### DIFF
--- a/dcompiler.dd
+++ b/dcompiler.dd
@@ -85,12 +85,12 @@ C:\&gt;unzip dmc.zip
 
     $(INSTALLATION_SCRIPT_HINT)
 
-    It will create
+    $(P It will create
     a $(D ~/dmd) directory with all the files in it.
 
     All the tools are command line tools, which means
     they are run from a console window.)
-    $(LI All tools can be run directly from the archive:
+    All tools can be run directly from the archive:
 $(CONSOLE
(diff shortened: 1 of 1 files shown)


dcompiler.dd
```diff
@@ -85,12 +85,12 @@ C:\&gt;unzip dmc.zip
...
-    It will create
+    $(P It will create
     a $(D ~/dmd) directory with all the files in it.
 
     All the tools are command line tools, which means
     they are run from a console window.)
-    $(LI All tools can be run directly from the archive:
+    All tools can be run directly from the archive:
 $(CONSOLE
 ~/dmd/linux/bin64/dmd
```
thewilsonator: This was the main reason the the bug report.

wilzbach: Look closer or at the generated output ;-)

Fixing the wrongly inserted parenthesis from https://github.com/dlang/dlang.org/pull/2491 makes this section Linux only as it was supposed to be.

Screenshot with this patch:

image

wilzbach:
Actually https://github.com/dlang/dlang.org/pull/2451 already solved this (it just got lost in the abyss), so we don't need this PR anymore.

PR #2569, opened 2019-02-06
[spec] Remove floating point NCEG operators from spec

kubo39:
According to https://dlang.org/deprecate.html, we can remove
floating point NCEG operators from specification.


andralex:
Thank you!

PR #2570, opened 2019-02-06
Update spec 19.13 virtual function this param

jblachly:
Spec section 19.13 (Virtual functions) currently reads:

7. Virtual functions all have a hidden parameter called the this reference, which refers to the class object for which the function is called.

8. Functions with Objective-C linkage has an additional hidden, unnamed, parameter which is the selector it was called with.

9. To avoid dynamic binding on member function call, insert base class name before the member function name. For example:

Revised to read:

7. **Non-static (i.e.,** virtual functions **and $(D final) member functions)** all have a hidden parameter called the $(I this) reference, which refers to the class object for which the function is called.

8. Functions with `Objective-C` linkage **have** an additional hidden,

9. To avoid dynamic binding on member function call, insert base class name before the member function name. **Note that when when invoked in this way, the member function will not be passed the hidden $(I this) parameter.** For example:


spec/function.dd
```diff
@@ -795,18 +795,22 @@ class Bar : Foo
 ------
 )
 
-        $(P Virtual functions all have a hidden parameter called the
-        $(I this) reference, which refers to the class object for which
-        the function is called.
+        $(P 
+            Non-static (i.e., virtual functions and $(D final) member
```
thewilsonator: you can use backticks for `$(D final)` now, \`final\`

spec/function.dd
```diff
@@ -795,18 +795,22 @@ class Bar : Foo
...
+            functions) all have a hidden parameter called the
+            $(I this) reference, which refers to the class object for which
```
thewilsonator: ditto for `$(I this)`

spec/function.dd
```diff
@@ -795,18 +795,22 @@ class Bar : Foo
...
-            Functions with `Objective-C` linkage has an additional hidden,
+            Functions with `Objective-C` linkage have an additional hidden,
             unnamed, parameter which is the selector it was called with.
         )
 
         $(P To avoid dynamic binding on member function call, insert
-        base class name before the member function name. For example:
+        base class name before the member function name. Note that when
+        when invoked in this way, the member function will not be passed
+        the hidden $(I this) parameter. For example:
```
thewilsonator: Are you sure this is the case?
given 

```d
class A { 
int a;
   void foo() {
       writeln(a);
   }; 
}

class B : A { override void foo() {}}
...
auto b = new B;
b.A.foo
```

A.foo definitely needs `this`, it will be adjusted yes, but its still needed.

jblachly: Interesting, as calling `A.foo()` statically does not add `this` and leads to compilation error (not sure what you mean by "it will be adjusted" ?) whereas `a.A.foo()` does add the hidden reference. Feel free to omit this hunk, or, one could clarify with example and counterexample.

thewilsonator: I think its better to omit it.

spec/function.dd
jblachly: Much nicer, thanks for letting me know. Currently, use of the macro matches the majority of `function.dd`

jblachly:
Are you able to exclude the third hunk using github's tools or do I need to resubmit the entire PR?

thewilsonator:
I can, but so can you. All you need to do is commit the inverse of the change and they will (partially) cancel. When I merge this all the needs to change is to squash it.

thewilsonator:
You have fallen foul of the infamous whitespace checker:
```
./spec/function.dd:798:        $(P 
```

jacob-carlborg:
Not sure if it's mentioned anywhere but static Objective-C methods are virtual.

jblachly:
Resurrected. Improving documentation shouldn't be so odious.

spec/function.dd
```diff
@@ -1026,18 +1026,23 @@ class Bar : Foo
...
             unnamed, parameter which is the selector it was called with.
         )
 
-        $(P To directly call a base member function, insert
-        base class name before the member function name. For example:
+        $(P To directly call a base member function (i.e., avoid dynamic
+        binding on member function call), insert base class name before the
+        member function name. Note that when when invoked in this way, the
+        member function will not be passed the hidden $(I this) parameter.
+        For example:
```
thewilsonator, line 1045: I think you're getting slightly confused here as to what the `statically` is referring to in the example below. `static` here means calling the function directly i.e. not calling through the function pointer int the virtual function table. This should not be confused with static member functions, which don't take a `this` reference.

think about what would happen if `B` were instead 
```
class B
{
    int a;
    int foo() { return a; }
}
```
and you tried to call `B.foo()` without the `this` parameter.

thewilsonator (changes requested):

spec/function.dd
```diff
@@ -1026,18 +1026,21 @@ class Bar : Foo
...
         $(P
             Functions with `Objective-C` linkage have an additional hidden,
             unnamed, parameter which is the selector it was called with.
         )
 
-        $(P To directly call a base member function, insert
-        base class name before the member function name. For example:
+        $(P To directly call a base member function (i.e., avoid dynamic
+        binding on member function call), insert the base class name before
+        the member function name. For example:
```
thewilsonator, line 1043: AFAIU, the principle reason for this is to call a function in a non-overridable way, of which avoiding a virtual function call is a nice side-effect.

spec/function.dd
jblachly: @thewilsonator this is corrected; it was a merging error when re-basing my branch as I removed this in response to your comment in 2019 ;)

https://github.com/dlang/dlang.org/pull/2570/commits/380da3c29eeeba5c4f7610daa81f4f6ce6a36c02

spec/function.dd
jblachly: Counterpoint, the purpose of avoiding dynamic binding on a member function call is to call a function in a non-overrideable way, so it is semantically and logically equivalent IMO

thewilsonator: Fair enough.

thewilsonator (approved):

thewilsonator:
```
./spec/function.dd:1029:        $(P 
```
otherwise looks  good.

PR #2572, opened 2019-02-08
rewrite section on struct equality comparison

WalterBright:
Make the definition more precise.
The idea that the struct holes are filled with 0 is not tenable anymore:

1. it's not compatible with the C/C++ ABI
2. it has efficiency issues
3. it interferes with struct slicing

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -300,39 +300,54 @@ $(GNAME EqualExpression):
 
     $(P Equality expressions compare the two operands for equality ($(D ==))
         or inequality ($(D !=)).
-        The type of the result is $(D bool). The operands
-        undergo the $(USUAL_ARITHMETIC_CONVERSIONS) to bring them to a common type before
-        comparison.
+        The type of the result is $(D bool).
     )
 
-    $(P If they are integral values or pointers, equality
-        is defined as the bit pattern of the type matches exactly.
+    $(P Inequality is defined as the logical negation of equality.)
+
+    $(P If the operands are integral values, the $(USUAL_ARITHMETIC_CONVERSIONS) are applied
+        to bring them to a common type before comparison. Equality is defined as the bit patterns
+        of the common type match exactly.
     )
 
-    $(P Equality for floating point types is more complicated. $(D -0) and
-        $(D +0) compare as equal. If either or both operands are NAN, then
-        both the $(D ==) returns false and $(D !=) returns $(D true). Otherwise, the bit
-        patterns are compared for equality.
+    $(P If the operands are pointers, equality is defined as the bit patterns of the operands
+        match exactly.
     )
 
-    $(P For complex numbers, equality is defined as equivalent to:)
+    $(P For float, double, and real values, the $(USUAL_ARITHMETIC_CONVERSIONS) are applied
+        to bring them to a common type before comparison.
+        The values $(D -0) and $(D +0) are considered equal.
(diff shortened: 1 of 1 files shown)


spec/expression.dd
```diff
@@ -300,39 +300,54 @@ $(GNAME EqualExpression):
...
-    )
+    $(P Inequality is defined as the logical negation of equality.)
 
-    $(P Equality for floating point types is more complicated. $(D -0) and
-        $(D +0) compare as equal. If either or both operands are NAN, then
-        both the $(D ==) returns false and $(D !=) returns $(D true). Otherwise, the bit
-        patterns are compared for equality.
+    $(P If the operands are integral values, the $(USUAL_ARITHMETIC_CONVERSIONS) are applied
+        to bring them to a common type before comparison. Equality is defined as the bit patterns
+        of the common type match exactly.
```
aG0aep6G: Not a native English speaker, but shouldn't it be "match**ing**"?

spec/expression.dd
```diff
@@ -300,39 +300,54 @@ $(GNAME EqualExpression):
...
     )
 
-    $(P For complex numbers, equality is defined as equivalent to:)
+    $(P If the operands are pointers, equality is defined as the bit patterns of the operands
+        match exactly.
```
aG0aep6G: ditto "matching"

spec/expression.dd
```diff
@@ -300,39 +300,54 @@ $(GNAME EqualExpression):
...
+    )
 
-        and inequality is defined as equivalent to:
+    $(P For complex numbers, equality is defined as equivalent to:
 
         ---
-        x.re != y.re || x.im != y.im
+        x.re == y.re && x.im == y.im
         ---
+    )
```
aG0aep6G: Having a code block in a paragraph like results in broken HTML which manifests as an empty paragraph number 7.

spec/expression.dd
```diff
@@ -300,39 +300,54 @@ $(GNAME EqualExpression):
...
+        An `opEquals()` can account for which of the overlapping fields contains valid data.
+        An `opEquals()` can override the default behavior of floating point NaN values
+        always comparing as unequal.
+        Be careful using `memcmp()` to implement `opEquals()` if:
+        $(UL
+        $(LI there are any alignment gaps)
+        $(LI if any fields have an `opEquals()`)
+        $(LI there are any floating point fields that may contain NaN or `-0` values)
+        )
+        )
```
aG0aep6G: Another empty paragraph (number 11) is generated here, because you put stuff into a paragraph that HTML doesn't like there.

spec/expression.dd
```diff
@@ -300,39 +300,54 @@ $(GNAME EqualExpression):
...
+        $(LINK2 https://dlang.org/spec/operatoroverloading.html#equals, `opEquals()` member function).
+        If an `opEquals()` is not provided, equality is defined as
+        the logical product of all equality
         results of the corresponding object fields.
-        If all struct fields use bitwise equality, the whole struct equality
-        could be optimized to one memory comparison operation (the existence of
-        alignment holes in the objects is accounted for, usually by setting them
-        all to 0 upon initialization).
+
+        $(IMPLEMENTATION_DEFINED The contents of any alignment gaps in the struct object.)
```
aG0aep6G: Is this supposed to say that it's implementation defined whether the gaps are compared for equality (in addition to the fields)? As it is, this says nothing about equality, so equality is still defined in terms of fields only.

PR #2574, opened 2019-02-09
[spec] safe asm statements can be used in safe functions

kubo39:
The specification of function says that the inline assembler is not
 allowed in safe functions, However, safe asm statements can be used in
 safe functions. I'm not sure but this behavior seems to be expected.


aG0aep6G:
> safe asm statements can be used in safe functions.

Changing the text to disallow `@system` asm statements is fine. But I'll note that `@safe` asm statements should always be invalid (unless the compiler can prove their safety, which it can't). What can be allowed are `@trusted` asm statements.

jacob-carlborg:
What is a "safe asm statements"?

thewilsonator:
`@safe asm { ... };`, no. They can be trusted though.

wilzbach:
> Can the compiler determine that?

FWIW it should be able to hard-code some trivial `asm` block as `@safe`:

```d
asm { nop; }
```

(This is often used for benchmarks as DMD's optimizer stops being greedy if there's explicit `asm`.)

PR #2576, opened 2019-02-13
traits, specify that `isSame` works with types

Basile-z:
To be accurate with the change in __traits semantics made by https://github.com/dlang/dmd/pull/9356

### DIFF
--- a/spec/traits.dd
+++ b/spec/traits.dd
@@ -1316,6 +1316,20 @@ void main()
 ---
 )
 
+        $(P If the two arguments are identical types, which includes the same
+        type qualifiers and the same type suffixes, true is returned.)
+
+$(SPEC_RUNNABLE_EXAMPLE_COMPILE
+---
+void main()
+{
+    // similarly to `is(T1 == T2)`, types with qualifiers and suffixes
+    static assert(__traits(isSame, immutable(char)[], string));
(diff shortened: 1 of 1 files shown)


spec/traits.dd
```diff
@@ -1349,6 +1352,12 @@ void test(alias pred)()
 
 void main()
 {
+
+    // similarly to `is(T1 == T2)`, types with qualifiers and suffixes
+    static assert(__traits(isSame, immutable(char)[], string));
+    static assert(!__traits(isSame, const(char)[], string));
+
+    // lambdas, common cases
```
ZombineDev: Can you move this example immediately after the new paragraph, as this example is already getting big?

Basile-z: I'll see but I'm afraid that this could break the paragraph numbers.

ZombineDev: There's an example between 1. and 2., so in theory at least it should work. But if it turns out that it's too much work, then no need to bother.

Basile-z: No this is definitively not too much work. I probably mismatch with another problem I've seen once while writing ddoc. I think that the issue was rather when you put code within a single paragraph.

Basile-z:
Well after all i think it's not required to mention the change.

PR #2578, opened 2019-02-15
Clarify what happens with overlarge shifts

WalterBright:


spec/expression.dd
```diff
@@ -550,13 +550,15 @@ $(GNAME ShiftExpression):
         $(D >)$(D >)$(D >) is an unsigned right shift.
     )
 
-    $(P It's illegal to shift by the same or more bits than the size of the
-        quantity being shifted:)
+    $(UNDEFINED_BEHAVIOR
+    It is undefined behavior to shift by a negative value or by the same or more bits
+    than the size of the quantity being shifted:
```
aG0aep6G: If bad shifts trigger undefined behavior, DMD needs to reject them in `@safe` code. It already does so when the right operand is a constant, but it needs to reject variables as well.

rainers:
Memory `@safe`ty is not compromised but I think having undefined behaviour with such a basic operation is undesirable. AFAICT it is motivated by having best performance for x86 processors but the optimizer might provide that with a sane definition, too ( (e.g. by detecting explicit `x << (c & 0x1f)`).

FWIW I just made a short search in other languages:
- C/C++: same undefined behaviour
- C#: explicitly applies modulo bit size
- Rust: adds runtime check
- Go: does not apply modulo bit size

aG0aep6G:
> Memory `@safe`ty is not compromised but I think having undefined behaviour with such a basic operation is undesirable.

`@safe` is defined in terms of undefined behavior, not memory safety.
https://dlang.org/spec/function.html#function-safety

rainers:
From what I read Walter usually considers `@safe` to only mean "memory safe", e.g. https://forum.dlang.org/post/mupb4h$1vc$1@digitalmars.com. The following paragraph in the spec also just lists dangerous memory operations.

I'd prefer a stronger definition, too, though, and a defined behaviour for the shift operators.

spec/expression.dd
thewilsonator: I don't think undefined behaviour is the right term here, unspecified is. Undefined is for things that you have no ability to reason what state the program is in. `1 << x` where `x` is invalid gives you an `int`, you don't know what value it holds, but that in itself doesn't make the program state invalid.

thewilsonator (dismissed):
Behaviour should be unspecified (not undefined) in the case:
```
int c,x = -1;
c <<= x;
```
 but disallowed as a static expression `1 << (-1)`. Examples using (u)long should also be added for where  `<< x` where `x<=63` is allowed

dkorpel:
The wrong promotion rule was fixed, so this PR is now about what happens with overlarge shifts. Ping @WalterBright

dkorpel:
During the DLF Monthly Meeting, we agreed it's not undefined behavior but implementation defined behavior.

dkorpel (approved):

PR #2583, opened 2019-02-17
Add reference to `toDelegate` in the spec.

veelo:
Documenting how function pointers can be converted to delegates, to clarify that no function overloads are necessary that replace delegates with function arguments. It complements the previous paragraph that shows that delegates can be initialized referencing any function, but they cannot be assigned like that; `toDelegate` fills that gap.

Motivated by https://forum.dlang.org/post/jueqsawmegwznoycdmjy@forum.dlang.org

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -2368,7 +2368,7 @@ void main()
 )
 
         $(P Function pointers can be passed to functions taking a delegate argument by passing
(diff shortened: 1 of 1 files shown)


thewilsonator:
Please squash. @wilzbach can we get `auto-merge-squash` for dlang.org please?

wilzbach:
Sure. Added

thewilsonator:
Thanks!

PR #2584, opened 2019-02-19
document ref-return-scope function parameters

WalterBright:
Part of DIP1000.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -1542,6 +1542,82 @@ class C
         $(P Template functions, auto functions, nested functions and lambdas can deduce
         the `return scope` attribute.)
 
+$(H3 $(LNAME2 ref-return-scope-parameters, Ref Return Scope Parameters))
+
+        $(P Parameters marked as `ref return scope` come in two forms:)
+
+---
+U xerxes(ref return scope V v);      // (1) ref and return scope
+ref U xerxes(ref return scope V v);  // (2) return ref and scope
+---
+
+        $(P The first form attaches the `return` to the `scope`, and has
+        $(LINK2 #return-scope-parameters, return scope parameter) semantics
+        for the value of the `ref` parameter.)
+
+        $(P The second form attaches the `return` to the `ref`, and has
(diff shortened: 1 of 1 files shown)


spec/function.dd
```diff
@@ -1542,6 +1542,33 @@ class C
         $(P Template functions, auto functions, nested functions and lambdas can deduce
         the `return scope` attribute.)
 
+$(H3 $(LNAME2 ref-return-scope-parameters, Ref Return Scope Parameters))
+
+        $(P Parameters marked as `ref return scope` come in two forms:)
+
+---
+U xerxes(ref return scope V v);      // (1) ref and return scope
+ref U xerxes(ref return scope V v);  // (2) return ref and scope
```
thewilsonator: This is still a bit cryptic to me, could you add some (il)legal uses/definitions please?

spec/function.dd
```diff
@@ -1542,6 +1542,82 @@ class C
...
+
+    int  val;
+    int* ptr;
+}
+
+int* foo1(return scope   ref S s);
+int  foo2(return scope   ref S s);
+
+ref int* foo3(return ref   scope S s);
+ref int  foo4(return ref   scope S s);
```
jacob-carlborg: Please avoid aligning the code. It will just require more work if new examples are added.

ZombineDev: I disagree, in this particular case it really helps to get the point across.

spec/function.dd
```diff
@@ -1542,6 +1542,82 @@ class C
...
+ref U xerxes(ref return scope V v);  // (2) return ref and scope
+---
+
+        $(P The first form attaches the `return` to the `scope`, and has
+        $(LINK2 #return-scope-parameters, return scope parameter) semantics
+        for the value of the `ref` parameter.)
+
+        $(P The second form attaches the `return` to the `ref`, and has
+        $(LINK2 #return-ref-parameters, return ref parameter) semantics
+        with additional
```
wilzbach: Don't use relative LINK2 in the specification. It doesn't work with Latex.

PR #2591, opened 2019-02-27
Fix confusing grammar in Assignment Overload spec

TJesionowski:
A previous commit (823d8a4) clarified that destructors are called upon assignment to an initialized object. In doing so it spliced the sentence in a very unnatural way. I've attempted to correct the sentence grammar while preserving the semantics.

It's not clear to me if calling the dtor on assignment to initialized objects is necessary, or if it's just an optional thing the default opAssign is meant to do. I'd appreciate clarification on this.

### DIFF
--- a/spec/struct.dd
+++ b/spec/struct.dd
@@ -1411,7 +1411,7 @@ $(H2 $(LEGACY_LNAME2 AssignOverload, assign-overload, Identity Assignment Overlo
 
         $(P While copy construction takes care of initializing
         an object from another object of the same type,
-        assignment is defined as copying the contents of a source 
+        assignment is defined as copying the contents of a source
         object over those of a destination object, calling the
         destination object's destructor if it has one in the process:
         )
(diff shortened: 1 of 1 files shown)


thewilsonator:
cc @RazvanN7

RazvanN7:
From my point of view, the sentence added by [1] should be deleted. If a struct defines a destructor it is used internally by the generated opAssign to destroy a temporary; for more information you can check [2] and the associated dmd code for buildOpAssign which is in src/dmd/clone.d in the dmd repo.

I don't think that we should mix compiler internals with public spec. [1] should not have been merged.

[1] https://github.com/dlang/dlang.org/commit/823d8a46d73b35f6f5836a636ca4f2169a0be99d
[2] https://dlang.org/phobos/dmd_clone.html#.buildOpAssign

ZombineDev:
BTW @RazvanN7, compared to C++, D's is behavior is a bit strange (even after reading `
buildOpAssign`'s docs): https://gist.github.com/ZombineDev/5132213eb6950bd9853fc903394aa4c5

RazvanN7:
I've replied on the example in your repo.

thewilsonator:
```
./spec/struct.dd:1593:        assignment is defined as copying the contents of a source 
```

PR #2593, opened 2019-03-01
Update spec for Issue 16657 and 11161

RazvanN7:
See https://github.com/dlang/dmd/pull/9289


thewilsonator:
So much for 72h then.

PR #2595, opened 2019-03-02

MartinNowak:


MartinNowak:
We could do sth. about the slow doc CI at some point. Over an hour waittime and it didn't even start.
Though the tool @CyberShadow has build is really good, and I'd prefer to keep it. So maybe he has an idea how to reduce latency and build time.

CyberShadow:
@MartinNowak I'll have a look. FYI, currently everything is temporarily running on my PC at home due to the issues with the server provider.

CyberShadow:
The problem is the ever-growing build time. What once took 5 minutes, now takes [almost half an hour](http://dtest.dlang.io/results/e835c564a315398d88f5801dcab47ffc487f5d1f/ef6b73bb2f1896f8c0b97290664794d893c3a603/build.log). It looks like there is something broken with the makefile dependencies - the `test` target causes all the documentation to be built all over again. @wilzbach I think you'd be the best person to have a look if you have time. 

If we can't improve build time, then parallelizing builds is the way forward. Currently DAutoTest does one build at a time, which causes long queues in busy periods. If BuildKite has a reasonable way to grab large artifacts with very many files, that would be one way to go, to avoid relying too much on the available computing power of the server DAutoTest is running on.

wilzbach:
I will try to have a look over the next weeks, but yeah the the build setup grew excessively :/

PR #2599, opened 2019-03-05
Update struct.dd to highlight copy constructor addition

RazvanN7:
As per le DIP [1].

[1] https://github.com/dlang/DIPs/blob/master/DIPs/DIP1018.md

### DIFF
--- a/spec/struct.dd
+++ b/spec/struct.dd
@@ -808,6 +808,178 @@ $(H2 $(LNAME2 field-init, Field initialization inside constructor))
         ---
         )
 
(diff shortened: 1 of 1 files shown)


thewilsonator:
```
./spec/struct.dd:985:    
```

Please can we automate this stupid test already.

andralex:
@thewilsonator @CyberShadow hmmm... failures in DAutoTest look odd indicating a broken archive. What could be happening?

CyberShadow:
curl failed (and thus its output was not a valid archive), just a transient network error it seems...

PR #2603, opened 2019-03-14
Update memory-safe-d.dd

RazvanN7:


thewilsonator:
```
./spec/memory-safe-d.dd:71:        
```

Why do we have this stupid rule?

RazvanN7:
We should be fine now.

spec/memory-safe-d.dd
```diff
@@ -68,6 +68,9 @@ $(H2 $(LNAME2 usage, Usage))
          $(P These attributes may appear after the formal parameter list, in which case they apply either to a method's `this` parameter, or to
          a free function's first parameter $(I iff) it is `ref`.
          `return` or `scope` is ignored when applied to a type that is not a low-level pointer.)
+         
+         $(P $(B Note:) Checks for `scope` parameters are currently enabled
```
ZombineDev: I think `return` falls in the same category as well. Part of the checking introduced by `-dip25` covers it, but I think there are some additional checks with `-dip1000` as well.

ZombineDev:
> Why do we have this stupid rule

Because everyone needs to have https://editorconfig.org/ setup on their editor of choice ;)

spec/memory-safe-d.dd
RazvanN7: It is a bit awkward since return checks work even if the function is not @safe when -dip1000 is passed

ZombineDev: Yeah, it's confusing :/

thewilsonator:
```
./spec/memory-safe-d.dd:71: 
```

Seriously can we please remove this stupid checker, I serves no purpose.

thewilsonator:
> Because everyone needs to have https://editorconfig.org/ setup on their editor of choice ;)

The GitHub online editor doesn't, which is where I make most of my dlang.org changes, and I'm assuming other do too. It's useless and provides net negative value. We should remove it.

ZombineDev:
> It's useless and provides net negative value. We should remove it.

Yes, I agree about the GitHub online editor. We should stop using it as it only causes problems.

thewilsonator:
> We should stop using it

GH editor or ws checker?

ZombineDev:
I have tried using the GH editor many times in the past years. The net result is that I always ended up loosing more time, than if I had cloned the repo and used a proper editor. For what is worth, GitLab's online IDE is much better.

thewilsonator:
For small PRs like this it is perfect, or it would be if it didn't get blocked by stupid things like the whitespace checker.

wilzbach:
> For small PRs like this it is perfect, or it would be if it didn't get blocked by stupid things like the whitespace checker

Hmm or automate submitting and editing PRs. This will save you a lot more time on the long run.
I use this here: https://github.com/wilzbach/git-tools

And sorry, no, I don't see the point of removing style checks just because the GH editor is bad.

thewilsonator:
> And sorry, no, I don't see the point of removing style checks just because the GH editor is bad.

Indeed. We should remove it because it serves no purpose and has net negative value.

RazvanN7:
Should be fine now.

>And sorry, no, I don't see the point of removing style checks just because the GH editor is bad

Isn't there a way to make GH highlight trailing spaces/tabs?

wilzbach:
>  serves no purpose and has net negative value.

Just because you don't see the value, doesn't mean there isn't any.
There are a plethora of documentation systems united here (Ddoc, Ddox, LaTeX, kindle build, ...) many of which are _white-space sensitive_ and we have run into issues with this in the past, which is why the check was introduced in the first place.

There are some other usability reasons here: https://softwareengineering.stackexchange.com/questions/121555/why-is-trailing-whitespace-a-big-deal

thewilsonator:
I don't find those linked opinions very convincing. We should fix the tools.

ZombineDev:
> We should fix the tools.

Yes, hence: https://github.com/dlang/dlang.org/pull/2603#issuecomment-472917046 ;)

jacob-carlborg:
I completely agree with Sebastian. If you're committing trailing whitespace I get the feeling you're not paying enough attention.

thewilsonator:
> Yes, hence: #2603 (comment) ;)

Thats is not what I meant.

thewilsonator:
> I completely agree with Sebastian. If you're committing trailing whitespace I get the feeling you're not paying enough attention.

It comes primarily from splitting lines that get too long, and the problem is not that I not paying enough attention, its that, by necessity I'm focused on the documentation updating, not the fact that the useless whitespace checker is going to be unhappy.

jacob-carlborg:
I don't understand why anyone would use the GitHub editor. Any of the core members should already have all the important repositories cloned. It takes less than 10 seconds to update to the latest master and create a new branch. These days GitHub will, when pushing a new branch, write out a URL that will create a new merge request for that branch. If you have a proper terminal you should be able to click on that URL directly.

jacob-carlborg:
Tailing whitespace has no meaning. Not a single byte should be added to a file that doesn't have any meaning.

thewilsonator:
> Tailing whitespace has no meaning.

Indeed, so it should be ignored.

> Not a single byte should be added to a file that doesn't have any meaning.

If you're worried about O(byte) bloat, you're missing the forrest for the trees.

RazvanN7:
Well, I have the dlang.org repository cloned, but I use it for bigger additions; sometimes I'm reading the spec and I encounter a typo; it is easier to simply click "Improve this page" than open a terminal, fetch latest master, find the file (this may take quite some time) and create the PR.

jacob-carlborg:
> If you're worried about O(byte) bloat, you're missing the forrest for the trees.

I don't. As I said, I'm worrying that people that doesn't care about trailing whitespace doesn't care enough.

thewilsonator:
Then you are missing the mountain the forest sits on for the trees. I would wager that not only is caring about whitespace not in direct correlation with general caring, but they are negatively correlated. If people are ignoring the noise (ws), then they have more mental bandwidth for the signal (the thing they actually care about: fixing the documentation). 

Its all fine and dandy for this to be automated but the editors that you use (or even better the infrastructure!), but I'm not going to bother with fiddling around on my machine for a 3 line PR. An I'm not alone with that: Razvan did, Andrei did. I don't think either of them are anywhere close to not caring enough. They want to get stuff done. So do I.

PR #2551, opened 2019-01-10
Bump versions

wilzbach:
Required for some PRs at DMD and while I was at it I thought I bump everything.

PR #2605, opened 2019-03-18
Add definition of full expression

andralex:
cc @WalterBright

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -4,10 +4,23 @@ $(SPEC_S Expressions,
 
 $(HEADERNAV_TOC)
 
+$(P This chapter defines the syntax, order of evaluation, and semantics of expressions. An expression is a sequence of operators and operands that specifies an evaluation.)
+
     $(P Expressions are used to compute values with a resulting type.
         These values can then be assigned,
         tested, or ignored. Expressions can also have side effects.
     )
+
+$(H2 $(LNAME2 general, Definitions and Terms))
+
+$(DDOC_ANCHOR define-full-expression)$(P $(B Definition) ($(DOUBLEQUOTE Full expression)): For any expression $(I expr) in a D program, we define the full expression of $(I expr) as follows. If $(I expr) parses as a subexpression of another expression $(I expr$(SUBSCRIPT 1)), then the full expression of $(I expr) is the full expression of $(I expr$(SUBSCRIPT 1)). Otherwise, $(I expr) is its own full expression.)
+
+$(P Each expression has a unique full expression.)
+
+Example: in the statement `return f() + g() * 2;`, the full expression of `g() * 2` is `f() + g() * 2`, but not the full expression of `f() + g()` because the latter is not parsed as a subexpression.
+
+Note: Although the definition is straightforward, a few subtleties exist related to function literals. In the statement `return (() => x + f())() * g();`, the full expression of `f()` is `x + f()`, not the expression passed to `return`. This is because the parent of `x + f()` has function literal type, not expression type.
+
 
 $(H2 $(LNAME2 order-of-evaluation, Order Of Evaluation))
 
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -12,7 +12,7 @@ of operators and operands that specifies an evaluation.)
         tested, or ignored. Expressions can also have side effects.
     )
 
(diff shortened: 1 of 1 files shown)


andralex:
Urgh. Is there a way to see a diff with wraparound enabled?

jacob-carlborg:
It's quite difficult to review lines that are these long.

jacob-carlborg:
Thanks.

CyberShadow:
> Urgh. Is there a way to see a diff with wraparound enabled?

M-x fill-paragraph, amend, force-push

In situations where newlines do not show up in rendered text, I found that adding them at sentence breaks makes sense for version controlled text (diffs are good), and it's not unlike how statements are generally one per line in programs.

andralex:
@CyberShadow I'm experimenting with Visual Studio Code...

spec/expression.dd
```diff
@@ -4,11 +4,31 @@ $(SPEC_S Expressions,
 
 $(HEADERNAV_TOC)
 
+$(P This chapter defines the syntax, order of evaluation, and semantics of expressions. An expression is a sequence
+of operators and operands that specifies an evaluation.)
+
```
WalterBright: "An expression is a sequence of operators and operands that specifies an evaluation.
The syntax, order of evaluation, and semantics of expressions are as follows."

(I dislike things that start with "This chapter defines...". It seems redundant.)

spec/expression.dd
```diff
@@ -4,11 +4,31 @@ $(SPEC_S Expressions,
...
     $(P Expressions are used to compute values with a resulting type.
         These values can then be assigned,
         tested, or ignored. Expressions can also have side effects.
     )
 
+$(H2 $(LNAME2 definitions-and-terms, Definitions and Terms))
+
+$(DDOC_ANCHOR define-full-expression)$(P $(B Definition) ($(DOUBLEQUOTE Full expression)): For any expression
+$(I expr) in a D program, we define the full expression of $(I expr) as follows. If $(I expr) parses as a
```
WalterBright: s/we define the full expression of $(I expr)/the full expression of $(I expr) is defined/

On a formal document, "we" should not be there.

andralex:
@WalterBright done

thewilsonator:
`./spec/expression.dd:27:Note: Although the definition is straightforward, a few subtleties exist related to function literals. In the 
`
Can we please get rid of this already?

andralex:
"This" being what?

thewilsonator:
The whitespace checker.

andralex:
@thewilsonator I guess it could fix the error itself. I introduced that because I was using the online editor.

BTW I see a bunch of errors in magic.d when looking at the doc builter log http://dtest.dlang.io/results/e63eaaa2e744e490663fb29307ffb6ff65b803de/b435eade3ae96f349e71710e47a353608fff6a8b/build.log @CyberShadow could they be fixed?

thewilsonator:
> I guess it could fix the error itself.

Yes it should.

> I introduced that because I was using the online editor.

Yes I do that all the time, the whitespace checker is a PITA.

> BTW I see a bunch of errors in magic.d when looking at the doc builter log

They look almost like the ones I had in https://github.com/dlang/dlang.org/pull/2453#issuecomment-443477237 (see the force pushes just below).

CyberShadow:
> @CyberShadow could they be fixed?

I think the authors of the relevant code are no longer maintaining it (DDox), so someone else would need to adopt / look into it. In any case, those errors are "benign", and indicate that the tool just failed to grok the syntax of that file.

wilzbach:
>  In any case, those errors are "benign", and indicate that the tool just failed to grok the syntax of that file.

Well, actually the errors are coming from libdparse trying to parse newer code from Phobos for Ddoc (<-not the c here) as we have to parse the AST again to be able to rewrite the unittests, so even if we drop Ddox, this would still be there.

See also: https://github.com/dlang/dlang.org/blob/master/ddoc/source/assert_writeln_magic.d

In theory, upgrading libdparse should do the trick, but last time I did I ran into issues: https://github.com/dlang/dlang.org/pull/2551

CyberShadow:
> actually the errors are coming from libdparse trying to parse newer code from Phobos for Ddoc

Thanks for the correction, I did mix that up. To be fair, libdparse is also no longer maintained by its original author, as with DDox.

PR #2606, opened 2019-03-19
Define rvalues and lvalues

andralex:
Did I miss something?

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -32,7 +32,7 @@ $(DDOC_ANCHOR define-lvalue)$(P $(B Definition) ($(DOUBLEQUOTE Lvalue)): The fol
 others, are called lvalue expressions or lvalues:)
 $(OL
 $(LI `this` inside `struct` and `union` member functions;)
-$(LI a simple symbol or the result of the $(I DotIdentifier) grammatical construct `.` (left side may be
+$(LI a variable or the result of the $(I DotIdentifier) grammatical construct `.` (left side may be
 missing) when the rightmost side of the dot is a variable,
 field (direct or `static`), function name, or invocation of a function that returns by reference;)
 $(LI the result of the following expressions:
@@ -59,7 +59,7 @@ $(DDOC_ANCHOR define-rvalue)$(P $(B Definition) ($(DOUBLEQUOTE Rvalue)): Express
 lvalues are rvalues.)
 
 $(P Note: Rvalues include all literals, special value keywords such as `__FILE__` and `__LINE__`,
-`enum` values, and the result of expressions not mentioned in the definition of lvalues above.)
+`enum` values, and the result of expressions not defined as lvalues above.)
 
 $(P The built-in address-of operator (unary `&`) may only be applied to lvalues.)
 
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -21,13 +21,47 @@ full expression of $(I expr$(SUBSCRIPT 1)). Otherwise, $(I expr) is its own full
 
 $(P Each expression has a unique full expression.)
 
-Example: in the statement `return f() + g() * 2;`, the full expression of `g() * 2` is `f() + g() * 2`, but not the
-full expression of `f() + g()` because the latter is not parsed as a subexpression.
+$(P Example: in the statement `return f() + g() * 2;`, the full expression of `g() * 2` is `f() + g() * 2`, but not the
+full expression of `f() + g()` because the latter is not parsed as a subexpression.)
 
-Note: Although the definition is straightforward, a few subtleties exist related to function literals. In the
+$(P Note: Although the definition is straightforward, a few subtleties exist related to function literals. In the
 statement `return (() => x + f())() * g();`, the full expression of `f()` is `x + f()`, not the expression passed
-to `return`. This is because the parent of `x + f()` has function literal type, not expression type.
+to `return`. This is because the parent of `x + f()` has function literal type, not expression type.)
 
+$(DDOC_ANCHOR define-lvalue)$(P $(B Definition) ($(DOUBLEQUOTE Lvalue)): The following expressions, and no
+others, are called lvalue expressions or lvalues:)
+$(OL
+$(LI `this` inside `struct` and `union` member functions;)
+$(LI a simple symbol or the result of the $(I DotIdentifier) grammatical construct `.` (left side may be
+missing) when the rightmost side of the dot is a variable,
+field (direct or `static`), function name, or invocation of a function that returns by reference;)
+$(LI the result of the following expressions:
+$(UL
+$(LI built-in unary operators `+` (when applied to an lvalue), `*`, `++` (prefix only), `--` (prefix only);)
+$(LI built-in indexing operator `[]` (but not the slicing operator);)
+$(LI built-in assignment binary operators, i.e. `=`, `+=`, `*=`, `/=`, `%=`, `&=`, `|=`, `^=`, `~=`,
+`<<=`, `>>=`, `>>>=`, and `^^=`;)
+$(LI the ternary operator $(I e) `?` $(I e$(SUBSCRIPT 1)) `:` $(I e$(SUBSCRIPT 2)) under the following
+circumstances:)
+$(OL
+    $(LI $(I e$(SUBSCRIPT 1)) and $(I e$(SUBSCRIPT 1)) are lvalues of the same type; OR)
+    $(LI One of $(I e$(SUBSCRIPT 1)) and $(I e$(SUBSCRIPT 2)) is an lvalue of type `T` and the other has
+    and `alias this` converting it to `ref T`;))
+$(LI user-defined operators if and only if the function called as a result of lowering returns
+by reference;)
+$(LI `mixin` expressions if and only if the compilation of the expression resulting from compiling
+the argument(s) to `mixin` is an lvalue;)
+$(LI `cast(U)` expressions applied to lvalues of type `T` when `T*` is implicitly convertible to `U*`;)
+$(LI `cast()` and `cast(`$(I qualifier list)`)` when applied to an lvalue.)
+)))
+
+$(DDOC_ANCHOR define-rvalue)$(P $(B Definition) ($(DOUBLEQUOTE Rvalue)): Expressions that are not
+lvalues are rvalues.)
+
+$(P Note: Rvalues include all literals, special value keywords such as `__FILE__` and `__LINE__`,
+`enum` values, and the result of expressions not mentioned in the definition of lvalues above.)
+
+$(P The built-in address-of operator (unary `&`) may only be applied to lvalues.)
 
 $(H2 $(LNAME2 order-of-evaluation, Order Of Evaluation))
 


spec/expression.dd
```diff
@@ -28,6 +28,52 @@ Note: Although the definition is straightforward, a few subtleties exist related
...
+    $(LI `null`, `true`, `false`, `$`, `this`, `super`;)
+    $(LI literals: integral, floating point, character, string, array, associative array, function;)
+    $(LI `enum` values;)
+    $(LI the result of the following expressions:)
+    $(UL
+        $(LI functions that return by value, whether explicitly invoked or as the result of lowering
+        user-defined operators;)
+        $(LI the ternary operator `?:` when the two choices have different types, or at least one is an 
+        rvalue;)
+        $(LI built-in binary operators `||`, `&&`, `|`, `^`, `&`, `==`, `!=`, `is`, `!is`, `<`, `<=`, `>`,
```
thewilsonator: I would reword that as the binary operators  [...] acting on builtin types.

spec/expression.dd
```diff
@@ -28,6 +28,52 @@ Note: Although the definition is straightforward, a few subtleties exist related
...
+        $(LI `is` expressions;)
+        $(LI the special keywords `__FILE__`, `__FILE_FULL_PATH__`, `__MODULE__`, `__LINE__`,
+        `__FUNCTION__`, `__PRETTY_FUNCTION__`, `__DATE__`, `__TIME__`, `__TIMESTAMP__`, `__VENDOR__`,
+        `__VERSION__`.)))
+
+$(DDOC_ANCHOR define-lvalue)$(P $(B Definition) ($(DOUBLEQUOTE Lvalue)): The following expressions, and no
+others, are called lvalue expressions or lvalues:)
+$(OL
+$(LI variable names;)
+$(LI function names;)
```
thewilsonator: s/names/symbols

spec/expression.dd
```diff
@@ -28,6 +28,52 @@ Note: Although the definition is straightforward, a few subtleties exist related
...
+$(LI the result of the $(I DotIdentifier) grammatical construct `.` when the rightmost side of the dot
+is the name of a field (whether direct or `static`);)
```
thewilsonator: ... and not an enum e.g. `struct a { enum b = 1; }`

spec/expression.dd
```diff
@@ -28,6 +28,52 @@ Note: Although the definition is straightforward, a few subtleties exist related
...
+$(LI the result of the following expressions:
+$(UL
+$(LI built-in unary operators `+`, `*`, `++` (prefix only), `--` (prefix only);)
```
thewilsonator: Is not so sure about '+' (I get why, but I'm not sure the compiler actually treats it that way)

spec/expression.dd
```diff
@@ -28,6 +28,52 @@ Note: Although the definition is straightforward, a few subtleties exist related
...
+$(LI built-in binary operators `[]`, `=`, `+=`, `*=`, `/=`, `%=`, `&=`, `|=`, `^=`, `~=`,
```
thewilsonator: `[]` isn't a binary operator.

spec/expression.dd
```diff
@@ -28,6 +28,52 @@ Note: Although the definition is straightforward, a few subtleties exist related
...
+`<<=`, `>>=`, `>>>=`, `^^=`;)
+$(LI the ternary operator `?:` when the two choices have identical types and are lvalues;)
```
thewilsonator: does `alias this` play any tricksies here?

spec/expression.dd
thewilsonator: fair enough.

spec/expression.dd
andralex: I see the grammar (same file, below) uses "identifier". I'll rephrase and wait for Walter.

spec/expression.dd
andralex: are enums fields? I think they're just enums that happen to be nested within the aggregate.

spec/expression.dd
andralex: I checked. See e.g. https://run.dlang.io/is/mPBBv1. But I realize the operand of `+` must be an lvalue, so I'll amend.

spec/expression.dd
andralex: Hrm, it actually does and in a not-super-consistent manner. Will update.

spec/expression.dd
thewilsonator: Oh, I missed the bit about fields.

andralex:
I've redone the definitions, focusing on lvalues and defining rvalues by complement. Otherwise things get very tedious, e.g. "lvalues are the ternary operator if this and that and the other; rvalues are the ternary operator if not this or not that or not the other" etc.

rainers:
You can grep https://github.com/dlang/dmd/blob/master/src/dmd/expression.d for `bool isLvalue()` to get the dirty details. Some have a comment to explain the special cases.

spec/expression.dd
```diff
@@ -21,13 +21,44 @@ full expression of $(I expr$(SUBSCRIPT 1)). Otherwise, $(I expr) is its own full
...
-to `return`. This is because the parent of `x + f()` has function literal type, not expression type.
-
+to `return`. This is because the parent of `x + f()` has function literal type, not expression type.)
+
+$(DDOC_ANCHOR define-lvalue)$(P $(B Definition) ($(DOUBLEQUOTE Lvalue)): The following expressions, and no
+others, are called lvalue expressions or lvalues:)
+$(OL
+$(LI `this` inside `struct` and `union` member functions;)
+$(LI the result of the $(I DotIdentifier) grammatical construct `.` (left side may be missing) when the
+rightmost side of the dot is the name of a field (whether direct or `static`), variable, or function;)
```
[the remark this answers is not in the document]
RazvanN7: Isn't the leftmost side supposed to be an lvalue?

spec/expression.dd
```diff
@@ -21,13 +21,44 @@ full expression of $(I expr$(SUBSCRIPT 1)). Otherwise, $(I expr) is its own full
...
+$(LI the result of the following expressions:
+$(UL
+$(LI built-in unary operators `+` (when applied to an lvalue), `*`, `++` (prefix only), `--` (prefix only);)
```
tgehr: Unary `+` indeed preserves lvalues, but I'm not convinced that it should.

spec/expression.dd
```diff
@@ -21,13 +21,44 @@ full expression of $(I expr$(SUBSCRIPT 1)). Otherwise, $(I expr) is its own full
...
+circumstances:)
+$(OL 
+    $(LI $(I e$(SUBSCRIPT 1)) and $(I e$(SUBSCRIPT 1)) are lvalues of the same type; OR)
+    $(LI One of $(I e$(SUBSCRIPT 1)) and $(I e$(SUBSCRIPT 2)) is an lvalue of type `T` and the other has
+    and `alias this` converting it to `ref T`.))
+$(LI `mixin` expressions if and only if the compilation of the expression resulting from compiling
+the argument(s) to `mixin` is an lvalue;)
+$(LI `cast(U)` expressions applied to lvalues of type `T` when `T*` is implicitly convertible to `U*`;)
+$(LI `cast()` and `cast(`$(I qualifier list)`)` when applied to an lvalue.)
+)))
```
tgehr: Missing:
- Identifiers referencing local variables.
- Results of `ref` function calls.

spec/expression.dd
```diff
@@ -21,13 +21,44 @@ full expression of $(I expr$(SUBSCRIPT 1)). Otherwise, $(I expr) is its own full
...
+$(P Note: Although the definition is straightforward, a few subtleties exist related to function literals. In the
 statement `return (() => x + f())() * g();`, the full expression of `f()` is `x + f()`, not the expression passed
-to `return`. This is because the parent of `x + f()` has function literal type, not expression type.
-
+to `return`. This is because the parent of `x + f()` has function literal type, not expression type.)
+
+$(DDOC_ANCHOR define-lvalue)$(P $(B Definition) ($(DOUBLEQUOTE Lvalue)): The following expressions, and no
+others, are called lvalue expressions or lvalues:)
+$(OL
+$(LI `this` inside `struct` and `union` member functions;)
```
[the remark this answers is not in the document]
andralex: @RazvanN7 this doesn't compile:

```d
class C { void f() { return fun(this); } }
void fun(ref C) {}
```

spec/expression.dd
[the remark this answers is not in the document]
andralex: >Is it still an rvalue if the rightmost side is a function? Maybe you meant ref/property function?

I think that's covered by function calls.

>Isn't the leftmost side supposed to be an lvalue?

No, e.g. this works:

```d
struct S { int field; }
void fun(ref int) {}
void gun() { fun(S().field); }
```

spec/expression.dd
andralex: @tgehr yah, documenting existing stuff instead of aspiring to perfection

andralex: On second thought, "+ is a no-op" is easier to explain and understand than "+ is almost a no-op, you see, it copies the thing into an rvalue".

spec/expression.dd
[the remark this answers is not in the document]
RazvanN7: You mean this is also fine? 

```d
struct S { int field; }

void main()
{
    int *p = &S().field;   
}
```

Then we should close https://issues.dlang.org/show_bug.cgi?id=11920

spec/expression.dd
andralex: Thanks! Those got removed during some text reshuffling.

spec/expression.dd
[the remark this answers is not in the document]
andralex: @RazvanN7 that's not fine because the reference escapes and outlives the temporary expression.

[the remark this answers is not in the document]
andralex: I was wrong, the case of member function calls wasn't covered. Fixed now, thanks.

andralex:
>You can grep https://github.com/dlang/dmd/blob/master/src/dmd/expression.d for bool isLvalue() to get the dirty details. Some have a comment to explain the special cases.

cc @RazvanN7 :)

andralex:
Updated to account for all feedback.

spec/expression.dd
tgehr: No, this makes no sense. Note that soon, + will perform integer promotion, and then promoted lvalues will most likely be rvalues. Much easier to just deprecate using +x as an lvalue completely, as otherwise either your explanation or your justification of the behavior is bound to become a lot more complicated. There is no need to innovate with unary +.

spec/expression.dd
```diff
@@ -21,13 +21,47 @@ full expression of $(I expr$(SUBSCRIPT 1)). Otherwise, $(I expr) is its own full
...
+the argument(s) to `mixin` is an lvalue;)
+$(LI `cast(U)` expressions applied to lvalues of type `T` when `T*` is implicitly convertible to `U*`;)
+$(LI `cast()` and `cast(`$(I qualifier list)`)` when applied to an lvalue.)
+)))
+
+$(DDOC_ANCHOR define-rvalue)$(P $(B Definition) ($(DOUBLEQUOTE Rvalue)): Expressions that are not
+lvalues are rvalues.)
+
+$(P Note: Rvalues include all literals, special value keywords such as `__FILE__` and `__LINE__`,
+`enum` values, and the result of expressions not mentioned in the definition of lvalues above.)
```
WalterBright: s/mentioned in the definition of/defined as/

spec/expression.dd
```diff
@@ -21,13 +21,47 @@ full expression of $(I expr$(SUBSCRIPT 1)). Otherwise, $(I expr) is its own full
...
 statement `return (() => x + f())() * g();`, the full expression of `f()` is `x + f()`, not the expression passed
-to `return`. This is because the parent of `x + f()` has function literal type, not expression type.
-
+to `return`. This is because the parent of `x + f()` has function literal type, not expression type.)
+
+$(DDOC_ANCHOR define-lvalue)$(P $(B Definition) ($(DOUBLEQUOTE Lvalue)): The following expressions, and no
+others, are called lvalue expressions or lvalues:)
+$(OL
+$(LI `this` inside `struct` and `union` member functions;)
+$(LI a simple symbol or the result of the $(I DotIdentifier) grammatical construct `.` (left side may be
```
WalterBright: `simple symbol` is not defined. It's erroneous anyway as a "simple symbol" that is a manifest constant is not an lvalue. I suggest replacing it with `variable`.

spec/expression.dd
andralex: done

andralex:
@WalterBright OK thanks pushed

spec/expression.dd
```diff
@@ -21,13 +21,47 @@ full expression of $(I expr$(SUBSCRIPT 1)). Otherwise, $(I expr) is its own full
...
+$(LI the result of the following expressions:
+$(UL
+$(LI built-in unary operators `+` (when applied to an lvalue), `*`, `++` (prefix only), `--` (prefix only);)
+$(LI built-in indexing operator `[]` (but not the slicing operator);)
+$(LI built-in assignment binary operators, i.e. `=`, `+=`, `*=`, `/=`, `%=`, `&=`, `|=`, `^=`, `~=`,
+`<<=`, `>>=`, `>>>=`, and `^^=`;)
+$(LI the ternary operator $(I e) `?` $(I e$(SUBSCRIPT 1)) `:` $(I e$(SUBSCRIPT 2)) under the following
+circumstances:)
+$(OL
+    $(LI $(I e$(SUBSCRIPT 1)) and $(I e$(SUBSCRIPT 1)) are lvalues of the same type; OR)
```
TurkeyMan: `1` and `2`??

andralex: @TurkeyMan fixed straight in master, thanks

PR #2607, opened 2019-03-20
Function Default Parameters are underspecified

WalterBright:
Default arguments were incompletely specified.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -1349,9 +1349,39 @@ void foo(int x, int y = 3)
 foo(4);   // same as foo(4, 3);
 ---
 
-        $(P Default parameters are evaluated in the context of the
-        function declaration.
-        If the default value for a parameter is given, all following
+        $(P Default parameters are resolved and semantically checked in the context of the
+        function declaration.)
+---
+module m;
+private immutable int b;
+pure void g(int a=b){}
+---
+---
+import m;
+int b;
+pure void f()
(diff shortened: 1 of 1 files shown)


aG0aep6G:
The commit message shouldn't say that this fixes issue 19645. DMD has to implement the new spec before the issue can be closed.

spec/function.dd
```diff
@@ -1349,9 +1349,41 @@ void foo(int x, int y = 3)
 foo(4);   // same as foo(4, 3);
 ---
 
-        $(P Default parameters are evaluated in the context of the
+        $(P Default parameters are resolved and semantically checked in the context of the
         function declaration.
-        If the default value for a parameter is given, all following
+---
```
aG0aep6G: Close the paragraph before starting a code section. HTML doesn't like `<pre>` in `<p>`.

WalterBright: ok

thewilsonator:
@WalterBright you need to change the commit message, thats what the bot looks for.

aG0aep6G:
@andralex, you've merged it with the bad commit message. I've cleaned up after you and reopened the issue.

thewilsonator:
@aG0aep6G Thanks!

andralex:
@aG0aep6G sorry, I assumed it'd been fixed already. Thanks for the fix.

PR #2610, opened 2019-03-25
Anonymous struct and union, specifies that they are more like attributes

Basile-z:


spec/struct.dd
```diff
@@ -137,6 +137,29 @@ S* p;     // ok, knowledge of members is not necessary
         $(BEST_PRACTICE They can be used to implement the
         $(LINK2 https://en.wikipedia.org/wiki/Opaque_pointer, PIMPL idiom).)
 
+$(H2 $(LNAME2 anon_struct_unions, Anonymous Structs and Unions))
+
+    $(P Anonymous structs and unions are implemented as attributes, which means
+    that their inner declarations are not nested in a new parent.)
```
thewilsonator: Is attributes the correct term here? I'd write it as 

 Declaration is anonymous structs and unions are transparent, i.e. they do not an additional scope.

Basile-z: I think it's correct, it's like 

```d
struct S
{
  @(SpecialOverlapingOrAlignAttribute)
  {
      /*Constrained DeclsDefs*/
  }
}
```

The same word is employed in the compiler [as a code comment](https://github.com/dlang/dmd/blob/0b317677dab38776c12602c4f618d08cd85bded3/src/dmd/parse.d#L3368).

But this is no a very important specification, maybe let's wait for a 3rd opinion.

thewilsonator: Oh thats what you meant by it. I agree it does make sense, but only if you know that. I'm fine with that wording if that example is given.

thewilsonator:
Looks like you messed up a rebasing.

Basile-z:
> Looks like you messed up a rebasing.

Yeah but that's in a DMD pr

ZombineDev:
@Basile-z why close?

PR #2631, opened 2019-04-21
Advance deprecation of delete expressions, class allocators, class deallocators: Error

JinShil:
Documentation update to accompany https://github.com/dlang/dmd/pull/9666


JinShil:
> What’s the change

If you look at https://dlang.org/deprecate.html you'll see the "Error" column is blank.  After this is pulled, it will be populated with the DMD version number in which dlang/dmd#9666 appears.  Right now I'm trying for 2.086, but I may have to update this PR to 2.087 if I can't get it done on time.

PR #2632, opened 2019-04-24
Add AliasDeclarationY rule for the function types

Basile-z:
This is the documentation part of https://github.com/dlang/dmd/pull/9667.

### DIFF
--- a/spec/declaration.dd
+++ b/spec/declaration.dd
@@ -348,6 +348,7 @@ $(GNAME AliasDeclarationX):
 $(GNAME AliasDeclarationY):
     $(GLINK_LEX Identifier) $(GLINK2 template, TemplateParameters)$(OPT) $(D =) $(GLINK StorageClasses)$(OPT) $(GLINK Type)
     $(GLINK_LEX Identifier) $(GLINK2 template, TemplateParameters)$(OPT) $(D =) $(GLINK2 expression, FunctionLiteral)
+    $(GLINK_LEX Identifier) $(GLINK2 template, TemplateParameters)$(OPT) $(D =) $(GLINK StorageClasses)$(OPT) $(GLINK BasicType) $(GLINK2 function, Parameters) $(GLINK2 function, MemberFunctionAttributes)$(OPT)
 )
 
     $(P $(I AliasDeclaration)s create a symbol that is an alias for another type,
@@ -398,6 +399,15 @@ t1.t v1;  // v1 is type int
 t2 v2;    // v2 is type int
 t3 v3;    // v3 is type int
 t4 v4;    // v4 is type int
+
+alias Fun = int(string p);
+int fun(string){return 0;}
+static assert(is(typeof(fun) == Fun));
+
+alias MemberFun1 = int() const;
+alias MemberFun2 = const int();
+// leading attributes apply to the func, not the return type
+static assert(is(MemberFun1 == MemberFun2));
 --------------------
 
         $(P


spec/declaration.dd
```diff
@@ -393,11 +394,15 @@ alias t1 = Foo2!(int);
 alias t2 = Foo2!(int).t;
 alias t3 = t1.t;
 alias t4 = t2;
+alias Fun = int(string p);
```
jacob-carlborg: Perhaps an example with function attributes as well.

Basile-z: done

spec/declaration.dd
```diff
@@ -398,6 +399,14 @@ t1.t v1;  // v1 is type int
...
 t3 v3;    // v3 is type int
 t4 v4;    // v4 is type int
+
+alias Fun = int(string p);
+int fun(string){return 0;}
+static assert(is(typeof(fun) == Fun));
+
+alias MemberFun1 = void() const;
+alias MemberFun2 = const void(); // leading storage classes are not part of the type
+static assert(is(MemberFun1 == MemberFun2));
```
jacob-carlborg: If the leading storage class is not part of the type, how can this be true?

Basile-z: It's just like when you declare a function. You can but it's recommended to put them before the body ;).

spec/declaration.dd
```diff
@@ -348,6 +348,7 @@ $(GNAME AliasDeclarationX):
 $(GNAME AliasDeclarationY):
     $(GLINK_LEX Identifier) $(GLINK2 template, TemplateParameters)$(OPT) $(D =) $(GLINK StorageClasses)$(OPT) $(GLINK Type)
     $(GLINK_LEX Identifier) $(GLINK2 template, TemplateParameters)$(OPT) $(D =) $(GLINK2 expression, FunctionLiteral)
+    $(GLINK_LEX Identifier) $(GLINK2 template, TemplateParameters)$(OPT) $(D =) $(GLINK StorageClasses)$(OPT) $(GLINK BasicType) $(GLINK Parameters) $(GLINK MemberFunctionAttributes)$(OPT)
```
ZombineDev: The `Parameters` link doesn't work as that part of the grammar is not defined in the current page. Try the following: `$(GLINK2 function, Parameters)`. Ditto for `MemberFunctionAttributes`.

spec/declaration.dd
```diff
@@ -398,6 +399,14 @@ t1.t v1;  // v1 is type int
 t2 v2;    // v2 is type int
 t3 v3;    // v3 is type int
 t4 v4;    // v4 is type int
+
```
ZombineDev: Perhaps out of scope for this PR, but I noticed that even though one can infer from the grammar that aliasing function and delegate pointer types is supported, there's no example of that and I think this paragraph may be a good place to add this. Ditto for function literals, which are in general a less known feature.

spec/declaration.dd
ZombineDev: Perhaps it would be a bit more clear if we say: "leading storage classes are interpreted as member function attributes", unless I'm missing something.

spec/declaration.dd
Basile-z: Someone can make another PR, i'll open an issue later to get track of this

Basile-z: https://issues.dlang.org/show_bug.cgi?id=19824

spec/declaration.dd
Basile-z: Yeah thanks, strange that the CI doesn't catch invalid links BTW.

WalterBright:
@ZombineDev language changes need to be approved by myself before they are merged.

ZombineDev:
@WalterBright apologies, I did not mean to overrule your decision in any way. I was operating under the assumption that you were already ok with the language change, since you did not raise any issues regarding it on your initial review of the dmd pull request, but only a comment regarding the implementation.

PR #2635, opened 2019-04-29
Introduce term qualifier-convertible

andralex:
Makes other topics easier to discuss (such as matching rules for function overloading).

### DIFF
--- a/spec/const3.dd
+++ b/spec/const3.dd
@@ -468,8 +468,10 @@ $(LATEX
 )
 
 $(P In the graph above, any directed path is a legal implicit conversion. No
-other qualifier combinations than the ones shown is valid. The same information
-is shown below in tabular format:)
+other qualifier combinations than the ones shown is valid. If a directed path
+exists between two sets of qualifiers, the types thus qualified are called
+$(GLOSSARY qualifier-convertible). The same information is shown below in tabular
+format:)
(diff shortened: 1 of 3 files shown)


thewilsonator:
```
./spec/function.dd:1075:        $(LI match with qualifier conversion (if the argument type is 
```

Seriously, can we get this to apply itself already?

spec/const3.dd
```diff
@@ -468,8 +468,10 @@ $(LATEX
 )
 
 $(P In the graph above, any directed path is a legal implicit conversion. No
-other qualifier combinations than the ones shown is valid. The same information
-is shown below in tabular format:)
+other qualifier combinations than the ones shown is valid. If a directed path
+exists between two sets of qualifiers, the types thus qualified are called
```
thewilsonator: What is meant by "a directed path [...] between two sets of qualifiers"? I would just use what you have written in the glossary instead.

wilzbach:
> Seriously, can we get this to apply itself already?

Well, no one is working on dlang-bot except occasionally me and Martin.

spec/const3.dd
andralex: There is a nice DAG on that page, sadly it does not have a number or a moniker. I'll look into it.

PR #2639, opened 2019-05-10
Add LDC to list of snap packages available to download

WebDrake:
As presented in my DConf 2019 lightning talk, there is a big discrepancy in usage between the DMD and DUB snap packages, and LDC.  One might well assume that this is because the first two are listed on the downoad page but LDC is not.

This patch corrects the discrepancy, and splits out the install commands into separate lines with helpful comments.


WebDrake:
This is broken (only the first line now appears).  I'll fix.

WebDrake:
OK, this now works AFAICS.  The problem was a `,` that was interpreted as a ddoc parameter separator.  I forget what the escape sequence is for a comma, but I changed the expression to avoid using it.

PR #2641, opened 2019-05-14
Update introduction to D language

dibyendumajumdar:
Revised the text describing the D language to reflect that D supports both manual and automatic memory management, as D code can be compiled and linked to form new libraries too.

### DIFF
--- a/spec/intro.dd
+++ b/spec/intro.dd
@@ -8,7 +8,7 @@ D programs are structured as modules that can be compiled separately and linked 
 to create native libraries or executables.)
 
 $(P This document is the reference manual for the D Programming Language. For more information and 
-other documents, see [dlang.org](https://dlang.org/).)
+other documents, see $(LINK2 https://dlang.org/, The D Language Website).)
 
 $(H2 Phases of Compilation)
 
--- a/spec/intro.dd
+++ b/spec/intro.dd
@@ -8,7 +8,7 @@ D programs are structured as modules that can be compiled separately and linked 
 to create native libraries or executables.)
 
 $(P This document is the reference manual for the D Programming Language. For more information and 
-other documents, see dlang.org.)
+other documents, see [dlang.org](https://dlang.org/).)
 
 $(H2 Phases of Compilation)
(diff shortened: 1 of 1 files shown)


spec/intro.dd
```diff
@@ -2,11 +2,13 @@ Ddoc
...
-        to native code that is combined with libraries and compiled C code by a linker to create a
-        native executable.
-        )
+$(P D is a general-purpose systems programming language with a C-like syntax that compiles to native code. 
+It is strongly typed and supports both automatic (garbage collected) and manual memory management. 
+D programs are structured as modules that can be compiled separately and linked with external libraries 
+to create native libraries or executables.)
+
+$(P This document is the reference manual for the D Programming Language. For more information and 
+other documents, see dlang.org.)
```
wilzbach: This text is also used to render the PDF and mobile book, so all links must use Ddoc macros. Though in this case I would simply skip the last sentence as it will be weird on paper?

dibyendumajumdar: I wasn't sure of the syntax for a link. I have updated it - is that the right syntax? I will have a look at the PDF output to see what this comes out as, but I would expect that PDFs should also show the link.

spec/intro.dd
```diff
@@ -2,11 +2,13 @@ Ddoc
...
-        to native code that is combined with libraries and compiled C code by a linker to create a
-        native executable.
-        )
+$(P D is a general-purpose systems programming language with a C-like syntax that compiles to native code. 
+It is strongly typed and supports both automatic (garbage collected) and manual memory management. 
+D programs are structured as modules that can be compiled separately and linked with external libraries 
+to create native libraries or executables.)
+
+$(P This document is the reference manual for the D Programming Language. For more information and 
+other documents, see [dlang.org](https://dlang.org/).)
```
wilzbach: You can not use markdown here yet. You need to use Ddoc macros like LINK2

dibyendumajumdar: Okay thanks - updated.

PR #2642, opened 2019-05-14
Add a paragraph about default construction

RazvanN7:
cc @wilzbach

### DIFF
--- a/spec/struct.dd
+++ b/spec/struct.dd
@@ -333,7 +333,7 @@ $(H2 $(LEGACY_LNAME2 StructLiteral, struct-literal, Struct Literals))
         struct literal, and all subsequent non-overlapping fields are default
         initialized.
         )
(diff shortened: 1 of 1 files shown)


RazvanN7:
Yep, I had an error. Retag please.

thewilsonator:
Done.

RazvanN7:
Thanks!

jmdavis:
Wait. This is wrong. Structs don't _have_ default constructors in D. They have an `init` value. And reading this PR, I don't think that you even meant default constructor. A default constructor is a constructor with no arguments that gets called when an object is constructed with no arguments, whereas I _think_ that what is being talked about here in the constructor that you automatically get for structs in D when you don't declare any. And even that isn't an actual constructor. IIRC no constructors show up in `allMembers` if you don't declare any. It's basically just using the constructor syntax for struct literals.

PR #2644, opened 2019-05-15
Deprecate scope as a type constraint on class declarations

JinShil:
Update to the deprecations table for https://github.com/dlang/dmd/pull/9799


JinShil:
ping DMD PR was merged.

PR #2646, opened 2019-05-17
Added a Conceptual models section. 

dibyendumajumdar:
Some initial content that includes Environments, Program structure and a revised version of the Phases
of Compilation renamed to Translation Phases.

### DIFF
--- a/spec/intro.dd
+++ b/spec/intro.dd
@@ -10,59 +10,51 @@ to create native libraries or executables.)
 $(P This document is the reference manual for the D Programming Language. For more information and
 other documents$(COMMA) see $(LINK2 https://dlang.org/, The D Language Website).)
 
-$(H2 Phases of Compilation)
+$(H2 Conceptual models)
 
-$(P The process of compiling is divided into multiple phases. Each phase has no
-dependence on subsequent phases. For example, the scanner is not perturbed by
-the semantic analyzer. This separation of the passes makes language tools like
-syntax directed editors relatively easy to produce. It also is possible to
-compress D source by storing it in $(SINGLEQUOTE tokenized) form.)
+$(H3 Environments)
+
+$(P The data-processing-system in which a D program is translated from source to executable format
+will be called the translation environment and the data-processing-system in which the translated
+D program is executed will be called the execution environment. The translation environment and
+the execution environment need not have the same system architecture.)
+
+$(H3 Program Structure)
+
+$(P A D program is kept in units called source files. Each source file contains a unit of translation
+called the module. D modules need not all be translated at the same time. Previously translated
+D modules may be individually preserved or in libraries. The separate modules of a program communicate
+by calls to functions whose identifers have external linkage, manipulation of objects whose identifers
+have external linkage, or manipulation of data files. Modules may be separately translated and then
+later linked to produce an executable program.)
+
+$(H3 Translation Phases)
+
+$(P D programs are translated in multiple phases.)
 
 $(OL
-        $(LI $(B source character set)$(BR)
 
-        The source file is checked to see what character set it is,
-        and the appropriate scanner is loaded. ASCII and UTF
-        formats are accepted.
+        $(LI $(B Lexical analysis)$(BR)
+
+        The source file is translated into a sequence of tokens.
         )
 
-        $(LI $(B script line) $(BR)
-
-        If the first line starts with "$(HASH)!", then that line
-        is ignored.
-        )
-
-        $(LI $(B lexical analysis)$(BR)
-
-        The source file is divided up into a sequence of tokens.
-        $(DDSUBLINK spec/lex, specialtokens, Special tokens)
-        are replaced with other tokens.
-        $(GLINK_LEX SpecialTokenSequence)s
-        are processed and removed.
-        )
-
-        $(LI $(B syntax analysis)$(BR)
+        $(LI $(B Syntax analysis)$(BR)
 
         The sequence of tokens is parsed to form syntax trees.
         )
 
-        $(LI $(B semantic analysis)$(BR)
+        $(LI $(B Semantic analysis)$(BR)
 
         The syntax trees are traversed to declare variables, load symbol tables, assign
         types, and in general determine the meaning of the program.
         )
 
-        $(LI $(B optimization)$(BR)
+        $(LI $(B Code generation)$(BR)
 
-        Optimization is an optional pass that tries to rewrite the program
-        in a semantically equivalent, but faster executing, version.
-        )
-
-        $(LI $(B code generation)$(BR)
-
-        Instructions are selected from the target architecture to implement
-        the semantics of the program. The typical result will be
-        an object file, suitable for input to a linker.
+        Following optional optimizations, instructions targetted for the execution
+        environment are emitted to implement the semantics of the D program.
+        The result is an object file suitable for input to a linker.
         )
 )
 
--- a/spec/intro.dd
+++ b/spec/intro.dd
@@ -10,33 +10,30 @@ to create native libraries or executables.)
 $(P This document is the reference manual for the D Programming Language. For more information and
 other documents$(COMMA) see $(LINK2 https://dlang.org/, The D Language Website).)
 
-$(H2 Conceptual models)
+$(H2 Environments)
 
-$(H3 Environments)
-
-$(P The data-processing-system in which a D program is translated from source to executable format
-will be called the translation environment and the data-processing-system in which the translated
-D program is executed will be called the execution environment. The translation environment and
+$(P The computer system in which a D program is compiled from source to executable format
+will be called the compilation environment and the computer system in which the compiled
+D program is executed will be called the execution environment. The compilation environment and
 the execution environment need not have the same system architecture.)
 
-$(H3 Program Structure)
+$(H2 Program Structure)
 
-$(P A D program is kept in units called source files. Each source file contains a unit of translation
-called the module. D modules need not all be translated at the same time. Previously translated
-D modules may be individually preserved or in libraries. The separate modules of a program communicate
-by calls to functions whose identifers have external linkage, manipulation of objects whose identifers
-have external linkage, or manipulation of data files. Modules may be separately translated and then
-later linked to produce an executable program.)
+$(P A D program is kept in units called source files. Each source file contains a unit of compilation
+called the module. D modules need not all be compiled at the same time. Previously compiled
+D modules may be individually preserved as object files or in libraries. Modules may be separately
+compiled and then later linked to produce an executable program.)
 
-$(H3 Translation Phases)
+$(H2 Phases of Compilation)
 
-$(P D programs are translated in multiple phases.)
+$(P D programs are compiled in multiple phases.)
 
 $(OL
 
         $(LI $(B Lexical analysis)$(BR)
 
-        The source file is translated into a sequence of tokens.
+        The source file is converted into a sequence of tokens. In particular, D programs
+        are not passed through a macro-preprocessor prior to compilation.
         )
 
         $(LI $(B Syntax analysis)$(BR)
@@ -53,8 +50,8 @@ $(OL
 
         $(LI $(B Optimization)$(BR)
 
-        Optimization is an optional phase during which the D program is translated into a
-        semantically equivalent, but more time or space efficient version.
+        Optimization is an optional phase during which the D program is converted into a
+        semantically equivalent, but more time and/or space efficient version.
         )
 
         $(LI $(B Code generation)$(BR)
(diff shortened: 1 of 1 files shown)


spec/intro.dd
```diff
@@ -8,59 +8,51 @@ $(SPEC_S Introduction,
...
-        in a semantically equivalent, but faster executing, version.
-        )
-
-        $(LI $(B code generation)$(BR)
+        $(LI $(B Code generation)$(BR)
 
-        Instructions are selected from the target architecture to implement
-        the semantics of the program. The typical result will be
-        an object file, suitable for input to a linker.
+        Following optional optimizations, instructions targetted for the execution
```
thewilsonator: Remove `optimal`.

spec/intro.dd
```diff
@@ -8,59 +8,51 @@ $(SPEC_S Introduction,
...
-        Optimization is an optional pass that tries to rewrite the program
-        in a semantically equivalent, but faster executing, version.
-        )
-
-        $(LI $(B code generation)$(BR)
+        $(LI $(B Code generation)$(BR)
 
-        Instructions are selected from the target architecture to implement
-        the semantics of the program. The typical result will be
-        an object file, suitable for input to a linker.
```
thewilsonator: The original is correct, you can make the compiler do other stuff, e.g. output assembly, LLVM textual IR / byte code, GPU kernels...

spec/intro.dd
```diff
@@ -8,59 +8,51 @@ $(SPEC_S Introduction,
...
         )
 
-        $(LI $(B semantic analysis)$(BR)
+        $(LI $(B Semantic analysis)$(BR)
 
         The syntax trees are traversed to declare variables, load symbol tables, assign
         types, and in general determine the meaning of the program.
         )
 
-        $(LI $(B optimization)$(BR)
```
thewilsonator: Please leave this section.

spec/intro.dd
```diff
@@ -8,59 +8,51 @@ $(SPEC_S Introduction,
...
+called the module. D modules need not all be translated at the same time. Previously translated
+D modules may be individually preserved or in libraries. The separate modules of a program communicate
+by calls to functions whose identifers have external linkage, manipulation of objects whose identifers
+have external linkage, or manipulation of data files. Modules may be separately translated and then
+later linked to produce an executable program.)
 
-        If the first line starts with "$(HASH)!", then that line
-        is ignored.
-        )
+$(H3 Translation Phases)
```
thewilsonator: Compilation phases

spec/intro.dd
```diff
@@ -8,59 +8,51 @@ $(SPEC_S Introduction,
...
 
-$(P The process of compiling is divided into multiple phases. Each phase has no
-dependence on subsequent phases. For example, the scanner is not perturbed by
-the semantic analyzer. This separation of the passes makes language tools like
-syntax directed editors relatively easy to produce. It also is possible to
-compress D source by storing it in $(SINGLEQUOTE tokenized) form.)
+$(H3 Environments)
 
-$(OL
-        $(LI $(B source character set)$(BR)
```
thewilsonator: Please keep this section, it probably doesn't belong in Compilation phases though

spec/intro.dd
```diff
@@ -8,59 +8,51 @@ $(SPEC_S Introduction,
...
+D program is executed will be called the execution environment. The translation environment and
+the execution environment need not have the same system architecture.)
 
-        The source file is checked to see what character set it is,
-        and the appropriate scanner is loaded. ASCII and UTF
-        formats are accepted.
-        )
+$(H3 Program Structure)
 
-        $(LI $(B script line) $(BR)
```
thewilsonator: Ditto

spec/intro.dd
dibyendumajumdar: Hi, did you mean `optional`? I put that in as presumably optimization is an optional step. The original text stated `Optimization is an optional pass`.

thewilsonator: Whoops, yes,

spec/intro.dd
dibyendumajumdar: Do you mean I should retain `typical`?

spec/intro.dd
dibyendumajumdar: I removed it because I felt that this description might only apply to DMD but not as clearly to the LLVM or GCC backends because as far as I know, in LLVM optimizations occur at various level even during instruction selection. So I was trying to avoid specifying how the DMD implementation works.

spec/intro.dd
thewilsonator: Yes.

spec/intro.dd
dibyendumajumdar: I renamed this to Translation Phases to keep it consistent with the terminology used in the C standard. The previous section I added uses the term 'translation' as it is used by the C standard.

spec/intro.dd
dibyendumajumdar: The reason I removed this point and the next is that they appear to dive into some details of lexical analysis. The lexical analysis process is described fully in a following section so picking a couple of the details and mentioning them here seems odd.

spec/intro.dd
thewilsonator: Hmm, DMD has its optimisation passes in weird places but the paragraph is vague enough to be technically correct.

spec/intro.dd
dibyendumajumdar: As above.

spec/intro.dd
dibyendumajumdar: Ok

spec/intro.dd
thewilsonator: Thats probably because the C standard calls the unit of compilation a "translation unit". I don't know what we call them, if anything.

spec/intro.dd
dibyendumajumdar: Ok

spec/intro.dd
thewilsonator: Hmm, OK.

spec/intro.dd
dibyendumajumdar: I used this to describe it: `A D program is kept in units called source files. Each source file contains a unit of translation called the module. D modules need not all be translated at the same time.`

ZombineDev:
@andralex @WalterBright you may want to have a look.

thewilsonator:
Its generally much more effective to email them. Particularly for the dlang.org repo

spec/intro.dd
```diff
@@ -10,59 +10,58 @@ to create native libraries or executables.)
 $(P This document is the reference manual for the D Programming Language. For more information and
 other documents$(COMMA) see $(LINK2 https://dlang.org/, The D Language Website).)
 
-$(H2 Phases of Compilation)
+$(H2 Conceptual models)
```
aliak00: What is conceptual models? Can we put a description under here of what is meant by conceptual models?

spec/intro.dd
```diff
@@ -10,59 +10,58 @@ to create native libraries or executables.)
...
-syntax directed editors relatively easy to produce. It also is possible to
-compress D source by storing it in $(SINGLEQUOTE tokenized) form.)
+$(H3 Environments)
 
-$(OL
-        $(LI $(B source character set)$(BR)
+$(P The data-processing-system in which a D program is translated from source to executable format
+will be called the translation environment and the data-processing-system in which the translated
+D program is executed will be called the execution environment. The translation environment and
+the execution environment need not have the same system architecture.)
```
aliak00: I think this paragraph is playing with words a bit too much and it seems the only purpose is to have a heading called "environments"? The first data-processing-system *is* the compiler, so just call it that and hoist this paragraph in to two sections: Compiler, and Execution environment (also makes it much more obvious IMO). And then if you must, throw in linker in the compiler to denote the part that create the exe, but maybe that's too much detail for now.

spec/intro.dd
aliak00: Also code generation and optimization can span multiple modules so maybe compilation phases is better? 🤷‍♂

spec/intro.dd
```diff
@@ -10,59 +10,58 @@ to create native libraries or executables.)
...
-        )
+$(H3 Program Structure)
 
-        $(LI $(B script line) $(BR)
+$(P A D program is kept in units called source files. Each source file contains a unit of translation
+called the module. D modules need not all be translated at the same time. Previously translated
+D modules may be individually preserved or in libraries. The separate modules of a program communicate
+by calls to functions whose identifers have external linkage, manipulation of objects whose identifers
+have external linkage, or manipulation of data files. Modules may be separately translated and then
+later linked to produce an executable program.)
```
aliak00: - s/the module/a module
- No need to say "D modules" every time. It's obvious.
- What does "individually preserved" mean?
- s/The separate modules of a program communicate by calls/Modules communicate via
- What is manipulation of data files?

spec/intro.dd
aliak00: Also, maybe this should be an H2? I'm not sure how this is a conceptual model

spec/intro.dd
```diff
@@ -10,59 +10,58 @@ to create native libraries or executables.)
...
-the semantic analyzer. This separation of the passes makes language tools like
-syntax directed editors relatively easy to produce. It also is possible to
-compress D source by storing it in $(SINGLEQUOTE tokenized) form.)
+$(H3 Environments)
 
-$(OL
-        $(LI $(B source character set)$(BR)
+$(P The data-processing-system in which a D program is translated from source to executable format
+will be called the translation environment and the data-processing-system in which the translated
+D program is executed will be called the execution environment. The translation environment and
```
joseph-wakeling-frequenz: I would suggest highlighting terms like `translation environment` and `execution environment` when first defining them, so they stand out as definitions.  Might even be worth explicitly defining a macro to highlight such definitions.

andralex:
+ @atilaneves

spec/intro.dd
atilaneves: Why would consistency with the C standard be a goal?

spec/intro.dd
atilaneves: What are conceptual models?

atilaneves:
I'm unsure of what the goal of this PR is.

spec/intro.dd
dibyendumajumdar: It isn't. It just happens that in this instance it was useful to borrow some terminology.

spec/intro.dd
dibyendumajumdar: The idea is that the certain aspects are being introduced conceptually. This PR covers the process of translation, but I would like to add memory model, object model and execution model in follow up PRs.

Basile-z: Yes this is abstruse. The old title made sense.

spec/intro.dd
```diff
@@ -10,59 +10,58 @@ to create native libraries or executables.)
...
+the execution environment need not have the same system architecture.)
 
-        The source file is checked to see what character set it is,
-        and the appropriate scanner is loaded. ASCII and UTF
-        formats are accepted.
-        )
+$(H3 Program Structure)
 
-        $(LI $(B script line) $(BR)
+$(P A D program is kept in units called source files. Each source file contains a unit of translation
```
Basile-z: > Each source file contains a unit of translation

What ? I'm sorry but again i see some abstruse writing.

dibyendumajumdar:
> 
> 
> I'm unsure of what the goal of this PR is.

The goal is to revise the description from being implementation specific to a more general description which enables us to introduce other aspects of the conceptual model. The existing introduction jumps into compilation phases. Why? Why not mention other aspects?

spec/intro.dd
dibyendumajumdar: Perhaps you are referring to the wrong title? It might be better to read the text rather than looking at the diff.

Basile-z: Indeed sorry. But that doesn't change the general feeling I have about your initiative. I'm afraid that the direction taken could make the documentation "profane resistant", by introducing abstruse technical terms. But your lucky I'm not a desion maker ;)

dibyendumajumdar:
Reviewers: please see my post a few days back: https://forum.dlang.org/post/ingeeweghupmnidpqljo@forum.dlang.org

Above gives you an idea of what I would like to achieve. This PR should be read in the context of that - it is the first PR in the series.

spec/intro.dd
dibyendumajumdar: Removed the heading Conceptual Models as it seems to have caused confusion.

spec/intro.dd
dibyendumajumdar: Simplified.

spec/intro.dd
dibyendumajumdar: Revised, please have a look.

PR #2647, opened 2019-05-18
Revert "Add a paragraph about default construction"

ZombineDev:
Reverts dlang/dlang.org#2642

See: https://github.com/dlang/dlang.org/pull/2642#issuecomment-492585252


ZombineDev:
What is disabled is the use of the "struct literal" syntax. Default constructor is the constructor with zero parameters. It can only be disabled by the user and is not something generated by the compiler as it can't have a body.

ZombineDev:
@RazvanN7 It would be better if I directly corrected the paragraph, but I don't have time at the moment. Please follow through with a corrected one later.

PR #2648, opened 2019-05-24
Fix typo in code

Bolpat:
> For backward compatibility [...] the rewrites `a.opSliceUnary!(op)(a, i, j)` [...] are tried instead, respectively.

To:

> For backward compatibility [...] the rewrites `a.opSliceUnary!(op)(i, j)` [note: without `a` in parameter list] [...] are tried instead, respectively.

The compiler does not do the former rewrite, and looks like a small error.

### DIFF
--- a/spec/operatoroverloading.dd
+++ b/spec/operatoroverloading.dd
@@ -199,7 +199,7 @@ $(H3 $(LNAME2 slice_unary_operators, Overloading Slice Unary Operators))
 
         $(P For backward compatibility, if the above rewrites fail and
         $(D opSliceUnary) is defined, then the rewrites
-        $(D a.opSliceUnary!(op)(a, i, j)) and
+        $(D a.opSliceUnary!(op)(i, j)) and
         $(D a.opSliceUnary!(op)) are tried instead, respectively.)
 
 $(H2 $(LEGACY_LNAME2 Cast, cast, Cast Operator Overloading))
--- a/spec/operatoroverloading.dd
+++ b/spec/operatoroverloading.dd
@@ -197,7 +197,7 @@ $(H3 $(LNAME2 slice_unary_operators, Overloading Slice Unary Operators))
         )
         )
 
-        $(P For backward compatibility, if the above rewrites fail and
+        $(P For backward compatibility, if the above rewrites fail to compile and
         $(D opSliceUnary) is defined, then the rewrites
         $(D a.opSliceUnary!(op)(i, j)) and
         $(D a.opSliceUnary!(op)) are tried instead, respectively.)


Bolpat:
How do I make the DAutoTest restart?

jacob-carlborg:
Run `git commit --amend` and force push the branch.

Bolpat:
I just don't get why the AutoTest fails. The changes does almost nothing, and nothing code related at all.

thewilsonator:
IDK either,  failing all else (which I guess you're at already) try randomly replacing `,`s with `$(COMMA)`.

thewilsonator:
I got nothin'.

CyberShadow:
Argh, PDF errors.

```
<recently read> \egroup 
                        
l.21144 }
         
? 
! Emergency stop.
<recently read> \egroup 
                        
l.21144 }
```

@andralex Any idea what's happening here?

andralex:
I'm stomped. There's no change in paragraph layout or anything, is there?

CyberShadow:
Here is the diff for the generated `.tex` files:

```diff
--- .generated_/dlangspec.tex	2019-07-05 00:50:15.683974308 +0000
+++ .generated/dlangspec.tex	2019-07-05 00:55:15.366033326 +0000
@@ -20867,7 +20867,7 @@
 \par
         For backward compatibility, if the above rewrites fail and
         \lstinline|opSliceUnary| is defined, then the rewrites
-        \lstinline|a.opSliceUnary!(op)(a, i, j)| and
+        \lstinline|a.opSliceUnary!(op)(i, j)| and
         \lstinline|a.opSliceUnary!(op)| are tried instead, respectively. \par
 \par
 \section{\hyperref[cast]{Cast Operator Overloading}}
```

lesderid:
Might be this? https://tex.stackexchange.com/a/229118/112633 (second part of answer)

CyberShadow:
Possibly, any idea how to apply that fix to our document?

lesderid:
(Hacky) fix might be to use `$TT(...)` instead of `$D(...)`?

CyberShadow:
I had a go converting `dlangspec.html` (generated for the eBook) to PDF using Weasyprint. Here is the result:

https://dump.thecybershadow.net/0734923ba355bf8f8ed27b5a07772a24/dlangspec-weasyprint.pdf

It looks decent, though small some things need fixing (like the TOC links).

@andralex Any thoughts on replacing the TeX-based PDF generation process with Weasyprint?

(I also tried wkhtmltopdf, but that one segfaults on the HTML input.)

lesderid:
@CyberShadow Unless I'm missing a font, inline code (`$D(...)`) is not using a monospace font, among other formatting issues making it look not quite right.

CyberShadow:
Right. Fixed version with the stylesheet loaded: https://dump.thecybershadow.net/60b7d7a3c800ccc102236d96e86661ec/dlangspec-weasyprint.pdf

andralex:
Gave it a close look. I inserted `\end{document}` after the change here and there in the document to see where it breaks.

This is a freak occurrence - the change shrinks the respective paragraph from three lines to two. In turn, due to some sort of a butterfly effect a major reflow is caused. The change is on page 417. Five pages down (page 422), there's an example with `struct F` and `void test()`. That example gets moved by the reflow all the way back to page 420, and spans a page break. That's where the problem is.

The listings package, which we use for code formatting, has a hindenbug that manifests itself sometimes when D code spans a page boundary. It has been a source of occasional annoyance with TDPL, and a source of occasional annoyance with dlang.org.

As a workaround, if the words "to compile" are added to the text after "fails" in that paragraph, the paragraph stays three lines and the major reformatting is not triggered.

As a long-term solution, moving away from LaTeX generation is a possibility. There's an even better HTML to PDF generator that @acehreli used for his D book, and produces beautiful print. From a maintenance perspective it seems a step backward though - LaTeX generation is the only major "non-html" backend we have, and helped us fix numerous bugs both in the ddoc engine and the document itself. I fear that retiring that would have us slowly reintroduce that kind of bugs.

A better solution would be to renounce the listings package - its main help is code coloring, but ddoc already has that capability (we currently mute the ddoc syntax coloring macros in LaTeX generation and rely on listings to do the coloring). Going with some simple `\color` and `\tt` macros seems like a good way to get rid of listings.

andralex:
BTW I never managed to reduce the bug to a simple enough variant (1-2 pages) to submit it. If anyone could do that, I think the listings package users would be helped.

CyberShadow:
I agree that it would be sad to give up on LaTeX, it's a good demonstration of what's possible with DDoc. My biggest concern with sticking to it is the bus factor, it doesn't seem like we have a lot of experts in this domain in our community, whereas HTML knowledge is very common. In any case, it's not an issue as long as you remain available to resolve such problems :) I would need to do a lot of research to be able to even begin to understand them.

PR #2649, opened 2019-05-25
add Memory Model and Object Model

WalterBright:
Not much but it's a start.

### DIFF
--- a/spec/intro.dd
+++ b/spec/intro.dd
@@ -67,6 +67,78 @@ $(OL
 )
 
 
+$(H2 Memory Model)
+
+    $(P A $(I byte) is the fundamental unit of storage. Each byte has 8 bits. Each byte has
+    a unique address. A $(I memory location) is a sequence of one or more bytes exactly large
+    enough to hold a scalar type. Multiple threads can access separate memory locations
+    without interference.
+    )
+
+    $(P Memory locations come in three groups:)
+
+    $(OL
+    $(LI $(I Thread local memory locations) are accessible from only one thread at a time.)
+    $(LI $(I Immutable memory locations) cannot be written to during its lifetime. Immutable
+    memory locations can be read from by multiple threads without synchronization.)
+    $(LI $(I Shared memory locations) are accessible from multiple threads.)
+    )
+
+    $(UNDEFINED_BEHAVIOR Allowing multiple threads to access a thread local memory
+    location results in undefined behavior.)
+
+    $(UNDEFINED_BEHAVIOR Writing to an immutable memory location during its lifetime
+    results in undefined behavior.)
+
+    $(UNDEFINED_BEHAVIOR Unless synchronized, writing to a shared memory location in
+    one thread while other threads read or write to it results in undefined behavior.)
+
+    $(P Execution of a single thread on thread local and immutable memory locations
+    is $(I sequentially consistent). This means the result of the operations
+    is the same as if it were executed in the order the operations appear in the program.
+    )
+
+    $(P A memory location can be transferred from thread local to immutable or shared
+    if there is only one reference to the location.)
+
+    $(P A memory location can be transferred from shared to immutable or thread local
+    if there is only one reference to the location.)
+
+    $(P A memory location can be temporarily transferred from shared to local if
+    synchronization is used to prevent any other threads from accessing the memory
+    location during the duration.)
+
+
+$(H2 Object Model)
+
+    $(P An $(I object) is created when one of the following is executed:
+    )
+
+    $(UL
+    $(LI a definition)
+    $(LI a $(GLINK2 expression, NewExpression))
+    $(LI a temporary is created)
+    $(LI changing which field of a union is active)
+    )
+
+    $(P An object spans a sequence of memory locations which may or may not
+    be contiguous. It exists during construction, its lifetime, and destruction.
+    Each object has a type, which is determined either statically or by runtime
+    type information.
+    The memory locations may include any combination of thread local, immutable, or
+    shared memory locations.
+    )
+
+    $(P Objects can be composed into a $(I composed object). The objects that make up
+    the composed object are $(I subobjects). An object that is not the subobject
+    of another object is a $(I complete object). The lifetime of a subobject
+    is always within the lifetime of its complete object.
+    )
+
+    $(P An object's address is the address of the first byte of the first memory
+    location for that object. Object addresses are distinct from each other
+    unless one of the objects is nested within the other.
+    )
 
 $(SPEC_SUBNAV_NEXT lex, Lexical)
 )


spec/intro.dd
```diff
@@ -67,6 +67,43 @@ $(OL
 )
 
 
+$(H2 Memory Model)
+
+    $(P A $(I byte) is the fundamental unit of storage. Each byte has 8 bits. Each byte has
```
atilaneves: This would exclude certain platforms.

thewilsonator:
Can I suggest this be a separate document? Having this in the intro seems akin to the rust folks putting object lifetime front and centre that Andrei talked about at dconf. This is important, but not really in a user-facing kind if way.

spec/intro.dd
thewilsonator: such micro controllers with e.g. bit-banding.

spec/intro.dd
```diff
@@ -67,6 +67,43 @@ $(OL
...
+    )
+
+    $(P Objects can be composed into a $(I composed object). The objects that make up
+    the composed object are $(I subobjects). An object that is not the subobject
+    of another object is a $(I complete object). The lifetime of a subobject
+    is always within the lifetime of its complete object.
+    )
+
+    $(P An object's address is the address of the first byte of the first memory
+    location for that object. Object addresses are distinct from each other
```
thewilsonator: How does this work for non-contiguous objects? Are non-contiguous objects always composed?

spec/intro.dd
WalterBright: We don't care about non-8 bit platforms. That doesn't actually precluded writing a D compiler specifically for such platforms, they just wouldn't be compliant.

For example, C supports platforms with 32 bit bytes. However, I can't think of any non-trivial C program that wouldn't need to be completely rewritten to work on such a machine. So it is kinda pointless to pretend to support them.

WalterBright:
@thewilsonator the spec is not meant as a tutorial. The memory model most definitely belongs in the spec. Putting it in front is because the rest of the spec relies on it.

spec/intro.dd
WalterBright: Non-contiguous just means there may be alignment holes.

thewilsonator: Might want to specify that then

spec/intro.dd
thewilsonator: cc @JinShil how much have you done with ARM cortex m3/m4?

JinShil: The ARM Cortex-M (a 32-bit microcontroller) has a feature called bit-banding where a 32-bit address addresses a single bit in memory.  It works for both SRAM and memory-mapped IO.  I rely on it to reduce code size and increase performance in my memory-mapped IO library:  

https://github.com/JinShil/stm32f42_discovery_demo/blob/master/source/stm32f42/mmio.d#L132-L142

I currently encapsulate using D's metaprogramming features:

spec/intro.dd
WalterBright: The alignment issue will be addressed later.

spec/intro.dd
WalterBright: I'd never heard of bit banding before, thanks for the links. I think it is a special case for ARM and outside the purview of the spec. Being outside of the spec, of course, doesn't prevent writing D code to work with bit banding.

thewilsonator:
> The memory model most definitely belongs in the spec

Definitely. What I meant is that it would be better off on the second page.

WalterBright:
> What I meant is that it would be better off on the second page.

I don't think that's a concern at the moment. We need the text to be correct and useful first. Probably a more useful organization of the material will be more clear after the text is created.

spec/intro.dd
```diff
@@ -67,6 +67,68 @@ $(OL
...
+    without interference.
+    )
+
+    $(P Memory locations come in three groups:)
+
+    $(OL
+    $(LI $(I Thread local memory locations) are accessible from only one thread at a time.)
+    $(LI $(I Immutable memory locations) cannot be written to during its lifetime. Immutable
+    memory locations can be read from by multiple threads without synchronization.)
+    $(LI $(I Shared memory locations) is accessible from multiple threads.)
```
tgehr: are accessible

spec/intro.dd
```diff
@@ -67,6 +67,68 @@ $(OL
...
+
+    $(UNDEFINED_BEHAVIOR Writing to an immutable memory location during its lifetime
+    results in undefined behavior.)
+
+    $(UNDEFINED_BEHAVIOR Unless synchronized, writing to a shared memory location in
+    one thread while other threads read or write to it results in undefined behavior.)
+
+    $(P Execution of a single thread on thread local and immutable memory locations
+    is $(I sequentially consistent). This means the result of the operations
+    is the same as if it were executed in the order the operations appear in the program.
```
tgehr: This needs to be improved. "Sequentially consistent" is a term that applies if there are _multiple threads_ accessing _shared memory locations_ and it says that the observed result is as if the operations of the threads were _interleaved_, where the operations from each single thread are kept in order. Right now, this document doesn't say anything about shared memory locations.

spec/intro.dd
```diff
@@ -67,6 +67,68 @@ $(OL
...
+    )
+
+    $(UL
+    $(LI a definition)
+    $(LI a $(GLINK2 expression, NewExpression))
+    $(LI a temporary is created)
+    $(LI changing which field of a union is active)
+    )
+
+    $(P An object spans a sequential sequence of memory locations which may or may not
```
tgehr: "sequential sequence" -> "sequence"

spec/intro.dd
```diff
@@ -67,6 +67,68 @@ $(OL
...
+    be contiguous. It exists during construction, its lifetime, and destruction.
+    Each object has a type, which is determined either statically or by runtime
+    type information.
```
tgehr: I'm pretty sure there can be memory locations whose type is not determined either statically or by runtime type information. What about a `cast(void*)new S` where S does not have a destructor? Maybe this could say while every object conceptually has a type, it may not possible to determine it from either static or runtime information.

spec/intro.dd
WalterBright: Sequentially consistent also applies even if there is only one thread. Of course we assume that, otherwise programs wouldn't work at all, but this sort of thing needs to be spelled out. (SC constrains what kinds of transformations the optimizer can apply.)

spec/intro.dd
WalterBright: I think in that case the data is "untyped". I anticipate a followon section which delves into object initialization and lifetimes.

WalterBright:
I added a section on converting between thread local / shared / immutable.

PR #2650, opened 2019-06-01
Revise the Introduction for greater accuracy and reduced redundancy

dibyendumajumdar:
Currently the introduction has a description of the phases of compilation.
A paragraph describes that each prior phase in unaffected by a subsequent phase.
This description is redundant as this is true for most compilers. Unless
more specific example can be given how D's phases differ from other languages
it is safer to not say anything specific here.

Secondly the phases of compilation goes into unnecessary and random details
of the lexical phase. This is not necessary.

The revised text tries to address these issues. It also adds some information
that may be useful to people who do not know D such as there is no
pre-processing phase.

Signed-off-by: Dibyendu Majumdar <mobile@majumdar.org.uk>

### DIFF
--- a/spec/intro.dd
+++ b/spec/intro.dd
@@ -10,59 +10,47 @@ to create native libraries or executables.)
 $(P This document is the reference manual for the D Programming Language. For more information and
 other documents$(COMMA) see $(LINK2 https://dlang.org/, The D Language Website).)
 
+$(H2 Program Structure)
+
+$(P A D program's unit of compilation is a module; each D source file corresponds to a single D module.
+D modules need not all be compiled at the same time. Previously compiled modules may be individually
+preserved as object files or in libraries, and later linked to produce an executable program.)
+
 $(H2 Phases of Compilation)
 
-$(P The process of compiling is divided into multiple phases. Each phase has no
-dependence on subsequent phases. For example, the scanner is not perturbed by
-the semantic analyzer. This separation of the passes makes language tools like
-syntax directed editors relatively easy to produce. It also is possible to
-compress D source by storing it in $(SINGLEQUOTE tokenized) form.)
+$(P D modules are compiled in multiple phases.)
 
 $(OL
-        $(LI $(B source character set)$(BR)
 
-        The source file is checked to see what character set it is,
-        and the appropriate scanner is loaded. ASCII and UTF
-        formats are accepted.
+        $(LI $(B Lexical analysis)$(BR)
+
+        The source file is converted into a sequence of tokens. In particular, D programs
+        are not pre-processed prior to compilation.
         )
 
-        $(LI $(B script line) $(BR)
+        $(LI $(B Syntax analysis)$(BR)
 
-        If the first line starts with "$(HASH)!", then that line
-        is ignored.
+        The sequence of tokens is parsed to form syntax trees. The syntax analysis
+        phase requires arbitrary lookahead of tokens.
         )
 
-        $(LI $(B lexical analysis)$(BR)
-
-        The source file is divided up into a sequence of tokens.
-        $(DDSUBLINK spec/lex, specialtokens, Special tokens)
-        are replaced with other tokens.
-        $(GLINK_LEX SpecialTokenSequence)s
-        are processed and removed.
-        )
-
-        $(LI $(B syntax analysis)$(BR)
-
-        The sequence of tokens is parsed to form syntax trees.
-        )
-
-        $(LI $(B semantic analysis)$(BR)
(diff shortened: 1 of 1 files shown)


spec/intro.dd
```diff
@@ -10,59 +10,47 @@ to create native libraries or executables.)
...
-
-$(P The process of compiling is divided into multiple phases. Each phase has no
-dependence on subsequent phases. For example, the scanner is not perturbed by
-the semantic analyzer. This separation of the passes makes language tools like
-syntax directed editors relatively easy to produce. It also is possible to
-compress D source by storing it in $(SINGLEQUOTE tokenized) form.)
+$(H2 Program Structure)
 
-$(OL
-        $(LI $(B source character set)$(BR)
```
atilaneves: Why delete the entries on source character set and script line?

spec/intro.dd
```diff
@@ -10,59 +10,47 @@ to create native libraries or executables.)
 $(P This document is the reference manual for the D Programming Language. For more information and
 other documents$(COMMA) see $(LINK2 https://dlang.org/, The D Language Website).)
 
-$(H2 Phases of Compilation)
-
-$(P The process of compiling is divided into multiple phases. Each phase has no
```
atilaneves: I think this paragraph is useful, why is it being deleted?

spec/intro.dd
dibyendumajumdar: Hi @atilaneves , I explained this in the previous pull request, here it is again. Basically to pick a couple of random details from a particular phase seems pointless as this is covered in more detail in the lexical analysis section. The description in the introduction ought to be a high level overview. These two points seemed very specific to me and do not add to the general overview of the process.

spec/intro.dd
dibyendumajumdar: Hi @atilaneves, I explained this in the forum post: https://forum.dlang.org/post/pxllknizhivuvajlnhps@forum.dlang.org. I am copying the contents below:

```
I feel this description is unclear, and it might just reflect how DMD is implemented. I haven't implemented a C++ parser but parsers I have worked with - such as for C - it is always the case that lexer doesn't get impacted by the semantic analysis. The standard process is to get a stream of tokens from the lexer and work with that. It is also conceivable that someone could create a "dumb" AST first for C++, and as a subsequent phase add semantic meaning to the AST, just as is done for D.

For now I propose to remove this paragraph until there is a better description available.
```

I did request Walter for a better explanation of how D is different from other compilers but he hasn't yet responded. Once  a better description is available then I shall add it back.

atilaneves: It definitly isn't the case for C++: `>>` could be a single token or two depending on whether it's the right shift operator or two closing template angle brackets.

dibyendumajumdar: Hi @atilaneves so there is a parsing ambiguity in C++ but how does that imply that lexical phase is perturbed by later phases? There are multiple ways to handle this ambiguity : for instance lexer can return two '>' tokens rather than '>>' token and let the AST phase handle it.

dibyendumajumdar: I am not sure I even understand what this is saying:

> Each phase has no dependence on subsequent phases. For example, the scanner is not perturbed by the semantic analyzer.

The C++ example you have given - how does it relate to this?

atilaneves: I don't know the exact details, but I think that in C++ lexing can't be done independently of parsing.

dibyendumajumdar: I am reluctant to keep the current wording as I don't know what it means. It would be clearer to state a) what kind of grammar D has (I asked about that but didn't get any answer), and b) something more concrete - such as: D's grammar is designed so that each lexical token unambiguously defines the production rule. 

I think that the way the ambiguity in C++ is handled in up to the implementation. It is possible to imagine an implementation that tokenizes '>>' as two '>' and lets a later phase handle '>>'. If  no one does it this way that doesn't make it impossible, it is likely it is less efficient to do it this way. 

In D too how are string mixins handled? Surely a mixin cannot be recognised until it is parsed, and then lexer has to be invoked on the mixin text? I see this dependency between phases.

atilaneves: You'd have to ask Walter, he's the expert.

dibyendumajumdar: Sure, asked already a couple of times on the forum, waiting for a reply.

The spec has to be technical and correct - we have to avoid vague or unsubstantiated claims, in particular if all they serve is to say D is better than X. The spec is not the place for comparisons; it just needs to describe the language with technical accuracy.

thewilsonator: > You'd have to ask Walter, he's the expert. 

You'd best email him.

PR #2658, opened 2019-06-13
Fixes 19944 - Unit test failures in std.file

Kriyszig:
Issue in focus: https://issues.dlang.org/show_bug.cgi?id=19944

The appending of `import std.stdio: write, writeln, writef, writefln;` to the code conflicts when trying to access `std.file.write`

Before:
alt text

After:
alt text

### DIFF
--- a/js/run_examples.js
+++ b/js/run_examples.js
@@ -28,7 +28,7 @@ function wrapIntoMain(code) {
         var codeOut = "void main()\n{\n";
         codeOut += "    import " + currentPackage + ";\n";
         // writing to the stdout is probably often used
-        codeOut += "    import std.stdio: write, writeln, writef, writefln;\n    ";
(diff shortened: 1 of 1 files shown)


CyberShadow:
@wilzbach 

Ugh, maybe just put the import outside `main`?

Kriyszig:
Will that work?
[Import Outside](https://run.dlang.io/is/pPAWjF)
[Without import of std.stdio.write](https://run.dlang.io/is/nFono8)

The former still gives an assertion error, the latter passes all tests.

CyberShadow:
Oh, because selective import has higher priority that a non-selective import in an inner scope.

Kriyszig:
Is there a better way to do this? I would be happy to modify this PR if there exists an alternate :octocat:

Kriyszig:
Is this supposed to auto merge? :octocat:

CyberShadow:
No, but thanks for the reminder.

Kriyszig:
Thanks :+1:

PR #2661, opened 2019-06-19
add methods for dealing with C++ namespaces

WalterBright:
As outlined in the discussion in https://github.com/dlang/dmd/pull/10031

### DIFF
--- a/spec/cpp_interface.dd
+++ b/spec/cpp_interface.dd
@@ -178,6 +178,61 @@ void main()
 }
 ------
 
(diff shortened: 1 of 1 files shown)


WalterBright:
> We should not be promoting the non-string version.

The string version is last because it is hackish and pretends that C++ namespaces do not introduce a namespace. Even so, the 3 methods are listed without editorial comment, and the user can do as he wishes.

thewilsonator:
Superseded by #2789

PR #2663, opened 2019-06-20
Add link to uninitializedArray in arrays

Alexibu:
From the language spec it appears as if there is no way to create an array and initialise its elements to something other than the element init value without having the values initialised twice.

### DIFF
--- a/spec/arrays.dd
+++ b/spec/arrays.dd
@@ -788,6 +788,10 @@ $(H3 $(LNAME2 void-initialization, Void Initialization))
         Void initializations are an advanced technique and should only be used
         when profiling indicates that it matters.
         )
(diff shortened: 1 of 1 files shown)


spec/arrays.dd
```diff
@@ -788,6 +788,10 @@ $(H3 $(LNAME2 void-initialization, Void Initialization))
         Void initializations are an advanced technique and should only be used
         when profiling indicates that it matters.
         )
+        $(P to void initialise the elements of dynamic array use
```
jacob-carlborg: `to` should have been capitalized. But I wasn't fast enough.

wilzbach: Also, shouldn't it be elements of __a__ dynamic array?

PR #2664, opened 2019-06-22
Specify the dotted expressions allowed as iasm operand

Basile-z:
This commits adds the new grammar allowing basic type properties as iasm operand (https://github.com/dlang/dmd/pull/10069) and specify what kind of dotted expressions are accepted.


Basile-z:
@thewilsonator  https://github.com/dlang/dmd/pull/10069 needs to be merged to match the spec now.

PR #2665, opened 2019-06-22
Turn on -preview=markdown

dgileadi:
This PR does three things:

1. Adds the `-preview=markdown` flag to the dlang.org Ddoc build.
2. Customizes the Ddoc-generated `OTHER_CODE` macro to support future instances of ` ``` some-language` code fences.
3. Fixes places in the docs that were unintentionally interpreted as markdown.

The fixes in item 3 were generally for:

1. Places that match Markdown's asterisk rules for emphasis. These places are generally in code examples, but since they aren't wrapped in fenced code blocks Ddoc isn't aware that it's supposed to treat them as code.
2. Backslashes that aren't in code blocks, and so are treated as escaping the next character.
3. Lines that start with `-` or `+`, which in Markdown starts a list item.
4. Lines that start with `>`, which in Markdown starts a quote block.


articles/ctod.dd
```diff
@@ -906,16 +906,16 @@ $(H4 The C Way)
        C has problems with the DOS file system because a \ is an escape in a string.
        To specifiy the file c:\root\file.c you would use:
 $(CCODE
-char file[] = "c:\\root\\file.c";
+`char file[] = "c:\\root\\file.c";`
```
jacob-carlborg: This will make the font bold. That will result in some code blocks look different from others. Not sure if that matters.

articles/ctod.dd
```diff
@@ -859,8 +859,8 @@ $(H4 The C Way)
        C initializes arrays by positional dependence. C99 fixes the issue:
 $(CCODE
```
ZombineDev: Can we used a fenced C code block here?

    ```C
    int a[3] = { 3,2,1 };
    /* ... */
    ```

ZombineDev: Ditto for the rest of the `$(CCODE ..)` changes.

articles/faq.dd
```diff
@@ -683,7 +683,7 @@ $(ITEM cpp_interface, Why doesn't D have an interface to C++ as well as C?)
         based on const and immutable.)
 
         $(LI C++ overloads operators in significantly different ways - for
-        example, operator[]() overloading for lvalue and rvalue is based on
+        example, operator\[]() overloading for lvalue and rvalue is based on
```
ZombineDev:
```suggestion
        example, `operator[]()` overloading for lvalue and rvalue is based on
```

changelog/2.074.0.dd
```diff
@@ -221,13 +221,13 @@ $(P
 It comes with a lot of new features
 $(UL
  $(LI $(LINK2 http://docs.algorithm.dlang.io/latest/mir_ndslice_topology.html, `mir.ndslice.topology`)
-        - Multidimensional `std.range` analog. Includes `bitwise`, `bitpack`, `zip`, `unzip`, `map`, `indexed` and many other features.)
+        \- Multidimensional `std.range` analog. Includes `bitwise`, `bitpack`, `zip`, `unzip`, `map`, `indexed` and many other features.)
```
ZombineDev: Perhaps the dash should be moved to the end of the previous line to avoid escaping it?

changelog/2.075.1.dd
```diff
@@ -28,7 +28,7 @@ $(LI $(BUGZILLA 17192): ParameterDefaults fails when parameter is called "i")
 )
 $(BUGSTITLE Phobos bugs,
 
-$(LI $(BUGZILLA 17667): regex$(LPAREN)[r".", r"[\$(LPAREN)\{[\]\}\$(RPAREN)]"]$(RPAREN);)
+$(LI $(BUGZILLA 17667): regex$(LPAREN)[r".", r"[\\$(LPAREN)\\{[\\]\\}\\$(RPAREN)]"]$(RPAREN);)
```
ZombineDev: Better wrap the whole thing in backticks, which will also remove the need for macros like `$(LPAREN)`.

changelog/2.079.0.dd
```diff
@@ -795,7 +795,7 @@ For example, on a Posix platform you can simply run:
 
 $(P
 $(CONSOLE
-sed "s/delete \(.*\);/__delete(\1);/" -i **/*.d
+`sed "s/delete \(.*\);/__delete(\1);/" -i **/*.d`
```
ZombineDev: Perhaps we should replace all uses of `$(CONSOLE ...)` with something along the lines of:

    ```console
    ```

changelog/index.dd
```diff
@@ -2,7 +2,7 @@ Ddoc
 
 $(SCRIPT
     var redir = window.location.href.replace(
-        /\/changelog\/index.html#(?:2\.|new2_)(\d+(?:\.\d+)?)$(DOLLAR)/,
+        /\\/changelog\\/index.html#(?:2\\.|new2_)(\\d+(?:\\.\\d+)?)$(DOLLAR)/,
         "/changelog/2.$(DOLLAR)1.html"
```
ZombineDev: Wow, using inline js with regex inside ddoc files looks like a bad idea. All JS code should be defined in `*.js` files and included automatically included inline if really necessary. That said, this out of the scope of your PR, so I'm fine with leaving it like this.

articles/ctod.dd
dgileadi: We certainly could, although we'd use a lowercase `c` to work with the existing CSS. The reason I didn't switch them  is because I wanted to minimize the amount of changes, and if we change one we'd probably want to change them all.

changelog/2.079.0.dd
dgileadi: Again, the reason I didn't make a change like this (and I was tempted to) is to minimize the amount of changes.

dgileadi:
> In addition to comments below, I also noticed that the output of some files changed like `lex.dd`, in which the list of escape sequences was now interpreted differently - IIRC from e.g. `\n` to `n`. In all those cases, the `$(D ...)` macro was used, so most likely the fix would be to use code blocks.

That worries me because I didn't see that behavior locally. It makes me wonder what else I might have missed. If it isn't too much trouble, could you let me know which files you saw that had issues?

CyberShadow:
@ZombineDev The documentation tester isn't showing any changes in lex.html, so I'm also confused why you would be seeing a difference in your output.

There are changes in `lex.verbatim`, but this is a partial preview of the intermediary expanded state of macros, and changes there don't necessarily map to changes in the final HTML render.

articles/ctod.dd
dgileadi: Once you pointed it out it bothered me enough that I went ahead and used ` ```c ` fences for these three items.

ZombineDev:
Yes, @CyberShadow, you're absolutely right!  I must have confused the verbatim vs html diffs. @dgileadi sorry for the false alarm.

ZombineDev:
@CyberShadow @aG0aep6G @wilzbach you guys may also want to have a look before we merge this PR. From my PoV it's good to go.

ZombineDev:
Arghh.. I meant to leave this open for others to review. @CyberShadow if you have any high-level concerns that cannot be easily addressed with a follow-up, feel free to revert this PR, so we can have more time to review.

CyberShadow:
No concerns, thank you @dgileadi for your work.

PR #2671, opened 2019-07-01
Update garbage.dd

vivianpiketmedia:
Added link to broader description on TLS in order to boost the readability for beginners.

### DIFF
--- a/spec/garbage.dd
+++ b/spec/garbage.dd
@@ -476,7 +476,7 @@ $(H2 $(LNAME2 precise_gc, Precise Heap Scanning))
 $(H2 $(LNAME2 precise_dataseg, Precise Scanning of the DATA and TLS segment))
 
     $(P $(B Windows only:) As of version 2.075, the DATA (global shared data)
-        and TLS segment (thread local data) of an executable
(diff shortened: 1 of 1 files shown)


spec/garbage.dd
```diff
@@ -476,7 +476,7 @@ $(H2 $(LNAME2 precise_gc, Precise Heap Scanning))
 $(H2 $(LNAME2 precise_dataseg, Precise Scanning of the DATA and TLS segment))
 
     $(P $(B Windows only:) As of version 2.075, the DATA (global shared data)
-        and TLS segment (thread local data) of an executable
+        and [TLS](https://hosting.review/web-hosting-glossary/#12) segment (thread local data) of an executable
```
lesderid: `TLS` refers to [thread-local storage](https://en.wikipedia.org/wiki/Thread-local_storage) here, not Transport Layer Security.

thewilsonator:
I don't  think markdown is enabled for the spec yet.

spec/garbage.dd
vivianpiketmedia: My mistake. Thanks.

PR #2672, opened 2019-07-01
Update statement.dd

vivianpiketmedia:
Added explanation on CPU (for beginners)

### DIFF
--- a/spec/statement.dd
+++ b/spec/statement.dd
@@ -1798,7 +1798,7 @@ $(GNAME AsmInstructionList):
 )
 
 $(P An asm statement enables the direct use of assembly language instructions.
-This makes it easy to obtain direct access to special CPU features without
+This makes it easy to obtain direct access to special [CPU](https://hosting.review/web-hosting-glossary/#19) features without
 resorting to an external assembler. The D compiler will take care of the
 function calling conventions, stack setup, etc.)
 


MoonlightSentinel:
I dont thinkt we should link to this website as the informations are ... misleading, at best.

```Central process unit (often referred to as “CPU”) is a measurement of how fast your computer hardware (or in this case - a server) can perform a task or operation.```

File: spec/statement.dd [hunk 1]
```diff
@@ -1798,7 +1798,7 @@ $(GNAME AsmInstructionList):
 )
 
 $(P An asm statement enables the direct use of assembly language instructions.
-This makes it easy to obtain direct access to special CPU features without
+This makes it easy to obtain direct access to special [CPU](https://hosting.review/web-hosting-glossary/#19) features without
```
MoonlightSentinel:
```suggestion
This makes it easy to obtain direct access to special [CPU](https://en.wikipedia.org/wiki/Central_processing_unit) features without
```

Why not link to Wikipedia instead?

vivianpiketmedia: I think this link https://hosting.review/web-hosting-glossary/#19 explains the term "CPU" in a language of beginners. I can say from my own perspective that it's better to explain it in easy words. That means they can understand this term better since that link also displays the useful illustrations.

aG0aep6G:
This is just spam.

spec/statement.dd
MoonlightSentinel: Easy words may be useful to explain for beginners.
But easy language does not change the fact that this statement is just **wrong**.

PR #2681, opened 2019-07-16
Add core.internal.array to module list

CyberShadow:
Follow-up to https://github.com/dlang/druntime/pull/2681

Overhaul the module list navigation tree generator:

- Explicitly specify internal packages
- Replace CSS hack for "Internal API" with delimiter macro
- Support nested inclusions and exclusions

CC @wilzbach


CyberShadow:
It's not great that we're maintaining an additional list of what modules to include or not in the navigation tree here. Ideally we would be reusing the module list from the makefiles (the same one which decides which `.d` files need to be rendered to `.html` files). Possibly by:

- parsing the list from the makefiles (e.g. the `mak/DOCS` format is pretty simple in Druntime)
- adding a target to the makefile which prints the list OSLT
- something like globbing the resulting list of `.html` files (by itself it would require two passes, and would require a clean checkout to avoid stale links)

posix.mak
```diff
@@ -205,6 +205,7 @@ MOD_EXCLUDES_PRERELEASE=$(addprefix --ex=, \
 	dmd.scanmach dmd.scanmscoff)
 
 MOD_EXCLUDES_LATEST=$(MOD_EXCLUDES_PRERELEASE)
+MOD_EXCLUDES_RELEASE=$(MOD_EXCLUDES_PRERELEASE)
```
CyberShadow: This looked like an omission but I'm not sure what its effect will be.

tools/modlist.d
```diff
@@ -4,147 +4,143 @@
...
+        foreach (pkg; packages)
+        {
+            TreeWriter writer;
+
+            void dump(ref Tree tree, string[] mod, bool inInternal, bool inIncluded)
+            {
+                inInternal = inInternal || mod in internalSet;
+                inIncluded = mod in includedSet ? true : mod in excludedSet ? false : inIncluded;
+                if (tree.exists && inInternal == dumpingInternal && inIncluded)
+                    if (mod.length > 1 || !tree.children) // std/package.d is currently unreachable!
```
CyberShadow: There's nothing one can click right now to see the `std/package.d` docs, because clicking on `std` expands/collapses the `std` tree.

CyberShadow:
Ping before this gets stale. @ZombineDev @wilzbach

PR #2682, opened 2019-07-17
fix Issue 19402 - specs for promotion rule of shift exp is wrong (without anything else)

dkorpel:
There already is #2578 attempting to fix it, but that is stalled because it also attempts to describe shift amounts out of range as undefined behavior, which is not agreed upon. In the mean time the blatant error in the spec persists, so this PR addresses only that so it can be merged sooner.


dkorpel:
Can this be merged now?

thewilsonator:
Whoops, yes.

PR #2689, opened 2019-08-13
Improve front page for small screens

WebFreak001:
adds scroll bars to non-fitting output & to readonly code starting at 500px

before:
unmodified frontpage
unmodified output

after:
modified frontpage
modified output

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1552,11 +1552,13 @@ textarea.d_code_stdin, textarea.d_code_args
     background: white;
     margin-left: 2px;
     outline: none;
+    hyphens: none;
 }
 
 pre.d_code_output {
     border: none;
     max-height: 30em;
+    overflow: auto;
 }
 
 div.d_code {margin: 0; padding: 0; background: #F5F5F5;}
@@ -1571,6 +1573,16 @@ div.d_code_output, div.d_code_stdin, div.d_code_args, div.d_code_unittest
 }
 
 .d_code_title {font-weight: bold;padding: 5px;}
+
+@media only screen and (max-width: 500px)
+{
+    pre.d_code
+    {
+        overflow-x: auto;
+        white-space: pre;
+        word-wrap: unset;
+    }
+}
 
 .CodeMirror-wrap {padding: 3px;}
 .CodeMirror {line-height: normal; border: 1px solid #CCC; background: #FCFCFC; height: auto}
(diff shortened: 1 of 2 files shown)


WebFreak001:
anything missing?

thewilsonator:
IDK, I'm not a web person, I'll leave it to your, judgement.

WebFreak001:
I think the changes should work on all pages with code because there are no other pages with variable width code blocks (they are full width) but otherwise the worst that can happen is that in a variable code width layout with long code (that would currently break) it will now overflow the entire code block or make it larger than the specified width in case of long code.

lesderid:
The first two commits look fine to me, but I don't understand why we need the last one.

WebFreak001:
@lesderid the last code fixed a bug that the overflow wouldn't happen until there is actually no space. So it would become wider than the specified 32em and push everything to the left and make it all smaller when the lines are very long.

lesderid:
@WebFreak001 Could you give an example of where this happens and with what resolution?

WebFreak001:
it happens with the "print hex dump example" when slowly resizing down (would probably happen on some phones in landscape mode)

lesderid:
I tried reproducing it with Chromium DevTools' resolution switcher, but didn't notice any major differences with that commit added.

I'm not much of a web person either, but I assume there is a reason that `max-width: 32em` for that div is currently behind `@media only screen and (max-width: 40em)`?

aG0aep6G:
> I tried reproducing it with Chromium DevTools' resolution switcher, but didn't notice any major differences with that commit added.

The issue is real in Firefox 68.0.2. To reproduce: open the new page in a large window, select a code sample with long lines, disable `max-width: 32em`, and then slowly make the window narrower. At some point a horizontal scroll bar should appear on the whole page, because the code box is wider than 32em.

> I'm not much of a web person either, but I assume there is a reason that `max-width: 32em` for that div is currently behind `@media only screen and (max-width: 40em)`?

Almost always, `max-width: 32em` is redundant when there's already `width: 32em`. But we have a special case here: the element also has `display: table-cell`. That means:

1. [Width is calculated differently for table cells.](https://drafts.csswg.org/css2/tables.html#width-layout) `width` can act similar to `min-width`. That's what happens to the code box.
2. ["[The] effect of [...] 'max-width' on [...] table cells [...] is undefined."](<https://drafts.csswg.org/css2/visudet.html#propdef-max-width>) In our case, the undefined effect seems to be that `max-width` works more effectively than `width` in restricting the width.

Instead of relying on the undefined behavior of `max-width: 32em`, I suggest one of these alternatives:

1. `body#Home #content > .intro > div > div { table-layout: fixed }` - Change the layout algorithm so that `width` is interpreted more strictly.
2. `.your-code-here-extra { width: 32em }` - Add `width` to a block element that also spans around the code but doesn't have `display: table-cell`.

PR #2693, opened 2019-08-28
Update host compiler to 2.087

JinShil:
We need 2.087 to support documenting public imports.

https://github.com/dlang/dmd/pull/9068
https://dlang.org/changelog/2.087.0.html


JinShil:
This looks ready to go.

Please note that this is likely to cause some disruption as public imports suddenly start appearing in users' documentation.  I'll do my best be attentive to any issues and solve them if I can.

FYI @wilzbach @ZombineDev @CyberShadow

JinShil:
We should arguably be using master for this, but I don't know that that's my call, and I'm not sure how to implement that in the makefile.  This was an easy fix and keeps development moving at least for a while.

ZombineDev:
Actually, I think it should already be using dmd master, if you look for the `$(DMD)` variable in `posix.mak`. Anyway bumping the `STABLE_DMD` version obviously doesn't hurt.

posix.mak
```diff
@@ -180,7 +180,7 @@ DRUNTIME_LATEST_DIR=$G/druntime-${LATEST}
 PHOBOS_LATEST_DIR=$G/phobos-${LATEST}
 
 # stable dub and dmd versions used to build dpl-docs
-STABLE_DMD_VER=2.085.0
+STABLE_DMD_VER=2.087.0
```
ZombineDev: Use the point release to avoid theoretically hitting some fixed bugs:
```suggestion
STABLE_DMD_VER=2.087.1
```

CyberShadow:
PR title is misleading, this is simply the version used to build the stable docs, not the host DMD version used to build DMD.

JinShil:
> PR title is misleading

I'm sorry. I didn't realize there was an intermediate step of bootstrapping the compiler. In fact, I still don't get it.  The build procedure seems to build master with a 2.079 bootstrap compiler but then downloads the STABLE_DMD_VER to build the docs.  Why? I don't know.

CyberShadow:
> I'm sorry

Don't worry about it.

> I didn't realize there was an intermediate step of bootstrapping the compiler.

This is optional. The official way to build DMD is to download an older binary release of DMD from dlang.org and use that. Bootstrapping is only needed when the user wishes to avoid using unauditable binaries of any kind, or when porting DMD to a new platform (where cross-compiling is not feasible).

>  The build procedure seems to build master with a 2.079 bootstrap compiler 

The DMD version used to build a host compiler is as old as possible (and is bumped as rarely as possible) to ensure that the number of steps for from-source bootstrapping is as small as possible.

> but then downloads the STABLE_DMD_VER to build the docs.

The stable docs are built with a stable (non-git) version of the compiler. A better question may be why we're not using the compiler version matching the version of the docs being built; the answer to that is probably to minimize the total amount of DMD versions downloaded and cached across many dlang.org builds.

JinShil:
@CyberShadow I understand that, but why is the procedure building DMD and then downloading STABLE_DMD_VER to build the docs?

CyberShadow:
We build master docs and stable docs on dlang.org. Master docs are built with master dmd, stable docs are built with "stable DMD".

JinShil:
Ok, I see now. Thank you.

CyberShadow:
The stable docs are not built with DMD master to ensure that they remain readable after DMD changes which change the DDoc output. There've been a few over the years, and we'll probably see more with upcoming additional Markdown support.

JinShil:
Should this actually use DMD_LATEST?

CyberShadow:
Possibly. I'm not sure why we're not doing that. @wilzbach Would you happen to know?

wilzbach:
In the past we needed to lock the DMD version used to build ddox as vibe.d couldn't be built with the latest release. For the same reason, it still makes sense imho to use a fixed version. Though now that we have the project tester, I guess we could use DMD_LATEST or even master.

OTOH we can't use 2.079, because that was too old and buggy for building ddox and friends. IIRC e.g. it has a dub version without the mirror support and I am not sure whether 2.079 even can build latest vibe.d.

wilzbach:
BTW markdown has already been enabled: https://github.com/dlang/dlang.org/pull/2665

CyberShadow:
> BTW markdown has already been enabled: #2665

That only enabled the opt-in for dlang.org.

PR #2699, opened 2019-09-05
Revert PR1914: `inout ref` parameters imply `return` attribute

JinShil:
Reverts #1914 

In support of https://github.com/dlang/dmd/pull/10390


thewilsonator:
cc @WalterBright

JinShil:
@WalterBright dlang/dmd#10390 needs your advise.

PR #2701, opened 2019-09-11
no semicolon after enum declaration

aG0aep6G:


aG0aep6G:
Spotted by Roberto Rosmaninho: <https://forum.dlang.org/post/xwnidepxlsrfosavzbqi@forum.dlang.org>

PR #2702, opened 2019-09-17
Use assert() rather than Error

WalterBright:


JinShil:
@WalterBright, Please see @Geod24's [comment on one of the druntime PRs](https://github.com/dlang/druntime/pull/2794#issuecomment-531389434).  Repeated here for convenience:

> `Error` and `assert()` do the same thing, we should just use one method. I'd like to eventually deprecate `Error`.

> No they don't. First, we have `checkaction` to change the behavior. Second, in release mode (and we distribute druntime / phobos in release mode), the `assert` you are putting lead to a HLT, without error message, while `throw new Error` will lead to sensible error messages.

I think it is a legitimate concern that deserves to be addressed.

RazvanN7:
@WalterBright what is your opinion with regards to @JinShil 's comment?

WalterBright:
To have a nice message, use checkaction to change it to what is desired. Throwing Errors should not be for dealing with program bugs.

mdparker (approved):

PR #2706, opened 2019-10-01
spec/interfaceToC update: add warning about non-toplevel uses of extern (C)

jrfondren:
The link is to https://issues.dlang.org/buglist.cgi?quicksearch=extern%20%28C%29 , with seven issues, five of which are expressions of surprise at the current behavior. I found these issues and the recent change to `extern (C)` in mixin templates after independently making my own expression of surprise about an undefined reference to a mangled 'puts' in:

```d
#! /usr/bin/env rdmd

void main() {
    import std.string : toStringz;
    extern (C) int puts(const(char)* s);

    puts("Hello, world!".toStringz);
}
```

### DIFF
--- a/spec/interfaceToC.dd
+++ b/spec/interfaceToC.dd
@@ -60,7 +60,7 @@ int myDfunction(char[] s)
         such as $(D extern (C)).)
         
         $(LI The strcmp declaration was toplevel. A nested declaration may well $(LINK2
-        https://issues.dlang.org/buglist.cgi?quicksearch=extern%20%28C%29 not do what you want),
+        https://issues.dlang.org/buglist.cgi?quicksearch=extern%20%28C%29, not do what you want),
         for example it may change calling conventions but not mangling.)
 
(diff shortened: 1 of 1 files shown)


jrfondren:
I can see that. It's written too much as if the speaker doesn't really know what nested scopes will do, because I don't understand the current behavior (from the mixin templates discussion, there's some use on Windows? Or it's a deliberate way to get C calling conventions with mangled names for functions that aren't actually implemented in C?).

I don't mind if someone else writes the line and adds it, for a consistent voice, and closes this PR. What I'd suggest is a directive to use `extern (C)` only at the module level, and more detail in the linkage attributes page.

How about

> The strcmp declaration is at module scope. Name mangling is performed for `extern (C)` symbols that are declared in nested scopes.

I guess it doesn't need an ("except for template mixins") since the template mixin spec already suggests the new behavior.

jrfondren:
Updated:

There are several things going on here:
...
- The C function was declared at module scope. A nested declaration with extern (C) will have C calling conventions but D name-mangling.

PR #2713, opened 2019-11-02
Deploy CircleCi build to Netlify

wilzbach:
Needs to be a branch at upstream to be able to test the upload.
Remaining steps for further work:
- [ ] kindle
- [ ] latex
- [ ] prerelease
- [ ] translate all `.htaccess` to Netlify rewrites

Preview: https://dlang.netlify.com

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,9 +1,22 @@
-version: 2
+version: 2.1
+workflows:
+  build_and_deploy:
+    jobs:
+      - build:
+          filters:
+            tags:
+              only: /.*/
+      - deploy:
+          requires:
+            - build
+          filters:
+            branches:
+              only: master
 jobs:
   build:
     working_directory: ~/dlang.org
     docker:
-      - image: circleci/node:4.8.2
+      - image: circleci/node
     parallelism: 1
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -19,3 +32,21 @@ jobs:
...
+      - image: circleci/node
+    steps:
+      - attach_workspace:
+          # Must be absolute path or relative path from working_directory
+          at: ~/dlang.org
+
+      - run:
+          command: |
+            npm i netlify-cli
+            npx netlify deploy --site "$NETLIFY_SITE_ID" --auth "$NETLIFY_ACCESS_TOKEN" --prod --dir=web
```
wilzbach: @zombinedev FYI: we could also run this on all builds without the `--prod` flag and then get a deployment preview. However, as we already have DAutoTest and there's not easy way to get the link back from the logs to GH, I'm not sure whether this is useful.

PetarKirov: We can save the output of `netlify deploy` to a file (or just stdin it) and grep for the URL. Then perhaps we can use @dlang-bot to post the link to the PR?

PR #2715, opened 2019-11-05
document hasCopyConstructor and hasPostblit traits

adamdruppe:
Companion to https://github.com/dlang/dmd/pull/10528


PetarKirov:
Thanks @adamdruppe!

PR #2716, opened 2019-11-08
Update extern(C++) grammar to support string namespaces

rmanthorpe:
The grammar is missing the string-style `extern(C++, "a", "b")` declaration. The `ArgumentList` isn't quite right here as it allows a trailing comma but the current implementation in dmd does not. However, this seems more like a bug in dmd since it's inconsistent with the rest of the grammar.

### DIFF
--- a/spec/attribute.dd
+++ b/spec/attribute.dd
@@ -90,6 +90,7 @@ $(GNAME LinkageType):
 
 $(GNAME NamespaceList):
     $(GLINK2 expression, ConditionalExpression)
+    $(GLINK2 expression, ConditionalExpression)$(COMMA)
     $(GLINK2 expression, ConditionalExpression)$(COMMA) $(I NamespaceList)
 )
 
--- a/spec/attribute.dd
+++ b/spec/attribute.dd
@@ -78,6 +78,7 @@ $(GRAMMAR
 $(GNAME LinkageAttribute):
     $(D extern) $(D $(LPAREN)) $(GLINK LinkageType) $(D $(RPAREN))
     $(D extern) $(D $(LPAREN)) $(D C++), $(GLINK2 declaration, IdentifierList) $(D $(RPAREN))
+    $(D extern) $(D $(LPAREN)) $(D C++), $(GLINK2 attribute, NamespaceList) $(D $(RPAREN))
 
 $(GNAME LinkageType):
     $(D C)
@@ -86,6 +87,10 @@ $(GNAME LinkageType):
     $(D Windows)
     $(D System)
     $(D Objective-C)
+
+$(GNAME NamespaceList):
+    $(GLINK2 expression, ConditionalExpression)
+    $(GLINK2 expression, ConditionalExpression)$(COMMA) $(I NamespaceList)
 )
 
         $(P D provides an easy way to call C functions and operating


spec/attribute.dd
```diff
@@ -77,6 +77,7 @@ $(GRAMMAR
 $(GNAME LinkageAttribute):
     $(D extern) $(D $(LPAREN)) $(GLINK LinkageType) $(D $(RPAREN))
     $(D extern) $(D $(LPAREN)) $(D C++), $(GLINK2 declaration, IdentifierList) $(D $(RPAREN))
+    $(D extern) $(D $(LPAREN)) $(D C++), $(GLINK2 declaration, ArgumentList) $(D $(RPAREN))
```
Hackerpilot: I think we may need to introduce a new rule here. The parser looks for a comma-separated list of conditional expressions, while ArgumentList is a comma-separated list of AssignExpression.

rmanthorpe: I'm not convinced. As it is it seems to happily parse an `AssignExpression`. For example `extern(C++, (a = "b")) int c();` parses fine, just like the identifier with an assign index works fine: `extern(C++, (d[e = f])) int g();`

Hackerpilot: Yes, `extern(C++, (a = "b")) int c();` parses, but `extern(C++, a = "b") int c();` does not, which is my point. Putting the assign expression inside the parenthesis makes it match the rule for PrimaryExpression, which causes it to match the rule for ConditionalExpression.

rmanthorpe: Ah you're quite right - this is not my forte. In that case, is the identifier list grammar wrong too? `extern(C++, d[e = f]) int g();` does not parse.

rmanthorpe:
@Hackerpilot does that look better now?

Hackerpilot:
I was going to suggest re-ordering things to make the rule left-recursive, but looking over the grammar I can see that the existing style is that rules are right-recursive. My only suggestion is to remove the redundant `NamespaceExpression` rule and use `ConditionalExpression` directly.

rmanthorpe:
Yes everything is right-recursive so I stuck with the same style. I've removed that redundant expression now.

rmanthorpe:
Bump @Hackerpilot

Hackerpilot:
I don't have the ability to merge this.

PetarKirov:
I can merge, but we need to figure out why the build fails first. I will force push, in order to restart it (assuming it was simply a transient failure).

rmanthorpe:
I rebased it just to see if that would help but now it's failing circleci too 🤦‍♂ I really have no idea why this would be. Any thoughts?

Geod24:
Rebased, squashed the commit, and (hopefully) fixed.

>  I really have no idea why this would be. Any thoughts?

I'm no DDOC expert but I haven't seen any other place in that page where `$(D ,)` is used. Since it's also used for macro arguments, that might be the source of the error, so I've removed the surrounding `$(D ,)`

rmanthorpe:
> Rebased, squashed the commit, and (hopefully) fixed.

Thanks I'm totally stumped by this!

Geod24:
> Thanks I'm totally stumped by this!

Well, me two. I found two errors, but it's still red. I'm not 100% sure the `$(D ,)` would lead to error, but what I found was that `GLINK` and `GLINK2` were inverted. `GLINK` is for links within the page. Since `ConditionalExpression` is in `expression.dd` and not `attribute.dd`, `$(GLINK2 expression, ConditonalExpression)` was needed. Additionally, `$(GLINK2)` was used to link to `declaration`'s `NamespaceList`, however `NamespaceList` is in `attribute`, not `declaration`, so we should use `GLINK` instead.

Now, even with that fixed, it is still red, and I have no clue why. I'm on Mac so I can't build ATM, going to give it another try later.

Geod24:
Yeah, I have absolutely no clue what is wrong. There is so much pre-processing going on that linking the error message with the input is near impossible. I tried a few things, including removing all links, and reducing this until it passes. It did pass when the `$(GNAME NamespaceList):` and following was removed, but I really can't see anything wrong with it.

@CyberShadow : Would it be possible to have artifacts of failing build be published ? I have no clue what was fed to the LaTeX compiler.

CyberShadow:
Thanks for looking into this.

> @CyberShadow : Would it be possible to have artifacts of failing build be published ? I have no clue what was fed to the LaTeX compiler.

It does do that if the *test* phase fails, but this is a build problem, and I think the latex input is a temporary file (not part of the output) so it wouldn't be accessible anyway.

The software running the build is based on Digger, so you should be able to reproduce build problems locally by running `digger build --with=website master+dlang.org#2716` (you may need to install some packages first).

Geod24:
Rebased on master, let's see if it is finally green

spec/attribute.dd
```diff
@@ -86,6 +87,10 @@ $(GNAME LinkageType):
     $(D Windows)
     $(D System)
     $(D Objective-C)
+
+$(GNAME NamespaceList):
```
MoonlightSentinel, line 91: Is this rule still required now that `extern(C++, "foo", )` allows trailing commas?

Geod24: Hum my grammar is a bit rusty, how would you express it ?
Alternatively feel free to fix it and force push to the remote and ping me for review. AFAICS the CI is happy now.

rmanthorpe: > AFAICS the CI is happy now

Finally! Thanks @Geod24

MoonlightSentinel: Nevermind, it uses `ConditionalExpression` which doesn't have correspongding list rule.

spec/attribute.dd
```diff
@@ -86,6 +87,10 @@ $(GNAME LinkageType):
...
+    $(GLINK2 expression, ConditionalExpression)
+    $(GLINK2 expression, ConditionalExpression)$(COMMA) $(I NamespaceList)
```
MoonlightSentinel, line 93: This allows trailing comma's:
```suggestion
    $(GLINK2 expression, ConditionalExpression)
    $(GLINK2 expression, ConditionalExpression)$(COMMA)
    $(GLINK2 expression, ConditionalExpression)$(COMMA) $(I NamespaceList)
```

(Don't think i can write on PR's here)

rmanthorpe: When did this change? I get an error when I leave a trailing comma.

MoonlightSentinel: Fixed that one recently in dlang/dmd#11090

Geod24:
Finally! Thanks a lot @rmanthorpe and @MoonlightSentinel . So glad we got rid of that PDF export...

PR #2719, opened 2019-11-26
fix issue 2968 by adding pragma(framework)

benjones:
For dmd PR: https://github.com/dlang/dmd/pull/10615

### DIFF
--- a/spec/pragma.dd
+++ b/spec/pragma.dd
@@ -62,6 +62,7 @@ $(P All implementations must support these, even if by just ignoring them:)
 $(UL
     $(LI $(LINK2 #inline, pragma inline))
     $(LI $(LINK2 #lib, pragma lib))
+    $(LI $(LINK2 #framework, pragma framework))
     $(LI $(LINK2 #linkerDirective, pragma linkerDirective))
     $(LI $(LINK2 #mangle, pragma mangle))
     $(LI $(LINK2 #msg, pragma msg))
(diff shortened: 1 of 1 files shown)


spec/pragma.dd
```diff
@@ -138,6 +139,17 @@ pragma(lib, "foo.lib");
...
 
+$(H3 $(LNAME2 framework, $(D pragma framework)))
+
+     $(P there must be one $(ASSIGNEXPRESSION) and it must evaluate at compile time to a string literal.
+     )
+-----------------
+pragma(framework, "someFramework");
+-----------------
+     $(IMPLEMENTATION_DEFINED
+     Typically, the parameters `-framework someFramework` will be passed to the linker
```
thewilsonator: might want to add "passed to the linker _if applicable_".

Geod24:
The DMD PR has been closed in favor of `linkerDirectives`

adamdruppe:
Just wished I had this again. linkerDirective only works on Windows. framework is a Mac feature.

kinke:
> linkerDirective only works on Windows

That's purely a DMD limitation. `pragma({lib,linkerDirective})` work as embedded Mach-O linker directives with LDC, on Windows *and* Mac.

adamdruppe:
Alas, extern(Objective-C) doesn't work on ldc, so we're at a useless intersection.

kinke:
I don't see how that's related - just saying that there's no object-file limitation to make the existing pragmas work with DMD for Mac too, which cover framework use cases and more.

adamdruppe:
it is related cuz the end goal here requires both, so there's no solution outside patching one or both compilers.

It isn't a difficult patch, but it is hard to realize one's IVY using D when something that was written *14 years ago* never got merged cuz "just use this other thing" and said other thing is also broken.

PR #2720, opened 2019-11-28
Clarify how contracts work

yshui:


yshui:
I totally misunderstood how contracts work. Please ignore this.

PR #2727, opened 2019-12-27
Clarify signed integer overflow by explicit example.

JohanEngelen:
Signed integers have wraparound semantics.
https://forum.dlang.org/thread/n23bo3$qe$1@digitalmars.com

@tgehr


JohanEngelen:
Note the fixup, merge with squashing commits.

PR #2733, opened 2020-01-13
Fix bug in code example

moon-chilled:


index.dd
```diff
@@ -640,7 +640,7 @@ import core.stdc.stdlib;
 void livingDangerously()
 {
     // Access to C's malloc and free primitives
-    auto buf = malloc(1024 * 1024);
+    auto buf = malloc(float.sizeof * 1024 * 1024);
```
aG0aep6G: This wasn't a bug. The extra memory you're allocating here isn't being used below.

PR #2734, opened 2020-01-13
Partially revert "Fix bug in code example"

MoonlightSentinel:
This reverts commit 8c3e5ea9d334864477c526a5cdddabbfc1b3946b.

`buf[0 .. 1024 * 1024]` creates a `void[]`  (length 1024 * 1024) which is then casted to a `float[]` (length ((1024 * 1024)) / float.sizeof)

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -640,11 +640,12 @@ import core.stdc.stdlib;
 void livingDangerously()
 {
     // Access to C's malloc and free primitives
-    auto buf = malloc(1024 * 1024);
+    enum bytes = float.sizeof * 1024 * 1024;
+    auto buf = malloc(bytes);
     // free automatically upon scope exit
     scope(exit) free(buf);
     // Interprets memory as an array of floats
-    auto floats = cast(float[]) buf[0 .. 1024 * 1024];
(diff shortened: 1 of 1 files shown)


MoonlightSentinel:
Optional: The second commit explains the resulting array length to avoid confusion for future readers.

thewilsonator:
cc @moon-chilled

moon-chilled:
Oops.  That's unlikely to be what's actually wanted, though.  Better solution:

        auto floats = (cast(float*)buf)[0 .. 1024 * 1024];

MoonlightSentinel:
> [...] That's unlikely to be what's actually wanted, though. [...]

You mean an `float[]` of length 1024 * 1024?

moon-chilled:
Yes.

I can't really see a case where you would want 1kb of floatage, unless you were doing some kind of bulk copy operation and your pages were really small.

MoonlightSentinel:
Okay, the the resulting array has 1024 * 1024 entries now.

moon-chilled:
LGTM.  Not a maintainer so can't approve.  @thewilsonator.

PR #2736, opened 2020-01-15
add FunctionAttributeKwd to grammar to factor out redundancy

WalterBright:
Having `nothrow` and `pure` appearing twice for function attributes should be tightened up.


WalterBright:
DAutoTest produces a veritable blizzard of errors, none of which seem to have anything to do with the changes I made.

RazvanN7 (approved):

PR #2739, opened 2020-01-23
Fix Issue 20500 - always pass -unittest and --DRT-testmode=run-main

wilzbach:
Requires 

- [x] https://github.com/dlang-tour/core/pull/743
- [x] https://github.com/dlang-tour/core-exec/pull/51

### DIFF
--- a/js/run.js
+++ b/js/run.js
@@ -101,7 +101,9 @@ var backends = {
     requestTransform: function(data) {
         var req = {
             source: data.code,
+            // always execute unittests and main for backwards compatibility with examples
             args: "-unittest",
+            runtimeArgs: "--DRT-testmode=run-main",
             compiler: dmdCompilerBranch
         }
         // only send set attributes


wilzbach:
Both changes are deployed now:

image

It can also be tested live.
Phobos: http://dtest.dlang.io/artifact/website-bd697944980567f6a4ca6682b542d0cbf1902dba-cc4d855be14f0c9d86e509c0341e91ee/web/phobos-prerelease/std_algorithm_setops.html
dlang.org: http://dtest.dlang.io/artifact/website-bd697944980567f6a4ca6682b542d0cbf1902dba-cc4d855be14f0c9d86e509c0341e91ee/web/index.html
http://dtest.dlang.io/artifact/website-bd697944980567f6a4ca6682b542d0cbf1902dba-cc4d855be14f0c9d86e509c0341e91ee/web/spec/function.html

Though to be fair I don't think `-unittest` is used often (at all?), so just removing it might have been an option as well. OTOH we now support runtime arguments and this is probably the more intuitive behavior, so we might as well support this.

PR #2740, opened 2020-01-25
Document version identifiers: iOS, TVOS and WatchOS

jacob-carlborg:


jacob-carlborg:
Related: https://github.com/ldc-developers/ldc/pull/3288.

jacob-carlborg:
Hmm, I don't see the actual error in the build log.

PR #2746, opened 2020-02-02
English improvements

mark-summerfield:


mark-summerfield:
I don't think the checklist really applies.

I'm learning D and having done the D-tour (to which I made various English improvements), and some textbook reading and some simple programs, I'm now reading the manual. And since I'm reading it anyway, I might as well improve the English as I go.

thewilsonator:
That checklist applies more to code changes rather than documentation changes, so no need to worry. Thanks!

spec/intro.dd
```diff
@@ -12,18 +12,18 @@ other documents$(COMMA) see $(LINK2 https://dlang.org/, The D Language Website).
...
+syntax directed editors relatively easy to create. It is also possible to
 compress D source by storing it in $(SINGLEQUOTE tokenized) form.)
 
 $(OL
         $(LI $(B source character set)$(BR)
 
-        The source file is checked to see what character set it is,
+        The source file is checked to see which encoding it is using,
         and the appropriate scanner is loaded. ASCII and UTF
-        formats are accepted.
```
Geod24: This whole paragraph is a bit superfluous TBH. ASCII is a subset of UTF-8 after all. So something as simple as "Any UTF-encoded source file is valid." might be enough, wouldn't it ?

spec/intro.dd
```diff
@@ -106,7 +106,7 @@ $(H2 Memory Model)
 
     $(P A memory location can be temporarily transferred from shared to local if
     synchronization is used to prevent any other threads from accessing the memory
-    location during the duration.)
+    location while the transfer is in effect.)
```
Geod24: I think it was meant to be "during the operation".
That sentence is a bit ambiguous to me: If the data held in that memory location has indirection, the transfer needs to be either a deep copy or retain `shared` on the inner types.

PR #2747, opened 2020-02-02
Minor English improvements + notes below

mark-summerfield:
1. I've replaced "Aka" with "alias". If that's not wanted I still think "Aka" is wrong and should be replaced by something else.
2. On lines 549-550 it says you can write a named character entity like this: `\name`. But I tried doing: `auto x = "\euro";` and it gave an error.
3. It is a real pity that `__DATE__` uses a US-specific format rather than the universal ISO8601. I hope an `__ISODATE__` and an `__ISOTIMESTAMP__` are added to repair this.
4. I don't know what the explanation for `__EOF__` means. Would "Tells the scanner to stop at this point in the source." be a correct alternative?
5. This page's very last example has the string `"foo\bar"`. Shouldn't this be `r"foo\bar"` or `"foo\\bar"`? And if not, a reason ought to be given.

### DIFF
--- a/spec/lex.dd
+++ b/spec/lex.dd
@@ -6,17 +6,17 @@ $(HEADERNAV_TOC)
 
 $(P The lexical analysis is independent of the syntax parsing and the semantic
 analysis. The lexical analyzer splits the source text up into tokens. The
-lexical grammar describes the syntax of those tokens. The grammar is designed to
+lexical grammar describes the syntax of these tokens. The grammar is designed to
 be suitable for high speed scanning and to make it easy to write a correct
-scanner for it. It has a minimum of special case rules and there is only one
+scanner. It has a minimum of special case rules and there is only one
 phase of translation.)
 
 $(H2 $(LNAME2 source_text, Source Text))
 
-$(P Source text can be in one of the following formats:)
+$(P Source text can be in any one of the following encodings:)
 
         $(LIST
-        ASCII,
+        ASCII (strictly, 7-bit ASCII),
         UTF-8,
         UTF-16BE,
         UTF-16LE,
@@ -43,10 +43,10 @@ be less than or equal to U+0000007F.)
 $(P The source text is decoded from its source representation into Unicode
 $(GLINK Character)s. The $(GLINK Character)s are further divided into: $(GLINK
 WhiteSpace), $(GLINK EndOfLine), $(GLINK Comment)s, $(GLINK
-SpecialTokenSequence)s, $(GLINK Token)s, all followed by $(GLINK EndOfFile).)
+SpecialTokenSequence)s, and $(GLINK Token)s, with the source terminated by an $(GLINK EndOfFile).)
 
-$(P The source text is split into tokens using the maximal munch technique,
-i.e., the lexical analyzer makes the longest token it can. For example
+$(P The source text is split into tokens using the maximal munch algorithm,
+i.e., the lexical analyzer makes the longest possible token. For example
 $(D >>) is a right shift token, not two greater than tokens. There are two
 exceptions to this rule:)
 
@@ -355,7 +355,12 @@ $(GNAME TokenString):
 $(H3 $(LNAME2 wysiwyg, Wysiwyg Strings))
 
         $(P
-        Wysiwyg ("what you see is what you get") quoted strings are enclosed by `r"` and `"`.
+        Wysiwyg ("what you see is what you get") quoted strings can be defined
+        using either of two syntaxes.
+        )
+
+        $(P
+        In the first form, they are enclosed between `r"` and `"`.
         All characters between
         the `r"` and `"` are part of the string.
         There are no escape sequences inside wysiwyg strings.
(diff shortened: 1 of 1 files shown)


thewilsonator:
1. While those are technically `alias`es they are they are almost never referred to as the underlying symbol, even by the compiler. e.g. if the type `immutable(char)[]` were to be part of an error message it would be displayed as `string`, so I think aka is fine there.

4. "Tells the _lexer_ to stop at this point in the source."

5. Most probably. `\b` is backspace and doesn't make much sense in a string.

mark-summerfield:
I still think use of "Aka" is poor practice. It is an abbreviation and not necessarily well known to non-native English speakers.

spec/lex.dd
```diff
@@ -355,7 +355,12 @@ $(GNAME TokenString):
 $(H3 $(LNAME2 wysiwyg, Wysiwyg Strings))
 
         $(P
-        Wysiwyg ("what you see is what you get") quoted strings are enclosed by `r"` and `"`.
+        Wysiwyg ("what you see is what you get") quoted strings can be defined
```
Geod24: Shouldn't it be all-caps (WYSIWYG) since it's an acronym ?

spec/lex.dd
```diff
@@ -355,7 +355,12 @@ $(GNAME TokenString):
...
+        using either of two syntaxes.
+        )
+        
+        $(P
+        They can be enclosed by `r"` and `"`.
```
Geod24: enclosed between ?
And the following might be redundant, but what do you think of:
```suggestion
        In the first form, they are enclosed between `r"` and `"`.
```
?

Geod24:
> On lines 549-550 it says you can write a named character entity like this: \name. But I tried doing: auto x = "\euro"; and it gave an error.

The correct syntax is: `auto x = "\&euro;";`

Geod24:
And thanks for your work!

spec/lex.dd
mark-summerfield: Yes, but lowercase and Titlecase are used elsewhere so this is a style decision that must be made by the documentation's owner or manager. I can just try to fix the English.

spec/lex.dd
```diff
@@ -542,7 +547,8 @@ $(H2 $(LNAME2 escape_sequences, Escape Sequences))
         $(D \U0001F603) represents the Unicode character U+1F603 (SMILING FACE
         WITH OPEN MOUTH).)
     $(TROW $(D \)$(I name), Named character entity from the HTML5
-        specification. See $(GLINK2 entity, NamedCharacterEntity).)
+        specification. These names begin with $(I &) and end with $(I ;), e.g., $D(&euro;).
+        See $(GLINK2 entity, NamedCharacterEntity).)
```
thewilsonator: You may need to escape the `&`, there seem to be some latex errors coming from those.

mark-summerfield: How do I escape &, I don't even know what markup language is being used; I've just tried to mimic what's there.

thewilsonator: The language is DDOC (see https://dlang.org/spec/ddoc). The escape for `&` is `&amp`.

thewilsonator: Or try `CODE_AMP` in place of `$(D &)`

adamdruppe: There is no correct answer here - this is ddoc's fatal design flaw. `&amp;` works if you are targeting HTML but is wrong for other targets like latex. The best you can do is whatever macro is defined in teh `.ddoc` file on your particular project.

Here it is `$(CODE_AMP)`.

thewilsonator:
Hmm, that still seems to fail. Maybe take that out and verify that is the case of the problem and do it in another PR?

thewilsonator:
Optimistically adding auto-merge

thewilsonator:
Hmm it failed, but not for the same reason:
```
http://downloads.dlang.org/releases/2.x/2.088.0/dmd.2.088.0.linux.tar.xz | tar -C ../generated/host_dmd-2.088.0 -Jxf - || rm -rf ../generated/host_dmd-2.088.0
../generated/host_dmd-2.088.0/dmd2/linux/bin64/dmd -conf=../generated/host_dmd-2.088.0/dmd2/linux/bin64/dmd.conf -of../generated/build build.d
../generated/build HOST_DMD="../generated/host_dmd-2.088.0/dmd2/linux/bin64/dmd" CXX="c++" OS=linux BUILD=release MODEL=64 AUTO_BOOTSTRAP="1" DOCDIR="" STDDOC="" DOC_OUTPUT_DIR="" MAKE="make" --called-from-make dmd
posix.mak:110: recipe for target 'dmd' failed
make[1]: *** [dmd] Terminated
posix.mak:611: recipe for target '.generated/dmd-2.090.0/generated/linux/release/64/dmd' failed
make: *** [.generated/dmd-2.090.0/generated/linux/release/64/dmd] Terminated
fatal: 'refs/ae-sys-d-cache/website-97da142fc48e01b0d2fbaf6055a95a21a4de9cb6-fc6a263ed9b6802d2e6ac01308d4159d' - not a valid ref
```

I'm not sure where that failed. @CyberShadow any guesses?

CyberShadow:
Something seems to have caused the host compiler to terminate while building the master compiler. I don't know what or why. Unless master is broken (which it doesn't seem to be), it is most likely a random failure (due to PID reuse or something of a similar nature). Should go away on a retry (rebase/amend and force-push).

Geod24:
Rebased on master, squashed the commits together, and removed a trailing whitespace on L361 of `lex.d`.

PR #2748, opened 2020-02-02
Minor English improvements

mark-summerfield:
1. What do lines 120/1 "If present, the ModuleDeclaration appears syntactically first in the source file" mean? Would something like "If present, the ModuleDeclaration must appear before any imports or code (but may be preceded by comments)." be clearer?
2. I would suggest adding a code comment to line 139: " // It is best to avoid -'s in file names".
3. I don't understand the para at lines 261-264 on scope searching.
4. In the static import section (lines 366+) should a distinction between compile-time and runtime be mentioned (if there is one)?
5. In the renamed imports section (lines 381+) some advice ought to be given as to when and why to use this -- or to recommend not to.

### DIFF
--- a/spec/module.dd
+++ b/spec/module.dd
@@ -117,8 +117,8 @@ $(P The $(I Identifier)s preceding the rightmost $(I Identifier) are the $(I Pac
 module is in. The packages correspond to directory names in the source file
 path. Package and module names cannot be $(GLINK_LEX Keyword)s.)
 
-$(P If present, the $(I ModuleDeclaration) appears syntactically first in the
-source file, and there can be only one per source file.)
+$(P If present, the $(I ModuleDeclaration) must be the first and only such declaration
+in the source file, and may be preceded only by comments.)
 
 $(P Example:)
 
--- a/spec/module.dd
+++ b/spec/module.dd
@@ -55,7 +55,7 @@ $(UL
 
         $(LI Modules do not inherit, they have no super modules, etc.)
 
-        $(LI Only one module per file.)
+        $(LI A file may contain only one module.)
 
         $(LI Module symbols can be imported.)
 
@@ -113,7 +113,7 @@ $(GNAME PackageName):
     $(GLINK_LEX Identifier)
 )
 
-$(P The $(I Identifier)s preceding the rightmost are the $(I Packages) that the
+$(P The $(I Identifier)s preceding the rightmost $(I Identifier) are the $(I Packages) that the
 module is in. The packages correspond to directory names in the source file
 path. Package and module names cannot be $(GLINK_LEX Keyword)s.)
 
@@ -127,9 +127,9 @@ module c.stdio; // module stdio in the c package
 ---------
 
 $(P By convention, package and module names are all lower case. This is because
-those names can have a one-to-one correspondence with the operating system's
-directory and file names, and many file systems are not case sensitive. All
-lower case package and module names will minimize problems moving projects
+these names can have a one-to-one correspondence with the operating system's
+directory and file names, and many file systems are not case sensitive. Using all
+lower case package and module names will avoid or minimize problems when moving projects
 between dissimilar file systems.)
 
 $(P If the file name of a module is an invalid module name (e.g.
@@ -139,7 +139,7 @@ $(P If the file name of a module is an invalid module name (e.g.
 module foo_bar;
 ---------
 
-$(P $(I ModuleDeclaration) can have an optional $(GLINK2 attribute,
+$(P A $(I ModuleDeclaration) can have an optional $(GLINK2 attribute,
 DeprecatedAttribute). The compiler will produce a message when the deprecated
 module is imported.)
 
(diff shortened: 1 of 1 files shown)


spec/module.dd
```diff
@@ -152,9 +152,9 @@ module bar;
 import foo;  // Deprecated: module foo is deprecated
 ---------
 
-$(P $(I DeprecatedAttribute) can have an optional $(ASSIGNEXPRESSION) argument to provide a
-more expressive message. The $(I AssignExpression) must evaluate at compile time to
-a string.)
+$(P A $(I DeprecatedAttribute) can have an optional $(ASSIGNEXPRESSION) argument to provide a
+more informative message. An $(I AssignExpression) must evaluate to a string at compile time.
+)
```
thewilsonator: I think the start of the second sentence is fine as is. i.e. keep the "The".

spec/module.dd
```diff
@@ -318,8 +318,8 @@ void test()
 $(H2 $(LNAME2 public_imports, Public Imports))
 
 $(P By default, imports are $(I private). This means that if module A imports
-module B, and module B imports module C, then names from C are visible only from
-B and not from A.)
+module B, and module B imports module C, then names inside C are visible only inside
+B and not inside A.)
```
thewilsonator: I think this is good as is.

spec/module.dd
```diff
@@ -127,9 +127,9 @@ module c.stdio; // module stdio in the c package
 ---------
 
 $(P By convention, package and module names are all lower case. This is because
-those names can have a one-to-one correspondence with the operating system's
-directory and file names, and many file systems are not case sensitive. All
-lower case package and module names will minimize problems moving projects
+these names can have a one-to-one correspondence with the operating system's
```
Geod24: I would drop the "can". It doesn't work in practice to have a different module name than path.

spec/module.dd
Geod24: True, not all assignexp needs to be CTFEable. But the change to the end of the sentence is 👍

spec/module.dd
```diff
@@ -177,8 +177,8 @@ import foo;  // Deprecated: module foo is deprecated - Please use foo2 instead.
     $(LI $(GLINK PackageName)s and $(GLINK ModuleName)s should be composed of the ASCII
     characters lower case letters, digits or `_` to ensure maximum portability and compatibility with
     various file systems.)
-    $(LI The file names for packages and modules should also be composed of
-    the ASCII characters lower case letters, digits or `_` and should not be a $(GLINK_LEX Keyword).)
+    $(LI The file names for packages and modules should be composed only of
```
Geod24: "must be composed of" ? We could use more RFC2119 in this text.

Geod24:
> Would something like "If present, the ModuleDeclaration must appear before any imports or code (but may be preceded by comments)." be clearer?

Yes. Although "code" seems a bit colloquial. What about "If present, the ModuleDeclaration must be the first declaration, and can only be preceded by comments."
Note that the module documentation has a special purpose, so the following is common:
```D
/**
    This module does stuff

    Description goes here
    Copyright: Me, 2020
*/
module foo
``

> would suggest adding a code comment to line 139

Sounds good, or a "Best practice" section. I personally prefer to phrase it as "file name must match module declaration", but your call.

> 3. I don't understand the para at lines 261-264 on scope searching.

Consider the following:
```D
void writeln(T...) (T args) { }
void main ()
{
   import std.stdio; // Contains `writeln`
   writeln("Hello World");
}
```

> In the static import section (lines 366+) should a distinction between compile-time and runtime be mentioned (if there is one)?

I'm not sure I understand the question. Symbols are always resolved at compile time.

This will not print anything, because of the two-pass approach. If it was a single pass, then the import inside `main` would provide the`writeln` symbol (as it is the "most enclosed").

> In the renamed imports section (lines 381+) some advice ought to be given as to when and why to use this -- or to recommend not to.

I would say (but that's my experience) that it is mostly useful when one needs to create a namespace. Such needs may arise from one's personal preference (e.g. wanting to have Go-style names) or to prevent conflict without requiring fully qualified name.

spec/module.dd
mark-summerfield: Using 'A' here is better English than 'The'.
So I'll leave you to change it if you insist:-)

spec/module.dd
mark-summerfield: I think that 'inside' is clearer. After all 'from' doesn't tell us where they originate (could be the imported module or one of the modules it publically imports).
Again, I'll leave you to revert (or presumably just reject).

spec/module.dd
thewilsonator: No, the `A` on line 155 is good. Its the `An` on line 156 that should stay a `The`. Sorry for the confusion.

spec/module.dd
mark-summerfield: I believe my version is clearer than the original. Quoting more RFC2119 is a separate matter.

spec/module.dd
```diff
@@ -113,12 +113,12 @@ $(GNAME PackageName):
...
 
-$(P The $(I Identifier)s preceding the rightmost are the $(I Packages) that the
+$(P The $(I Identifier)s preceding the rightmost $(I Identifier) are the $(I Packages) that the
 module is in. The packages correspond to directory names in the source file
 path. Package and module names cannot be $(GLINK_LEX Keyword)s.)
 
-$(P If present, the $(I ModuleDeclaration) appears syntactically first in the
-source file, and there can be only one per source file.)
+$(P If present, the $(I ModuleDeclaration) must be the first and only such declaration
+in the source file, and may be preceded only by comments.)
```
thewilsonator: This isn't technically true. `#line` directives may appear before the module declaration.

PR #2750, opened 2020-02-05
English improvements + sexism fix etc

mark-summerfield:
This article explains why D's `const` is the way it is.
But what this article does *not* answer -- and ought to, or at least provide a link to -- is why and how and *when* to use `const` and `immutable` (and `in` and `enum`!).

### DIFF
--- a/articles/const-faq.dd
+++ b/articles/const-faq.dd
@@ -32,19 +32,19 @@ $(ITEM const, Why does D have const?)
 
         $(LI It makes function interfaces more self-documenting. Without
         transitive const, for all pointer/reference parameters one must rely on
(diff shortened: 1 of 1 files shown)


articles/const-faq.dd
```diff
@@ -32,19 +32,19 @@ $(ITEM const, Why does D have const?)
 
         $(LI It makes function interfaces more self-documenting. Without
         transitive const, for all pointer/reference parameters one must rely on
-        the documentation (which is always missing, out of date, or wrong). Note
+        the documentation (which is may be missing, out of date, or wrong). Note
```
MoonlightSentinel:
```suggestion
        the documentation (which may be missing, out of date, or wrong). Note
```

PR #2756, opened 2020-02-19
Clarify 'body' keyword deprecation

Geod24:
The deprecation was reverted in DMD PR 10763.


deprecate.dd
```diff
@@ -92,6 +92,12 @@ $(H4 Rationale)
         which led to frequent trouble for people interfacing with other languages
         (e.g. javascript) or auto-generating code.
     )
+    $(P Not that given the large amount of code relying on the 'body' keyword
```
thewilsonator: Should that be `Note` not `not`?

Geod24: Right, fixed

PR #2757, opened 2020-02-19
Add documentation for pragma(crt_{con,de}structor)

Geod24:
```
It was implemented as part of DMD PR 7182 but never documented.
Note that significant fixes came later, e.g. DMD PR 10562.
```

Testing here because the docs don't build on OSX.

### DIFF
--- a/spec/pragma.dd
+++ b/spec/pragma.dd
@@ -60,6 +60,8 @@ $(H2 $(LEGACY_LNAME2 Predefined-Pragmas, predefined-pragmas, Predefined Pragmas)
 $(P All implementations must support these, even if by just ignoring them:)
 
 $(UL
+    $(LI $(LINK2 #crtctor, pragma crt$(UNDERSCORE)constructor))
+    $(LI $(LINK2 #crtdtor, pragma crt$(UNDERSCORE)destructor))
     $(LI $(LINK2 #inline, pragma inline))
     $(LI $(LINK2 #lib, pragma lib))
     $(LI $(LINK2 #linkerDirective, pragma linkerDirective))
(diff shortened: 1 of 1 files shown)


Geod24:
Any idea what I did wrong ? The CI output isn't very user-friendly.

Geod24:
```
LaTeX Warning: Hyper reference `predefined-pragmas' on page 156 undefined on in
put line 6868.

! Missing $ inserted.
<inserted text> 
                $
l.6876     \item pragma crt_
                            constructor
? 
! Emergency stop.
<inserted text> 
                $
l.6876     \item pragma crt_
                            constructor
!  ==> Fatal error occurred, no output PDF file produced!
Transcript written on .generated/dlangspec.log.
posix.mak:552: recipe for target '/dev/shm/dtest/work/repo/dlang.org/web/dlangspec.pdf' failed
```

Is what I'm seeing

thewilsonator:
Maybe it doesn't like the underscore?

thewilsonator:
That seems to work.

Geod24:
Yup, but the name of the pragma is `crt_constructor`... I'll reduce it to see what macro triggers it.

thewilsonator:
I know. 

I don't think that is a macro problem though, `_` is a latex thing. see .e.g. 
https://tex.stackexchange.com/questions/349869/how-to-write-in-latex-without-it-complaining
https://tex.stackexchange.com/questions/48632/underscores-in-words-text/48633

adamdruppe:
I know I drop this comment in a lot of these threads, but this is ddoc's fundamental design flaw. I pointed this out *eight years ago* and pointed out it isn't even that hard to fix... yet these problems remain due to simple stubbornness, so I had to create a new doc system myself which doesn't suffer from this misdesign.

We should call out how much time this poor design wastes and how many potential contributors it turns way every chance we get.

Geod24:
@adamdruppe : Do you happen to have a fix / workaround for this case ? 👼

adamdruppe:
I think it will work to modify this ESCAPES macro in latex.ddoc:

https://github.com/dlang/dlang.org/blob/master/spec/latex.ddoc#L91

```
ESCAPES=/&/&/ /</</ />/>/
```

Add `/_/\_/` to the end of that. That *should* do it in the `---` delimited and backtick surrounded code samples. Outside code samples, you must use a separate `$(UNDERSCORE)` macro instead of plain `_`. I see there is one defined in that latex.ddoc file as well. If there isn't a corresponding definition in the HTML output file (idk where that is), one will need to be added or ddoc will just disappear your characters.

this ESCAPES macro is (mis, but close enough) documented here btw
https://dlang.org/spec/ddoc.html

That ESCAPES thing is only applied inside backticks and dash-dash-dash code samples... and I *think* it can do any single byte character.

the latex `\_` might not quite work either, it might need to be `\underscore` or whatever else, I don't know latex very well.

thewilsonator:
All green.

Geod24:
Good, it means the issue does not trigger in backtips (it would be weird). I'll do one more test and squash everything, then ping for merge.

adamdruppe:
All green, but the underscores are deleted in the backtick samples. The ESCAPES thing might just help with that though, just don't forget that they were deleted in the previous debug commit.

spec/pragma.dd
```diff
@@ -70,6 +72,66 @@ $(UL
...
+pragma(crt_constructor)
+extern(C) void initializer() { initCount += 1; }
+-----------------
+
+    $(P It is useful for system programming and interfacing with C/C++,
+        for example to allow for initialization of the runtime when loading a DSO,
+        or as a simple replacement for $(D shared static this) in
+        $(LINK2 betterc.html, betterC mode).
+    )
+    $(P A module may contain any number of functions annotated with `crtconstructor`.
```
adamdruppe: here's a missing underscore in the backtick.

Geod24:
> All green, but the underscores are deleted in the backtick samples. 

Yeah I was testing to reduce the issue. First made it green by deleting all underscores, proving @thewilsonator 's theory, then checked if `$(UNDERSCORE)` works in a macro (just in one place, since it's just a test) and if underscore in backtips works, both tests succeeded. Now testing if underscore in `$(D ...)` macro works.
If that last test is successful, it means I'll just have to surround the pragma name with `$(D ...)` or backtips and be done with it. If it does not, I'll just surround them with backtips (which would slightly differ from the rest of the module, as `$(D ...)` is used for the `LINK2` and `LNAME2` macros.

adamdruppe:
aight. It should work fine in the `D` macro since that continues to process macros inside (contrast to backtick and `---` which don't).

anyway just wanted to make sure it wasn't an oversight.

btw thx for documenting this, it is one of those things i could swear I saw in PRs but weren't in teh docs a while ago so good to have it here.

im off to bed tonight tho ttyl

Geod24:
Finally fixed & ready to go

PR #2760, opened 2020-02-24
document printf format validation

WalterBright:
Rewrote the section about calling `printf` to support its implementation https://github.com/dlang/dmd/pull/10812

### DIFF
--- a/spec/betterc.dd
+++ b/spec/betterc.dd
@@ -84,6 +84,7 @@ $(H2 $(LNAME2 retained, Retained Features))
     $(LI `switch` with strings)
     $(LI `final switch`)
     $(LI `unittest`)
+    $(LI $(DDSUBLINK spec/interfaceToC, calling_printf, `printf` format validation))
     )
 
 $(H3 $(LNAME2 unittests, Running unittests in `-betterC`))
--- a/spec/interfaceToC.dd
+++ b/spec/interfaceToC.dd
@@ -208,30 +208,79 @@ extern (C)
 ---
 
 
-$(H2 $(LNAME2 calling_printf, Calling printf()))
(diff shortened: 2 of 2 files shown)


spec/betterc.dd
```diff
@@ -84,6 +84,7 @@ $(H2 $(LNAME2 retained, Retained Features))
     $(LI `switch` with strings)
     $(LI `final switch`)
     $(LI `unittest`)
+    $(DDSUBLINK spec/interfaceToC, calling_printf, `printf` format validation)
```
PetarKirov: Needs to be wrapped in `$(LI ... )` as it's part of `$(OL ..)` (ordered html list). I'll fix it myself.

PetarKirov: (Done: https://github.com/dlang/dlang.org/compare/175ea2ce01a84fb99d28ccb2d78b70196e1e949f..09364b2ea0a09e7d312682b473cabd4863af49f7)

spec/interfaceToC.dd
```diff
@@ -208,30 +208,79 @@ extern (C)
...
+
+$(H3 Non-Standard Format Specifiers)
+
+        $(P Non-Standard format specifiers will be rejected by the compiler.
+        Since the checking is only done for formats as string literals,
+        non-Standard ones can be used:
+        )
+---
+string format = "value: %K\n";
+printf(format, value);
```
adamdruppe: I very much doubt that will actually compile. Have you tested it?

I believe it will have to be `format.ptr`...

moon-chilled: Probably better to make the original `format` variable `const char*` so it gets properly null-terminated.

PR #2766, opened 2020-03-25
Indent continuation lines with 2 indentations

belka-ew:
It is a D Style proposal. The suggestion is to state explicitly how continuation lines should be indented.

One logical line can be written as two or more lines (so called line continuation or line wrapping). One example is a long argument list, that can be made more readable if written in multiple lines. And I propose to use two indentation levels (8 spaces) for line wrapping.

So:
```d
my_func_name(arg1,
        arg2);
```
instead of

```d
my_func_name(arg1,
    arg2);
```

### Rationale

- This is how dfmt formats the code and it can't be configured: https://github.com/dlang-community/dfmt/issues/257
- Phobos or at least some Phobos maintainers seem to enforce this guideline (e.g. see the remarks to this PR: https://github.com/dlang/phobos/pull/7105)

### DIFF
--- a/dstyle.dd
+++ b/dstyle.dd
@@ -27,6 +27,16 @@ $(UL
     $(LI Use spaces instead of hardware tabs.)
 
     $(LI Each indentation level will be four columns.)
+
+    $(LI For line breaks inside parentheses the continuation shall be indented
+         two indentation levels beyond the first line.
+
+-------------------------------
+myFunction(foo,
+        bar,
+        baz);
+-------------------------------
+    )
 )
 
 $(H3 $(LNAME2 naming_conventions, Naming Conventions))


FeepingCreature:
Does this also apply to call chains? Ie.

```
    arg.foo
        .bar // like this
            .baz; // or like this?
```

If not, maybe describe the affected lines better.

mdparker:
How about:

"Continuation lines shall be indented two indentation levels beyond the first line."

Then provide a three-line example for clarity.

belka-ew:
I added a clarification that only the breaks inside parentheses are affected.

RazvanN7 (approved):

PetarKirov:
@RazvanN7 Changes to the DStyle guide can't be made solely based on one's preference. Only changes match the currently established practices can be "codified" via a lightly reviewed PR. Otherwise such changes need the consensus of the majority of active contributors.
That said, such changes don't need a DIP, since this is not change that affects any users of the language, only the contributors.

PR #2768, opened 2020-04-10
function.dd: add notion of safe interface

WalterBright:
Expanded description of trusted functions to include the notion that such have a _safe interface_.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -2884,9 +2884,59 @@ $(H3 $(LNAME2 safe-functions, Safe Functions))
         so they can be corrected.
         )
 
+$(H4 Safe External Functions)
+
+        $(P External functions don't have a function body visible to the compiler:
+        )
+        ---
+        @safe extern (C) void play();
+        ---
+        and so safety cannot be verified automatically.
+
+        $(BEST_PRACTICE Explicitly set an attribute for external functions rather
+        than relying on default settings.)
+
 $(H3 $(LNAME2 trusted-functions, Trusted Functions))
 
         $(P Trusted functions are marked with the $(CODE @trusted) attribute.)
+
+        $(P Trusted functions act as an interface between safe and system functions.
+        A trusted function provides a safe interface. A $(I Safe Interface) is one that:
+        )
+
+        $(UL
+
+        $(LI Does not invalidate pointers that are passed to it. For example,
+        ---
+        @system void free(void* p);
+        ...
+        free(p);
+        *p = 3;  // p is invalid
+        ---
+        is not a safe interface because `free(p)` means `p` is no longer a valid
+        pointer when it returns. `free` can only be `@system`.)
+
+        $(LI Does not attempt to access memory beyond the bounds of the memory object
+        a parameter points to. For example:
+        ---
+        @system size_t strlen(char* p);
+        ---
+        is not a safe interface because the implementation of `strlen` advances `p` until
+        it finds a `null` character, and this can be arbitrarily beyond the `char` that `p`
+        purportedly points to. `strlen` can only be `@system`. Any function that traverses
+        a C string passed as an argument can only be `@system`.
+        )
+
+        $(LI Does not have array bounds passed as a parameter separate from the pointer
+        type. For example:
+        ---
+        @system void* memcpy(void* dst, void* src, size_t nbytes);
+        ---
+        is not a safe interface because the extent of `dst` and `src` is set by `nbytes`, and
+        so cannot be verified by the compiler. Such functions can only be `@system`.
+        )
+
+        )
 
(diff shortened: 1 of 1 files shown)


spec/function.dd
```diff
@@ -2884,10 +2884,60 @@ $(H3 $(LNAME2 safe-functions, Safe Functions))
...
+        @system size_t strlen(char* p);
+        ---
+        is not a safe interface because the implementation of `strlen` advances `p` until
+        it finds a `null` character, and this can be arbitrarily beyond the `char` that `p`
+        purportedly points to. `strlen` can only be `@system`. Any function that traverses
+        a C string passed as an argument can only be `@system`.
+        )
+
+        $(LI Has array bounds passed as a separate parameter rather than as part of the pointer
+        type. For example:
```
Herringway, line 2930: This third item seems to be the opposite of what it should be. Shouldn't it be that array bounds as a separate parameter are unsafe, while the other is safe?

spec/function.dd
```diff
@@ -2884,10 +2884,60 @@ $(H3 $(LNAME2 safe-functions, Safe Functions))
...
         )
 
+$(H4 Safe External Functions)
+
+        $(P External functions don't have a function body visible to the compiler:
+        ---
+        @safe extern (C) void play();
+        ---
+        and so safety cannot be verified automatically.
+        )
```
aG0aep6G, line 2894: Don't put code blocks into `$(P ...)`. HTML doesn't allow it. The empty second paragraph in the output is because of this.

spec/function.dd
```diff
@@ -2884,10 +2884,60 @@ $(H3 $(LNAME2 safe-functions, Safe Functions))
...
+
+        $(BEST_PRACTICE Explicitly set an attribute for external functions rather
+        than relying on default settings.)
```
aG0aep6G, line 2897: A "best practice" doesn't absolve DIP 1028 silently changing the default for the worse.

spec/function.dd
```diff
@@ -2884,10 +2884,60 @@ $(H3 $(LNAME2 safe-functions, Safe Functions))
...
+
 $(H3 $(LNAME2 trusted-functions, Trusted Functions))
 
         $(P Trusted functions are marked with the $(CODE @trusted) attribute.)
 
+        $(P Trusted functions act as an interface between safe and system functions.
+        A trusted function provides a safe interface. A $(I Safe Interface) is one that:
+
+        $(UL
```
aG0aep6G, line 2906: Same thing as with the code block above: Do not put `$(UL ...)` into `$(P ...)`. HTML doesn't allow it. The empty third paragraph in the output is because of this.

spec/function.dd
```diff
@@ -2884,10 +2884,60 @@ $(H3 $(LNAME2 safe-functions, Safe Functions))
...
+        type. For example:
+        ---
+        @system void* memcpy(void* dst, void* src, size_t nbytes);
+        ---
+        is not a safe interface because the extent of `dst` and `src` is set by `nbytes`, and
+        so cannot be verified by the compiler. Such functions can only be `@system`.
+        )
+
+        )
+        )
```
aG0aep6G, line 2939: I suggest merging this whole new section into the one below it ("Trusted functions are guaranteed to [...]"). The existing paragraph already says that `@trusted` functions must be safe. Maybe add the phrase "Safe Interface" there. And there is already a block of examples of things that can't be `@trusted`. Maybe add your examples there.

spec/function.dd
WalterBright: good catch

spec/function.dd
WalterBright: ok

spec/function.dd
WalterBright: This is not about DIP 1028.

spec/function.dd
WalterBright: ok

spec/function.dd
WalterBright: It's better in the trusted section because many people misunderstand what trusted is for.

aG0aep6G: > It's better in the trusted section

I'm not saying to move it out of the trusted section. Just merge it better with the rest of it. The text is somewhat repeating itself right now.

Instead of this structure:

* your definition of `@trusted` with "safe interface"
* examples of things that can't be `@trusted`
* existing definition of `@trusted`
* more examples of things that can't be `@trusted`

Do this:

* Expand existing definition of `@trusted` to include "safe interface".
* Add new examples next to existing examples.

spec/function.dd
aG0aep6G: > This is not about DIP 1028.

Sure.

PR #2773, opened 2020-04-18
safe interfaces, safe values, safe aliasing

aG0aep6G:
This is based on Walter's #2768, but I got carried away so this heavily remixes and expands the original.

The goal here is being able to tell exactly what `@safe` and `@trusted` functions can
assume on their inputs and what they must ensure for their outputs.

@WalterBright, you probably want to have a look.
@dkorpel, as the author of the upcoming [DIP on `@system` variables](https://github.com/dlang/DIPs/pull/179) you might be interested.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -2836,16 +2836,13 @@ $(H2 $(LNAME2 nogc-functions, No-GC Functions))
 
 $(H2 $(LNAME2 function-safety, Function Safety))
 
-        $(P $(I Safe functions) are functions that are statically checked
-        to exhibit no possibility of
-        $(DDSUBLINK glossary, undefined_behavior, $(I undefined behavior)).
-        Undefined behavior is often used as a vector for malicious
-        attacks.
-        )
-
 $(H3 $(LNAME2 safe-functions, Safe Functions))
 
         $(P Safe functions are marked with the $(CODE @safe) attribute.)
+
+        $(P Safe functions have $(RELATIVE_LINK2 safe-interfaces, safe
+        interfaces). An implementation must enforce this by restricting the
+        function's body to operations that are known safe.)
 
         $(P The following operations are not allowed in safe
         functions:)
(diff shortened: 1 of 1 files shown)


aG0aep6G:
This has been sitting idle for over a month. Anyone?

spec/function.dd
```diff
@@ -2952,6 +2960,227 @@ $(H3 $(LNAME2 system-functions, System Functions))
...
+        )
+
+$(H3 $(LNAME2 safe-aliasing, Safe Aliasing))
+
+    $(P When one memory location is accessible with two different types, that
+    aliasing is considered safe in these cases:)
+    $(OL
+        $(LI both types are `const` or `immutable`; or)
+        $(LI one of the types is mutable while the other is a `const`-qualified
+            $(DDSUBLINK spec/type, basic-data-types, basic data type); or)
```
dkorpel, line 3150: It would be nice if there were some general rule that avoided explicitly mentioning basic data types and static arrays with length 0. 
I'm not sure if the list is complete currently, if I interpret it correctly, a `struct Int {int x;}` may not alias with an `int y`, but two raw `int`s can?

spec/function.dd
```diff
@@ -2836,17 +2836,14 @@ $(H2 $(LNAME2 nogc-functions, No-GC Functions))
 
 $(H2 $(LNAME2 function-safety, Function Safety))
 
-        $(P $(I Safe functions) are functions that are statically checked
-        to exhibit no possibility of
-        $(DDSUBLINK glossary, undefined_behavior, $(I undefined behavior)).
```
dkorpel, line 2841: I think it's still useful to mention that safe functions cannot exhibit undefined behavior.

spec/function.dd
```diff
@@ -2952,6 +2960,227 @@ $(H3 $(LNAME2 system-functions, System Functions))
         $(P System functions are $(B not) covariant with trusted or safe functions.
         )
 
+        $(BEST_PRACTICE When in doubt, mark `extern (C)` and `extern (C++)` functions as
+        `@system` when their implementations are not in D, as the D compiler will be
+        unable to check them. Most of them are `@safe`, but will need to be manually
```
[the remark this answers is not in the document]
aG0aep6G: > I don't like this sentence because a. saying something about 'most functions not implemented in D' is a bit dubious and b. `@safe` has no clear definition for implementations in other languages.

Yeah, that's Walter's. I agree, that the claim about "most" foreign functions having safe interfaces is dubious. And even if they do, they should marked `@trusted`, not `@safe`.

But it's a "best practice" (i.e. ultimately meaningless), and it correctly says: when in doubt, `@system`. So I feel like that's good enough for me.

spec/function.dd
```diff
@@ -2952,6 +2960,227 @@ $(H3 $(LNAME2 system-functions, System Functions))
...
+
+        $(P A static array is safe when all its elements are safe. Regardless
+        of the element type, a static array with length zero is always safe.)
+
+        $(P An associative array is safe when all its keys and elements are
+        safe.)
+
+        $(P A struct/union instance is safe when:)
+        $(OL
+            $(LI the values of its accessible fields are safe, and)
```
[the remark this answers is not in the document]
aG0aep6G: > What counts as a non-accessible field?

It's meant in terms of `public`, `private`, etc. For example, a `private` field is not accessible, unless your code is in the same module. Maybe making "accessible" a link to <https://dlang.org/spec/attribute.html#visibility_attributes> could make this clearer?

This provides some support for the current practice of using `private` to hide unsafe implementations. With this rule, at least using those types is kosher. Of course, the implementations are still breaking the rules. Which is where `@system` variables come in. And when those land, it might make sense to revisit this.

spec/function.dd
aG0aep6G: > It would be nice if there were some general rule that avoided explicitly mentioning basic data types and static arrays with length 0.
> I'm not sure if the list is complete currently,

I think I just gave up at some point with aliasing. I don't think that zero-length arrays are particularly noteworthy, but they're simple to reason about. Structs are hard, with anonymous unions, `align`, and whatnot.

> if I interpret it correctly, a `struct Int {int x;}` may not alias with an `int y`, but two raw `int`s can?

That's right. But the struct case should be added. It even passes `@safe` checks. And at least any aliasing that (correctly) passes as `@safe` should be considered "safe aliasing" by the spec.

spec/function.dd
aG0aep6G: > I think it's still useful to mention that safe functions cannot exhibit undefined behavior.

That's moved to the definition of safe interfaces. Safe functions have safe interfaces. Functions with safe interfaces cannot exhibit undefined behavior.

I don't think it's particularly useful to repeat that aspect of safe interfaces in the definition of safe (and trusted) functions. But I wouldn't mind if someone added it back.

thewilsonator:
@dkorpel Thanks for the extra review. Please open another PR if you feel there is anything that needs to be changed.

PR #2775, opened 2020-04-18
Added missing argument in compile command example

blackbird806:
The current compile line example does not works because the C++ sample code use std. The option -L-lstdc++ was missing in the compile line example, which is needed in order to compile the sample code.


PetarKirov:
Thanks, @blackbird806!

PR #2777, opened 2020-04-23
Spec: Add FuncAttrLive

kubo39:
See also https://github.com/dlang/dmd/pull/11049

### DIFF
--- a/spec/abi.dd
+++ b/spec/abi.dd
@@ -536,6 +536,7 @@ $(GNAME FuncAttr):
     $(GLINK FuncAttrScope)
     $(GLINK FuncAttrTrusted)
     $(GLINK FuncAttrSafe)
+    $(GLINK FuncAttrLive)
 )
 
     $(P Function attributes are emitted in the order as listed above.)
@@ -567,6 +568,9 @@ $(GNAME FuncAttrTrusted):
 
 $(GNAME FuncAttrSafe):
     $(B Nf)
+
+%(GNAME FuncAttrLive):
+    %(B Nm)
 
 $(GNAME Parameters):
     $(GLINK Parameter)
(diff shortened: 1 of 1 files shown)


PetarKirov:
Can you also add `Nk` for return parameters (https://github.com/dlang/druntime/commit/87fbb2885462f6a869161aa581efa21231d693d3)? 

BTW, I wonder why there is need for both `Parameter` and `Parameter2`. Instead it would be simpler if we defined it as:

```
Parameter:
    Type
    J Parameter     // out
    K Parameter     // ref
    L Parameter     // lazy
    M Parameter   // scope
    Nk Parameter  // return
```

@ibuclaw @WalterBright what do you think?

spec/abi.dd
```diff
@@ -581,6 +585,7 @@ $(GNAME Parameter2):
     $(B J) $(GLINK Type)     $(GREEN // out)
     $(B K) $(GLINK Type)     $(GREEN // ref)
     $(B L) $(GLINK Type)     $(GREEN // lazy)
+    $(B Nk) $(GLINK Type)     $(GREEN // return)
```
PetarKirov, line 588: I think that if we keep the current split between `Parameter` and `Parameter2`, `Nk` should go next to `scope` in `Parameter`.

PetarKirov: (Because both `scope` and `return` can be combined with `out` and `ref`.)

kubo39: Yes, I missed. Thanks for pointing out!

PetarKirov: No worries, and excuse me for not being specific the first time I suggested adding `Nk` to the docs!

PetarKirov:
I'll wait for the dmd pull request to be merged before merging this one.

PR #2783, opened 2020-05-03
Add AVR version identifier to documentation

ErnyTech:
See https://github.com/dlang/dmd/pull/11094


JohanEngelen:
Can you be more specific in the description?
For example, add "8-bit Atmel AVR Microcontrollers" to make clear this does not include the AVR32 cpus.

ErnyTech:
@JohanEngelen Done

PR #2788, opened 2020-05-06
Overhaul the donate page

mdparker:
Revise the foundation/donate page to clarify how donations can be targeted, how the different funds are used, and how to set task bounties.


mdparker:
So what's this then? From DAutoTest:

"../dmd/src/dmd/doc.d(368,34): Error: file `"default_ddoc_theme.ddoc"` cannot be found or not in a path specified with `-J`"

thewilsonator:
Not sure, I've seen this in a couple of other PRs. No idea what cases it though.

PR #2790, opened 2020-05-07
Add documentation for @live functions

WalterBright:
Since it's in the compiler, it needs documentation.

### DIFF
--- a/spec/ob.dd
+++ b/spec/ob.dd
@@ -1,0 +1,309 @@
+Ddoc
+
+$(SPEC_S Live Functions,
+
+$(HEADERNAV_TOC)
+
+$(B Experimental, Subject to Change)
+
+$(P If a memory object has only one pointer to it, that pointer is the $(I owner)
+of the memory object. With the single owner, it becomes straightforward to
+manage the memory for the object. It also becomes trivial to synchronize access
+to that memory object among multiple threads, because it can only be accessed
+by the thread that controls that single pointer.
+)
+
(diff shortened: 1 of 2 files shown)


12345swordy:
I think that @atilaneves needs admin access now, sense that andralex has step down from leadership.

spec/ob.dd
```diff
@@ -0,0 +1,309 @@
...
+$(BLOCKQUOTE At any point in the program, for each memory object,
+there is exactly one live mutable pointer to it or all the live pointers to
+it are read-only.)
+
+$(P Function declarations annotated with the `@live` attribute are checked for compliance with
+the Ownership/Borrowing rules. The checks are run after other semantic processing is complete.
+The checking does not influence code generation.
+)
+
+$(P Whether a pointer is allocates memory using the GC or some other
```
baryluk, line 91: Some grammatical error here with `is allocates`?

WalterBright: fixed

spec/ob.dd
```diff
@@ -0,0 +1,309 @@
...
+if non-`@live` functions call `@live` functions, arguments passed are
+expected to follow `@live` conventions.
+)
+
+$(P It will not detect attempts to dereference `null` pointers or possibly
+`null` pointers. This is unworkable because there is no current method
+of annotating a type as a non-`null` pointer.
+)
+
+$(H4 Tracked Pointers)
```
kubo39, line 112: Is it H3, not H4?

PR #2791, opened 2020-05-08
class allocator receives size_t not uint

baryluk:
When running gdc 9.3.0:

```
decl.d:45:12: error: allocator decl.XClass.new first argument must be type size_t, not uint
```

### DIFF
--- a/spec/class.dd
+++ b/spec/class.dd
@@ -997,7 +997,7 @@ $(GNAME Allocator):
 $(P A class member function of the form:)
 
 ------
-new(uint size)
+new(size_t size)
 {
     ...
 }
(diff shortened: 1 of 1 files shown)


spec/class.dd
```diff
@@ -997,15 +997,15 @@ $(GNAME Allocator):
...
+new(size_t size)
 {
     ...
 }
 ------
 
         is called a class allocator.
         The class allocator can have any number of parameters, provided
-        the first one is of type uint.
+        the first one is of type size_t.
```
thewilsonator, line 1008: try 
```
the first one is of type $(D size_t).
```

kubo39: I guess this is causing an error in the PDF generator.

PR #2795, opened 2020-05-11
merge stable

MartinNowak:
- update download and changelog for v2.092.0


MartinNowak:
Not sure what broke the ddoc build here @wilzbach, sth. about the string import path of dmd.

wilzbach:
See https://github.com/dlang/dmd/pull/11125#issuecomment-626901284

PR #2796, opened 2020-05-11
Remove issuestats badge

kubo39:
(Edited): The issuestats service is no longer up.


kubo39:
I think this is good to go!

PR #2808, opened 2020-06-02
[class.dd] Move Objective-C method specifics to subsection

ntrel:
The Objective-C method specifics (added in #2540) make it hard to read the docs for people not interested in Objective-C linkage. This moves the details to a subsection.

Also:
* Mention that member functions are also called methods (which is used in some places in class.dd already).
* 2 minor tweaks to surrounding class docs.


jacob-carlborg:
Fine by me.

PR #2814, opened 2020-06-05
Fix spec wrt. the align attribute

kinke:
The semantics were changed with https://github.com/dlang/dmd/pull/6754.

Resolves https://issues.dlang.org/show_bug.cgi?id=20754.

### DIFF
--- a/spec/attribute.dd
+++ b/spec/attribute.dd
@@ -221,7 +221,8 @@ struct S
     int b;    // placed at offset 4
     long c;   // placed at offset 8
 }
-auto sz = S.sizeof;  // 16
+static assert(S.alignof == 8);
+static assert(S.sizeof == 16);
 --------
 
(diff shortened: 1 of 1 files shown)


veelo:
I am in favour of https://github.com/dlang/dmd/pull/6754, but I don’t think the language should be changed silently (https://forum.dlang.org/post/dsklfejzyfmytckeuhby@forum.dlang.org). Since dmd has worked this way for more than three years now I guess one could argue that we have exceeded the deprecation period anyway, but I’m not sure everybody would see it that way. A change log entry would be in order, maybe?

kinke:
A changelog entry for something that was changed/fixed 3 years ago? C'mon. The previous scheme was just plain wrong, resulting in stuff like `T.alignof = 4, T.sizeof = 10` etc.

PR #2817, opened 2020-06-07
adding keyring hint to the install.sh help page

AntonOks:
Add the keyring update info, which is a continuous reoccurring confusion, see also https://forum.dlang.org/post/tmnixwomwogmgjopsjqp@forum.dlang.org

### DIFF
--- a/install.dd
+++ b/install.dd
@@ -309,7 +309,7 @@ $(CONSOLE
 
 $(H3 $(LNAME2 update, Update))
 
-Update the installer itself.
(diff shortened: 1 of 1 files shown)


AntonOks:
> 
> 
> Thanks! Though we really should improve the installer to auto-update or detect when the keyring/script has been upgraded.

true.
can you point me to the install.sh script?

wilzbach:
https://github.com/dlang/installer/blob/master/script/install.sh

PR #2832, opened 2020-07-04
Improve specs for `__traits(isSame)` over two tuples

NilsLankila:
Related bug fix_ https://github.com/dlang/dmd/pull/11373


NilsLankila:
ping, compiler change is merged now.

PR #2844, opened 2020-08-12
ddoc: Better define section names

jblachly:
Clarify difference in section and section name. Also explicilty describe that section names may not contain spaces, but that underscores will be replaced with spaces.

### DIFF
--- a/spec/ddoc.dd
+++ b/spec/ddoc.dd
@@ -199,9 +199,15 @@ $(H3 $(LNAME2 sections, Sections))
 
 $(P
 The document comment is a series of $(I Section)s.
-A $(I Section) is a name that is the first non-blank character on
+A $(I Section name) is a name that is the first non-blank character on
 a line immediately followed by a ':'. This name forms the section name.
 The section name is not case sensitive.
+)
+
+$(P
+Section names do not contain whitespace. Underscores will be replaced with spaces.
+For example, a section entitled 'New API' would be written with having its first line
+'New_API:'
 )
 
 $(P


spec/ddoc.dd
```diff
@@ -199,11 +199,17 @@ $(H3 $(LNAME2 sections, Sections))
 
 $(P
 The document comment is a series of $(I Section)s.
-A $(I Section) is a name that is the first non-blank character on
+A $(I Section name) is a name that is the first non-blank character on
 a line immediately followed by a ':'. This name forms the section name.
 The section name is not case sensitive.
```
Hackerpilot, line 204: This seems to not actually be true. I just tried writing `section_name:` vs `Section_name:`, and only the upper-case version is treated as a section name by ddoc in DMD 2.093.0.

jblachly: ddoc, while a fantastic idea, leaves a lot to be desired. At least we are moving toward markdown.

spec/ddoc.dd
```diff
@@ -199,11 +199,17 @@ $(H3 $(LNAME2 sections, Sections))
...
 )
 
+$(P
+Section names do not contain whitespace. Underscores will be replaced with spaces.
+For example, a section entitled 'New API' would be written with having its first line
+'New_API:'
```
mdparker, line 210: "written with having its first line" is rather awkward. And it isn't accurate. The section name doesn't need to be the first line, but at the front of the first line. So something like:

"For example, the first line of a section titled 'New API' would begin with 'New_API:'."

mdparker (changes requested):

spec/ddoc.dd
jblachly: Strike "with" and change "would" to "could" makes it both grammatical and semantically accurate.

jblachly:
The response to this PR is a symptom of something bigger

PR #2845, opened 2020-08-16
Expand on function parameters documentation, add -preview=in

Geod24:
Underline the difference between argument and parameter.
Split the first sentence between type constructors and storage classes.
Finally, add documentation on what `in` will means with `-preview=in`.

Waiting on dlang/dmd#11000

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -1219,39 +1219,58 @@ $(H2 $(LNAME2 parameters, Function Parameters))
 
 $(H3 $(LNAME2 param-storage, Parameter Storage Classes))
 
-        $(P Parameter storage classes are $(D in), $(D out),
-        $(D ref), $(D lazy), $(D const), $(D immutable), $(D shared),
-        $(D inout) or
-        $(D scope).
+        $(P Parameter storage classes are $(D in), $(D out), $(D ref), $(D lazy), $(D scope) and $(D inout).
+        Parameters can also take the type constructors $(D const), $(D immutable) and $(D shared).)
+
+        $(P $(D in), $(D ref), $(D out) and $(D lazy) are mutually exclusive. The first three are used to
+        denote input, input/output, and output parameters, respectively.
         For example:
         )
 ------
-int foo(in int x, out int y, ref int z, int q);
+int read(in char[] input, ref size_t count, out int errno);
 ------
 
-        $(P x is $(D in), y is $(D out), z is $(D ref), and q is none.
-        )
+        $(P In this example, $(D input) will only be read and no reference to it will be kept around,
+        while $(D count) will be read and written to, and $(D errno) will be set to a value from
+        within the function.
+        This approach gives a semantic meaning to the parameters and allows the compiler to potentially
+        optimize the generated code.)
 
-        $(UL
-        $(LI The function declaration makes it clear what the inputs and
-        outputs to the function are.)
-        $(LI It eliminates the need for IDL (interface description language) as a separate language.)
-        $(LI It provides more information to the compiler, enabling more
-        error checking and
-        possibly better code generation.)
+        $(P When talking about function parameters, the $(I parameter) is what the function prototype defines,
+        and the $(I argument) is the value that will $(I bind) to it.
+        For example, using the previous definition of $(D read):
         )
+------
+void main()
+{
+    size_t a = 42;
+    int b;
+    int r = read("Hello World", a, b);
+}
+------
+       $(P In this example, the argument $(D "Hello World") gets bind to $(D input),
+       $(D a) gets bind to $(D count) and $(D b) to $(D errno).
+       )
 
     $(TABLE_2COLS Parameter Storage Classes,
     $(THEAD Storage Class, Description)
-    $(TROW $(I none), parameter becomes a mutable copy of its argument)
-
-    $(TROW $(D in), defined as $(D scope const).  However $(D in) has not yet been properly
-    implemented so it's current implementation is equivalent to $(D const).  It is recommended
-    to avoid using $(D in) until it is properly defined and implemented.  Use $(D scope const)
-    or $(D const) explicitly instead.)
-    $(TROW $(D ref), parameter is passed by reference)
-    $(TROW $(D out), parameter is passed by reference and initialized upon function entry with the default value
-    for its type)
+    $(TROW $(I none), The parameter will be a mutable copy of its argument)
+    $(TROW $(D in), The parameter is an input to the function. Input parameters behaves as if they have
+    the $(D const scope) storage classes. Input parameters may be passed by reference by the compiler.
+    Unlike $(D ref) parameters$(COMMA) $(D in) parameters can bind to both lvalues and rvalues (such as literals).
+    Types that would trigger a side effect if passed by value (such as types with postblit$(COMMA)
+    copy constructor$(COMMA) or destructor)$(COMMA) and types which cannot be copied$(COMMA)
+    e.g. if their copy constructor is marked as $(D @disable)$(COMMA) will always be passed by reference.
+    Dynamic arrays$(COMMA) classes$(COMMA) associative arrays$(COMMA) function pointers$(COMMA) and delegates
+    will always be passed by value$(COMMA) to allow for covariance.
+    If the type of the parameter does not fall in one of those categories$(COMMA)
+    whether or not it is passed by reference is implementation defined$(COMMA) and the backend is free
+    to choose the method that will best fit the ABI of the platform.
+    $(B Note: This requires the $(D -preview=in) switch$(COMMA) available in v2.094.0 or higher.))
+    $(TROW $(D ref), The parameter is an $(I input/output) parameter$(COMMA) passed by reference.
+    It is expected that the function will both read and write to this parameter.)
+    $(TROW $(D out), The argument must be an lvalue$(COMMA) which will be passed by reference and initialized
+    upon function entry with the default value (`T.init`) of its type)
 
     $(TROW $(D scope), $(ARGS
     The parameter must not escape the function call


Geod24:
@ibuclaw @kinke : Does the doc work for you, too ?

kinke:
Wrt. params and args, I usually explain that args are used at the call sites, and params by the callee.

Wrt. `in` and new `-preview=in`, it's IMO best to think of it as being passed by ref, similar to C++ `const T&`, and potentially being optimized to pass-by-value. That's important because of aliasing and the ref-param potentially being mutated via another param, global state or another thread.

spec/function.dd
```diff
@@ -1219,39 +1219,57 @@ $(H2 $(LNAME2 parameters, Function Parameters))
...
+        optimize the generated code.)
 
-        $(UL
-        $(LI The function declaration makes it clear what the inputs and
-        outputs to the function are.)
-        $(LI It eliminates the need for IDL (interface description language) as a separate language.)
-        $(LI It provides more information to the compiler, enabling more
-        error checking and
-        possibly better code generation.)
+        $(P When talking about function parameter, the $(I parameter) is what the function prototype defines,
```
ibuclaw, line 1239: When talking about function parameters?

spec/function.dd
```diff
@@ -1219,39 +1219,57 @@ $(H2 $(LNAME2 parameters, Function Parameters))
...
+    the $(D const scope) storage classes. Input parameters can be passed by reference.
+    If they get passed by reference, unlike with $(D ref), rvalues (such as literals) can be bound to them.
+    Types that would trigger a side effect if passed by value (such as types with postblit,
+    copy constructor, or destructor), and types which cannot be copied, e.g. if their copy constructor is
+    marked as $(D @disable), will always be passed by reference. Dynamic arrays, classes,
+    associative arrays, function pointers and delegates will always be passed by value,
+    to allow for covariance. If a parameter is of a type that does not fall in one of those categories,
+    whether or not it is passed by reference is implementation defined, and the backend is free
+    to choose the method that will best fit the ABI of the platform.
+    $(B Note: This require the $(D -preview=in) switch available in v2.094.0 or higher.))
```
ibuclaw, line 1268: Something is really, really wrong here for the website to be rendered list [this](http://dtest.dlang.io/artifact/website-94abd4f028d79da8757bca9d1cf8d5d3e7f68de3-3a34ed1ce175aef09784ac4a45b2da1f/web/spec/function.html#param-storage).

ibuclaw: Looks like the problem is your use of commas.  There should be a ddoc macro to use instead.

spec/function.dd
```diff
@@ -1219,39 +1219,57 @@ $(H2 $(LNAME2 parameters, Function Parameters))
...
+int read(in char[] input, ref size_t count, out int errno);
 ------
 
-        $(P x is $(D in), y is $(D out), z is $(D ref), and q is none.
-        )
+        $(P In this example, $(D input) will only be read and no reference to it will be kept around,
+        while $(D count) will be read and written to, and $(D errno) will be set to a value from
+        within the function.
+        This approach gives a semantic meaning to the parameters and allow the compiler to potentially
+        optimize the generated code.)
```
ibuclaw, line 1237: allows the compiler to generate code that has more aggressive optimizations based on these assumptions. ?

spec/function.dd
ibuclaw: Should be `$(COMMA)`.

ibuclaw (dismissed):
Needs some adjustments.

spec/function.dd
Geod24: I'm not sure what the difference between those two sentences are, in practice ? I also try to stay away from terms which could have a negative connotation, such as "aggressive".

Geod24:
Fixed. Thanks for the tip about `$(COMMA)` @ibuclaw !

spec/function.dd
ibuclaw: I think I was trying to phrase it in a more direct manner, to make it clear that circumventing the guarantee is not a wrong-code compiler bug.

Geod24: I'm not sure I know which guarantee you refer to ? Maybe "allow the compiler to potentially" => "could allow the compiler to" ?

spec/function.dd
```diff
@@ -1219,39 +1219,59 @@ $(H2 $(LNAME2 parameters, Function Parameters))
...
-    $(TROW $(D in), defined as $(D scope const).  However $(D in) has not yet been properly
-    implemented so it's current implementation is equivalent to $(D const).  It is recommended
-    to avoid using $(D in) until it is properly defined and implemented.  Use $(D scope const)
-    or $(D const) explicitly instead.)
-    $(TROW $(D ref), parameter is passed by reference)
-    $(TROW $(D out), parameter is passed by reference and initialized upon function entry with the default value
-    for its type)
+    $(TROW $(I none), The parameter will be a mutable copy of its argument)
+    $(TROW $(D in), The parameter is an input to the function. Input parameters behaves as if they have
+    the $(D const scope) storage classes. Input parameters can be passed by reference.
```
AndrejMitrovic, line 1259:
```suggestion
    the $(D const scope) storage classes. Input parameters may be passed by reference by the compiler.
```

The distinction is important: the user has no way of controlling this behavior, and "can be passed" reads like the user has control over the behavior.

spec/function.dd
```diff
@@ -1219,39 +1219,59 @@ $(H2 $(LNAME2 parameters, Function Parameters))
...
+    If they get passed by reference$(COMMA) unlike with $(D ref)$(COMMA) rvalues (such as literals)
+    can be bound to them.
```
AndrejMitrovic, line 1261: This sentence was very strange. It said "if the parameter is already passed by reference" .. "then rvalues can bind to them". I suggest:

```suggestion
    Unlike $(D ref) parameters$(COMMA), `$(D in)` parameters can bind to both lvalues and rvalues (such as literals).
```

spec/function.dd
```diff
@@ -1219,39 +1219,59 @@ $(H2 $(LNAME2 parameters, Function Parameters))
...
+    Types that would trigger a side effect if passed by value (such as types with postblit$(COMMA)
```
AndrejMitrovic, line 1262: What happens to rvalues with copy constructors? I think this section _implies_ that we're passing in an lvalue. But maybe we should document this explicitly - "If the parameter is an lvalue and its type would trigger.."?

spec/function.dd
```diff
@@ -1219,39 +1219,59 @@ $(H2 $(LNAME2 parameters, Function Parameters))
...
+    copy constructor$(COMMA) or destructor)$(COMMA) and types which cannot be copied$(COMMA)
+    e.g. if their copy constructor is marked as $(D @disable)$(COMMA) will always be passed by reference.
+    Dynamic arrays$(COMMA) classes$(COMMA) associative arrays$(COMMA) function pointers and delegates
```
AndrejMitrovic, line 1265:
```suggestion
    Dynamic arrays$(COMMA) classes$(COMMA) associative arrays$(COMMA) function pointers$(COMMA) and delegates
```

Oxford comma, please.

spec/function.dd
```diff
@@ -1219,39 +1219,59 @@ $(H2 $(LNAME2 parameters, Function Parameters))
...
+    will always be passed by value$(COMMA) to allow for covariance.
+    If a parameter is of a type that does not fall in one of those categories$(COMMA)
```
AndrejMitrovic, line 1267:
```suggestion
    If a parameter's type does not fall in one of those categories$(COMMA)
```

Alternatively:

```suggestion
    If the type of the parameter does not fall in one of those categories$(COMMA)
```

spec/function.dd
```diff
@@ -1219,39 +1219,59 @@ $(H2 $(LNAME2 parameters, Function Parameters))
...
+    whether or not it is passed by reference is implementation defined$(COMMA) and the backend is free
```
AndrejMitrovic, line 1268:
```suggestion
    the choice of whether or not it is passed by reference is implementation defined$(COMMA) and the backend is free
```

spec/function.dd
```diff
@@ -1219,39 +1219,59 @@ $(H2 $(LNAME2 parameters, Function Parameters))
...
+    to choose the method that will best fit the ABI of the platform.
+    $(B Note: This require the $(D -preview=in) switch available in v2.094.0 or higher.))
```
AndrejMitrovic, line 1270:
```suggestion
    $(B Note: This requires the $(D -preview=in) switch$(COMMA) available in v2.094.0 or higher.))
```

spec/function.dd
```diff
@@ -1219,39 +1219,59 @@ $(H2 $(LNAME2 parameters, Function Parameters))
...
+    $(TROW $(D ref), The parameter is an $(I input/output) parameter$(COMMA) passed by reference.
+    It is expected that the function will both read and write to this parameter.)
```
AndrejMitrovic, line 1272:
```suggestion
    It is expected that the function may both read from and write to this parameter.)
```

spec/function.dd
```diff
@@ -1219,39 +1219,59 @@ $(H2 $(LNAME2 parameters, Function Parameters))
...
+    $(TROW $(D out), The argument must be an lvalue$(COMMA) which will be passed by reference and initialized
+    upon function entry with the default value for its type)
```
AndrejMitrovic, line 1274:
```suggestion
    upon function entry with the `.init` value of its type)
```

I'm not 100% sure this will be clearer, so maybe you can ignore this suggestion.

AndrejMitrovic (changes requested):
Few nits!

spec/function.dd
Geod24: Maybe using both, e.g. "the default value (`T.init`) of its type" ?

spec/function.dd
Geod24: Then the "expected" part is weird no ?

spec/function.dd
Geod24: Technically, `-preview=in` is available in v2.093.0 as well but has a different meaning (just `scope const`).

spec/function.dd
Geod24: rvalues are always constructed in the caller's scope. So if it has copy constructor, postblit, whatever, it just follows the rules.

spec/function.dd
AndrejMitrovic: Hmm yeah. I guess what I was trying to say is `ref` doesn't imply the function absolutely will do both reads and writes.

Anyway, leave your docs as-is.

spec/function.dd
Geod24: Just a note: backtips and `$(D ...)` are redundant. Other than that, LGTM.

spec/function.dd
Geod24: I'll go with the second one as I think it might be easier for non-native speaker.

spec/function.dd
Geod24: Hum, I'm not sure about this, it feels redundant ?

spec/function.dd
Geod24: Do you want any change to this text ?

Geod24:
Addressed @AndrejMitrovic 's review, 3 outstanding comments.

AndrejMitrovic (approved):

spec/function.dd
AndrejMitrovic: No it's alright.

Geod24:
Ping

Geod24:
Self merging because it's been a while and the beta is out. Feel free to raise a PR and tag me if something is wrong.

PR #2846, opened 2020-08-16
add spec for the @disable new(); allocator relic

NilsLankila:


adamdruppe:
Is that actually implemented still? I thought it was removed (but would be very happy to see it still here, it is useful)

NilsLankila:
Yes, I've managed to keep it when working on the final deprecation step, see https://github.com/dlang/dmd/pull/11581.

deprecate.dd
```diff
@@ -12,7 +12,6 @@ $(SPEC_S Deprecated Features,
         $(THEAD Feature,                                                          Spec,  Dep,    Error,  Gone)
         $(TROW $(DEPLINK body keyword),                                           2.075, &nbsp;, &nbsp;, &nbsp;)
         $(TROW $(DEPLINK Hexstring literals),                                     2.079, 2.079, 2.086, &nbsp;)
-        $(TROW $(DEPLINK Class allocators and deallocators),                      &nbsp;, 2.080, 2.087, &nbsp;)
```
ibuclaw, line 15: Just need to update the gone column, no need to remove the documented deprecation.

spec/module.dd
```diff
@@ -634,6 +633,27 @@ $(GNAME MixinDeclaration):
         $(GLINK DeclDefs), and is compiled as such.
         )
 
+$(H2 $(LEGACY_LNAME2 AllocatorRelic, allocator-relic, Allocator Relic))
```
ibuclaw, line 636: I don't see there any need to call it a relic, either in the grammar or documentation, unless `@disable new()` itself is up for deprecation.

NilsLankila: Sorry but I really like this rule name. from the Cambridge dictionnary

```d
relic
 noun
  /ˈrelik/
something left from a past time
```

That matches perfectly to what it is.

ibuclaw: By calling it a relic, to me this is implicitly saying "don't use", or at the very least discourages use (why would you use a relic feature in your codebase?)

NilsLankila: A relic is not something  you drop. It's cultural inheritage. It's something you keep because it tells something about history and culture.

ibuclaw: A relic conjures up that it is an artefact, vestige, or an archaism in the language.  But as far as I can tell, there is no reason to tell people that they should not use `@disable new()`.

NilsLankila: Suggest me another rule name. Discussing about semantics will lead to nothing, although, to be clear, I really liked my proposal.

deprecate.dd
```diff
@@ -116,56 +115,6 @@ $(H4 Rationale)
...
-            destroy(obj);
-        }
-
-        Foo foo = make!Foo();
-        if (foo !is null) dispose(foo);
-        ---
-    )
-$(H4 Rationale)
-    $(P Classes should not be responsible for their own (de)allocation strategy.
-    )
```
wilzbach, line 167: No this needs to be kept. The entire point of this document is to provide historic rationals.

spec/module.dd
MoonlightSentinel: Maybe it's a cultural/language thing but I would read "relic" (in the context of software docs) as a soft deprecation.
(Read as "it exists but we don't dare to remove it").

Maybe use a neutral name based on the actual effect of `@disable new`? Smth. like `Disabling dynamic allocation`, ...

wilzbach: Yes, I'm sorry, but I agree that relic is a terrible name for the reasons stated above.

Why not name it what it actually is. A few examples: "NewClassDisabler", "ClassAllocationOptOut", "DefaultClassAllocationRule".

spec/module.dd
```diff
@@ -634,6 +633,27 @@ $(GNAME MixinDeclaration):
...
+
+$(GRAMMAR
+$(GNAME AllocatorRelic):
+    $(D new) $(D $(LPAREN)) $(D $(RPAREN)) $(D ;)
+)
+
+    $(P This is a relic of the old D1-style class and struct custom $(D new) operator.
+        Its sole purpose is to be prefixed with $(D @disable) so that $(D new) operator
+        is not permited for the aggregate that contains the allocator relic, for example
```
wilzbach, line 645: permitted

NilsLankila: "NewClassDisabler" does not work. `@disable` is checked during semantics checks, it is still grammatically correct to have `new();` without the attribute.

spec/module.dd
ibuclaw: > Suggest me another rule name. Discussing about semantics will lead to nothing, although, to be clear, I really liked my proposal.

I think the following:
1. Use LEGACY_LNAME2 as you currently are.
2. Keep the grammar rule name as `Allocator`.
3. While the spec shouldn't be used to sell features, try to keep to strictly saying what the feature is, rather allude to what it was.

For instance (re-using some of the existing documentation):
> A class member function of the form
>
> ------
> @disable new();
> ------
>
> is used to explicitly disallow the $(D new) operator being used to allocate an object instance.
> For example, to enforce the usage of $(D std.allocators.make())  on classes or encourage
> stack allocation for structs.

I assume that `@disable new()` is permitted for structs too?

spec/module.dd
NilsLankila: let's go for `NewDeclaration` then.

12345swordy:
And the author is gone.

adamdruppe:
we could fork the PR if need be

12345swordy:
There is no reason to left this PR open if isn't going to be merge.

RazvanN7:
This PR is already linked in dlang/dmd#11581 and the author is gone. If anyone wants to adopt this, we can reopen.

PR #2848, opened 2020-09-05
Remove deprecated preview=markdown & adapt documentation

MoonlightSentinel:
CC @wilzbach

### DIFF
--- a/spec/ddoc.dd
+++ b/spec/ddoc.dd
@@ -512,10 +512,6 @@ $(P
  +/
 ------------------------------------
 
-$(P
-$(EM Enable support for backticks, tildes and other code languages with the `-preview=markdown` compiler flag.)
-)
-
 $(H3 $(LNAME2 inline_code, Inline Code))
 
 $(P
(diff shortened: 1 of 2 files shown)


spec/ddoc.dd
```diff
@@ -513,7 +513,7 @@ $(P
 ------------------------------------
 
 $(P
-$(EM Enable support for backticks, tildes and other code languages with the `-preview=markdown` compiler flag.)
+$(EM Enable support for backticks, tildes and other code languages with the `-preview=markdown` compiler flag - enabled by default in 2.094.)
 )
```
wilzbach, line 517: Why not simply remove this comment?

MoonlightSentinel: Wasn't sure if the docs should cater towards older compilers

wilzbach: No, they don't need to. We have doc archives for that.

MoonlightSentinel: Done

wilzbach (approved):

PR #2850, opened 2020-09-09
Fixing FreeBSD info link

Ezneh:
This change fixes the info tooltip from "Linux setup and usage information" to "FreeBSD setup and usage information".
Also fixes the url from https://dlang.org/dmd-linux.html to https://dlang.org/dmd-freebsd.html


thewilsonator (approved):

Ezneh:
Seems like I forgot to change the line [87](https://github.com/Ezneh/dlang.org/blob/ac4d1fcde74d536834acd10ea8c9ce62ddc899f9/download.dd#L87) of this file.

PR #2860, opened 2020-09-27
update release schedule

MartinNowak:
- extend to end of 2021
- shift releases by 1 month after 2.094 delay and to avoid beta during holiday season


wilzbach (approved):
In the past activity always seemed to be higher over the holidays as people get time off, but if it's a time concern for you, I certainly don't want to block it.

PetarKirov (approved):
LGTM as well.

PR #2863, opened 2020-10-08
Show comments on code example

Imperatorn:
I'm trying to think like a complete newbie. 
I think we often forget that some people actually come to D as a first language.

1. tee = Perform a transparent operation
image

2. walkLength = optimized version of length
3. byLine = Get an input range set up to read one line at a time

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -46,7 +46,7 @@ void main()
     auto count = stdin
         //Get an input range set up to read one line at a time
         .byLine
-        //Perform a transparent operation 
+        //Perform a transparent operation (as in the shell command tee)
         .tee!(l => sum += l.length)
         //Optimized version of length
         .walkLength; 
--- a/index.dd
+++ b/index.dd
@@ -43,8 +43,13 @@ void main()
     import std.range, std.stdio;
 
     auto sum = 0.0;
-    auto count = stdin.byLine
-        .tee!(l => sum += l.length).walkLength;
+    auto count = stdin
+        //Get an input range set up to read one line at a time
+        .byLine
+        //Pipe input range to output range
+        .tee!(l => sum += l.length)
+        //Optimized version of length
+        .walkLength; 
 
     writeln("Average line length: ",
         count ? sum / count : 0);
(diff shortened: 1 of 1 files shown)


index.dd
```diff
@@ -43,8 +43,13 @@ void main()
...
 
     auto sum = 0.0;
-    auto count = stdin.byLine
-        .tee!(l => sum += l.length).walkLength;
+    auto count = stdin
+        //Get an input range set up to read one line at a time
+        .byLine
+        //Perform a transparent operation 
+        .tee!(l => sum += l.length)
+        //Optimized version of length
```
thewilsonator, line 51: this comment is misleading. In this case `.length` is not available since it isn't known in advance.

index.dd
```diff
@@ -43,8 +43,13 @@ void main()
     import std.range, std.stdio;
 
     auto sum = 0.0;
-    auto count = stdin.byLine
-        .tee!(l => sum += l.length).walkLength;
+    auto count = stdin
+        //Get an input range set up to read one line at a time
+        .byLine
+        //Perform a transparent operation
```
thewilsonator, line 49: you might want to add something like "as in the unix command `tee`"

thewilsonator (commented):
`

index.dd
Imperatorn: How would you describe walkLength instead?

thewilsonator:
```
./index.dd:52:        .walkLength;
```

index.dd
thewilsonator: It iterates the range and counts for each iteration. I'm not sure how to best word that.

Imperatorn: What about 
"Best-effort implementation of length" or "Best-effort implementation of length for ranges"?

Or maybe remove the comment? 🤔

thewilsonator: I think it would be fine with no comment, `walkLength` is a lot easier to figure out from the context than `tee`.

thewilsonator (approved):

PR #2867, opened 2020-10-16
Spec PR for `__traits(getCppNamespaces, sym)`

thewilsonator:
see https://github.com/dlang/dmd/pull/11863 for the corresponding dmd PR

### DIFF
--- a/spec/traits.dd
+++ b/spec/traits.dd
@@ -60,6 +60,7 @@ $(GNAME TraitsKeyword):
     $(GLINK getOverloads)
     $(GLINK getParameterStorageClasses)
     $(GLINK getPointerBitmap)
+    $(GLINK getCppNamespaces)
     $(GLINK getProtection)
     $(GLINK getTargetInfo)
     $(GLINK getVirtualFunctions)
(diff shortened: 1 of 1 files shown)


spec/traits.dd
```diff
@@ -1127,6 +1128,22 @@ void main()
...
 )
 
+$(H2 $(GNAME getCppNamespaces))
+	$(P The argument is a symbol.
+	The result is a tuple of strings, possibly empty, that correspond to the namespaces the symbol resides in.
+	)
+
+$(SPEC_RUNNABLE_EXAMPLE_COMPILE
+---
+extern(C++, "ns")
```
thewilsonator, line 1138: SPEC_RUNNABLE_EXAMPLE_COMPILE insets a `void main(){}` around this code block resulting in `namespaces are not allowed here`, which macro should I use instead?

TurkeyMan: Test also with a non-string namespace:
```d
extern(C++, ns2) void fun();
```

TurkeyMan: Actually, if this is just for the spec, just make sure the unit tests for the implementation test with string and non-string namespace.

Geod24:
@thewilsonator : Plz fix

thewilsonator:
@Geod24 any ideas for https://github.com/dlang/dlang.org/pull/2867#discussion_r506046660 ? Otherwise I'll just axe it.

Geod24:
Can't you do it without macro ? Using `---` will give you a code block.

thewilsonator:
Should hopefully be green now.

Geod24 (approved):

PR #2872, opened 2020-10-27
Document issue 21340 - `extern(C++, (expression))` now accepts empty tuples

thewilsonator:
see also https://github.com/dlang/dmd/pull/11905

### DIFF
--- a/spec/cpp_interface.dd
+++ b/spec/cpp_interface.dd
@@ -186,6 +186,13 @@ module ns;
 extern (C++, `ns`)
 {
     int foo() { return 1; }
+}
+---
+    $(P or with a compile time computed expression that resolves to a, possibly empty, tuple of strings.
+      If the expression removes to an empty tuple it is equivalent to `extern (C++)`)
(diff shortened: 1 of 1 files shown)


thewilsonator:
@Geod24 ping for spec PR

spec/cpp_interface.dd
```diff
@@ -186,6 +186,13 @@ module ns;
 extern (C++, `ns`)
 {
     int foo() { return 1; }
+}
+---
+    $(P or with a compile time computed expression that resolves to a, possibly empty, tuple of strings.
+      If the expression removes to an empty tuple it is equivalent to `extern (C++)`)
```
Geod24, line 192:
```suggestion
    $(P Any expression that resolves to either a tuple of strings or an empty tuple is accepted.
      When the expression resolves to an empty tuple, it is equivalent to `extern (C++)`)
```

Geod24 (approved):
Slight rephrase, oLGTM

PR #2873, opened 2020-11-02
Rename grammar entities for more descriptive names

Bolpat:
**Nothing concerning the compiler is intended to be changed by this PR**
(If this change happens to entail that the compiler must act differently, this is an error!)

This PR replaces grammar entities for more descriptive names. The replacements are in this table:
| Former name | New Name |
|-----------|-----------|
| `IdentifierList` | `QualifiedIdentifier` |
| `QualifiedIdentifierList` | `MixinQualifiedIdentifier` |
| `BasicTypeX` | `FundamentalType` |
| `BasicType2` | `TypeSuffixes` |
| `BasicType2X` | `TypeSuffix` |
| `AutoDeclarationX` | `AutoAssignments` |
| `AutoDeclarationY` | `AutoAssignment` |
| `AliasDeclarationX` | `AliasAssignments` |
| `AliasDeclarationY` | `AliasAssignment` |
| `AltDeclaratorX` | `AltDeclaratorInner` |

Rationale for changing  `IdentifierList` and `QualifiedIdentifierList`:
* They aren't lists. All other `*List`s are comma-separated; examples include `DeclaratorIdentifierList`, `ArgumentList`, `BaseClassList`, and `NamespaceList` (non-exhaustive).
* In the case of changing `IdentifierList` to `QualifiedIdentifier`, note that `Quaified*` is used in the spec with exactly that meaning. (If you come from a C++ background, their use of “qualified” is not present in D.)
* In the case of `QualifiedIdentifierList`, apart not being a list of `QualifiedIdentifier`s, prefixing it with `Mixin` is justified because it *is* specific to mixins. `MixinQualifiedIdentifier` seemed like a decent and descriptive choice to me.

Rationale for changing type stuff:
* `BasicTypeX` is any of the keyword types like `int`, `double`, etc. Very fundamental, I guess.
* `BasicType2X` really is used as the suffix of types. It is the `*` for pointer types, `[]` for slice types, etc. It is only used in grammar rules to appear behind some `BasicType` in one way or another.
* As the new name suggests, `TypeSuffixes` is iterated `TypeSuffix`es.

Concerning the renaming of `*X` and `*Y` to `Assignments` and `Assignment`: That's what it is.

The `AltDeclaratorX` is the stuff in parentheses in an `AltDeclaration`. It really is the *inner* part of an `AltDeclaration`.


thewilsonator (approved):
Thanks, are the old names still referenced in `parse.d`? If so please update them too.

PR #2881, opened 2020-11-17
improve definition of assign expression behaviour

Alexibu:
The spec appears to not specify that an assign expression must be evaluated after it's operands.
It also helps to define that it must do the same thing as an asignment of the binary operator.

I believe this is the intention of the language, so that it has the same operator behaviour as C++.

See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=97843

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -261,6 +261,18 @@ $(GNAME AssignExpression):
     occurs. The resulting expression is a modifiable lvalue.
     )
 
+	$(P Operands of an assign expression are evaluated before the assign expression. 
+	The expression)
+        --------------
+        a op= b
+        --------------
+
+        is exactly the same as:
+
+        --------------
+        a = a op b
+        --------------
+
     $(UNDEFINED_BEHAVIOR
     If either operand is a reference type and one of the following:
     $(OL
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -75,6 +75,8 @@ subexpression `h()` has no smallest short-circuit expression.
 
 $(H2 $(LNAME2 order-of-evaluation, Order Of Evaluation))
 
+$(P Operands are evaluated before operators are executed. Arguments are evaluated before functions are called.)
+
 $(P Built-in prefix unary expressions `++` and `--` are evaluated as if lowered (rewritten) to
 assignments as follows: `++expr` becomes `((expr) += 1)`, and `--expr` becomes `((expr) -= 1)`.
 Therefore, the result of prefix `++` and `--` is the lvalue after the side effect has been
@@ -260,18 +262,6 @@ $(GNAME AssignExpression):
     operand, and the value is the value of the left operand after assignment
     occurs. The resulting expression is a modifiable lvalue.
     )
-
-	$(P Operands of an assign expression are evaluated before the assign expression. 
-	The expression)
-        --------------
-        a op= b
-        --------------
-
-        is exactly the same as:
-
-        --------------
-        a = a op b
-        --------------
 
     $(UNDEFINED_BEHAVIOR
     If either operand is a reference type and one of the following:


PetarKirov (approved):

PetarKirov:
@ibuclaw @kinke @WalterBright do you guys agree?

spec/expression.dd
```diff
@@ -261,6 +261,18 @@ $(GNAME AssignExpression):
...
+	$(P Operands of an assign expression are evaluated before the assign expression. 
+	The expression)
+        --------------
+        a op= b
+        --------------
+
+        is exactly the same as:
+
+        --------------
+        a = a op b
```
thewilsonator, line 273: what if `a` has side-effects?

spec/expression.dd
```diff
@@ -261,6 +261,18 @@ $(GNAME AssignExpression):
     occurs. The resulting expression is a modifiable lvalue.
     )
 
+	$(P Operands of an assign expression are evaluated before the assign expression. 
+	The expression)
+        --------------
+        a op= b
+        --------------
+
+        is exactly the same as:
```
PetarKirov, line 270: @thewilsonator I suppose this would be a more accurate wording:
```suggestion
        is rewritten to:
```
Which would address the question of side effects.

spec/expression.dd
PetarKirov: Isn't that the point? Whether or not `a` has side effects, it would be evaluated once:

```
exp1 op= exp2;
```

=>

```
{
    auto tmp1 = exp1;
    auto tmp2 = exp2;
    tmp1 = tmp1 op tmp2;
}

That is, unless `exp1` resolves to a type with custom `opAssign`, or `opOpAssign`.

File: spec/expression.dd (hunk 2 above)
Comment (thewilsonator) [reply spec/expression.dd:270]: Is it though? I thought it would be rewritten to `auto [ref] __tmp = a; __tmp = __tmp op b;`, the number of evaluation is important.

Comment (PetarKirov) [reply spec/expression.dd:270]: Ah, I thought that was clear. Yeah, I guess it makes sense to specify this more accurately.

**Comment by thewilsonator** [comment]
cc @JohanEngelen I know we had a couple of bugs w.r.t. this or sometime similar in LDC.

File: spec/expression.dd (hunk 1 above)
Comment (ibuclaw) [reply spec/expression.dd:273]: Then in most cases, a can't be an lvalue in an op= expression, unless the result of a is a stable reference (iirc, `*a op= b`)

File: spec/expression.dd (hunk 2 above)
Comment (ibuclaw) [reply spec/expression.dd:270]: One exception to that rewrite though is array concatenation.
```
a ~= b
// a = a ~ b;
```
Because evaluating `a` changes its underlying size, so it would be conceptually
```
auto tmp = b
extend(&a)[beg .. end] = tmp;
```

Comment (Alexibu) [reply spec/expression.dd:270]: Why does evaluating `a` change its size ? Isn't the size change part of the execution of `~=`, not the evaluation of operands ?

Comment (ibuclaw) [reply spec/expression.dd:270]: Because in `a op= b`, operand `a` can only be evaluated once.  Separating the size change from the evaluation means you'll have to compute it twice.

Comment (kinke) [reply spec/expression.dd:270]: > Is it though?

Nope, this is definitely wrong, see e.g. https://github.com/dlang/dmd/blob/a9e5b0db723f8edaf80df28f93257edf06d3f17b/test/runnable/evalorder.d#L59-L61.

**Comment by ibuclaw** [comment]
So a little further down the page, you already have a section on [assignment operator expressions](https://dlang.org/spec/expression.html#assignment_operator_expressions), would it be better just to clarify the wording of that passage to avoid redundancy?

File: spec/expression.dd (hunk 2 above)
Comment (Alexibu) [reply spec/expression.dd:270]: I translated that test for eval order  to the `~=` operator:
```

auto mul11ret3(T)(ref T s)
{
    s ~= 11;
    return [3];
}

void main()
{
	static auto test3(int[] val) { (val ~= 7) ~= mul11ret3(val); return val; }
	assert(test3([2]) == [2,7,11,3]);
}
```

DMD fails with [2,11,7]
gdc 10.2 fails with core.exception.OutOfMemoryError.

Comment (ibuclaw) [reply spec/expression.dd:270]: > I translated that test for eval order to the `~=` operator:
> gdc 10.2 fails with core.exception.OutOfMemoryError.

Excellent, that means garbage is passed as the destination slot in `append(&dst, mul11ret3(val))` instead of the address to `val`.

**Comment by Alexibu** [comment]
The part about `a op= b` being exactly equal to 'a = a op b` was both incorrect and redundant because of 10.5,2.
I've modified the PR to what I think is missing.
Is it possible that this is super obvious but omittied from the spec ?
I had thought it was a basic assumption, but in discussing why `a ~= a.sum` didn't work I realise it is a topic of debate, and not clarified by the spec.
There is probably a lot more not documented. For example see https://en.cppreference.com/w/cpp/language/eval_order

**Comment by JohanEngelen** [comment]
See: https://github.com/dlang/dlang.org/pull/1429

**Comment by WalterBright** [comment]
> See: #1429

Yes. In particular, see @andralex 's comment:

> I'd say a correct formulation goes as follows:
The expression `a @= b`, where `@` stands for any of the binary operators, 
is evaluated as if it were lowered as follows:
```
(function ref(ref x, y) { return x = cast(typeof(x)) (x @ y); })(a, b)
```

This is a much clearer specification, as the text in this PR was rather baffling to me, and I had to study the examples to see what was going on.

WalterBright (changes requested):
See my comment.

Alexibu:
Closed PR because #1429 will clarify.

PR #2885, opened 2020-11-28
Fix casing/spelling of TargetOS

WalterBright:
Required for https://github.com/dlang/dmd/pull/12005


thewilsonator (approved):

WalterBright:
Of course, this fails because it is both blocking #12005 and is blocked by #12005.

Shows the folly of having a project import files from another project.

WalterBright:
Either this or https://github.com/dlang/dlang.org/pull/2885 has to be pulled first, because they are mutually dependent. I recommend this.

PR #2887, opened 2020-12-04
spec/traits.dd: Document __traits(getTargetInfo, "cppStd")

ibuclaw:


spec/traits.dd
```diff
@@ -1203,6 +1203,7 @@ version (CppRuntime_Microsoft)
 
         $(UL
         $(LI $(D "cppRuntimeLibrary") - The C++ runtime library affinity for this toolchain)
+        $(LI $(D "cppStd") - The version of C++ standard supported in $(D extern$(LPAREN)C++$(RPAREN) code)
```
thewilsonator, line 1206: Specifically it is the numerical value of `__cplusplus`, please make that clear (e.g. with an example or text that says that)

ibuclaw: Added `equivalent to the __cplusplus macro in a C++ compiler`.

thewilsonator (approved):

PR #2893, opened 2020-12-07
Improve wording of Virtual Functions section

WalterBright:


spec/function.dd
```diff
@@ -722,32 +722,30 @@ $(H2 $(LNAME2 property-functions, Property Functions))
...
+        $(P Static member functions are not virtual.
         )
 
-        $(P Functions marked as $(D final) may not be overridden in a
-        derived class, unless they are also $(D private).
-        For example:
+        $(P Member functions which are $(D private) or $(D package) are not virtual.
+        )
+
+        $(P Functions with non-D linkage are not virtual.
```
thewilsonator, line 741: What about C++? 
Or it this the case where the linkage of the function differs from the class.

WalterBright: This was copied from the previous description, but is incorrect. Will fix.

thewilsonator (approved):

PR #2894, opened 2020-12-08
add Ddoc spec macros RATIONALE and NOTE

WalterBright:
Instead of them being ad-hoc, and allowing for custom styling of each. Also makes it clear that they are not formally part of the spec.

### DIFF
--- a/spec/spec.ddoc
+++ b/spec/spec.ddoc
@@ -9,10 +9,17 @@ $(COMMENT While not part of the spec, best practices offers advice on how to bes
 BEST_PRACTICE= $(DIVC spec-boxes best-practice, $(B Best Practices:) $0)
 
 $(COMMENT Identifies implementation-defined behavior in the spec)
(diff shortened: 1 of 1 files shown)


spec/spec.ddoc
```diff
@@ -13,6 +13,13 @@ IMPLEMENTATION_DEFINED=$(DIVC spec-boxes implementation-defined, $(B Implementat
 
 $(COMMENT Identifies undefined-defined behavior in the spec)
 UNDEFINED_BEHAVIOR=$(DIVC spec-boxes undefined-behavior, $(B Undefined Behavior): $0)
+
+$(COMMENT While not part of the spec, an explanation for something unexpected)
+RATIONALE= $(DIVC spec-boxes rationale, $(B RationaleBest Practices:) $0)
```
adamdruppe, line 18: missing space there? between Rationale and Best or just delete a word

WalterBright: Copy pasta error alert!

WalterBright: fixed

thewilsonator (approved):

PR #2898, opened 2020-12-08
small improvements to Safe Values section wording

WalterBright:


spec/function.dd
```diff
@@ -3139,10 +3139,11 @@ $(H3 $(LNAME2 safe-values, Safe Values))
...
         possible bit patterns are safe.)
 
-        $(P A pointer is safe when it is `null` or:)
+        $(P A pointer is a safe value when:)
         $(OL
-            $(LI it can be dereferenced validly, and)
-            $(LI the value of the pointee is safe.)
+            $(LI it is `null`, or)
+            $(LI it points to a memory object that is live, and)
+            $(LI the pointed to value in that memory object is safe.)
```
aG0aep6G, line 3146: Does that mean "(null or live) and value safe" or "null or (live and value safe)"? We know it's the latter, but the order of operations isn't well-defined in English (as far as I know). I suggest not mixing "and"s and "or"s in the same list.

WalterBright: fixed

atilaneves (approved):

PR #2904, opened 2020-12-09
fix Compile Time Function Execution section

WalterBright:


spec/function.dd
```diff
@@ -2666,10 +2666,11 @@ void main()
...
 
-    $(P Functions which are both portable and free of global side-effects can be
-    executed at compile time. In certain contexts, such compile time execution
-    is guaranteed. It is called Compile Time Function Execution (CTFE) then.
-    The contexts that trigger CTFE are:)
+    $(P In contexts where a compile time value is required, functions
+    can be used to compute those values. This is called $(I Compile Time Function
+    Execution), or $(I CTFE).)
+
+    $(P These contexts include:)
```
aG0aep6G, line 2673: "Include" implies that there are other CTFE contexts not listed here. But the spec should describe the language completely, so the list here should be exhaustive.

atilaneves: I agree with @aG0aep6G and think this should be changed to "these contexts are".

RazvanN7 (approved):

RazvanN7:
@WalterBright mind the @aG0aep6G 's comment and this is good to go.

spec/function.dd

PR #2908, opened 2020-12-11
fix Local Static Variables section

WalterBright:


thewilsonator (approved):

WalterBright:
No clue in the log why CyberShadow/DAutoTest fails.

PR #2916, opened 2020-12-13
improve Parameter Storage Classes section

WalterBright:
1. Fixed enumeration of storage classes to match order entries in table.
2. Moved `in` documentation to its own paragraph.
3. Better example, more thorough explanation.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -1248,75 +1248,86 @@ $(H2 $(LNAME2 parameters, Function Parameters))
 
 $(H3 $(LNAME2 param-storage, Parameter Storage Classes))
 
-        $(P Parameter storage classes are $(D in), $(D out), $(D ref), $(D lazy), $(D scope) and $(D inout).
-        Parameters can also take the type constructors $(D const), $(D immutable) and $(D shared).)
+        $(P Parameter storage classes are $(D in), $(D out), $(D ref), $(D lazy), `return` and $(D scope).
+        Parameters can also take the type constructors $(D const), $(D immutable), $(D shared) and `inout`.
+        )
 
-        $(P $(D in), $(D ref), $(D out) and $(D lazy) are mutually exclusive. The first three are used to
+        $(P $(D in), `out`, $(D ref) and $(D lazy) are mutually exclusive. The first three are used to
         denote input, input/output, and output parameters, respectively.
         For example:
         )
-------
-int read(in char[] input, ref size_t count, out int errno);
-------
 
-        $(P In this example, $(D input) will only be read and no reference to it will be kept around,
-        while $(D count) will be read and written to, and $(D errno) will be set to a value from
-        within the function.
-        This approach gives a semantic meaning to the parameters and allows the compiler to potentially
-        optimize the generated code.)
+        ---
+        int read(in char[] input, ref size_t count, out int errno);
 
-        $(P When talking about function parameters, the $(I parameter) is what the function prototype defines,
-        and the $(I argument) is the value that will $(I bind) to it.
-        For example, using the previous definition of $(D read):
+        void main()
+        {
+            size_t a = 42;
+            int b;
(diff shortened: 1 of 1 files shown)


spec/function.dd
```diff
@@ -1248,75 +1248,86 @@ $(H2 $(LNAME2 parameters, Function Parameters))
 
 $(H3 $(LNAME2 param-storage, Parameter Storage Classes))
 
-        $(P Parameter storage classes are $(D in), $(D out), $(D ref), $(D lazy), $(D scope) and $(D inout).
-        Parameters can also take the type constructors $(D const), $(D immutable) and $(D shared).)
+        $(P Parameter storage classes are $(D in), $(D out), $(D ref), $(D lazy), `return` and $(D scope).
```
RazvanN7, line 1251: Why is `return` not `$(D return)`?

spec/function.dd
```diff
@@ -1248,75 +1248,86 @@ $(H2 $(LNAME2 parameters, Function Parameters))
...
+        Parameters can also take the type constructors $(D const), $(D immutable), $(D shared) and `inout`.
```
RazvanN7, line 1252: ditto for `inout`

spec/function.dd
```diff
@@ -1248,75 +1248,86 @@ $(H2 $(LNAME2 parameters, Function Parameters))
...
+        )
 
-        $(P $(D in), $(D ref), $(D out) and $(D lazy) are mutually exclusive. The first three are used to
+        $(P $(D in), `out`, $(D ref) and $(D lazy) are mutually exclusive. The first three are used to
```
RazvanN7, line 1255: ditto for `out`?

spec/function.dd
```diff
@@ -1248,75 +1248,86 @@ $(H2 $(LNAME2 parameters, Function Parameters))
...
+
+        $(TROW $(D const), argument is implicitly converted to a const type)
+        $(TROW $(D immutable), argument is implicitly converted to an immutable type)
+        $(TROW $(D shared), argument is implicitly converted to a shared type)
+        $(TROW $(D inout), argument is implicitly converted to an inout type)
+        )
+
+$(H3 $(LNAME2 in-params, In Parameters))
+
+        $(P The parameter is an input to the function. Input parameters behaves as if they have
```
RazvanN7, line 1317: Input parameters **behave**

spec/function.dd
```diff
@@ -1248,75 +1248,86 @@ $(H2 $(LNAME2 parameters, Function Parameters))
...
+        the $(D const scope) storage classes. Input parameters may be passed by reference by the compiler.
+        Unlike $(D ref) parameters$(COMMA) $(D in) parameters can bind to both lvalues and rvalues (such as literals).
+        Types that would trigger a side effect if passed by value (such as types with postblit$(COMMA)
```
RazvanN7, line 1320: suggested change: "... such as types with copy constructors, postblit ...". Put the copy constructor first so that people see it as a better alternative than postblit.

spec/function.dd
```diff
@@ -1248,75 +1248,86 @@ $(H2 $(LNAME2 parameters, Function Parameters))
...
+        copy constructor$(COMMA) or destructor)$(COMMA) and types which cannot be copied$(COMMA)
+        e.g. if their copy constructor is marked as $(D @disable)$(COMMA) will always be passed by reference.
```
RazvanN7, line 1322: **they** will always ...

RazvanN7 (approved):

spec/function.dd
WalterBright: They're equivalent.

WalterBright: same for the rest.

spec/function.dd

spec/function.dd

spec/function.dd

PR #2917, opened 2020-12-13
improve Function Overloading section

WalterBright:
Minor wording improvements. The clause at line 1166 was removed because it is incorrect.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -1103,10 +1103,10 @@ $(H2 $(LNAME2 inline-functions, Inline Functions))
 
 $(H2 $(LNAME2 function-overloading, Function Overloading))
 
-        $(P Functions are overloaded based on how well the arguments
-        to a function can match up with the parameters.
-        The function with the $(I best) match is selected.
-        The levels of matching are:
+        $(P $(I Function overloading) occurs when two or more functions in the same scope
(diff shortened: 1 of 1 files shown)


spec/function.dd
```diff
@@ -1117,27 +1117,26 @@ $(H2 $(LNAME2 function-overloading, Function Overloading))
...
-        and $(CODE g()) by taking the parameter types of $(CODE f()),
+        Partial ordering is determined for functions $(I f)
+        and $(I g) by taking the parameter types of $(I f),
         constructing a list of arguments by taking the default values
-        of those types, and attempting to match them against $(CODE g()).
-        If it succeeds, then $(CODE g()) is at least as specialized
-        as $(CODE f()).
+        of those types, and attempting to match them against $(I g).
+        If it succeeds, then $(I g) is at least as specialized
+        as $(I f).
```
PetarKirov, line 1139: I think we should use code blocks throughout for the function names `f` and `g`
```suggestion
        as `f`.
```

PetarKirov (approved):

spec/function.dd
WalterBright: Since the functions in the example were not f and g, I decided to use the more abstract mathematical notation.

PR #2918, opened 2020-12-13
improve Function Inheritance and Overriding section

WalterBright:
Use fewer words and more pedantic language.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -883,8 +883,8 @@ void main()
 
 $(H3 $(LNAME2 function-inheritance, Function Inheritance and Overriding))
 
-        $(P A function in a derived class with the same name and parameter
-        types as a function in a base class overrides that function:)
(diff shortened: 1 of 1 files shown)


spec/function.dd
```diff
@@ -1037,9 +1031,8 @@ void test()
 }
 ------
 
-        $(P If a derived class overrides a base class member function with different
-        $(GLINK FunctionAttributes), the missing attributes will be
-        automatically compensated by the compiler.)
+        $(P An overriding function inherits any unspecified $GLINK FunctionAttributes)
```
aG0aep6G, line 1034:
```suggestion
        $(P An overriding function inherits any unspecified $(GLINK FunctionAttributes)
```

WalterBright: fixed

atilaneves (approved):

PR #2868, opened 2020-10-16
add new feature AliasAssign to spec

WalterBright:
Implementation: https://github.com/dlang/dmd/pull/11846

PR #2919, opened 2020-12-17
spec: add AliasReassignment

WalterBright:
Implementation: https://github.com/dlang/dmd/pull/11846/

### DIFF
--- a/spec/declaration.dd
+++ b/spec/declaration.dd
@@ -502,6 +502,44 @@ a = 2;         // sets `S.i` to `2`
 b = 4;         // sets `S.j` to `4`
 -----------
 
+
+$(H2 $(LNAME2 alias-reassignment, Alias Reassignment))
+
+        $(GRAMMAR
+        $(GNAME AliasReassignment):
+            $(GLINK_LEX Identifier) $(D =) $(GLINK StorageClasses)$(OPT) $(GLINK Type)
+            $(GLINK_LEX Identifier) $(D =) $(GLINK2 expression, FunctionLiteral)
+            $(GLINK_LEX Identifier) $(D =) $(GLINK StorageClasses)$(OPT) $(GLINK BasicType) $(GLINK2 function, Parameters) $(GLINK2 function, MemberFunctionAttributes)$(OPT)
+        )
+
+        $(P An alias declaration inside a template can be reassigned a new value.)
+
+        ---
+        template staticMap(alias F, T...)
+        {
+            alias A = AliasSeq!();
+            static foreach (t; T)
+                A = AliasSeq!(A, F!T); // alias reassignment
+            alias staticMap = A;
+        }
+        ---
+
+        $(P The $(I Identifier) must resolve to a lexically preceding $(GLINK AliasDeclaration).
+        Both must be members of the same $(GLINK2 template, TemplateDeclaration).
+        )
+
+        $(P The right hand side of the $(I AliasReassignment) replaces the right hand side of the
+        $(I AliasDeclaration).
+        )
+
+        $(P Once the $(I AliasDeclaration) has been referred to in any context other than the
+        right hand side of an $(I AliasReassignment) it can no longer be reassigned.
+        )
+
+        $(RATIONALE Alias reassignment can result in faster compile times and lowered memory consumption,
+        and requires significantly simpler code than the alternative recursive method.)
+
+
 $(H2 $(LNAME2 extern, Extern Declarations))
 
 $(P Variable declarations with the storage class $(D extern) are not allocated


spec/declaration.dd
```diff
@@ -502,6 +502,44 @@ a = 2;         // sets `S.i` to `2`
...
+        $(P The right hand side of the $(I AliasReassignment) replaces the right hand side of the
+        $(I AliasDeclaration).
+        )
+
+        $(P Once the $(I AliasDeclaration) has been referred to in any context other than the
+        right hand side of an $(I AliasReassignment) it can no longer be reassigned.
+        )
+
+        $(RATIONALE Alias reassignment can result in significantly faster compile times
+        and lowered memory consumption than the alternative recursive method.)
```
ibuclaw, line 540: Citation? I was under the impression that this is currently inconclusive.

12345swordy:
@atilaneves there is a lot of spec pull requests that @WalterBright  has made that needs review from you as you took over @andralex leadership position on this. Can you spare some time and review this PR and other PR likes this that Walter has made?

spec/declaration.dd
atilaneves: So was I.

WalterBright: I don't need a test to know that:
```
malloc(5);
```
is faster and uses less memory than:
```
malloc(5);
malloc(5);
```
:-)

maxhaton: https://github.com/dlang/phobos/pull/7756#issuecomment-764048209 Some plots on this. The performance increase is measurable but not particularly dramatic for `staticMap`

mdparker:
@atilaneves The text looks good editorially. I'm good for merging this if you and @ibuclaw are good now with the objection you raised.

Imperatorn:
Merge?

spec/declaration.dd
BorisCarvajal: > I don't need a test to know that:
> 
> ```
> malloc(5);
> ```
> 
> is faster and uses less memory than:
> 
> ```
> malloc(5);
> malloc(5);
> ```
> 
> :-)


But that's exactly what `static foreach` is doing, it will `syntaxCopy` the `foreach` body N-times.
The rationale should simply be to provide an imperative alternative for constructing alias sequences vs the recursive way.
The performance is just an implementation detail. I hope it gives more performance though.

maxhaton:
> Merge?

Needs WalTila approval. 

@WalterBright @atilaneves 

It's been something like 7 months now, needs a decision. 

There are probably lessons to be learned from how this has been handled.

I say just merge, if something bad happens due to Alias Assign down the line it will likely be due to the implementation, and if we decide it's bad language design then the design itself can be deprecated reasonably smoothly as its use should be both fairly limited and mostly by the hands of people who know what they're doing and are in either maintainers or in direct contact with maintainers.

atilaneves:
Does anyone have numbers on the actual impact on compile-times due to this?

mdparker:
@maxhaton So what's the verdict? Is the performance claim good to go?

andralex:
FWIW in all my real-world testing (building large apps) AliasAssignment has been a win or break even, with huge gains in code size and readability.

mdparker:
Then I propose:

```
$(RATIONALE Alias reassignment can result in faster compile times and lowered memory consumption, and requires significantly fewer lines of code than the alternative recursive method.)
```

andralex:
> Then I propose:
> 
> ```
> $(RATIONALE Alias reassignment can result in faster compile times and lowered memory consumption, 
> and requires significantly fewer lines of code than the alternative recursive method.)
> ```

s/fewer lines of/simpler/

mdparker:
Paging @WalterBright.

mdparker:
> s/fewer lines of/simpler/

Then perhaps also s/requires/results in/.

maxhaton:
> FWIW in all my real-world testing (building large apps) AliasAssignment has been a win or break even, with huge gains in code size and readability.

I agree/concur with Andrei. This feature is still a little raw (Different syntax? It's a bit weird as is in the sense that it's not quite an assignment given the restrictions), but the code-size wins are really nice, and the blow-up losses seem to be an implementation problem which we can nail down (they are _awful_ when you hit them, but we're talking staticMaps over thousands and thousands of elements here).

atilaneves:
> FWIW in all my real-world testing (building large apps) AliasAssignment has been a win or break even, with huge gains in code size and readability.

How big was the median win?

andralex:
> > FWIW in all my real-world testing (building large apps) AliasAssignment has been a win or break even, with huge gains in code size and readability.
> 
> How big was the median win?

Say some 5%-7% but on a rather noisy setup that measured only the whole project bottom line. Let me know if precision is needed and I can set up a more rigorous experimental rig.

CyberShadow:
Either the compiler behavior needs to be put behind a `-preview` flag (I'm guessing not doable because Phobos is using this) or this needs to be merged, because right now the reference implementation's behavior does not correspond to the official specification.

We can help prevent incidents like these from happening in the future by requiring (i.e. enforcing as part of CI) that the compiler test suite / Phobos / etc. are valid according to the official grammar. This will require first making a pull request to dlang.org, describing the language change, which only then will allow any changes to the reference implementation (DMD) to be merged.

spec/declaration.dd
mdparker: @WalterBright The consensus appears to be to revise this bit as:

> $(RATIONALE Alias reassignment can result in faster compile times and lowered memory consumption, 
and requires significantly simpler code than the alternative recursive method.)

mdparker (dismissed):

mdparker (approved):

spec/declaration.dd
```diff
@@ -502,6 +502,44 @@ a = 2;         // sets `S.i` to `2`
...
 -----------
 
+
+$(H2 $(LNAME2 alias-reassignment, Alias Reassignment))
+
+        $(GRAMMAR
+        $(GNAME AliasReassignment):
+            $(GLINK_LEX Identifier) $(D =) $(GLINK StorageClasses)$(OPT) $(GLINK Type)
+            $(GLINK_LEX Identifier) $(D =) $(GLINK2 expression, FunctionLiteral)
+            $(GLINK_LEX Identifier) $(D =) $(GLINK StorageClasses)$(OPT) $(GLINK BasicType) $(GLINK2 function, Parameters) $(GLINK2 function, MemberFunctionAttributes)$(OPT)
```
CyberShadow, line 512: Sorry, how is this different from the `AliasAssign` rule added in #2868?

Furthermore, `AliasReassignment` is not referenced anywhere in the spec, which effectively makes this syntax not a part of the language. What does this grammar block describe exactly?

@WalterBright

spec/declaration.dd
```diff
@@ -502,6 +502,44 @@ a = 2;         // sets `S.i` to `2`
 b = 4;         // sets `S.j` to `4`
 -----------
 
+
+$(H2 $(LNAME2 alias-reassignment, Alias Reassignment))
+
+        $(GRAMMAR
+        $(GNAME AliasReassignment):
+            $(GLINK_LEX Identifier) $(D =) $(GLINK StorageClasses)$(OPT) $(GLINK Type)
```
CyberShadow, line 510: `Type` is declared in `type.dd`, not in this file.

spec/declaration.dd
CyberShadow, line 512: Ditto for `BasicType`.

CyberShadow: @WalterBright ping

CyberShadow: @WalterBright ping

PR #2925, opened 2020-12-23
simd: add clarification that implicit vector casts are reinterpretations

WalterBright:


WalterBright:
Motivation: https://issues.dlang.org/show_bug.cgi?id=20112

PetarKirov (approved):

PR #2926, opened 2020-12-29
Exclude news about `-vtemplates` in 2.095 already added in 2.094

nordlow:


Geod24 (approved):
Thanks!
For context: https://forum.dlang.org/post/eaebihvpzdrzmqqfgavu@forum.dlang.org

PR #2933, opened 2021-01-22
Fix 21292: Link directly to https s3 downloads

Geod24:
Companion PR: dlang/downloads.dlang.org#6

### DIFF
--- a/download.dd
+++ b/download.dd
@@ -238,9 +238,9 @@ Macros:
     SBTN=$(SPANC sig_btn,$(BTN $1,$+)<br>$(BTN $1.sig,sig))
     BTN=<a href="$1" class="btn">$+</a>
     H3I=<h3 class="download">$0</h3>
-    DLSITE=http://downloads.dlang.org/releases/2.x/$(DMDV2)/$0
-    B_DLSITE=http://downloads.dlang.org/pre-releases/2.x/$(B_DMDV2)/$0
-    N_DLSITE=http://downloads.dlang.org/nightlies/dmd-master/$0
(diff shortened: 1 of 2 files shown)


wilzbach (approved):
I would really prefer if we moved to releases to something that's easier to work with, but I guess this is good enough and eases the pains for the time being.

Geod24:
@wilzbach : So do I, but the issue is orthogonal. We currently have users that click on a download link, and get absolutely nothing. Not even a message from Chrome saying "This file couldn't be downloaded because...". It's embarrassing and needs to be fixed yesterday.

Geod24:
This didn't work, because the files are stored by year internally, not by release...

download.dd
```diff
@@ -238,9 +238,9 @@ Macros:
     SBTN=$(SPANC sig_btn,$(BTN $1,$+)<br>$(BTN $1.sig,sig))
     BTN=<a href="$1" class="btn">$+</a>
     H3I=<h3 class="download">$0</h3>
-    DLSITE=http://downloads.dlang.org/releases/2.x/$(DMDV2)/$0
-    B_DLSITE=http://downloads.dlang.org/pre-releases/2.x/$(B_DMDV2)/$0
-    N_DLSITE=http://downloads.dlang.org/nightlies/dmd-master/$0
+    DLSITE=https://s3.us-west-2.amazonaws.com/downloads.dlang.org/releases/2.x/$(DMDV2)/$0
+    B_DLSITE=https://s3.us-west-2.amazonaws.com/downloads.dlang.org/pre-releases/2.x/$(B_DMDV2)/$0
+    N_DLSITE=https://s3.us-west-2.amazonaws.com/downloads.dlang.org/dmd-master/2.x/$0
```
Geod24, line 243: Copy pasta error

PR #2934, opened 2021-01-24
Allow empty comments and string literals in grammar

tim-dlang:


spec/lex.dd
```diff
@@ -1078,7 +1078,7 @@ $(GNAME SpecialTokenSequence):
     $(D # line) $(GLINK IntegerLiteral) $(GLINK Filespec) $(GLINK EndOfLine)
 
 $(GNAME Filespec):
-    $(B ") $(GLINK Characters) $(B ")
+    $(B ") $(GLINK Characters)$(OPT) $(B ")
```
PetarKirov, line 1081: I wasn't sure about this, but looking at the frontend, the implementation seems to be in agreement:
https://github.com/dlang/dmd/blob/b7fe99e9621b75ca47db021fbc7df7500ff8753a/src/dmd/lexer.d#L2399

PetarKirov (approved):

PR #2936, opened 2021-01-28
Revert "Indent continuation lines with 2 indentations"

PetarKirov:
Reverts dlang/dlang.org#2766

Motivation:

Changes to the DStyle guide can't be made solely based on one's preference. Only changes match the currently established practices can be "codified" via a lightly reviewed PR. Otherwise DStyle changes need the consensus of the majority of active contributors.
That said, such changes don't need a DIP, since this is not change that affects any users of the language, only the contributors.

### DIFF
--- a/dstyle.dd
+++ b/dstyle.dd
@@ -27,16 +27,6 @@ $(UL
     $(LI Use spaces instead of hardware tabs.)
 
     $(LI Each indentation level will be four columns.)
-
-    $(LI For line breaks inside parentheses the continuation shall be indented
-         two indentation levels beyond the first line.
-
--------------------------------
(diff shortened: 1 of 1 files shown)


Geod24 (approved):

RazvanN7:
That seemed like what we were doing anyway so I did not see any harm in having a written guideline of it. However, it folks prefer a different style, I'm fine with reverting this.

PetarKirov:
As a policy, I think it's generally safe to merge reverts to controversial PRs. In the worst case, we can simply propose the change again.

@RazvanN7 I'd like to to have this topic clarified in the DStyle, but we need the approval of more people.

PR #2942, opened 2021-02-11
Improve pragma inline section

WalterBright:


spec/pragma.dd
```diff
@@ -149,15 +149,15 @@ pragma(inline)
         )
         $(LI
 ---
-pragma(inline, boolean_expression)
+pragma(inline, AssignExpression)
 ---
-        If the expression evaluates to false the functions are never inlined, otherwise they are always inlined.
+        The $(GLINK2 expression, AssignExpression) is evaluated and must have a type that can be converted
+        to a boolean.
+        If the result is false the functions are never inlined, otherwise they are always inlined.
```
RazvanN7, line 156: Is it worthwhile to specify that the assign expression must have the ability to be evaluated at compile time?

WalterBright: I don't think it's necessary, as it makes no sense if not inlined. Since they are all evaluated at compile time, saying that would be better at the top of the page than here.

RazvanN7 (approved):

PR #2943, opened 2021-02-11
Improve pragma lib and pragma linkerDirective

WalterBright:
Putting `$(P ...)` inside an IMPLEMENTATION_DEFINED macro causes it to fail to format properly.

### DIFF
--- a/spec/pragma.dd
+++ b/spec/pragma.dd
@@ -190,8 +190,8 @@ pragma(lib, "foo.lib");
 -----------------
 
     $(IMPLEMENTATION_DEFINED
-    Typically, the string literal specifies the file name of a library file. This name
-    is inserted into the generated object file, or otherwise is passed to the linker,
+    The string literal specifies the file name of a library file. This name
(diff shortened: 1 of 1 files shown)


RazvanN7 (approved):

spec/pragma.dd
```diff
@@ -204,9 +204,8 @@ pragma(linkerDirective, "/FAILIFMISMATCH:_ITERATOR_DEBUG_LEVEL=2");
 -----------------
 
     $(IMPLEMENTATION_DEFINED
-    $(P The string literal specifies a linker directive to be embedded in the generated object file.)
-
-    $(P Linker directives are only supported for MS-COFF output.)
+    The string literal specifies a linker directive to be embedded in the generated object file.
```
RazvanN7, line 207: I think that the sentence needs to be wrppaed in a `($P )` . This might be the cause of the documentation tester.

WalterBright: `$(IMPLEMENTATION_DEFINED )` is used without `$(P )` in many places in the spec.

PR #2944, opened 2021-02-11
improve pragma mangle and pragma msg

WalterBright:


spec/pragma.dd
```diff
@@ -216,32 +216,47 @@ $(H3 $(LNAME2 mangle, $(D pragma mangle)))
...
------------------
-pragma(msg, "compiling...", 1, 1.0);
------------------
+    ---
+    pragma(msg, "compiling...", 6, 1.0); // prints "compiling...61.0" at compile time
+    ---
 
-    $(IMPLEMENTATION_DEFINED The arguments are typically presented to the user during compilation,
-    such as by printing them to the standard error stream.)
+    $(IMPLEMENTATION_DEFINED What form the message takes, and how it is presented to the user.
```
RazvanN7, line 247:
```suggestion
    $(IMPLEMENTATION_DEFINED The form the message takes and how it is presented to the user. For example, the message could be printed to the standard error stream.
```

spec/pragma.dd
```diff
@@ -216,32 +216,47 @@ $(H3 $(LNAME2 mangle, $(D pragma mangle)))
...
------------------
-pragma(mangle, "body")
-extern(C) void body_func();
------------------
 
 
 $(H3 $(LNAME2 msg, $(D pragma msg)))
 
-    $(P Constructs a message from the $(I ArgumentList).)
+    $(P Each $(ASSIGNEXPRESSION) is evaluated at compile time and then combined into a message.)
```
RazvanN7, line 241:
```suggestion
    $(P Each argument, which is an $(ASSIGNEXPRESSION), is evaluated at compile time and then combined with the others into a message.)
```

RazvanN7 (commented):
Looks good except for the minor comments.

spec/pragma.dd
WalterBright: See line 248.

RazvanN7: Yes, I was suggesting to reword the entire paragraph.

spec/pragma.dd
WalterBright: I did a shorter version, leaving out the `argument, which is an` because it's redundant.

(Note: the spec isn't a tutorial. Ideally, it should have zero unnecessary and/or redundant words in it. If it can be rephrased to say the same with with fewer words, that's even better.)

spec/pragma.dd
WalterBright: My way has fewer words :-)

RazvanN7 (approved):

PR #2946, opened 2021-02-12
Revert "Added a note to the doc saying that the feature is not yet implemented."

ljmf00:
This reverts commit 8b96e90cd7afb965ab908dfe24b6df1514dbc702.

Static Initialization of AAs is implemented in D. This note is misleading and
people can interpret this wrongly.

---

Reference-to: #2623

In fact, this code doesn't work on global scope:

```d
immutable long[string] aa = [
  "foo": 5,
  "bar": 10,
  "baz": 2000
];

unittest
{
    assert(aa["foo"] == 5);
    assert(aa["bar"] == 10);
    assert(aa["baz"] == 2000);
}
```

But inside a function, static initialization perfectly works:

```d
void main() {
    immutable long[string] aa = [
      "foo": 5,
      "bar": 10,
      "baz": 2000
    ];

    assert(aa["foo"] == 5);
    assert(aa["bar"] == 10);
    assert(aa["baz"] == 2000);
}
```

The example code could be changed to reflect the working example and this should be rather an issue.

### DIFF
--- a/spec/hash-map.dd
+++ b/spec/hash-map.dd
@@ -337,8 +337,6 @@ $(H2 $(LNAME2 advanced_updating, Advanced updating))
 
 $(H2 $(LNAME2 static_initialization, Static Initialization of AAs))
 
-        NOTE: Not yet implemented.
-
         ---------
         immutable long[string] aa = [
           "foo": 5,


adamdruppe:
Your example doesn't show static initialization. That's a runtime function call and thus a dynamic initialization.

ljmf00:
> Your example doesn't show static initialization. That's a runtime function call and thus a dynamic initialization.

Sure, it uses the runtime under the hood.

If you take this example:
```d
void main() {
    enum long[string] aa = [
      "foo": 5,
      "bar": 10,
      "baz": 2000
    ];

    assert(aa["foo"] == 5);
    assert(aa["bar"] == 10);
    assert(aa["baz"] == 2000);
}
```

It works because of CTFE. Surely it's not completed, but my point here is maybe changing the spec to reflect the actual behavior because, AFAIK, this syntax is only shown here and it's misleading and misunderstood by many people.

adamdruppe:
The way I like to phrase it to people is AA's can't cross the divide between compile time and run time. You can use them in CTFE, you can use them in runtime execution contexts, you just can't construct one at compile time then use it at run time (or obviously not vice versa).

The example inside a function works since that's all runtime context. Outside the function though, the initializer is run at CT and the variable is RT, so that's the divide it can't cross.

And this is indeed just an implementation limitation.


I do agree the wording should be clearer just I don't think deleting it without more examples and/or explanation is the right thing to do.

spec/hash-map.dd
```diff
@@ -337,8 +337,6 @@ $(H2 $(LNAME2 advanced_updating, Advanced updating))
 
 $(H2 $(LNAME2 static_initialization, Static Initialization of AAs))
 
-        NOTE: Not yet implemented.
```
maxhaton, line 340: Firstly, "not yet implemented" is terrible for a language specification - so getting rid of it is good. 

This behaviour needs to be better described - also a note on the word behaviour: In the context of a language *specification* the aim is not to describe what the language does in reality, but what it should do in order for it to be consistent in some desirable way.

https://dlang.org/spec/declaration.html#global_static_init The spec declares that this behaviour should be implemented for D to be compliant with its own specification in the sense that this declaration should be evaluable at compile time - in this case the expression is evaluable but for whatever (druntime) reason cannot be assigned to anything global or static.

The behaviour needs to be elaborated upon and the non-compliance noted in a more concrete way with an example.

maxhaton (changes requested):
Nonetheless, thanks (I had never read this part of the spec)

RazvanN7:
ping @ljmf00

ljmf00:
Hi,

> ping @ljmf00

Thanks for pinging me.

I'm so sorry for leaving my contributions hanging in dlang, dlang-community and dlang-tour organizations. I have a lot of contributions pending actually and I'm so sorry for that, but I've been not so motivated to contribute to open-source during my final course year as it leads me to some overwelming and a lot of work to do. Although I want to keep contributing, as soon as I'm free from my school assessments, and be more active in the community but right now my brain is not capable of handling such external tasks.

Anyway, I added a commit.

spec/hash-map.dd
ljmf00: I elaborated a bit more, similarly to static initialization of normal arrays. I also added a more elaborated note about not being implemented on dmd compiler.

RazvanN7:
Thanks @ljmf00 . No worries, hope you will get back once your timetable loosens up.

ljmf00:
> Thanks @ljmf00 . No worries, hope you will get back once your timetable loosens up.

I can discuss this on my free times tho, but don't expect me to answer quickly.

RazvanN7:
@maxhaton care to re-review?

PR #2956, opened 2021-02-26
Add -preview=shortenedMethods to language spec

pbackus:
See dlang/dmd#11833.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -102,6 +102,7 @@ $(GRAMMAR
 $(GNAME FunctionBody):
     $(GLINK SpecifiedFunctionBody)
     $(GLINK MissingFunctionBody)
+    $(GLINK ShortenedFunctionBody)
 
 $(GNAME FunctionLiteralBody):
     $(GLINK SpecifiedFunctionBody)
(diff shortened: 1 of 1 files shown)


adamdruppe:
Thank you for taking action!

spec/function.dd
```diff
@@ -115,8 +116,14 @@ $(GNAME MissingFunctionBody):
     $(D ;)
     $(GLINK FunctionContracts)$(OPT) $(GLINK InOutContractExpression) $(D ;)
     $(GLINK FunctionContracts)$(OPT) $(GLINK InOutStatement)
+
+$(GNAME ShortenedFunctionBody):
+    $(D =>) $(GLINK AssignExpression) $(D ;)
```
maxhaton, line 121: The parser accepts the fully Expression rather than AssignExpression, so one of them is wrong/(needs changing). 

This could also use an example as well (just a cleaned up testcase would be fine)

pbackus: Added an example.

Re: Expression vs AssignExpression, AssignExpression is what's used in the grammar for lambdas, which this is supposed to match. And the only other thing included in Expression is CommaExpression, which would be rejected by semantic analysis anyway. So I think it's the parser that's wrong here.

maxhaton (approved):

schveiguy:
https://github.com/dlang/dmd/pull/12241

CyberShadow:
@pbackus Currently https://github.com/dlang/dmd/blob/master/test/compilable/shortened_methods.d#L9 doesn't parse with the official grammar, would you be willing to have another look at this?

CyberShadow:
Added 403bdbc43fa1186fc501fd6bb07fe65dc684ad57 to #3021.

PR #2958, opened 2021-03-03
merge Function Literals and Lambda sections

WalterBright:
These sections duplicated much of the grammar and text, implying Function Literals are somehow different from Lambdas, when in fact the only difference is slightly shorter syntax. The PR merges them into one grammar.


WalterBright:
I also added handy links to function_literals from other files that referenced lambdas.

RazvanN7 (approved):

PR #2959, opened 2021-03-04
improve Function Literal section

WalterBright:
Now that Function Literal and Lambda sections have been merged, I took another pass through it to:
1. add necessary examples, such as explaining the role of `ref` in the grammar
2. make the prose more precise
3. draw more equivalences between the various ways to write a function literal
4. point out when function literals become templates
5. show use of function literals as template alias parameters
6. remove some redundant text
7. move clarifications to NOTE
8. explain the peculiar omission of `x { return statement; }` syntax

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -1691,11 +1691,7 @@ $(GNAME FunctionLiteralBody):
         $(GLINK ParameterWithAttributes) or $(GLINK ParameterWithMemberAttributes)
         can be used to specify the parameters for the function. If these are
         omitted, the function defaults to the empty parameter list $(D ( )).
-        The type of a function literal is pointer to function or
-        pointer to delegate.
-        If the keywords $(D function) or $(D delegate) are omitted,
-        it is inferred from whether $(I FunctionLiteralBody) is actually
-        accessing to the outer context.
+        The type of a function literal is a delegate or a pointer to function.
     )
 
     $(P For example:)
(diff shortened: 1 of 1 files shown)


spec/expression.dd
```diff
@@ -1749,8 +1746,22 @@ $(GNAME FunctionLiteralBody):
         }
         -------------
 
-    $(P and the following where the return type $(D int) and
-        $(D function)/$(D delegate) are inferred:)
+    $(P The use of `ref` declares the return value is done by reference:)
```
mdparker, line 1749: "is done by" should be "is returned by".

Alternatively, "the returned value is passed by reference"

spec/expression.dd
```diff
@@ -1760,11 +1771,11 @@ $(GNAME FunctionLiteralBody):
...
 
-            abc( (int c) { return 6 + b; } );  // inferred to delegate
-            def( (int c) { return c * 2; } );  // inferred to function
-          //def( (int c) { return c * b; } );  // error!
+            abc( int(int c) { return 6 + b; } );  // inferred to delegate
+            def( uint(uint c) { return c * 2; } ); // inferred to function
+          //def( int(int c) { return c * b; } );  // error!
             // Because the FunctionLiteralBody accesses b, then the function literal type
-            // is inferred to delegate. But def cannot receive delegate.
+            // is inferred to delegate. But def cannot accept a delegate argument.
```
mdparker, line 1778: "def" should be formatted as inline code.

mdparker (changes requested):

mdparker (approved):

PR #2962, opened 2021-03-06
type: mostly fix up links

WalterBright:
This is just a first pass at improving this page. It is a very aged file little changed since D started.

1. add useful links
2. spit arrays into static arrays and dynamic arrays
3. add missing interfaces
4. remove redundant enum section
5. better explanation of void and real types
6. move string exposition to where other alias types are


mdparker (approved):
LGTM

PR #2963, opened 2021-03-06
interface.dd: Improve first section

WalterBright:
1. Identify implementation defined interfaces.
2. remove unnecessary words


mdparker:
Looks like these same changes are also in #2962.

WalterBright:
I goofed. Review the other one.

PR #2965, opened 2021-03-06
Do some spring cleaning on code_coverage

maxhaton:


maxhaton:
@mdparker

articles/code_coverage.dd
```diff
@@ -35,12 +35,12 @@ $(LI It can be used to track down why a particular section of code
 $(LI Since execution counts are given for each line, it is possible
      to use the coverage analysis to reorder the basic blocks in
      a function to minimize jmps in the most used path, thus
-     optimizing it.)
-)
+     optimizing it. This is known as $(LINK2 https://en.wikipedia.org/wiki/Profile-guided_optimization, profile-guided optimization)
```
mdparker, line 38: Your addition is fine, but the existing sentence is in need of revision. That 'thus optimizing it' at the end is disruptive. And it's I suggest removing it and rewriting as: "Since execution counts are given for each line, a possible optimization is to use the coverage analysis to reorder the basic blocks in a function to minimize jmps in the most used path."

mdparker (changes requested):

articles/code_coverage.dd
maxhaton: OK will do (if you're able to push to my branch from your end feel free since I'm in bed). Minor point (subject of my blog post coming ASAP): This is one of the only optimization you can really act blindly on - i.e. the branch predictor is very good but needs a hand for that last few percent.

maxhaton: I've reworded it.

articles/code_coverage.dd
```diff
@@ -34,8 +34,10 @@ $(LI It can be used to track down why a particular section of code
 
 $(LI Since execution counts are given for each line, it is possible
      to use the coverage analysis to reorder the basic blocks in
-     a function to minimize jmps in the most used path, thus
-     optimizing it. This is known as $(LINK2 https://en.wikipedia.org/wiki/Profile-guided_optimization, profile-guided optimization)
+     a function to minimize branches in the most used path, reducing stress
+     on the processor's pipeline (A run-of-the-mill X86 processor can usually handle 48 branches in its pipeline) - this technique
```
mdparker, line 38: Two things with this line. The "A" in the parentheses should be lowercase. And the dash before "this technique" should either be an em dash, a semicolon, or just a new sentence. I don't believe there's a macro for th HTML mdash entity, so a semi-colon or a new sentence would be easier.

mdparker (changes requested):

articles/code_coverage.dd
```diff
@@ -35,12 +35,12 @@ $(LI It can be used to track down why a particular section of code
...
      a function to minimize jmps in the most used path, thus
-     optimizing it.)
-)
+     optimizing it. This is known as $(LINK2 https://en.wikipedia.org/wiki/Profile-guided_optimization, profile-guided optimization)
+))
 
 $(P Experience with code coverage analyzers show that they dramatically
 reduce the number of bugs in shipping code.
-But it isn't a panacea, a code coverage analyzer won't help with:)
+But it isn't a panacea, a code coverage analyzer won't help (Unlike Valgrind or the sanitizers available in GCC and LLVM) with:)
```
mdparker, line 45: Looks like I missed this one earlier: "Unlike" should be "unlike".

mdparker (changes requested):

articles/code_coverage.dd
maxhaton: Now a new sentence.

articles/code_coverage.dd
maxhaton: done

maxhaton:
@mdparker anything else? I have some more coming soon

mdparker (approved):

PR #2967, opened 2021-03-07
exorcise the word 'you' from the spec

WalterBright:
I'm sure you will agree that if you remove you from the text, you will find that the text almost always is shorter and you can see it reads better without it.

### DIFF
--- a/spec/arrays.dd
+++ b/spec/arrays.dd
@@ -666,14 +666,15 @@ array.length = i;
 ---------
 )
 
-        $(P Picking a good initial guess is an art, but you usually can
-        pick a value covering 99% of the cases.
+        $(P Base selection of the initial size on expected common
+        use cases, which can be determined by instrumenting the code,
+        or simply using good judgement.
         For example, when gathering user
         input from the console - it's unlikely to be longer than 80.
         )
 
-        $(P Also, you may wish to utilize the $(D reserve)
-        function to pre-allocate array data to use with the append operator.)
+        $(P The $(D reserve)
+        function expands an array's capacity for use by the append operator.)
 
 $(SPEC_RUNNABLE_EXAMPLE_COMPILE
 ---------
--- a/spec/const3.dd
+++ b/spec/const3.dd
@@ -370,7 +370,7 @@ struct S
     }
 }
 ---
-    $(P To make the return type $(D_KEYWORD immutable), you need to surround the return type with parentheses:
+    $(P To make the return type $(D_KEYWORD immutable), surround it with parentheses:
     )
 
 ---
@@ -381,7 +381,7 @@ struct S
     }
 }
 ---
-    $(P To make both the return type and the method $(D_KEYWORD immutable), you can write:
+    $(P To make both the return type and the method $(D_KEYWORD immutable), write:
     )
 ---
 struct S
--- a/spec/ddoc.dd
+++ b/spec/ddoc.dd
@@ -631,8 +631,8 @@ The generated links to D symbols are relative if they have the same root package
 as the module being documented. If not, their URLs are preceded by a
 `$(DOLLAR)(DDOC_ROOT_pkg)` macro, where `pkg` is the root package of the symbol
 being linked to. Links to D symbols are generated with a `$(DOLLAR)(DOC_EXTENSION)`
-macro after the module name. So the generated URL for `[Object]` in the above
-example is as if you had written:
+macro after the module name. Then the generated URL for `[Object]` in the above
+example is as if it had been written:
 )
 
 ---
@@ -640,7 +640,7 @@ $(DOLLAR)(DOC_ROOT_object)object$(DOLLAR)(DOC_EXTENSION)#.Object
 ---
 
 $(P
-You may define your own `DOC_ROOT_` macros for any external packages you wish
+`DOC_ROOT_` macros can be defined for any external packages
 to link to using a $(LINK2 #macros, Macros section).
 )
 
@@ -813,7 +813,7 @@ underscores or hyphens:
 $(P
 As with $(LINK2 #lists, lists), note that the initial `*` in the example above
 will be stripped because it is part of a documentation comment that is delimited
-with asterisks, so you need at least three subsequent asterisks.
+with asterisks. At least three subsequent asterisks are needed.
 )
 
 $(P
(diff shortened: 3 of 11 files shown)


spec/arrays.dd
```diff
@@ -666,14 +666,15 @@ array.length = i;
...
+        use cases, which can be determined by instrumenting the code,
+        or simply using good judgement.
         For example, when gathering user
         input from the console - it's unlikely to be longer than 80.
         )
 
-        $(P Also, you may wish to utilize the $(D reserve)
-        function to pre-allocate array data to use with the append operator.)
+        $(P The $(D reserve)
+        function is used to pre-allocate array data to use with the append operator.)
```
mdparker, line 677: To say that it preallocates data implies there's some .init stuff going on. And it doesn't need to be passive. I would say:

"The $(D reserve) function expands an array's capacity for use by the append operator."

spec/const3.dd
```diff
@@ -370,7 +370,7 @@ struct S
     }
 }
 ---
-    $(P To make the return type $(D_KEYWORD immutable), you need to surround the return type with parentheses:
+    $(P To make the return type $(D_KEYWORD immutable), surround the return type with parentheses:
```
mdparker, line 373: The second use of "the return type" here is awkward. "surround it with" should be fine.

spec/ddoc.dd
```diff
@@ -631,16 +631,16 @@ The generated links to D symbols are relative if they have the same root package
 as the module being documented. If not, their URLs are preceded by a
 `$(DOLLAR)(DDOC_ROOT_pkg)` macro, where `pkg` is the root package of the symbol
 being linked to. Links to D symbols are generated with a `$(DOLLAR)(DOC_EXTENSION)`
-macro after the module name. So the generated URL for `[Object]` in the above
-example is as if you had written:
+macro after the module name. Then the generated URL for `[Object]` in the above
+example is as if it was written:
```
mdparker, line 635: "was" -> "had been"

spec/ddoc.dd
```diff
@@ -813,7 +813,7 @@ underscores or hyphens:
 $(P
 As with $(LINK2 #lists, lists), note that the initial `*` in the example above
 will be stripped because it is part of a documentation comment that is delimited
-with asterisks, so you need at least three subsequent asterisks.
+with asterisks - at least three subsequent asterisks are needed.
```
mdparker, line 816: The - needs to be an em dash, which is longer than a regular dash and should have no spaces between it and the words before and after. `&#8212;` or `&mdash;` should do it: `"asterisks&#8212;at"` becomes "asterisks&#8212;at". That means a macro is warranted.

spec/enum.dd
```diff
@@ -281,7 +281,7 @@ template Bar(T)
 
     // ... use of 'size' at compile time ...
 
-    // If you take the address of Foo!T.size, it would also go into exe file.
+    // Taking the address of Foo!T.size also causes it to go into exe file.
```
mdparker, line 284: "into exe file" -> "into the exe file"

spec/garbage.dd
```diff
@@ -447,27 +447,27 @@ $(H2 $(LNAME2 gc_config, Configuring the Garbage Collector))
...
         has to make pessimistic assumptions if a memory slot can contain both a pointer or
         an integer value, it will still be scanned (e.g. in a `union`).
     )
 
-    $(P If you use the GC memory functions from `core.memory`, and plan to use it
-        for data with a mixture of pointers and non-pointer data you should pass the
-        TypeInfo of your allocated struct, class or type as the optional parameter.
+    $(P To use the GC memory functions from `core.memory`
+        for data with a mixture of pointers and non-pointer data, pass the
+        TypeInfo of the allocated struct, class or type as the optional parameter.
```
mdparker, line 460: IMO, we really should be using the Oxford comma everywhere. I'd put a comma after "class" in "class or type". We probably aren't using it everywhere, but I think we should insert it when we see it's missing.

spec/garbage.dd
```diff
@@ -447,27 +447,27 @@ $(H2 $(LNAME2 gc_config, Configuring the Garbage Collector))
...
         ---------
         struct S { size_t hash; Data* data; }
         S* s = cast(S*)GC.malloc(S.sizeof, 0, typeid(S));
         ---------
 
     $(P $(RED Attention:) Enabling precise scanning needs slightly more caution with
-        type declarations. For example, if you reserve a buffer as part of a struct and later
-        emplace an object instance with references to other allocations into this memory,
-        you must not use basic integer types to reserve the space. Doing so will cause the
+        type declarations. For example, if reserving a buffer as part of a struct and later
```
mdparker, line 468: "if reserving" -> "when reserving"

spec/interfaceToC.dd
```diff
@@ -307,7 +307,7 @@ $(H2 $(LNAME2 structs_and_unions, Structs and Unions))
         do the right shift and masks.
         )
 
-    $(P D does not support declaring variables of anonymous struct types. In such a case you can define a named struct in D and make it private:)
+    $(P D does not support declaring variables of anonymous struct types. In such a case define a named struct in D and make it private:)
```
mdparker, line 310: "In such a case" needs a comma after "case".

spec/template.dd
```diff
@@ -391,9 +391,9 @@ immutable(S)
 )
 
     $(P This is especially useful when used with inheritance. For example,
-        you might want to implement a final base method which returns a derived
-        class type. Typically you would return a base type, but this won't allow
-        you to call or access derived properties of the type:)
+        consider implementing a final base method which returns a derived
+        class type. Typically this would return a base type, but this won't allow
```
mdparker, line 395: Replacing "you" with "this" here has rendered the "this" in "but this won't" redundant and awkward. The "won't" was already awkward anyway, as it doesn't jibe with the "would" in the preceding clause. Something about "not allow" feels off to me in this context also. I suggest:

"but this won't allow" -> "but that would prohibit"

spec/template.dd
```diff
@@ -391,9 +391,9 @@ immutable(S)
 )
 
     $(P This is especially useful when used with inheritance. For example,
-        you might want to implement a final base method which returns a derived
-        class type. Typically you would return a base type, but this won't allow
-        you to call or access derived properties of the type:)
+        consider implementing a final base method which returns a derived
```
mdparker, line 394: Given the context of the sentence, "consider implementing" is a bit confusing. It sounds like a recommendation to implement something. That put me in the wrong frame for the rest of the sentence and I had to shift gears. "consider the implementation of" sounds more like you're asking the reader to ponder a hypothetical (which is what you're doing).

mdparker (changes requested):

mdparker (approved):
LGTM

PR #2969, opened 2021-03-07
spec: Add Built-in Type Aliases heading

ntrel:
The new heading groups 3 subheadings together for easier navigation.

### DIFF
--- a/spec/type.dd
+++ b/spec/type.dd
@@ -364,7 +364,7 @@ $(GNAME MixinType):
         ---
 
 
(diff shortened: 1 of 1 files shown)


mdparker:
Given the format of the other headings on the page that refer to "types" (Derived Data Types, User-Defined Types, Mixin Types), I believe "Aliased Types" would be a more appropriate heading.

ntrel:
@mdparker OK, updated.

mdparker (approved):
LGTM

PR #2970, opened 2021-03-07
spec: Move `typeof` from declaration.dd to type.dd

ntrel:
Although `typeof` can be used in declarations, it is more logical to define it in `type.dd` because it produces a type. It can be used in expressions, not just declarations. It should be easier for people to find in `type.dd`.

No changes to content.


mdparker (approved):
I wasn't sure about this, but Walter was already thinking about doing it. All good.

PR #2971, opened 2021-03-07
spec/type.dd: Fix GLINKs, add Type Conversions heading

ntrel:
The new heading groups 4 subheadings together for easier navigation.

GLINK can only be used for relative links.


mdparker (changes requested):
LGTM

mdparker (approved):

PR #2972, opened 2021-03-07
spec/function.dd: Tweak inout section

ntrel:
Rename *Inout Functions* to *Inout Type Modifier*, because it isn't talking about `inout` on a method. Keep old anchor for link compatibility.

Format keywords.


spec/function.dd
```diff
@@ -670,28 +671,28 @@ $(H2 $(LNAME2 inout-functions, Inout Functions))
...
-    $(P The $(D_KEYWORD inout) forms a wildcard that stands in for
-        any of mutable, const, immutable, inout, or inout const. When the
-        function is called, the inout of the return type is changed to whatever
-        the mutable, const, immutable, inout, or inout const status of the
-        argument type to the parameter inout was.
+    $(P The $(D_KEYWORD inout) token forms a wildcard that stands in for
+        any of mutable, `const`, `immutable`, `inout`, or `inout const`. When calling
+        the function, the `inout` of the return type is changed to whatever
+        the mutable, `const`, `immutable`, `inout`, or `inout const` status of the
+        argument type to the `inout` parameter was.
```
mdparker, line 678: While we're here, might as well fix the awkward wording of this sentence. At a minimum, removing "whatever" and the trailing "was" improves it.

A bigger issue is the use of "status". As worded, by listing all the possible states before the word "status", it looks like it's describing the status of each of *those* states, and not saying that one of those states will be applied. We'd be better off with this:

"the `inout` status of the return type is changed to that of the argument type to the `inout` parameter., one of [list all the states here]."

spec/function.dd
```diff
@@ -644,11 +644,12 @@ $(H2 $(LNAME2 auto-ref-functions, Auto Ref Functions))
         auto ref int foo(double x) { return x; }   // error, cannot convert double to int
         ---
 
-$(H2 $(LNAME2 inout-functions, Inout Functions))
+$(LNAME2 inout-functions, Inout Functions)
+$(H2 $(LNAME2 inout-type-modifier, Inout Type Modifier))
```
mdparker, line 648: We've been referring to "input functions" for a long while now. We also refer to "auto" functions and "auto ref" functions, even though we're technically talking about return types. Given that `inout` also is all about return types, I don't see that this particular change is appropriate.

mdparker (changes requested):

spec/function.dd
ntrel: I think `inout` can be useful inside a function body without using it in the return type, but maybe that's an uncommon case. I'll revert the change.

spec/function.dd
ntrel: Good point, I've rewritten it now.

spec/function.dd
ntrel: > inout can be useful inside a function body without using it in the return type

Now documented this too.

spec/function.dd
```diff
@@ -670,11 +670,15 @@ $(H2 $(LNAME2 inout-functions, Inout Functions))
         ---
         )
 
-    $(P The $(D_KEYWORD inout) token forms a wildcard that stands in for
-        any of mutable, `const`, `immutable`, `inout`, or `inout const`. When calling
-        the function, the `inout` of the return type is changed to whatever
-        the mutable, `const`, `immutable`, `inout`, or `inout const` status of the
-        argument type to the `inout` parameter was.
+    $(P The $(D_KEYWORD inout) keyword forms a wildcard that stands in for
+        one of these modifiers: mutable, `const`, `immutable`, `inout`, or `inout const`.
```
mdparker, line 674: There is no mutable modifier, so I don't think it's accurate to put in a list of "one of these modifiers". I think it's fine to drop "these modifiers" and just leave it at that. Or, alternatively, use "states" instead.

spec/function.dd
```diff
@@ -670,11 +670,15 @@ $(H2 $(LNAME2 inout-functions, Inout Functions))
...
+        When calling the function, the `inout` modifier of the return type is changed to
```
mdparker, line 675: I think "state" is better here in place of "modifier" as well.

spec/function.dd
```diff
@@ -670,11 +670,15 @@ $(H2 $(LNAME2 inout-functions, Inout Functions))
...
+        match one of the above modifiers in the argument type to the `inout` parameter.
```
mdparker, line 676: I think this line is difficult to parse. One way to read it is that all of the above modifiers are "in" the argument type and one of them is being picked. Then the "to" could be read as "the modifier of the argument type is being matched to the parameter". I know the intended point, but I still stumbled over it on the first pass. And I know the "to" is from the original sentence. It was awkward then, but more obvious now with your new wording.

I think if you replace "match one of the above modifiers in" to "that of" and insert "passed" before "to", it's more readable:

"the inout state of the return type is changed to that of the argument type passed to the inout parameter".

mdparker (changes requested):

mdparker (approved):

PR #2973, opened 2021-03-08
move Type grammar from declaration.dd to type.dd

WalterBright:
It's kinda where one would expect to find it.

Didn't change anything else other than adjusting links.


mdparker (approved):
LGTM

ntrel:
Great, I was thinking of doing this too. BTW this has broken some links, I'm working on a pull to fix them.

PR #2985, opened 2021-03-24
Make .identifier searchable

Imperatorn:
People have been asking about .identifier and have trouble searching for it or even know what to search for

They see stuff like this in code
.myVar

And have no idea what it is.

I've just added "dot before identifier" to increase the probability of finding it

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -1387,7 +1387,7 @@ $(GNAME PrimaryExpression):
 
 $(H3 $(LNAME2 identifier, .Identifier))
 
-    $(P $(IDENTIFIER) is looked up at module scope, rather than the current
+    $(P $(IDENTIFIER) (with a dot before identifier) is looked up at module scope, rather than the current
         lexically nested scope.
     )
 


thewilsonator:
> And have no idea what it is.

`.` is the D scope resolution operator, like `::` is in C++.  see e.g. https://stackoverflow.com/questions/75213/scope-resolution-operator-without-a-scope

Imperatorn:
> 
> 
> > And have no idea what it is.
> 
> `.` is the D scope resolution operator, like `::` is in C++. see e.g. https://stackoverflow.com/questions/75213/scope-resolution-operator-without-a-scope

Yeah, maybe something like "scope resolution operator" could've been added as well, I don't know

Imperatorn:
Why are the CyberShadow build failing?

thewilsonator:
```
[boostrap] Downloading compiler http://downloads.dlang.org/releases/2.x/2.095.0/dmd.2.095.0.linux.tar.xz
curl: (28) Operation too slow. Less than 1024 bytes/sec transferred the last 30 seconds
xz: (stdin): Unexpected end of input
tar: Unexpected EOF in archive
tar: Unexpected EOF in archive
tar: Error is not recoverable: exiting now
ERROR: bootstrapping failed.
```

network problems, I think.

thewilsonator:
force push to restart.

spec/expression.dd
```diff
@@ -1387,7 +1387,7 @@ $(GNAME PrimaryExpression):
 
 $(H3 $(LNAME2 identifier, .Identifier))
 
-    $(P $(IDENTIFIER) is looked up at module scope, rather than the current
+    $(P $(IDENTIFIER) (with a dot before identifier) is looked up at module scope, rather than the current
```
maxhaton, line 1390: Either *identifier* should be emphasized or this should say "with a dot before the identifier". FWIW This kind of language isn't ideal for the language specification.

Imperatorn: Add "the"

RazvanN7: How does this make it more searchable?

RazvanN7: Maybe we can reprhase to : "The dot identifier construction enables the user to specify that the identifier should be looked up starting from the module scope, not from the current scope. More information on look up rules could be found here. "

Imperatorn: > How does this make it more searchable?

Because you cannot literally search for "." but you can search for "dot", for example "dot before identifier" etc.

For examle if you were to search for "dlang . before identifier" you would get someting like this:
https://dlang.org/spec/declaration.html
https://dlang.org/spec/version.html
https://dlang.org/spec/lex.html
https://dlang.org/spec/attribute.html
https://dlang.org/spec/traits.html
https://dlang.org/spec/pragma.html

And if you were to follow those links trying to search for ". before identifier" or "leading ." etc you would not find anything regarding using a dot before identifier.

You would have to know beforehand what it was, like a "module scope operator" or similar to be able to find it, hence, adding something to make it more searchable.

aG0aep6G: I think y'all aren't aware that `.identifier` is also described here:
https://dlang.org/spec/module.html#module_scope_operators

I suggest to replace the whole sentence here with "See [link to module.html].". And on module.html, change "leading ." to "leading dot".

Imperatorn: > Maybe we can reprhase to : "The dot identifier construction enables the user to specify that the identifier should be looked up starting from the module scope, not from the current scope. More information on look up rules could be found here. "

Yeah, something like that.

Imperatorn: > I think y'all aren't aware that `.identifier` is also described here:
> https://dlang.org/spec/module.html#module_scope_operators
> 
> I suggest to replace the whole sentence here with "See [link to module.html].". And on module.html, change "leading ." to "leading dot".

Correct, we could link there or change the description. Otherwise we would get a catch 22 where you would have to know what "." meant before trying to search for it. Just a small change to help new users.

aG0aep6G: > I suggest to replace the whole sentence here with "See [link to module.html].". And on module.html, change "leading ." to "leading dot".

Easier done than explained: #2988

Imperatorn: Nice. It should be repeated in this file as well tho

aG0aep6G: > Nice. It should be repeated in this file as well tho

I disagree. If you think that expression.html is the better location, move the description from module.html over. Duplicated content just goes out of sync.

Imperatorn: As long as you find the information it's ok

PR #2986, opened 2021-03-25
Fixed wrong info about `long` C++ mangling

dukc:
Real mangling of `long` does not currently match the spec. According to a comment at https://github.com/dlang/druntime/pull/3400 , it's the spec that should change.

### DIFF
--- a/spec/cpp_interface.dd
+++ b/spec/cpp_interface.dd
@@ -770,12 +770,12 @@ $(H2 $(LNAME2 data-type-compatibility, Data Type Compatibility))
 
     $(TROW
     $(ARGS $(B long)),
-    $(ARGS $(B long long))
+    $(ARGS $(B long) if it is 64 bit wide, otherwise $(B long long))
     )
 
     $(TROW
     $(ARGS $(B ulong)),
-    $(ARGS $(B unsigned long long))
+    $(ARGS $(B unsigned long) if it is 64 bit wide, otherwise $(B unsigned long long))
     )
 
     $(TROW
--- a/spec/cpp_interface.dd
+++ b/spec/cpp_interface.dd
@@ -770,12 +770,12 @@ $(H2 $(LNAME2 data-type-compatibility, Data Type Compatibility))
 
     $(TROW
     $(ARGS $(B long)),
-    $(ARGS $(B long) if it is 64 bit wide, otherwise $(B long long))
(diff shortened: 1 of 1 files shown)


dukc:
cc @WalterBright  @andralex

spec/cpp_interface.dd
```diff
@@ -770,12 +770,12 @@ $(H2 $(LNAME2 data-type-compatibility, Data Type Compatibility))
 
     $(TROW
     $(ARGS $(B long)),
-    $(ARGS $(B long long))
+    $(ARGS $(B long) if it is 64 bit wide, otherwise $(B long long))
```
Geod24, line 773: I would suggest a slightly different wording:
```suggestion
    $(ARGS $(B long) on 64 bits platforms, $(B long long) otherwise)
```

Just wanted to make it more compact, also "if it is 64 bit wide" can be misinterpreted as "long is 64 bits wide", which it always is.

dukc: That would be true on POSIX, but not on Windows. according to `core.stdc.config`, Windows D `long` == Windows C++ `long`, in both 32-bit and 64-bit.

dukc:
@WalterBright @andralex decision still needed

dukc:
Decision still needed. I want to get this merged or rejected so I can move forward with https://github.com/dlang/druntime/pull/3400.

RazvanN7 (approved):

spec/cpp_interface.dd
mdparker, line 773: "64 bit wide" -> "64 bits wide"

spec/cpp_interface.dd
```diff
@@ -770,12 +770,12 @@ $(H2 $(LNAME2 data-type-compatibility, Data Type Compatibility))
...
     )
 
     $(TROW
     $(ARGS $(B ulong)),
-    $(ARGS $(B unsigned long long))
+    $(ARGS $(B unsigned long) if it is 64 bit wide, otherwise $(B unsigned long long))
```
mdparker, line 778: "64 bit wide" -> "64 bits wide"

mdparker (changes requested):

mdparker:
@dukc Please fix the two issues I pointed out and I'll merge this.

PR #2990, opened 2021-04-02
add @safe restrictions on types with invariants

WalterBright:
See https://github.com/dlang/dmd/pull/12326 and https://issues.dlang.org/show_bug.cgi?id=21665

### DIFF
--- a/spec/declaration.dd
+++ b/spec/declaration.dd
@@ -467,14 +467,27 @@ $(GNAME VoidInitializer):
         used before it is set, undefined program behavior will result.
         )
 
-        $(UNDEFINED_BEHAVIOR If a void initialized variable's value is
-        used before it is set, the behavior is undefined.
+        $(IMPLEMENTATION_DEFINED If a void initialized variable's value is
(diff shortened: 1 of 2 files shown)


RazvanN7 (approved):

spec/declaration.dd
```diff
@@ -467,14 +467,27 @@ $(GNAME VoidInitializer):
         used before it is set, undefined program behavior will result.
         )
 
-        $(UNDEFINED_BEHAVIOR If a void initialized variable's value is
-        used before it is set, the behavior is undefined.
+        $(IMPLEMENTATION_DEFINED If a void initialized variable's value is
```
PetarKirov, line 470: Please don't mix distinct changes like this in the same commit. Making the values of `void` initialized variables implementation defined has little to do with the topic of this PR. Allowing access to uninitialized memory has no place in a language that strives to be safe. If you disa "implementation defined" should be changed to "unspecified", as no implementation is ever going to define the exact bit pattern, unless it is zero, which would defeat the purpose of the feature.

PR #2995, opened 2021-04-28
First pass at bringing float.dd up to scratch

maxhaton:
This file is very old - the use of the term "internal temporary" could potentially predate the term "CTFE" (Discussion suggests the dates could line up?) so working out what was originally referred to is slightly difficult (Regardless, the - removed - detail as to performance is not really true outside extremely scalar code anyway). I didn't remove the note about x87 (yet, haven't checked whether it actually does this or not at runtime)


maxhaton:
@mdparker

mdparker (approved):
LGTM

PR #2996, opened 2021-05-01
Remove inappropriate NewExpressionWithArgs

linkrope:
D Grammar defines that a `PrimaryExpression` can be a `NewExpressionWithArgs` (instead of just a `NewExpression`). The arguments, however, are optional: `new Class` (without arguments) is a valid expression. The implementation in parse.d also has a single function `parseNewExp` that allows the arguments to be missing.
While `NewExpressionWithArgs` is just wrong for `PrimaryExpression`, I removed `NewExpressionWithArgs` because it has no further use.


RazvanN7 (approved):
Thanks!

PR #2998, opened 2021-05-06
Fix Issue 21161 - [Variadic Templates] uses outdated example from D1

maxhaton:


articles/variadic-function-templates.dd
```diff
@@ -132,16 +132,16 @@ $(H3 The D Look Ma No Templates Solution)
 ------
 import core.vararg;
 import core.stdc.stdio;
-import std.format;
-
+//For reference, this function signature is from D1.0
```
Geod24, line 135: The point of this example is to show a template-less approach. Removing the FQN from `doFormat` makes sense but mentioning D1 is just confusing IMO. If you really want it mentioned, something like:
"While the standard library favors variadic templates over runtime type information, this approach can be taken whenever variadic templates are not practical, such as when passing a delegate is needed"

maxhaton: feedback incorporated, thoughts?

maxhaton:
@mdparker

Geod24 (approved):

PR #3001, opened 2021-05-10
statement.dd: Add example for tuple expansion in foreach

MoonlightSentinel:


spec/statement.dd
```diff
@@ -815,6 +815,41 @@ $(H3 $(LEGACY_LNAME2 foreach_with_ranges, foreach-with-ranges, Foreach over Stru
         }
         ---
 
+    $(P Multiple loop variables are allowed if the `front` property returns a tuple
+        whose size corresponds to the number of variables. Each variable is assigned
+        to the corresponding value in the tuple.
+    )
+
+        ---
+        import std.typecons : Tuple;
```
dkorpel, line 824: The example suggests `foreach` with multiple loop variables expects a Phobos Tuple.
As far as I understand it, it looks for a 'compiler tuple' as constructed with `AliasSeq`, and `Tuple` converts to that using `alias this`.

MoonlightSentinel: Phobos `Tuple` is just a common example of a tuple, I didn't want to conflate a discussion of tuple implementations into this paragraph. But I'll add a clarification.

PR #3012, opened 2021-05-27
Update IRC channel (from Freenode to Libera.Chat)

CyberShadow:
Following most open-source projects, we are moving our official IRC presence from Freenode to Libera.Chat.

Some users have chosen to move to OFTC instead. It is not mentioned here but I updated https://wiki.dlang.org/ to mention OFTC as well.


Geod24 (approved):

ibuclaw (approved):
FYI to all reviewers.  We have registered ourselves as a project, and are in charge of the #d and #dlang namespaces.

As we never had a "registered" presence on FreeNode, I think it is good to take advantage of the perks we gain from being on Libera.Chat

PR #3018, opened 2021-06-11
Update year on release schedule page

andrejp88:
Better late than never :)


thewilsonator (approved):

Geod24:
Or maybe remove the year altogether, and just say "The release schedule is as follow:" since it includes the year ?

PR #3019, opened 2021-06-11
Grammar fixes identified during attempt to convert it to tree-sitter

CyberShadow:
The conversion program: https://github.com/CyberShadow/tree-sitter-d/tree/generated/generator

TODO:
- [x] Drop a7ddca3
- [x] Drop a182054880b0d4e7a4dd5437982012128224f918
- [x] Rebase 4663823


CyberShadow:
OK, I think this is a good spot for a review/merge of everything so far.

The work above is what was needed to produce a [working (albeit lightly tested) tree-sitter grammar](https://github.com/CyberShadow/tree-sitter-d/). Most of these are fixes to DDoc syntax, grammar syntax, improvements to make the grammar more machine-readable and such. More fixes will follow in the future, but they will likely be of a different kind than most of the above (i.e. fixes in the actual grammar by clarifying ambiguous definitions or adding missing syntax variants).

CC @WalterBright @RazvanN7 @mdparker (looking at the git log I believe you are most involved in maintaining the spec lately).

CyberShadow:
TODO:
- Drop a7ddca3
- Rebase 4663823

mdparker:
I don't consider myself qualified to review the actual grammar changes, but the rest of it (syntax, English), LGTM. If @WalterBright doesn't check in, maybe @MoonlightSentinel or @Geod24 can?

spec/function.dd
```diff
@@ -105,7 +104,7 @@ $(GNAME FunctionBody):
     $(GLINK ShortenedFunctionBody)
 
 $(GNAME FunctionLiteralBody):
-    $(GLINK SpecifiedFunctionBody)
+    $(GLINK BlockStatement)
```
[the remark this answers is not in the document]
MoonlightSentinel: On a side note, this rule should probably be moved to `expression.d` because it's only used in the function literal section

spec/lex.dd
```diff
@@ -13,6 +13,24 @@ phase of translation.)
 
 $(H2 $(LNAME2 source_text, Source Text))
 
+$(GRAMMAR
+$(GNAME SourceFile):
+    $(GLINK ByteOrderMark) $(GLINK Module)$(OPT)
```
MoonlightSentinel, line 18: `$(GLINK Module)` yields a dead link.

MoonlightSentinel: Does a leading byte-order mark really preclude a following shebang?
Otherwise this rule could be simplified to
```
ByteOrderMark (opt) Shebang (opt) Module (opt)
```

spec/lex.dd
```diff
@@ -265,8 +290,9 @@ $(GNAME StringLiteral):
 
     $(GLINK HexString)
     $(GLINK DelimitedString)
-    $(GLINK TokenString)
+    $(GLINK TokenString))
```
MoonlightSentinel, line 293: The closing brace should appear on the next line s.t. it's consistent with the other rules.

spec/lex.dd
```diff
@@ -332,11 +358,33 @@ $(GNAME StringPostfix):
...
+    $(B [)
+    $(B <)
+    $(GLINK Identifier)
+
+$(GNAME MatchingDelimiter):
+    $(B $(RPAREN))
+    $(B })
+    $(B ])
+    $(B >)
+    $(GLINK Identifier))
```
MoonlightSentinel, line 375: This rule could be modified to enforce the correct delimiters for almost all combinations except `Identifier`.

spec/class.dd
```diff
@@ -905,13 +902,13 @@ $(GNAME SharedStaticDestructor):
...
 $(GRAMMAR
-$(GNAME ClassInvariant):
+$(GNAME Invariant):
     $(D invariant $(LPAREN) $(RPAREN)) $(GLINK2 statement, BlockStatement)
     $(D invariant) $(GLINK2 statement, BlockStatement)
     $(D invariant $(LPAREN)) $(GLINK2 expression, AssertArguments) $(D $(RPAREN) ;)
 )
 
-    $(P $(I ClassInvariant) specify the relationships among the members of a class instance.
+    $(P Class $(I Invariant) specify the relationships among the members of a class instance.
```
RazvanN7, line 911:
```suggestion
    $(P Class $(I Invariant)s specify the relationships among the members of a class instance.
```

spec/class.dd
```diff
@@ -1009,7 +1006,6 @@ $(H2 $(LNAME2 allocators, Class Allocators))
 $(B Note): Class allocators are deprecated in D2.
 $(GRAMMAR
 $(GNAME Allocator):
-    $(D new) $(GLINK2 function, Parameters) $(D ;)
```
RazvanN7, line 1012: Even though class allocators are obsolete it seems that the parser does not complain about the following:

```d
class A
{
    new(size_t size);    
    
}
```

spec/class.dd
```diff
@@ -1089,7 +1085,6 @@ assert(foo.x == int.init);  // object is still accessible
 
 $(GRAMMAR
 $(GNAME Deallocator):
-    $(D delete) $(GLINK2 function, Parameters) $(D ;)
     $(D delete) $(GLINK2 function, Parameters) $(GLINK2 function, FunctionBody)
```
RazvanN7, line 1088: Hm, it seems that delete declarations are no longer accepted by the parser.

```d
class A
{
    delete(void *p) {}  
}
```

This fails with a parser error since  2.091.1.
Since class allocators/deallocators both end up with compilation errors, maybe it's just better if we
remove them altoghether from the spec?

adamdruppe:
On Thu, Jun 24, 2021 at 03:40:16AM -0700, Razvan Nitu wrote:
> > @@ -1009,7 +1006,6 @@ $(H2 $(LNAME2 allocators, Class Allocators))
>  $(B Note): Class allocators are deprecated in D2.
>  $(GRAMMAR
>  $(GNAME Allocator):
> -    $(D new) $(GLINK2 function, Parameters) $(D ;)
> 
> Even though class allocators are obsolete it seems that the parser does not complain about the following:


note that it is specifically still allowed to do ***@***.*** new(...);`

adamdruppe:
omg that message got wtf'd

but it is the "allocator relic" thing, with at disable new.

spec/class.dd
CyberShadow: This deletion comes from this commit:
```
spec: Remove definitions redundant with MissingFunctionBody

$(D ;) is already a possible FunctionBody (as MissingFunctionBody).
```

spec/class.dd
CyberShadow: For this specific line deletion, please see above.

For the general topic of describing old language syntax, we can use `GDEPRECATED`.

There are some use cases to allow the grammar specify also previous versions of the language. If the grammar is used for anything other than develop a conforming implementation, such as code analysis tooling or (as is my case) syntax highlighting / editor support. Such tools have reasons to support previous versions of the language too.

We can use CSS or such to make removed / deprecated parts of the grammar clearly display as such (e.g. using 50% transparency or strikethrough).

CyberShadow:
@MoonlightSentinel @RazvanN7 Thank you for the review. I updated the PR with these changes:

- Rebased to reorder and fold away reverts
- Separated changes which actually change the grammar and moved them to #3021, leaving only changes which fix syntax and other obvious errors
- Changed more occurrences of consecutive `$(D ...)` to represent contiguous characters to use a single `$(D ...)`
- Added a commit which expands the definition of `DelimitedString` to describe its nested structure (when using `()/{}/[]/<>` delimiters) within the grammar (in #3021)
- Added a commit which consistently puts the closing `)` of `GRAMMAR` blocks on its own line (and adjusted all commits here to do the same)
- Added commits which repair broken links (and fixed new broken links in the commits here)
- `$(I Invariant)s`

spec/lex.dd
CyberShadow: Yes, both the byte order mark and shebang want to be the very first bytes in the physical file. So, you can't have both.

RazvanN7 (approved):

CyberShadow:
Rebased to fix merge conflict with #3025.

Geod24 (approved):
Very nice! A few things:
- In commit "spec/lex.dd: Don't use $(D ...) for Unicode characters", there is the following message:
> $(I ...) makes the backslash look weird, so use $(B ...) (freed up by the previous commit) instead.
  I guess it's slightly outdate ?
  
- "spec: Use $(D ...) instead of $(COMMA)": Why ? Simple curiosity here, as I'm the author of the code.
- There's a lot of consistency fixes. Do you think you could amend/create a README or CONTRIBUTING so that all this knowledge is more prominently displayed / accessible ? Can be a new PR.
- A previous version of this had parenthesis fixes. Was there bug reports opened for DMD's accept-invalid ?

Feel free to self merge when ready (I assume you might want to fix the commit message).

CyberShadow:
> In commit "spec/lex.dd: Don't use $(D ...) for Unicode characters", there is the following message:

Well spotted, yes, it is outdated. I will remove it.

> "spec: Use $(D ...) instead of $(COMMA)": Why ? Simple curiosity here, as I'm the author of the code.

`$(COMMA)` is necessary to insert a literal comma in cases where it would otherwise be interpreted as an argument separator for multi-argument macros. As neither `GRAMMAR` nor `D` are multi-argument macros (i.e. they use `$0` and not `$1`, `$2` etc.), it would be synonymous with just using a comma. Furthermore, we would need to wrap it in `$(D ...)` to indicate that it is a token.

> There's a lot of consistency fixes. Do you think you could amend/create a README or CONTRIBUTING so that all this knowledge is more prominently displayed / accessible ? Can be a new PR.

I think the less documentation contributors need to read, the better. We can check some things in CI, and I hope that by making everything consistent once, it becomes easier to remain consistent simply by following the existing conventions. Some things like `$(B/D/I ...)` distinctions are more subtle though, maybe separate macros could be used instead of those.

> A previous version of this had parenthesis fixes. Was there bug reports opened for DMD's accept-invalid ?

I'm guessing you mean https://github.com/dlang/dlang.org/pull/3017 . I believe DMD does warn on unbalanced parentheses if you run it with both `-D` and `-w`, however, nobody looks at the warnings.

Geod24:
Bad bot!

PR #3022, opened 2021-06-19
spec: traits: add support for function calls when using getParameterStorageClasses

ljmf00:

### DIFF
--- a/spec/traits.dd
+++ b/spec/traits.dd
@@ -1089,7 +1089,7 @@ $(H2 $(GNAME getParameterStorageClasses))
 
     $(P
         Takes two arguments.
-        The first must either be a function symbol, or a type
+        The first must either be a function symbol, a function call, or a type
         that is a function, delegate or a function pointer.
         The second is an integer identifying which parameter, where the first parameter is
         0.
@@ -1106,6 +1106,13 @@ static assert(__traits(getParameterStorageClasses, foo, 0)[1] == "ref");
 static assert(__traits(getParameterStorageClasses, foo, 1)[0] == "scope");
 static assert(__traits(getParameterStorageClasses, foo, 2)[0] == "out");
 static assert(__traits(getParameterStorageClasses, typeof(&foo), 3)[0] == "lazy");
+
+int* p, a;
+int b, c;
+
+static assert(__traits(getParameterStorageClasses, foo(p, a, b, c), 1)[0] == "scope");
+static assert(__traits(getParameterStorageClasses, foo(p, a, b, c), 2)[0] == "out");
+static assert(__traits(getParameterStorageClasses, foo(p, a, b, c), 3)[0] == "lazy");
 ---
 )
 


ljmf00:
What happened :eyes: ? Just heisenberg's ?

spec/traits.dd
```diff
@@ -1067,7 +1067,7 @@ $(H2 $(GNAME getParameterStorageClasses))
 
     $(P
         Takes two arguments.
-        The first must either be a function symbol, or a type
+        The first must either be a function symbol, a function call, or a type
```
thewilsonator, line 1070: can you please add an example to the `SPEC_RUNNABLE_EXAMPLE_COMPILE` section following this addition.

ljmf00: Done, this is also being tested on https://github.com/dlang/dmd/pull/12712 now.

spec/traits.dd
```diff
@@ -1084,6 +1084,13 @@ static assert(__traits(getParameterStorageClasses, foo, 0)[1] == "ref");
 static assert(__traits(getParameterStorageClasses, foo, 1)[0] == "scope");
 static assert(__traits(getParameterStorageClasses, foo, 2)[0] == "out");
 static assert(__traits(getParameterStorageClasses, typeof(&foo), 3)[0] == "lazy");
+
+int* p, a;
+int b, c;
+
+static assert(__traits(getParameterStorageClasses, foo(p, a, b, c), 1)[0] == "scope");
+static assert(__traits(getParameterStorageClasses, foo(p, a, b, c), 2)[0] == "out");
+static assert(__traits(getParameterStorageClasses, foo(p, a, b, c), 3)[0] == "lazy");
```
RazvanN7, line 1093:
```
__stdin.d(15): Error: first argument to `__traits(getParameterStorageClasses, foo(p, a, b, delegate int() pure nothrow @nogc @safe => c), 1)` is not a function
__stdin.d(15):        while evaluating: `static assert((__error)[0] == "scope")`
__stdin.d(16): Error: first argument to `__traits(getParameterStorageClasses, foo(p, a, b, delegate int() pure nothrow @nogc @safe => c), 2)` is not a function
__stdin.d(16):        while evaluating: `static assert((__error)[0] == "out")`
__stdin.d(17): Error: first argument to `__traits(getParameterStorageClasses, foo(p, a, b, delegate int() pure nothrow @nogc @safe => c), 3)` is not a function
__stdin.d(17):        while evaluating: `static assert((__error)[0] == "lazy")`
```

I don't know which compiler version the documentation tester uses, but it doesn't seem to pick up the latest version. You could try a repush to see if it picks up the latest dmd and in case it doesn't you could change the examples to not be `SPEC_RUNNABLE_EXAMPLE_COMPILE`

ljmf00: Done

RazvanN7 (approved):

PR #3028, opened 2021-07-01
ImportC: add an outline

WalterBright:
This adds enough of an outline that PRs for improvements can be done in parallel without needing to constantly rebase.

### DIFF
--- a/spec/importc.dd
+++ b/spec/importc.dd
@@ -4,7 +4,166 @@ $(SPEC_S ImportC,
 
 $(HEADERNAV_TOC)
 
-$(P Documentation for ImportC coming soon!)
+    $(NOTE: This document is a work in progress.)
+
+    $(P ImportC is a C compiler embedded into the D implementation.
+    Its purpose is to enable direct importation of C files, without
+    needing to manually prepare a D file corresponding to the the declarations
+    in the C file. It enables directly compiling C files into modules, and
+    linking them in with D code to form an executable. It can be used
+    as a C compiler to compile and link 100% C programs.
+    )
+
+    $(P This is not a C reference manual nor programming tutorial.
+    It describes the specifics of the dialect of C that
+    ImportC is, and how to use it effectively.)
+
+
+$(H2 $(LNAME2 dialect, ImportC Dialect))
+
+    $(P There are many versions of C. ImportC is designed to be an
+    implementation of (LINK2 https://en.wikipedia.org/wiki/C11_(C_standard_revision), ISO/IEC 9899:1999),
+    which will hereafter be referred to as $(B C11).
+    References to the C11 Standard will be C11 followed by the paragraph number.
+    Prior versions, such as C99, C89, and K+R C, are not supported.
+    )
+
+    $(P Adjustment to the ImportC dialect is made to match the
+    behavior of the C compiler that the D compiler is matched to. For
+    example, on Win32 D is matched to the Digital Mars C compiler,
+    and can be matched to the Visual C compiler using the $(TT -mscoff32)
+    switch. Win64 D is matched to the Visual C compiler.
+    On Posix targets, the matching C compiler is Gnu C or Clang C.
+    )
+
+    $(P Further adjustment is made to take advantage of some of the D
+    implementation's capabilities.)
+
+    $(P This is all covered in the rest of this document.)
+
+
+$(H2 $(LNAME2 command-line, Invoking ImportC))
+
+    $(P The ImportC compiler can be invoked in two ways:)
+
+    $(UL
+    $(LI directly via the command line)
+    $(LI indirectly via importing a C file)
+    )
+
+    $(H3 ImportC Files on the Command Line)
+
+    $(H3 Importing C Files)
+
+
+$(H2 $(LNAME2 preprocessor, Preprocessor))
+
+    $(P ImportC does not have a preprocessor. It is designed to compile C
+    files after they have been first run through the C preprocessor.)
+
+    $(H3 Digital Mars C Preprocessor)
+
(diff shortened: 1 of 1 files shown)


spec/importc.dd
```diff
@@ -4,7 +4,164 @@ $(SPEC_S ImportC,
 
 $(HEADERNAV_TOC)
 
-$(P Documentation for ImportC coming soon!)
+    $(P ImportC is a C compiler embedded into the D implementation.
+    It's purpose is to enable direct importation of C files, without
```
CyberShadow, line 8:
```suggestion
    Its purpose is to enable direct importation of C files, without
```

spec/importc.dd
```diff
@@ -4,7 +4,164 @@ $(SPEC_S ImportC,
...
+
+    $(H3 ImportC Files on the Command Line)
+
+    $(H3 Importing C Files)
+
+
+$(H2 $(LNAME2 preprocessor, Preprocessor))
+
+    $(P ImportC does not have a preprocessor. It is designed to compile C
+    files after they have been first run through the C preprocessor.)
```
CyberShadow, line 60: Is this temporary? I thought that being able to include system headers was a big reason for this entire endeavour.

CyberShadow (commented):
Since this is going to get deployed to dlang.org immediately upon merging, maybe there should be a line somewhere mentioning that this document is still a work in progress.

CC @ibuclaw @mdparker

spec/importc.dd
WalterBright: Atila and I have plans in that direction, but they're not in the near future.

spec/importc.dd
WalterBright: done

WalterBright:
> there should be a line somewhere

Added at the beginning.

WalterBright:
@CyberShadow ?

CyberShadow:
Yes? LGTM but @ibuclaw knows more about betterC and @mdparker known more about copy-editing than me.

WalterBright:
I was hoping for an approval.

WalterBright:
As that would mean then multiple people can work on it without stepping on each other.

CyberShadow:
> I was hoping for an approval.

It would be mostly vacuous. I never used betterC, and I haven't participated in its development.

> As that would mean then multiple people can work on it without stepping on each other.

There are ways to contribute even to an open pull request, such as the GitHub suggestions feature, or sending a pull request to your pull request branch.

CyberShadow (approved):
Approving vacuously, as it's not a critical topic

spec/importc.dd
ibuclaw: > Is this temporary? I thought that being able to include system headers was a big reason for this entire endeavour.

Granted that we don't know all imports until after semantic3, how will that possibly scale? Other than being N times slower than a C compiler for needing to invoke CPP on every import seen.

Not to forget that system headers are littered with dialect-specific alternative keywords (`const`, `__const`, `__const__`) and ISO C extensions (ISO 18661-3:2015), so there's got to be a middle-man in-between the header and importC to [sanitize things](https://github.com/ibuclaw/importC/blob/main/src/keywords.c.in).

spec/importc.dd
```diff
@@ -4,7 +4,166 @@ $(SPEC_S ImportC,
...
+$(H2 $(LINK2 d-side, ImportC from D's Point of View))
+
+    $(P There is no one-to-one mapping of C constructs to D constructs, although
+    it is very close. What follows is a description of how the D side views
+    the C declarations that are imported.)
+
+    $(H3 Module Name)
+
+    $(P The module name assigned to the ImportC file is the filename stripped
+    of its path and extension. This is just like the default module name assigned
```
ibuclaw, line 119: Do we need to reassure that same name, different directory won't conflict? I.e importing both math.h and bits/math.h.

WalterBright: It works exactly the same way as the default module name. It's the same code. I didn't change a line of code for ImportC.

spec/importc.dd
WalterBright: > Other than being N times slower than a C compiler for needing to invoke CPP on every import seen.

It's not happening once per .h file. It's happening once per .h file *along with its #includes*, so it is exactly like what a C compiler would do.

> so there's got to be a middle-man in-between the header and importC

I added a section on wrapping files in a later PR, which addresses that issue.

ibuclaw: > It's not happening once per .h file. It's happening once per .h file along with its #includes, so it is exactly like what a C compiler would do.

A C compiler only calls `cpp` _once_, not once per thing.

What I'm refering to is:
```
import header1; // header1.h preprocessed and imported before semantic1
int fn()
{
    import header2; // header2.h preprocessed and imported during semantic3
}
```
That is to say, things won't scale well if we make the D compiler call the preprocessor too.

WalterBright: > A C compiler only calls cpp once, not once per thing.

Yes, I know how C compilers work :-)

> What I'm refering to is:

What I'm saying is that for:
```
import header1;
int fn()
{
    import header1;
}
```
then `header1` is only processed **once**. Just as in D, `header1` can be imported again and again, yet it is only processed once. (Unlike in C.) Now, if the user has *n* `.c` files, and each `#include`'s `stdio.h`, then `studio.h` will be preprocessed *n* times, just as in C.

There is no reliable way to treat `#include` files as modules (I've tried, and so have others), and ImportC is not an attempt to fix any of C's problems (we have DasBetterC for that).

WalterBright: In other words, I do not know of any case where an ImportC file will be preprocessed by D more often than a C compiler would, and many cases when it would be less.

PR #3033, opened 2021-07-04
spec/lex.dd: Describe Delimiter/MatchingDelimiter in grammar

CyberShadow:


spec/lex.dd
```diff
@@ -340,7 +340,37 @@ $(GNAME StringPostfix):
     $(B d)
 
 $(GNAME DelimitedString):
-    $(B q") $(I Delimiter) $(GLINK WysiwygCharacters)$(OPT) $(I MatchingDelimiter) $(B ")
+    $(B q") $(GLINK Delimiter) $(GLINK WysiwygCharacters)$(OPT) $(GLINK MatchingDelimiter) $(B ")
```
dkorpel, line 343: There should be a newline before and after `$(GLINK WysiwygCharacters)$(OPT)`, though that might be out of the scope of this PR.

dkorpel (approved):

spec/lex.dd
```diff
@@ -340,7 +340,37 @@ $(GNAME StringPostfix):
...
+$(GNAME Delimiter):
+    $(GLINK Identifier)
+
+$(GNAME MatchingDelimiter):
+    $(GLINK Identifier)
+
+$(GNAME ParenDelimitedCharacters):
+    $(GLINK WysiwygCharacter)
+    $(GLINK WysiwygCharacter) $(GSELF ParenDelimitedCharacters)
+    $(B $(LPAREN)) $(GSELF ParenDelimitedCharacters)$(OPT) $(B $(RPAREN))
```
dkorpel, line 358: Shouldn't this rule come first so it's preferred over `WysiwygCharacter`, since a ( is also a `WysiwygCharacter`?

CyberShadow: I haven't noticed this pattern elsewhere. If anything, the order seems to be more generic to less generic (e.g. in `NestingBlockCommentCharacter` and `DoubleQuotedCharacter`).

dkorpel: I thought the ordering of rules was used to express operator precedence in the parsing grammar. Thinking about the lexical grammar, ordering isn't actually the issue, but rather that `WysiwygCharacter` should not match the delimiters, or else the delimiters could optionally be skipped and interpreted as part of the string / comment. This is what the dgrammar in PEGGED does:
https://github.com/PhilippeSigaud/Pegged/blob/9084c497d29e816f69182e1215bfbbca07fbdf8a/examples/dgrammar/src/pegged/examples/dgrammar.d#L993

I'm not sure how your tree-sitter converter handles this, but these should be rejected:
```D
/+ +/ +/
"abc"def"
```

CyberShadow: Yes, for these situations I currently apply manual patches. I did do one fix of this kind to the official grammar, `TokenNoBraces` (for [`TokenString`](https://github.com/dlang/dlang.org/pull/3036/files)).

Some things are surprisingly difficult to describe exactly in this way, even simple block comments:

RazvanN7 (approved):

PR #3041, opened 2021-07-04
spec/module.dd: Specify that Module may contain zero DeclDefs

CyberShadow:


spec/module.dd
```diff
@@ -6,6 +6,7 @@ $(HEADERNAV_TOC)
 
 $(GRAMMAR
 $(GNAME Module):
+    $(GLINK ModuleDeclaration)
     $(GLINK ModuleDeclaration) $(GLINK DeclDefs)
```
dkorpel, line 10: Why not use `$(GLINK DeclDefs) $(OPT)`?

CyberShadow: This is a case of `A or B or A B`, so you would need to choose which of `A` or `B` to make optional. Stylistically, the spec also seems to be conservative in its use of `$(OPT)`, perhaps because it was added later as a necessity to prevent some definitions from being unreasonably complicated. As such, just listing all three possibilities seems to be the most straight-forward way to express this rule.

Geod24 (approved):

PR #3043, opened 2021-07-04
spec/declaration.dd: Add AtAttribute to StorageClass

CyberShadow:
According to DMD, @-attributes may occur wherever storage classes are parsed (see `Parser.parseStorageClasses`).

### DIFF
--- a/spec/declaration.dd
+++ b/spec/declaration.dd
@@ -90,6 +90,7 @@ $(GNAME StorageClasses):
 $(GNAME StorageClass):
 $(MULTICOLS 5,     $(GLINK2 attribute, LinkageAttribute)
     $(GLINK2 attribute, AlignAttribute)
+    $(GLINK2 attribute, AtAttribute)
     $(D deprecated)
     $(D enum)
     $(D static)


dkorpel:
This, along with https://github.com/dlang/dlang.org/pull/3042, closes https://issues.dlang.org/show_bug.cgi?id=20503 I think

Geod24 (approved):

spec/declaration.dd
```diff
@@ -90,6 +90,7 @@ $(GNAME StorageClasses):
 $(GNAME StorageClass):
 $(MULTICOLS 5,     $(GLINK2 attribute, LinkageAttribute)
     $(GLINK2 attribute, AlignAttribute)
+    $(GLINK2 attribute, AtAttribute)
```
MoonlightSentinel, line 93: Fine for now, but I don't think that this (and `align`) belongs into `StorageClass`.
They should probably be extracted into a different rule and combined a few layers above.

MoonlightSentinel (approved):

spec/declaration.dd
CyberShadow: Yes, attributes and storage classes are somewhat of a mess that will probably not be easy to untangle.

PR #3048, opened 2021-07-04
spec/function.dd: Add default values for anonymous parameters

CyberShadow:


dkorpel (commented):
Why is the `BasicType` case there in the first place, considering any `BasicType` can be parsed through `Type`? It looks like you can reduce it to three cases.

CyberShadow:
As I understand `BasicType` is essentially `Type` without `TypeSuffixes` (which I think are a left-over from the C-like syntax of variable declarations). Because `Declarator` already includes `TypeSuffixes`, but anonymous parameters don't have the part of `Declarator` for the parameter name, therefore the current rule.

dkorpel:
Makes sense. Apparently, C-style declarations like `int a[]` are still allowed in the grammar although dmd rejects it (even inside `version(none)`), so those should be removed at some point. `Parameter` can then be simplified to `ParameterAttributes Type Identifier(opt) ParameterSuffix(opt)`. In the meantime, this is a good minimal changes fix.

dkorpel (approved):

Geod24 (approved):

PR #3049, opened 2021-07-04
spec/function.dd: Remove "ref" from "return ref"

CyberShadow:
Return-ref parameters don't need to specifically be declared with the exact token sequence `return ref`. For example, `ref return` also works, as does e.g. `return scope int* a`.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -56,7 +56,7 @@ $(GNAME InOut):
     $(D lazy)
     $(D out)
     $(D ref)
-    $(RELATIVE_LINK2 return-ref-parameters, $(D return ref))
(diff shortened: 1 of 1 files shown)


spec/function.dd
```diff
@@ -56,7 +56,7 @@ $(GNAME InOut):
     $(D lazy)
     $(D out)
     $(D ref)
-    $(RELATIVE_LINK2 return-ref-parameters, $(D return ref))
+    $(RELATIVE_LINK2 return-ref-parameters, $(D return))
```
dkorpel, line 59: The link to return-ref-parameters is now confusing since it can equally well mean return-scope now, maybe remove it?

dkorpel (approved):

PR #3051, opened 2021-07-04
spec/iasm.dd: Expand definition of Register and Register64

CyberShadow:
These were previously defined as a set of tokens loosely arranged in a table. Unfortunately, this layout was ambiguous, and if interpreted literally (or by a program) in the same way as the rest of the grammar, they would mean that the tokens on the same line actually have to appear one after another in the source file in the specified order.

As such, we have to sacrifice a bit of human-readability to achieve machine readability. If necessary, the former can be later improved using CSS or additional macros, such as `MULTICOLS`.


dkorpel (approved):

ibuclaw:
@CyberShadow do you support iasmgcc grammar in treesitter?

CyberShadow:
Not yet. Looks like item # 10 here: https://github.com/CyberShadow/tree-sitter-d/issues/3

Would you happen to have a description of the syntax, maybe in the same grammar format as used by the spec?

ibuclaw:
> Would you happen to have a description of the syntax, maybe in the same grammar format as used by the spec?

Posted in the issue.

PR #3055, opened 2021-07-04
spec/iasm.dd: Add definition for Opcode

CyberShadow:
The full set of supported opcodes is probably an implementation detail, so we should at least include definitions for the syntax used to lex them.


CyberShadow:
The full list is actually here:

https://dlang.org/spec/iasm.html#supported_opcodes

But, I don't think it would make sense to restrict this list to whatever DMD supports.

Geod24 (approved):

PR #3059, opened 2021-07-10
spec/function.dd: Document optional contracts for ShortenedFunctionBody

CyberShadow:
See e.g. https://github.com/dlang/dmd/blob/21c3c5ab24ef2bf42822e7a41ab152df3f0434a4/test/compilable/shortened_methods.d#L9

Block contracts are allowed, as long as the last contract is using the expression syntax.

I think I should have submitted this as part of the previous batch, but for some reason it fell through.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -118,6 +118,7 @@ $(GNAME MissingFunctionBody):
 
 $(GNAME ShortenedFunctionBody):
     $(D =>) $(GLINK2 expression, AssignExpression) $(D ;)
+    $(GLINK FunctionContracts)$(OPT) $(GLINK InOutContractExpression) $(D =>) $(GLINK2 expression, AssignExpression) $(D ;)
 )
 
         $(P Examples:)


dkorpel:
Huh, using contracts with `=>` syntax like this looks like a weird combination. It's not allowed in expression lambdas with `=>` syntax, but you can apply `=>` to traditional functions with contracts such as:

```D
int f() in {
    // checks
} do {return 3;}
```

But you have to do:
```D
int f() in {
    // checks
} in(true) => 3;
```

I think it's actually better if the implementation gets changed to the spec instead of the spec to the implementation. shortenedMethods is in this weird state of being a preview feature without formal approval AFAIK, but still in the official spec. Not sure what to do here.

CyberShadow:
CC @adamdruppe

adamdruppe:
These working with contracts was an intentional feature, though I never tested it with the block contract syntax! So I could go either way there, I think a block contract ought to have a block body anyway for code style.

But since it is a function declaration, all features of function declarations ought to be possible. UDAs, const, and yes, contracts.

CyberShadow:
Do you think you could improve the implementation in DMD to make it work as expected with block contracts?

adamdruppe:
On Sun, Jul 11, 2021 at 05:32:59AM -0700, Vladimir Panteleev wrote:
> Do you think you could improve the implementation in DMD to make it work as expected with block contracts?

maybe, it is a question of finding the time. it is probably just adding
a check for => after the do keyword whereas right now i think i check for
it only beside the do keyword.

maxhaton (approved):

CyberShadow:
On pause as this project keeps getting bumped off of the top of my list, will need to revisit it as time allows.

PR #3060, opened 2021-07-10
spec/lex.dd: Specify DelimitedString mandatory line breaks in grammar

CyberShadow:
This implements a suggestion from a previous PR: https://github.com/dlang/dlang.org/pull/3033#discussion_r663481518

We can't describe (within the grammar) that `Delimiter` and `MatchingDelimiter` must match, nor that they can't appear within the line, but this brings us a iota closer towards accuracy.

### DIFF
--- a/spec/lex.dd
+++ b/spec/lex.dd
@@ -340,7 +340,7 @@ $(GNAME StringPostfix):
     $(B d)
 
 $(GNAME DelimitedString):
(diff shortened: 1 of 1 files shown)


dkorpel (approved):

dkorpel:
> We can't describe (within the grammar) that `Delimiter` and `MatchingDelimiter` must match, nor that they can't appear within the line, but this brings us a iota closer towards accuracy.

We also might need to figure out how to exclude delimiter characters (newline in this case) from WysiwygCharacter

PR #3061, opened 2021-07-10
Fix issue 20557 - add StringPostfix after DelimitedString and TokenString

dkorpel:


CyberShadow (approved):
Also I realized I forgot to cover the punctuation-but-not-an-opener case when expanding `DelimitedString`, as in the `q"!test!"w;` example.

PR #3062, opened 2021-07-10
spec/iasm.dd: Add the GDC inline assembler grammar

CyberShadow:
```
The root definition (GccAsmStatement) is purposefully not referred to
anywhere from the rest of the grammar. This allows implementations
(grammar consumers) to opt-in by replacing or augmenting AsmStatement
with GccAsmStatement as appropriate.
```

Adapted from here: https://github.com/CyberShadow/tree-sitter-d/issues/3#issuecomment-873658827

Draft because it seems there still is some code apparently using GCC assembler syntax which does not conform to the grammar added here so far: https://github.com/CyberShadow/tree-sitter-d/issues/3#issuecomment-877692161

### DIFF
--- a/spec/iasm.dd
+++ b/spec/iasm.dd
@@ -1100,6 +1100,50 @@ $(H3 $(LNAME2 simd, SIMD))
 
     $(P SSE, SSE2, SSE3, SSSE3, SSE4.1, SSE4.2 and AVX are supported.)
 
+$(H2 $(LNAME2 gcc, GCC syntax))
+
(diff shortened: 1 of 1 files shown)


ibuclaw:
I've checked, and that syntax is [deprecated](https://github.com/dlang/dmd/pull/10820) and will be removed in a future version.  The enclosing parentheses around the operand expression is a requirement of the syntax.

WalterBright (approved):

PR #3067, opened 2021-07-16
ImportC: add section on enums

WalterBright:


spec/importc.dd
```diff
@@ -112,6 +112,26 @@ $(H2 $(LNAME2 preprocessor-directives, Preprocessor Directives))
     directives are normally embedded in the output of C preprocessors.)
 
 
+$(H2 $(LNAME2 implementation, Implementation))
+
+    $(P The implementation defined characteristics of ImportC are:)
```
maxhaton, line 117: Is this actually implementation defined - I was under the impression that the standard says they should be always of type `int` anyway, but other compilers don't comply (that being said I don't have an up to date copy of the standard)

RazvanN7 (approved):

PR #3069, opened 2021-07-19
explain difference between ImportC and BetterC

WalterBright:
Because people will ask!

### DIFF
--- a/spec/betterc.dd
+++ b/spec/betterc.dd
@@ -63,6 +63,10 @@ Hello betterC
     be tricky, hence removing D's GC from the equation may be worthwhile sometimes.)
     )
 
+    $(P BetterC and $(LINK2 https://dlang.org/spec/importc.html) are very different.
+    ImportC is an actual C compiler. BetterC is a subset of D that relies only on the
+    existence of the C Standard library.)
+
 
 $(H2 $(LNAME2 retained, Retained Features))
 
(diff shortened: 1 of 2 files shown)


spec/betterc.dd
```diff
@@ -63,6 +63,10 @@ Hello betterC
     be tricky, hence removing D's GC from the equation may be worthwhile sometimes.)
     )
 
+    $(P BetterC and $(LINK2 https://dlang.org/spec/importc.html) are very different.
+    ImportC is an actual C compiler. BetterC is a subset of D that relies only on the
+    existance of the C Standard library.)
```
dkorpel, line 68: existence

dkorpel (approved):

spec/betterc.dd
dkorpel: Then again, the compiler still likes to emit `memcpy`, `memcmp` and `memset` calls, so I suppose it's true that it relies on the existence of the C Standard library even though it doesn't require it.

WalterBright: BetterC still depends on the C Standard library for things like its assert implementation. Besides, it's not meant to be a thorough review of what it can do, the link suffices for that.

adamdruppe: yeah it assumes at least a few pieces of the c stdlib. pretty easy to stub out though. (...but then again so is the D runtime...)

PR #3072, opened 2021-07-22
Remove documentation on class deallocators

ibuclaw:
This was removed from the parser in [2020-01](https://github.com/dlang/dmd/pull/10727) by @Geod24.

FYI @CyberShadow in-case you were using this in your tree-sitter grammar.


CyberShadow:
If they are *removed* from the frontend, then they are OK to remove from the grammar too.

I would like to restore a few things (using the `GDEPRECATED` macro) that have been deprecated in DMD but completely removed in the grammar, because we can't build tools based on the grammar that would support this deprecated but still valid code.

maxhaton (approved):

PR #3073, opened 2021-07-26
switch expression and integral promotions

WalterBright:
Clarify what switch expressions are accepted, and when integral promotions on it are done. This documents existing behavior.

### DIFF
--- a/spec/statement.dd
+++ b/spec/statement.dd
@@ -1103,9 +1103,17 @@ $(GNAME StatementNoCaseNoDefault):
     $(GLINK ScopeBlockStatement)
 )
 
-        $(P $(EXPRESSION) is evaluated. The result type T must be
-        of integral type or $(CODE char[]), $(CODE wchar[]) or $(CODE dchar[]).
-        The result is
+        $(P $(I Expression) is evaluated.
+        If the type of the $(I Expression) is an enum,
+        it is repeatedly converted
+        to its $(GLINK2 enum, EnumBaseType) until it isn't. Then,
+        if the type is of integral type,
+        the $(I Expression) undergoes $(DDSUBLINK spec/type, integer-promotions, Integer Promotions).
+        Then, the type of the $(I Expression) must be either of integral type or
+        a static or dynamic array of $(CODE char), $(CODE wchar), or $(CODE dchar).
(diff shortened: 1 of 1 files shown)


PetarKirov (approved):

PetarKirov:
> WalterBright requested a review from andralex

I missed that. @andralex feel free to open a follow-up PR, if you want to change the wording.

spec/statement.dd
```diff
@@ -1103,9 +1103,17 @@ $(GNAME StatementNoCaseNoDefault):
     $(GLINK ScopeBlockStatement)
 )
 
-        $(P $(EXPRESSION) is evaluated. The result type T must be
-        of integral type or $(CODE char[]), $(CODE wchar[]) or $(CODE dchar[]).
-        The result is
+        $(P $(I Expression) is evaluated.
+        If the type of the $(I Expression) is an enum,
```
andralex, line 1107: The rule to follow here is, anything that is not an English word must have some annotation other than plain text. The word "enum" is not in the dictionary, so it needs to be in code font.

spec/statement.dd
```diff
@@ -1103,9 +1103,17 @@ $(GNAME StatementNoCaseNoDefault):
...
+        it is repeatedly converted
+        to its $(GLINK2 enum, EnumBaseType) until it isn't. Then,
+        if the type is of integral type,
```
andralex, line 1110: A type does not have a type. So "if the type is integral" would be an alternative.

spec/statement.dd
```diff
@@ -1103,9 +1103,17 @@ $(GNAME StatementNoCaseNoDefault):
...
+        the $(I Expression) undergoes $(DDSUBLINK spec/type, integer-promotions, Integer Promotions).
+        Then, the type of the $(I Expression) must be either of integral type or
```
andralex, line 1112: "the type of. [...] must be [...] of integral type" again does not work, so perhaps: "Then, the resulting type must be either integral or ..."

andralex:
I've added a few notes, thanks.

WalterBright:
@andralex see https://github.com/dlang/dlang.org/pull/3075

PR #3075, opened 2021-07-26
minor edits to #3703

WalterBright:
In response to @andralex 's comments on https://github.com/dlang/dlang.org/pull/3073

### DIFF
--- a/spec/statement.dd
+++ b/spec/statement.dd
@@ -1107,12 +1107,12 @@ $(GNAME StatementNoCaseNoDefault):
 )
 
         $(P $(I Expression) is evaluated.
-        If the type of the $(I Expression) is an enum,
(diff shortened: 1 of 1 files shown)


spec/statement.dd
```diff
@@ -1107,12 +1107,12 @@ $(GNAME StatementNoCaseNoDefault):
 )
 
         $(P $(I Expression) is evaluated.
-        If the type of the $(I Expression) is an enum,
+        If the type of the $(I Expression) is $(CODE enum),
```
PetarKirov, line 1110: Why wrap `enum` in `$(CODE)`, instead of in `` ` ``s?
```suggestion
        If the type of the $(I Expression) is `enum`,
```

maxhaton (approved):

PR #3076, opened 2021-07-26
specify @trusted functions must have @safe interface

WalterBright:
Fix omission from https://github.com/dlang/dlang.org/pull/2453#commitcomment-53328593

### DIFF
--- a/spec/memory-safe-d.dd
+++ b/spec/memory-safe-d.dd
@@ -38,7 +38,7 @@ $(H2 $(LNAME2 usage, Usage))
         $(P `@trusted` functions have all the capabilities of `@system` functions but may be called from
         `@safe` functions. For this reason they should be very limited in the scope of their use. Typical uses of
         `@trusted` functions include wrapping system calls that take buffer pointer and length arguments separately so that
-        @safe` functions may call them with arrays.)
+        `@safe` functions may call them with arrays. `@trusted` functions must have a $(DDSUBLINK spec/function, safe-interfaces, safe interface).)
 
         $(P `@safe` functions have a number of restrictions on what they may do and are intended to disallow operations that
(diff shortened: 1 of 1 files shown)


joseph-wakeling-frequenz:
Thanks @WalterBright!

FWIW here is the phrasing that was deleted from the spec when changes were made to reflect -dip1000:
https://github.com/dlang/dlang.org/pull/2453/files#diff-291012f1595e64cfa146ae8b4283c199ce2bc2b6bf83ce94818f330121a08da9L30-L31

spec/memory-safe-d.dd
```diff
@@ -38,7 +38,7 @@ $(H2 $(LNAME2 usage, Usage))
         $(P `@trusted` functions have all the capabilities of `@system` functions but may be called from
         `@safe` functions. For this reason they should be very limited in the scope of their use. Typical uses of
         `@trusted` functions include wrapping system calls that take buffer pointer and length arguments separately so that
-        @safe` functions may call them with arrays.)
+        @safe` functions may call them with arrays. `@trusted` functions must have an `@safe` interface.)
```
joseph-wakeling-frequenz, line 41: It might be better to phrase this as:

```
`@trusted` functions must have a safe interface
```

or:

```
`@trusted` functions must have a memory-safe interface
```

to avoid confusion between safe and `@safe` (if that makes sense).  Cf. how things were phrased in the original version of the spec (link in separate comment by me).

jmh530: Would it be useful to provide an example of a "memory-safe interface"?

aG0aep6G: > Would it be useful to provide an example of a "memory-safe interface"?

We already have examples on another page: <https://dlang.org/spec/function.html#safe-interfaces>. Don't repeat all that. Just link there.

dkorpel (approved):

PR #3079, opened 2021-07-29
[DSSv3] Issue 21637: trailing comma in import list

dorianverna17:
Added an example of the functionality implemented in the pull request https://github.com/dlang/dmd/pull/12917 at the section regarding selective imports

### DIFF
--- a/spec/module.dd
+++ b/spec/module.dd
@@ -418,6 +418,30 @@ void main()
 ---
 )
 
+$(P There can also be a trailing comma at the end of the import list:)
+
+$(RUNNABLE_EXAMPLE
+---
+void foo()
+{
+	import std.stdio: writeln, write,;
+	write("foo");
+}
+
+void bar()
+{
+	import std.stdio: writeln,;
+	writeln("bar");
+}
+
+int main() {
+	foo();
+	bar();
+	return 1;
+}
+---
+)
+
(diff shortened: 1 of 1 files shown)


dkorpel (commented):
I don't think this is worth bringing attention to in the specification. Trailing commas are allowed in general (argument list, array literal, enum members, case statement) and there are no examples there either.

dorianverna17:
Should I close this pull request @dkorpel ?

dkorpel:
> Should I close this pull request @dkorpel ?

The change should still be reflected in the grammar:
https://github.com/dlang/dlang.org/blob/0a90480d5a6967eabbf8f4a7e62709c9d51d55b6/spec/module.dd#L196-L199

You can use this PR to do that. For reference, here's how ParameterList is specified:

https://github.com/dlang/dlang.org/blob/0a90480d5a6967eabbf8f4a7e62709c9d51d55b6/spec/function.dd#L33-L37

dkorpel (commented):
From the implementation it seems you actually modified `ImportBindList`, not `ImportList`, so I pointed you to the wrong rule. Also, your implementation allows an empty bind list, but the grammar does not unless you add `$(OPT)` behind `ImportBindList` (similar to ParameterList as well).

PR #3080, opened 2021-07-30
fix Issue 3345: Static and nonstatic methods with the same name shoul…

WalterBright:
…d be allowed

This is how it was always designed, but never spec'ed.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -1347,6 +1347,33 @@ void test()
         specialized than a function without.
         )
 
+        $(P A static member function can be overloaded with a member
+        function. The struct, class
+        or union of the static member function is inferred from the
+        type of the `this` argument.)
+
+        ---
+        struct S {
+            void eggs(int);
+            static void eggs(long);
+        }
+        S s;
+        s.eggs(0);  // calls void eggs(int);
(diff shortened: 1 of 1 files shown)


spec/function.dd
```diff
@@ -1347,6 +1347,33 @@ void test()
...
+        struct T {
+            void bacon(int);
+            static void bacon(int);
+        }
+        T t;
+        t.bacon(0);  // error: ambiguous
+        T.bacon(0);  // error: ambiguous
+        ---
+
+        $(RATIONALE  This so a static member function that doesn't need
```
RazvanN7, line 1375:
```suggestion
        $(RATIONALE  A static member function that doesn't need
```

spec/function.dd
```diff
@@ -1347,6 +1347,33 @@ void test()
         specialized than a function without.
         )
 
+        $(P A static member function can be overloaded with a member
+        function. The struct class
```
RazvanN7, line 1351:
```suggestion
        function. The struct, class
```

RazvanN7 (changes requested):

WalterBright:
@RazvanN7 done

RazvanN7 (approved):

PR #3082, opened 2021-08-04
spec: DIP1029: Add `throw` as Function Attribute

ljmf00:


dukc (changes requested):
Better be consistent with rest of the function attributes. Change the already existing section `$(D nothrow) Attribute` to `$(D throw) and $(D nothrow) attributes`.

Also the other attributes in this file only have links to `function.dd` where the real descriptions are. Follow that pattern.

PR #3093, opened 2021-08-25
zero size structs with extern (C) attribute

WalterBright:
Clear up documentation to correspond with https://github.com/dlang/dmd/pull/13009


spec/struct.dd
```diff
@@ -62,7 +62,9 @@ $(H2 $(LNAME2 struct_layout, Struct Layout))
     the first field and the start of the object.
     )
 
-    $(P Structs with no fields of non-zero size (aka $(I Empty Structs)) have a size of one byte.)
+    $(P A struct with no fields of non-zero size is an $(I Empty Struct).
+    An Empty Struct with a $(DDSUBLINK spec/attribute, linkage, Linkage Attribute) of `extern (C)` has a size of zero.
+    Otherwise, its size is one byte.)
```
dukc, line 67: This is wrong. I think you wanted to say that empty `extern (C) struct` size is the same as with the underlying C compiler. Also you might want to add what happens if the C compiler simply disallows empty structs?

spec/struct.dd
```diff
@@ -95,8 +99,9 @@ $(H2 $(LNAME2 struct_layout, Struct Layout))
     attributes to describe an exact match. Using a $(DDSUBLINK spec/version, static-assert, Static Assert)
     to ensure the result is as expected.)
     $(LI Although the contents of the padding are often zero, do not rely on that.)
-    $(LI Avoid using empty structs when interfacing with C and C++ code.)
-    $(LI Avoid using empty structs as parameters or arguments to variadic functions.)
+    $(LI Avoid using Empty Structs when interfacing with C++ code that declares them with `extern "C"`.
+    Add a one byte dummy field to those structs.)
```
dukc, line 103: Confusing. It's irrelevant if the C++ code declares then as `extern (C)` or not. And on D side, you can declare the `struct` as `extern (D)` or `extern (C++)` to accomplish the same - no need for the dummy field.

dukc (changes requested):

dukc:
ping @WalterBright

spec/struct.dd
WalterBright: C11 6.7.2.1-8 says "If the struct-declaration-list does not contain any named members, either directly or via an anonymous structure or anonymous union, the behavior is undefined."

Undefined behavior is not disallowed.

spec/struct.dd
WalterBright: You're right that empty C++ structs have sizeof 1 even for `extern "C++"` structs.

WalterBright: I removed the line about `extern "C"`.

spec/struct.dd
WalterBright: If we encounter a C compiler with different behavior, we'll adjust the spec.

ibuclaw: From what I recall, the problem being that gcc and clang don't agree with each other _on the same platform_. So you could choose to imitate gcc on Linux, but then you're abi incompatible with clang on Linux - which someone might instead choose to use.

But that's all by the by.

dkorpel (approved):

spec/struct.dd
dukc: > Undefined behavior is not disallowed.

The point is, undefined behaviour allows refusal to compile. A reasonable C compiler could emit an error when encountering an empty struct.

> If we encounter a C compiler with different behavior, we'll adjust the spec.

Better, but why not do it right away?

PR #3095, opened 2021-09-04
Update arrays.dd

MrcSnm:
Mention new expression for initializing dynamic array length

### DIFF
--- a/spec/arrays.dd
+++ b/spec/arrays.dd
@@ -779,22 +779,20 @@ $(H3 $(LNAME2 default-initialization, Default Initialization))
         $(LI Dynamic arrays are initialized to having 0 elements.)
         $(LI Associative arrays are initialized to having 0 elements.)
         )
-        
- $(H3 $(LNAME2 length-initialization, Length Initialization))
+
+$(H3 $(LNAME2 length-initialization, Length Initialization))
         
         $(P The $(D new) expression can be used to start a dynamic array
         with a specified length by specifying its type and then using the
         () syntax:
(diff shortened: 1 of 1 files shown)


dukc (changes requested):
I think this one better befits the "Array Initialization" section rather than the "Array properties" section.

MrcSnm:
> I think this one better befits the "Array Initialization" section rather than the "Array properties" section.

That section is actually `Setting Dynamic Array Length`. I didn't made into Array Initialization section because I saw it was only mentioning initializing it with data, and not its size.

maxhaton:
```
./spec/arrays.dd:782:        
./spec/arrays.dd:784:        
./spec/arrays.dd:789:        
./spec/arrays.dd:796:
```

Whitespace checker says jump, we have to say "how high?".

MrcSnm:
Did my best to edit it correctly, but idk how it really should be. Writing Ddoc is too strange for me yet

spec/arrays.dd
```diff
@@ -780,6 +780,20 @@ $(H3 $(LNAME2 default-initialization, Default Initialization))
         $(LI Associative arrays are initialized to having 0 elements.)
         )
 
+$(H3 $(LNAME2 length-initialization, Length Initialization))
+
```
maxhaton, line 784:
```suggestion

```
Little bit of trailing whitespace here

dukc (approved):

maxhaton (approved):

PR #3097, opened 2021-09-08
max number of cases in CaseRange is 257

WalterBright:
as it was left unspecified.

### DIFF
--- a/spec/statement.dd
+++ b/spec/statement.dd
@@ -1128,6 +1128,7 @@ $(GNAME StatementNoCaseNoDefault):
 
         $(P A $(I CaseRangeStatement) is a shorthand for listing a series
         of case statements from $(I FirstExp) to $(I LastExp).
+        The maximum number of such case statements is 257.
         )
 
 


spec/statement.dd
```diff
@@ -1128,6 +1128,7 @@ $(GNAME StatementNoCaseNoDefault):
 
         $(P A $(I CaseRangeStatement) is a shorthand for listing a series
         of case statements from $(I FirstExp) to $(I LastExp).
+        The maximum number of such case statements is 257.
```
dkorpel, line 1131: The maximum number of such case statements *per CaseRangeStatement*. 
This is accepted by dmd: `case 0: .. case 256: case 257: .. case 512:`

UplinkCoder:
should that be 256?

dkorpel (commented):
Is this something the language needs to nail down, or is this an implementation limitation? Maybe specify that an implementation must support *at least* a range of up to 257 cases.

spec/statement.dd
UplinkCoder: I am pretty sure it's an off-by-one error in the check.
either it should be a power-of-two or not be limited at all.

dkorpel: It is, see https://github.com/dlang/dmd/pull/13058
There is no reason the limit needs to be a power of 2, we could make it 300 if we want to. Lowering it would be a breaking change (although an unlikely one).

maxhaton: Why have a limit in the specification at all (i.e. why actually pick a number beyond making an implementation choose one)?

dkorpel:
@WalterBright Do you still want to add this to the spec?

dkorpel:
Closing this for now because of inactivity and because this doesn't need to be specified.

PR #3102, opened 2021-09-26
Remove unsupported mechanism for keeping pointers alive

schveiguy:
As noted by @WalterBright on the D forums, only `GC.addRoot` is a supported mechanism for keeping a pointer to GC data alive, you may not rely on stack references.

See https://forum.dlang.org/post/siol11$2hic$1@digitalmars.com

### DIFF
--- a/spec/interfaceToC.dd
+++ b/spec/interfaceToC.dd
@@ -108,9 +108,6 @@ $(H2 $(LNAME2 storage_allocation, Storage Allocation))
         $(LI Making a copy of the data using core.stdc.stdlib.malloc() and passing
         the copy instead.)
 
-        $(LI Leaving a pointer to it on the stack (as a parameter or
-        automatic variable), as the garbage collector will scan the stack.)
-
         $(LI Leaving a pointer to it in the static data segment, as the
         garbage collector will scan the static data segment.)
 


UplinkCoder:
Perhaps mention addRoot?

schveiguy:
It is mentioned.

UplinkCoder (approved):
LGTM

UplinkCoder:
my bad I didn't look at the whole page.

maxhaton:
I'm in favour of this change for now, but it may be worth mentioning that the advanced user can add a reference to the stack manually. The issue is really a misunderstanding of the toolchain and stack.

schveiguy:
> the advanced user can add a reference to the stack manually

How? I've tried quite a bit, and it's not really possible to do reliably.

schveiguy:
TBH, I would prefer a straight mechanism to say "this thing must stay live until this line". I don't really care how it's done.

WalterBright:
It would be better to add the fuller explanation I wrote in the n.g. For instance, if you're going to actually use the reference later, it is not necessary to use addRoot.

schveiguy:
> For instance, if you're going to actually use the reference later, it is not necessary to use addRoot.

As I [mentioned](https://forum.dlang.org/post/sinn8p$586$1@digitalmars.com), this still doesn't change anything. I can use it later, and it still may be collected early. Only certain "usages" are going to keep it allocated, and only if the optimizer doesn't grow intelligence to prove that it actually doesn't need it (or thinks it doesn't need it).

e.g.:

```d
auto c = new C;
someCode(); // might collect c
useLater(c); // doesn't necessarily help
```

I think it is important to either say what specific usage is *guaranteed by the language* to keep the reference scannable, or get rid of the notion that you can do anything to stop it from being collected aside from using `GC.addRoot`.

WalterBright:
> I think it is important to either say what specific usage is guaranteed by the language to keep the reference scannable

Sure. Shouldn't be too hard to do - it has to be a use that cannot be done at compile time.

schveiguy:
```d
void useLater(C) { writeln("not compile time"); }
```

schveiguy:
The definition is that the usage cannot be optimized out, it has nothing to do with compile time. And what "optimized out" means is implementation dependent. There are some usages I can write that work on DMD, but not ldc. There are some that work on DMD and LDC, but not GDC.

With inlining a possibility, you can't guarantee usage by making it a parameter. LDC can inline virtual functions if it knows what the object's type must be. It is really really hard to find a way to get this to stay live.

WalterBright:
`useLater(C)` is not using `C`, so is not reliable. Passing a runtime value around is not using it.

WalterBright:
> The definition is that the usage cannot be optimized out, it has nothing to do with compile time. And what "optimized out" means is implementation dependent.

Yes, but a runtime value *cannot* be optimized away regardless of how advanced the optimizer gets, and hence is reliable. For example, reading the time of day, then printing it.

WalterBright (changes requested):
It's good, but not sufficient. See my other comments.

schveiguy:
`useLater` might only be a noop due to static introspection. But that's not how it looks anyway, it looks to the user like he used it.

This is really hard to get right, and is setting a subtle trap for someone if they are depending on that usage keeping it alive.

D is not the only language with these traps. This is why C# and Go have special primitives to keep it alive. Java, as far as I know, doesn't have a keepalive primitive, but still has the trap.

> a runtime value _cannot_ be optimized away regardless of how advanced the optimizer gets

It can already be optimized away, if the optimizer can prove that it has a certain value (or the later usage will result in no code because of its value).

But this is kind of nit-picking here. If you agree the current spec is not correct, but think there is a better way to write it, for sure, write that update and I can close this one.

schveiguy:
As an example of a value that is not visible at compile time:

```d
class C
{
   const size_t x;
   this(size_t x) { this.x = x;}
}

void main(string[] args)
{
   auto c = new C(args.length);

   clobberStack(); // makes sure there are no references to c in the stack
   GC.collect();

   writeln(c.x);
}
```

Here, c.x is not readable at compile time. However, the compiler knows that `c.x` must be the same as `args.length` (which hasn't changed), so it logically can change the last line effectively to:

```d
writeln(args.length);
```

Thereby eliminating any need to store `c`, and possibly collecting it via the earlier collection cycle. Would this be a savings in terms of performance? I don't know. But I would consider it a valid optimization that a compiler *might* take.

LDC already does some optimizations that are magic like this. It is super surprising. How many times do people write benchmarking code that the LDC optimizer completely eliminates because it's not doing anything useful?

schveiguy:
I'm just trying to convey how difficult it is to define what "ensure the pointer is used later" really means in terms of how a user can reliably accomplish this, and what the pitfalls might be.

moon-chilled:
> I'm just trying to convey how difficult it is to define what "ensure the pointer is used later" really means

Yes.  In general, the compiler is permitted to remove loads and stores to/from memory if it can prove that the program would behave _as if_ it had performed them.  But reachability is ~an implementation detail, so it doesn't affect legality of such removals.

The only exceptions AFAIK are volatile and (where supported) asm constraints.  _Everything_ else can be elided and does not count as a 'use' (atomic is more complicated, but still can still be elided).  IMO the latter is pretty much ideal; and [using](https://github.com/schveiguy/keepalive/blob/master/source/keepalive.d#L37) it with GDC generates optimal code.

But using addRoot is, as you mention on the ng, not exactly cheap, and it violates the principle of 'you only pay for GC when you allocate'.  Whereas `scope(exit)volatileRead(...)` has minimal cost and ought to be supported in all cases, so why not recommend that?

PR #3103, opened 2021-09-27
Mention type indifference for mixin name conflicts

christoph-ehm:
As I have tested in Online D Editor, neither type nor function signature of symbols is considered when symbol names are treated as conflicts, unfortunately.
Overloaded functions with different signatures are conflicting as well as variabels with function names and probably other names like template instantiations and type definitions too.


adamdruppe:
i think this is ok to clarify but note that the old text saying "declaration with name" is the same same thing im pretty sure

PR #3106, opened 2021-10-05
Updated D/C++ template comparisons up to C++20.

Victarus:
The comparisons were pre-C++11, and since C++ has developed greatly since then this was due for an update.

To avoid having a crowded table listing every standard or only listing the latest-and-greatest, I opted to just have a single C++ column that calls out differences when present.

I also made the indentation consistent since I was having trouble figuring out where sections began/ended.

### DIFF
--- a/articles/template-comparison.dd
+++ b/articles/template-comparison.dd
@@ -2,8 +2,8 @@ Ddoc
 
 $(COMMUNITY Template Comparison,
 
-$(P C++ pioneered templates and template metaprogramming, and continues
-to improve on it with C++0x.
+$(P C++ pioneered templates and template metaprogramming and has continued
+to improve on it.
 The D programming language is the first to comprehensively reengineer
 templates based on the C++ experience.
 )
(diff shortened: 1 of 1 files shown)


articles/template-comparison.dd
```diff
@@ -175,58 +189,89 @@ template&lt;&gt; class factorial&lt;1&gt;
...
+    SlowFoo foo = slow_foo(i);
+  use_foo(foo); // $(ERROR foo is undeclared)
+}
+
+template&lt;class T, int maxLength&gt; class HashTable {
+    // $(ERROR cannot use 'if' outside of a function)
+    if constexpr (maxLength < 0xFFFE)
+        alias CellIdx = ushort;
+    else
+        alias CellIdx = uint;
```
thewilsonator, line 254: these should be a `using`, no?

thewilsonator (commented):
Otherwise looks OK, will wait for others to give it a look over.

articles/template-comparison.dd
Victarus: Yes, fixed.

thewilsonator (approved):

bostikforever:
Sorry to place this as a comment here, but:
1. I have not been able to create an account on the dlang issue tracking system with my email account (gmail is "blocked because of spam".
2. This PR is relevant enough, and github issues are not enabled for this project.
3. As I indicated in the last paragraph, it's not clear how the issue should be resolved (hence why I have written this vs opening a PR).

I am happy to be redirected to somewhere else where I can post this, as long as this does not require looking for a new email provider.

I have had this written in April this year but did not figure out how to get it out. I have updated it a bit to post now.

@maxhaton @thewilsonator

> ### Quite unfair characterisation of (modern) C++
>
> I am slowly getting introduced to D coming from  a C++ background and I have a lot of admiration for the design of the D language.  I however think that comparison with other languages should be fair in the sense that they should be written how a programmer in that language today would write the snippets being compared. This would probably complicate the maintenance of these kind of comparisons as languages change all the time and such comparison snippets would need to be updated. Still, I think other languages should be painted in the best light for the purpose of comparison.
>
>  In this [section](https://dlang.org/articles/cpptod.html#metatemplates ) we are presented with code from [here]( http://www.eptacom.net/pubblicazioni/pub_eng/paramint.html) as a solution to the stated problem of creating an integral type with a minimum compile-time-specified bit length. This article is pre standard C++ (98). Moreover, it is not clear if this ever compiled (I could not get it to compile using the —std=c++98 flag on available compilers) .
>The section also makes a statement about C++ which has not been true in over a decade: “_A preprocessor macro is also needed to make up for the lack of template `typedefs`._” 
>
>Here’s a more accurate (as of C++11) version of the above code (that actually compiles): https://godbolt.org/z/acKWxjE8e 
>
> However even this version is very dated. As of C++17, I’ll probably write something like this using compile time expressions:  https://godbolt.org/z/aGWqT8cfo
> We can see that this C++17 version comes very close to the D version both in style and clarity. I would still accept that the D version is more direct, “say what you mean”, than the C++ version which relies on a bit of indirection via `decltype` to get where we want to be. 
> It's possible that C++20 improves this particular snippet further, however my knowledge of what C++20 is limited.
>
>(For completeness, here’s a modified D version from the same section: https://godbolt.org/z/511hrK9eG)  
>
>**I suggest that the article be updated to use the C++17 snippet, or if the use of modern C++ would seriously weaken the point being made then a removal of that section would be in order.**

PR #3107, opened 2021-10-07
ImportC: disallow implicit function declaration

WalterBright:
As use of them became invalid 22 years ago.


ibuclaw:
> As use of them became invalid 22 years ago.

This came up when compiling zlib though, which made use of an undeclared `fseek`.

PR #3118, opened 2021-11-06
Update broken examples

Imperatorn:
body is deprecated and text.txt is not readable on run.dlang.io which is suboptimal

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -529,7 +529,7 @@ void main()
     // Code is executed in any case upon scope exit.
     scope(exit) { writeln("Exiting main."); }
     // File is closed deterministically at scope's end.
-    foreach (line; File("text.txt").byLine())
+    foreach (line; File("/etc/issue").byLine())
     {
         writeln(line);
     }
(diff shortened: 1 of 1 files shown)


index.dd
```diff
@@ -529,7 +529,7 @@ void main()
     // Code is executed in any case upon scope exit.
     scope(exit) { writeln("Exiting main."); }
     // File is closed deterministically at scope's end.
-    foreach (line; File("text.txt").byLine())
+    foreach (line; File("/etc/issue").byLine())
```
PetarKirov, line 532: `/etc/...` is not very portable (at least not on Windows), so how about:
```suggestion
    foreach (line; File(__FILE_FULL_PATH__).byLine())
```

PetarKirov (dismissed):
Looks good otherwise:

index.dd
Imperatorn: Much better, I'll change it

thewilsonator (approved):

PR #3119, opened 2021-11-07
spec: interfaceToC: remove duplicate ulong type on the table

ljmf00:

---

By reading https://dlang.org/spec/interfaceToC.html, I found out that `ulong` is duplicated and the first one seems wrong/incomplete.

**EDITED:** I also removed extra newlines.


ljmf00:
I think [this](https://dlang.org/articles/ctod.html#types) article is also wrong. Can someone confirm?

```
long               =>        int
unsigned long      =>        uint
```
This is not entirely right since `long` and `unsigned long` is dependent on the byte size.

Geod24:
> This is not entirely right since long and unsigned long is dependent on the byte size.

It should probably point to the right Druntime definition, too.

ljmf00:
> > This is not entirely right since long and unsigned long is dependent on the byte size.
> 
> It should probably point to the right Druntime definition, too.

Changed

articles/ctod.dd
```diff
@@ -88,8 +88,8 @@ unsigned short     =&gt;        ushort
 wchar_t            =&gt;        wchar
```
PetarKirov, line 88: @ljmf00 Can you also fix this:

```suggestion
wchar_t            =&gt;        core.stdc.stddef.wchar_t
```

(For reference: https://github.com/dlang/druntime/blob/v2.098.0/src/core/stdc/stddef.d#L27-L36)

ljmf00: I didn't notice it, thanks. Fixed

PetarKirov (approved):
Thanks!

PR #3122, opened 2021-11-16
Add  __traits(initSymbol) to the spec

MoonlightSentinel:
The traits was upstreamed from LDC in dlang/dmd#13298

### DIFF
--- a/spec/traits.dd
+++ b/spec/traits.dd
@@ -77,6 +77,7 @@ $(GNAME TraitsKeyword):
     $(RELATIVE_LINK2 isSame, $(D isSame))
     $(RELATIVE_LINK2 compiles, $(D compiles))
     $(RELATIVE_LINK2 toType, $(D toType))
+    $(RELATIVE_LINK2 initSymbol, $(D initSymbol))
 
 $(GNAME TraitsArguments):
     $(GLINK TraitsArgument)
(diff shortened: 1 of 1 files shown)


spec/traits.dd
```diff
@@ -1713,6 +1714,39 @@ $(H2 $(GNAME toType))
 
         $(RATIONALE Provides the inverse operation of the $(DDSUBLINK property, mangleof, `.mangleof`) property.)
 
+$(H2 $(GNAME initSymbol))
+
+        $(P Takes a single argument, which must evaluate to a `class` or `struct` type.
+            Returns a `const(void)[]` that points to either the initializer symbol of the supplied
+            type or `null` if it is $(RELATIVE_LINK2 isZeroInit, zero-initialized). The referenced symbol
```
kinke, line 1721: `__traits(isZeroInit, <class>)` is always true (=> class *ref*). All classes have init symbols (vptr is definitely not null).

MoonlightSentinel: Right. Removed the link

spec/traits.dd
```diff
@@ -1713,6 +1714,39 @@ $(H2 $(GNAME toType))
 
         $(RATIONALE Provides the inverse operation of the $(DDSUBLINK property, mangleof, `.mangleof`) property.)
 
+$(H2 $(GNAME initSymbol))
+
+        $(P Takes a single argument, which must evaluate to a `class` or `struct` type.
+            Returns a `const(void)[]` that points to either the initializer symbol of the supplied
+            type or `null` if it is zero-initialized. The referenced symbol holds the initial state
+            of any instance of the supplied type.
```
ljmf00, line 1722: Can we refer to something about the length of the returned slice? From the example below, I understood that `__traits(initSymbol, C)` returns the correct length of the initializer memory even if zero-initialized and that matches the implementation, although this wording can be interpreted wrongly, IMO.

MoonlightSentinel: Added a more explicit description for the `ptr` and `length` properties of  the returned slice.

ljmf00 (approved):
I don't know much about DDoc structure-wise, but it looks good to me

thewilsonator (approved):

PR #3123, opened 2021-11-17
Fix Issue 22518 - return without scope/ref not specified

dkorpel:


RazvanN7 (approved):

RazvanN7:
Darn it! I thought the dmd PR is merged. Revert: https://github.com/dlang/dlang.org/pull/3137

dkorpel:
It's okay, the new plan is to deprecate `return` without `scope` or `ref`, so this will be revived at some point anyway.

PR #3124, opened 2021-11-20
spec: abi: add function-local parent symbols to the mangling grammar

ljmf00:
Related to the issue #20565.

---
This is needed in order for the LLVM demangler be in sync with the compiler's implementation. See https://reviews.llvm.org/D111415 .

CC @Geod24

### DIFF
--- a/spec/abi.dd
+++ b/spec/abi.dd
@@ -379,6 +379,7 @@ $(GNAME Namechars):
 $(GRAMMAR
 $(GNAME LName):
     $(GLINK Number) $(GLINK Name)
+    $(GLINK Number) $(B __S) $(GLINK Number)    $(GREEN // function-local parent symbols)
 
 $(GNAME Number):
     $(GLINK Digit)


spec/abi.dd
```diff
@@ -379,6 +379,7 @@ $(GNAME Namechars):
 $(GRAMMAR
 $(GNAME LName):
     $(GLINK Number) $(GLINK Name)
+    $(GLINK Number) $(B __S) $(GLINK Number)    $(GREEN // fake parent symbols)
```
WebFreak001, line 382: are comments like this used anywhere else? Would think a separate `$(GNAME FakeParentName):` section would be more consistent if there are existing tools parsing this spec file.

ljmf00: > are comments like this used anywhere else? Would think a separate $(GNAME FakeParentName): section would be more consistent if there are existing tools parsing this spec file.

Yes, there is, here is an example.

image

I believe that a separate node is overkill, at least for that purpose for one branch. I think it is easy to skip those as they are marked specifically different with a green tag, although another rationale could come up

PetarKirov (approved):

PR #3125, opened 2021-11-22
spec: class: add __ctor() and __dtor() aliases

ljmf00:


ljmf00:
Should we also refer to this on `struct`s ?

spec/class.dd
```diff
@@ -371,6 +371,9 @@ $(GNAME Constructor):
         }
         ------
 
+        $(P The compiler may generate a $(D __ctor) alias that matches with
```
ljmf00, line 374: I'm not sure if the right wording is `may`, `should` or omit it. Should we enforce this to be implemented? I think so

FFY00: Is the compiler required to generate these aliases to comply to the spec? Or are they just an optional thing?

ljmf00: > Is the compiler required to generate these aliases to comply to the spec? Or are they just an optional thing?

Technically yes, since e.g. `__ctor` is mentioned in several specification examples but this can be considered an implementation detail. You, as a user, probably don't require this, unless doing e.g. custom class allocators, custom traits or other advanced features. This is mostly used internally by the D runtime hooks and standard library traits, although I think it is useful to make it not implementation-defined to ease the user for the reasons I stated above.

spec/class.dd
```diff
@@ -729,6 +732,11 @@ $(GNAME Destructor):
         by the gc.
         )
 
+        $(P The compiler may generate $(D __dtor) and $(D __xdtor) alias that
+        matches the class destructor definition and inclusive destructors,
```
maxhaton, line 736: is the term "inclusive destructor" defined elsewhere in the spec? If not I think it should have a definition somewhere. 

I note that `__xpostblit()`'s description in the struct page of the spec mentions the creation of aliases. 
>void __xpostblit(). The field and aggregated postblits, although generated for a struct, are not actual struct members. In order to be able to call them, the compiler internally creates an alias, called __xpostblit which is a member of the struct and which points to the generated postblit that is the most inclusive.

However, I note that this wording seems to be descriptive rather than prescriptive, which isn't ideal for the language specification.

dkorpel:
There's no linked bugzilla issue, so what's the idea behind this PR? It's already specified that identifiers starting with a double underscore are reserved.

ljmf00:
> There's no linked bugzilla issue, so what's the idea behind this PR? It's already specified that identifiers starting with a double underscore are reserved.

Yes, but this has a special purpose, and as I said above, it is noted in some specification examples such as traits and structs. We should also follow the same struct section, which specifies the behaviour of `__postblit` and `__xpostblit`.

spec/class.dd
ljmf00: I added the `inclusive destructor` concept to the glossary. Please review it.

About the `__xpostblit()`, I think it is a bit out of scope of this PR, although I can write a separate one fixing that wording.

spec/class.dd
```diff
@@ -371,6 +371,9 @@ $(GNAME Constructor):
         }
         ------
 
+        $(P The compiler generate a $(D __ctor) alias that matches with
```
RazvanN7, line 374:
```suggestion
        $(P The compiler generates a $(D __ctor) alias that matches with
```

spec/class.dd
```diff
@@ -371,6 +371,9 @@ $(GNAME Constructor):
...
+        constructors definitions.)
```
RazvanN7, line 375:
```suggestion
        constructor definitions.)
```

RazvanN7: I don't really understand what you mean by this sentence. Could you please elaborate?

PetarKirov:
@ljmf00 can you please rebase your PR?

ljmf00:
> @ljmf00 can you please rebase your PR?

Yes, I'll add to my TODO list.

glossary.dd
```diff
@@ -258,6 +258,11 @@ void test()
         syntax) refers to the ability to call a function as if it were a method of
         its first argument. For example `funct(arg)` may be written as
         `arg.funct()`.)
+
+        $(DT $(LNAME2 inclusive_destructor, inclusive-destructor, Inclusive
```
ntrel, line 262:
```suggestion
        $(DT $(LNAME2 inclusive-destructor, Inclusive
```

ntrel (changes requested):

glossary.dd
```diff
@@ -258,6 +258,11 @@ void test()
...
+        Destructor))
+        $(DD An inclusive destructor is a destructor that includes destruction
+        of the object itself and its fields, if reaching its lifetime.)
```
pbackus, line 265:
```suggestion
        of the object itself and its fields.)
```
Destructors can be called early (for example, by `destroy`), and this definition applies regardless of when it is called.

RazvanN7:
ping @ljmf00

PR #3129, opened 2021-11-26
Fix portability of the book builder

MoonlightSentinel:
The addition of the book builder introduced the dependency on a `dmd` / `rdmd` found on the path. This broke multiple CI's:

1. DMD nightlies are built with ldc and hence have no `dmd` installed by default
2. Installer CI has no system-wide rdmd installation

---

```
Use STABLE_DMD building/running book/build.d

Relying on the `STABLE_DMD` release ensures that the book can be built
even if there is no `rdmd` / `dmd` found on the path.
```

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -383,10 +383,13 @@ docs-prerelease: dmd-prerelease druntime-prerelease phobos-prerelease apidocs-pr
 
 docs : docs-latest docs-prerelease
 
-html : $(ALL_FILES) makebook
+html : html_files makebook
 
-makebook:
-	(cd book && rdmd build.d 6)
+html_files : $(ALL_FILES)
+
+makebook : book/build.d | $(STABLE_DMD)
+	cd book && PATH="$(STABLE_DMD_BIN_ROOT):$$PATH" rdmd build.d 6
+
 html-verbatim: $(addprefix $W/, $(addsuffix .verbatim,$(PAGES_ROOT)))
 
 verbatim : html-verbatim phobos-prerelease-verbatim


posix.mak
```diff
@@ -383,10 +383,16 @@ docs-prerelease: dmd-prerelease druntime-prerelease phobos-prerelease apidocs-pr
 
 docs : docs-latest docs-prerelease
 
-html : $(ALL_FILES) makebook
+html : html_files makebook
+
+html_files : $(ALL_FILES)
```
MoonlightSentinel, line 388: Moving this into a different target ins't required but improves debugging the scripts.

maxhaton (approved):

posix.mak
```diff
@@ -383,10 +383,13 @@ docs-prerelease: dmd-prerelease druntime-prerelease phobos-prerelease apidocs-pr
...
+
+makebook : book/build.d | $(STABLE_DMD)
+	cd book && PATH="$(STABLE_DMD_BIN_ROOT):$$PATH" rdmd build.d 6
```
ljmf00, line 391: I'm not too familiar with GNU Makefile, but why subshell is no longer needed? does every command run in a subshell by default?

ljmf00: With a bit of research, I found out that this is indeed redundant

ljmf00 (approved):

lionello:
This doesn't appear to have solved the CI build issue: https://github.com/dlang/installer/pull/489#issuecomment-980709464

PR #3131, opened 2021-11-27
spec: pragma: fix ctr_destructor documentation

ljmf00:

### DIFF
--- a/spec/pragma.dd
+++ b/spec/pragma.dd
@@ -152,15 +152,15 @@ $(H3 $(LNAME2 crtctor, $(D pragma crt_constructor)))
 
 $(H3 $(LNAME2 crtdtor, $(D pragma crt_destructor)))
 
-    $(P `pragma(crt_destructor)` works the same as `pragma(crt_destructor)` except:)
+    $(P `pragma(crt_destructor)` works the same as `pragma(crt_constructor)` except:)
 
-    $(P Annotates a function so it is run after the D runtime library is terminated
-        and before the C runtime library is terminated.
-        Calling C's `exit()` function also causes the annotated functions to run.
+    $(OL
+        $(LI Annotates a function so it is run after the D runtime library is terminated
+             and before the C runtime library is terminated.
+             Calling C's `exit()` function also causes the annotated functions to run.)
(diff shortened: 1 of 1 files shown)


spec/pragma.dd
```diff
@@ -152,7 +152,7 @@ $(H3 $(LNAME2 crtctor, $(D pragma crt_constructor)))
 
 $(H3 $(LNAME2 crtdtor, $(D pragma crt_destructor)))
 
-    $(P `pragma(crt_destructor)` works the same as `pragma(crt_destructor)` except:)
+    $(P `pragma(crt_destructor)` works the same as `pragma(crt_constructor)` except:)
```
maxhaton, line 155: Can you merge this section with the following paragraph, I think it's a big ugly. Might as well improve it while we're here.

ljmf00: What do you mean, join `pragma crt_constructor` and `pragma crt_destructor` altogether?

maxhaton:
"y works the same as x, except blah blah" is crap technical writing. It
should just say "y has such and such a behaviour".

spec/pragma.dd
ibuclaw: See how the numbering is done in 9.1.1.2 of the spec.
https://dlang.org/spec/pragma.html#crtctor

ljmf00: Yes, I thought about nesting it to make it clearer, although I'm not getting the point of @maxhaton , can you elaborate a bit more? Do you mean I have to duplicate the points where X and Y have in common?

ibuclaw: > Yes, I thought about nesting it to make it clearer, although I'm not getting the point of @maxhaton , can you elaborate a bit more? Do you mean I have to duplicate the points where X and Y have in common?

Join the two paragraphs together instead of separate numbered points is my interpretation.

RazvanN7:
ping @ljmf00

spec/pragma.dd
ljmf00: Joined using `OL` and `LI` macros.

ljmf00:
> ping @ljmf00

Thanks!

dkorpel (approved):

PR #3133, opened 2021-12-04
Old d derivations

dukc:
Links to my repo that might be useful for a Nix user who wants to use a different version of DMD compared to what happens to be on NixPkgs repository.


dukc:
Some extra commits here, closing for now

dukc:
New attempt here: #3134

PR #3135, opened 2021-12-05
Fix Issue 18960 - Function parameter requires name with default value 

sorin-gabriel:
Closing the issue by adding clarifications in the spec. The behaviour mentioned in the issue actually seems reasonable (so it's not a bug), but the docs don't mention this anywhere.

### DIFF
--- a/spec/declaration.dd
+++ b/spec/declaration.dd
@@ -431,6 +431,40 @@ a = 2;         // sets `S.i` to `2`
 b = 4;         // sets `S.j` to `4`
 -----------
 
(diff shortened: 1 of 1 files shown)


rikkimax:
After reviewing my ticket, I cannot place what exactly I was referencing in the spec. I should have linked it.

Therefore I'm going to close it as invalid because I'm pretty sure I miss interpreted something related to function parameters.

However your example is good, just drop the reference to the ticket.

RazvanN7 (approved):

PR #3138, opened 2021-12-09
Update arrays.dd

kurt-krueckeberg:
Referring to a "slice operator" is confusing since there isn't really a special, separate "slice operator". You change values in the underlying array by assignment to the entire slice:

int[] a;
// snip...
int[] b;
b = a[1..4];
b = 0; // Sets a[1] = 0, a[2] = 0 and a[3] = 0. 

and you can also use the array index operator when assigning to a slice

int[] a;
// snip...
int[] b;
b = a[1..4];
b[0..2] = 0; // Sets a[1] = 0, a[2] = 0 

The operator used in the first example is the array- and slice-assignment operator. The second example also uses the array- and slice-index. Neither example uses a separate, special "slice operator".

### DIFF
--- a/spec/arrays.dd
+++ b/spec/arrays.dd
@@ -229,7 +229,7 @@ writeln(b[7]);      // 10
 
 $(H2 $(LNAME2 array-copying, Array Copying))
 
-        $(P When the slice operator appears as the left-hand side of an
+        $(P When the slice appears as the left-hand side of an
         assignment expression, it means that the contents of the array are the
         target of the assignment rather than a reference to the array.
         Array copying happens when the left-hand side is a slice, and the
@@ -284,7 +284,7 @@ assert(s == [2, 3, 3, 4]);
 
 $(H2 $(LNAME2 array-setting, Array Setting))
 
-        $(P If a slice operator appears as the left-hand side of an assignment
+        $(P If a slice appears as the left-hand side of an assignment
         expression, and the type of the right-hand side is the same as the
         element type of the left-hand side, then the array contents of the
         left-hand side are set to the right-hand side.


dkorpel (changes requested):
I don't think this is right

spec/arrays.dd
```diff
@@ -229,7 +229,7 @@ writeln(b[7]);      // 10
 
 $(H2 $(LNAME2 array-copying, Array Copying))
 
-        $(P When the slice operator appears as the left-hand side of an
+        $(P When the slice appears as the left-hand side of an
```
dkorpel, line 232: The paragraph explains the difference between:
```D
int[] a, b;
b[] = a; // slice operator, copy elements of a to b
b = a; // slice without operator, make b refer to the same data as a
```
Removing 'operator' is not correct here, the `[]` is important.

dkorpel (changes requested):

spec/arrays.dd
kurt-krueckeberg: Yes, my proposed wording change may not be the best.  But is there a **slice operator** in D?  Isn't `[]` the array subscript operator in D just like it is in C and C++? So  the actual operators in the expression

```d
b[] = a; // slice operator, copy elements of a to b
```
are the  `=` and `[]` operators? Where is the **slice operator** in the expression?  I don't know precisely how to improve the comment, but maybe simply saying

```d
b[] = a; // copies elements of a to b
```
is clearer than mentioning a non-extant **slice operator**. I'm new to D and coming from C++ but the _only_ **slice operator** that I am aware of is the tilde (~) operator described in the article [D Slices](https://dlang.org/articles/d-array-article.html) used for concatenation and appending.

kurt-krueckeberg: The section [Operator Overloading](https://dlang.org/spec/operatoroverloading.html#slice_unary_operators) of the  [Operator Overloading](https://dlang.org/spec/operatoroverloading.html) page of the Language Reference does explain that the  unary `[]` operator is overloaded for slices.

Obviously `[]` is the index operator for arrays and slices, but this simple explanation seems to be missing from the documentation. The only explanation of the unary `[]` operator of the D languange seems to be [Index Expressions](https://dlang.org/spec/expression.html#IndexExpression), and it is a very technical explanation.

dkorpel: 'Slice operator' may be poorly defined in the spec.

> Where is the slice operator in the expression? 

The `=` is the operator, and the `[]` on the left hand side makes `=` a slice operator. You can change the wording or clear it up with a better explanation, but removing 'operator' from that sentence without anything else makes it worse.

kurt-krueckeberg: Yes, I agree that simply removing 'operator' is not an improvement.

dkorpel: Are you going to change it?

kurt-krueckeberg: No.

PR #3139, opened 2021-12-10
fix Issue 19136 - is expressions don't work as documented

sorin-gabriel:
As per @adamdruppe 's [explanation](https://github.com/dlang/dmd/pull/13387#issuecomment-986288394), I edited the description for `is ( Type Identifier : TypeSpecialization )` in the spec to match `is ( Type : TypeSpecialization )`.


adamdruppe (approved):
yeah i think this is the right way to describe it and matches what it actually deos

RazvanN7 (approved):

PR #3149, opened 2021-12-24
[spec] Document `noreturn`

ntrel:
See also: https://github.com/dlang/DIPs/blob/master/DIPs/accepted/DIP1034.md

### DIFF
--- a/spec/type.dd
+++ b/spec/type.dd
@@ -566,8 +566,7 @@ $(H3 $(LNAME2 noreturn, $(D noreturn)))
     optimize such code accordingly.)
 
     $(P A function that never returns has the return type `noreturn`. This can
-    occur due to an infinite loop or always throwing an exception.
-    This is particularly useful when inferring types.)
+    occur due to an infinite loop or always throwing an exception.)
 
 $(SPEC_RUNNABLE_EXAMPLE_COMPILE
 ---
--- a/spec/type.dd
+++ b/spec/type.dd
@@ -558,6 +558,38 @@ $(H3 $(LNAME2 string, $(D string)))
 
     $(P A $(DDSUBLINK spec/arrays, strings, $(I string) is a special case of an array.))
 
+$(H3 $(LNAME2 noreturn, $(D noreturn)))
+
+    $(P `noreturn` is the $(LINK2 https://en.wikipedia.org/wiki/Bottom_type, bottom type)
+    which can implicitly convert to any type, including `void`.
+    A value of type `noreturn` will never be produced and the compiler can
+    optimize such code accordingly.)
+
+    $(P A function that never returns has the return type `noreturn`. This can
+    occur due to an infinite loop or always throwing an exception.
+    This is particularly useful when inferring types.)
+
+$(SPEC_RUNNABLE_EXAMPLE_COMPILE
+---
+noreturn abort(const(char)[] message);
+
+int example(int i)
+{
+    if (i < 0)
+    {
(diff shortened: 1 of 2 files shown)


maxhaton:
@dkorpel

ntrel:
As documented in the DIP, I tried a throwing lambda, function pointers with other return types accepting a noreturn function, and template inference for `[]` to `T[]` and `null` to `T*` but none of these things work yet with stable dmd. Shall I file enhancements for these?

BTW it would be good to merge this pull to stable as well once it is approved.

ntrel:
> it would be good to merge this pull to stable 

I have split out the changes to #3150 for `main` return type because that isn't stable yet.

MoonlightSentinel:
> As documented in the DIP, I tried a throwing lambda, function pointers with other return types accepting a noreturn function, and template inference for `[]` to `T[]` and `null` to `T*` but none of these things work yet with stable dmd. Shall I file enhancements for these?

Sure. Note that there's already a [draft PR for `throw` expressions](dlang/dmd#13162) .

> BTW it would be good to merge this pull to stable as well once it is approved.

This will happen anyways with the regular master - stable merge and there is no rush given the state of `noreturn` (very WIP).

spec/type.dd
```diff
@@ -558,6 +558,38 @@ $(H3 $(LNAME2 string, $(D string)))
...
+{
+    if (i < 0)
+    {
+        // abort does not return, so it doesn't need to produce an int
+        int val = abort("less than zero");
+    }
+    // ternary expression's common type is still int
+    return i != 0 ? 1024 / i : abort("calculation went awry.");
+}
+---
```
MoonlightSentinel, line 586: This / another example should probably include usages of `noreturn` outside of function return types, i.e. variables.

ntrel: @MoonlightSentinel Do you have an idea for an example? The DIP only seems to mention `noreturn[] storage;` for `MapResult`, and stable dmd doesn't seem to have the enhanced compile-time understanding of this (which might make a good example).

MoonlightSentinel: Unlike void, `noreturn` can be used everywhere a "normal" type may appear, i.e. variables, members, parameters, ... . 

Maybe add an example where the "return value" of a `noreturn` variable is stored in a variable?

ntrel:
@MoonlightSentinel 
> This will happen anyways with the regular master - stable merge

Ok.

> there is no rush given the state of noreturn (very WIP).

It would be good to merge this and the dependent `main` PR before the next release. It seems what I have documented is quite stable. I haven't added a variable example because I don't know if this is stable enough for a meaningful example.

spec/type.dd
```diff
@@ -558,6 +558,38 @@ $(H3 $(LNAME2 string, $(D string)))
...
+$(H3 $(LNAME2 noreturn, $(D noreturn)))
+
+    $(P `noreturn` is the $(LINK2 https://en.wikipedia.org/wiki/Bottom_type, bottom type)
+    which can implicitly convert to any type, including `void`.
+    A value of type `noreturn` will never be produced and the compiler can
+    optimize such code accordingly.)
+
+    $(P A function that never returns has the return type `noreturn`. This can
+    occur due to an infinite loop or always throwing an exception.
+    This is particularly useful when inferring types.)
```
dkorpel, line 570: I don't think sentence adds value, it's unspecific.

ntrel: Removed now.

dkorpel (approved):

PR #3150, opened 2021-12-26
[spec] `main` can be declared with `auto` or `noreturn`

ntrel:
Add grammar for `main` now it can be declared with `auto` and
`noreturn`:
https://issues.dlang.org/show_bug.cgi?id=22419
https://issues.dlang.org/show_bug.cgi?id=22113

(This is for the 2.099.0 release).

Also document `string[]` parameter.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -2962,8 +2962,8 @@ $(H2 $(LNAME2 main, $(D main()) Function))
 
         $(GRAMMAR
         $(GNAME MainFunction):
-            $(GLINK MainReturnDecl) main $(D $(LPAREN)$(RPAREN)) $(GLINK2 statement, BlockStatement)
-            $(GLINK MainReturnDecl) main $(D $(LPAREN)string[] args$(RPAREN)) $(GLINK2 statement, BlockStatement)
+            $(GLINK MainReturnDecl) $(D main$(LPAREN)$(RPAREN)) $(GLINK2 statement, BlockStatement)
+            $(GLINK MainReturnDecl) $(D main$(LPAREN)string[]) $(GLINK_LEX Identifier)$(D $(RPAREN)) $(GLINK2 statement, BlockStatement)
 
         $(GNAME MainReturnDecl):
             $(D void)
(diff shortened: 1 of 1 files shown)


spec/function.dd
```diff
@@ -2957,14 +2957,19 @@ $(H2 $(LNAME2 main, $(D main()) Function))
         It gets called after all the module initializers are run, and
         after any unittests are run.
         After it returns, all the module destructors are run.
-        $(D main()) must be declared using one of the following forms:
+        $(D main()) must be declared as follows:
```
MoonlightSentinel, line 2960: Only mentioning the function signature is somehwat misleading w.r.t. `auto` because further restrictions apply to the inferred return type (must be `void`, `int` or `noreturn`). It would be good to be explicit and mention it in an additional paragraph below the grammar.

MoonlightSentinel:
> This pull depends on #3149 for the `noreturn` link.

Introduce the anchor in this PR?

spec/function.dd
ntrel: I've documented that now.

MoonlightSentinel: Thanks

spec/function.dd
```diff
@@ -2954,19 +2954,36 @@ $(H3 $(LNAME2 anonymous, Anonymous Functions and Anonymous Delegates))
...
-        $(LI `int main(string[] args) { ... }`)
+        $(LI If `main` returns `void`, the OS will receive a zero value on success.)
+        $(LI If `main` returns `void` or `noreturn`, the OS will receive a non-zero
+        value on abnormal termination, such as an uncaught exception.)
+        $(LI If `main` is declared as `auto`, the inferred return type must be
+        one of `void`, `int` and `noreturn`.)
         )
 
+        $(P If the $(D string[]) parameter is declared, the parameter will hold any
+        arguments passed to the program by the OS. The first argument is typically
```
MoonlightSentinel, line 2984: The arguments for a `D main` are preprocessed and Druntime removes `--DRT...` options (unless they are preceeded by `--`) IIRC.

ntrel: Thanks, updated.

thewilsonator (approved):

PR #3156, opened 2022-01-01
[expression.dd] Improve formatting for comparison operators

ntrel:
* Fix closing `<p>` tags before `<div>`, lists or code examples to prevent empty lines in HTML.
* Add *Aggregate Equality* subheading under *Equality Expressions* & move paragraph about arrays above it.
* Add *Array Comparisons* subheading under *Relational Expressions*.
* Improve example showing runtime null object equality error.
* Move paragraph about `o1.opCmp(o2)` under existing *Class Comparisons* subheading.

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -507,18 +507,18 @@ $(GNAME EqualExpression):
         Otherwise, the bit patterns of the common type are compared for equality.
     )
 
-    $(P For complex numbers, equality is defined as equivalent to:)
-
-        ---
-        x.re == y.re && x.im == y.im
-        ---
-
     $(P For static and dynamic arrays, equality is defined as the
         lengths of the arrays
         matching, and all the elements are equal.
     )
 
-    $(H3 $(LNAME2 aggregate_equality, Aggregate Equality))
+    $(DDOC_DEPRECATED For complex numbers, equality is defined as equivalent to:)
+
+        ---
+        x.re == y.re && x.im == y.im
+        ---
+
+    $(H3 $(LNAME2 class_struct_equality, Class & Struct Equality))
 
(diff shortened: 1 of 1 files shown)


spec/expression.dd
```diff
@@ -507,18 +507,25 @@ $(GNAME EqualExpression):
         Otherwise, the bit patterns of the common type are compared for equality.
     )
 
-    $(P For complex numbers, equality is defined as equivalent to:
+    $(P For complex numbers, equality is defined as equivalent to:)
```
maxhaton, line 511: We have deprecated built-in complex numbers, perhaps this sentence should be reworded or removed?

ntrel: I've indicated this now.

spec/expression.dd
```diff
@@ -677,13 +670,24 @@ $(H3 $(LEGACY_LNAME2 class_comparisons, class-comparisons, Class comparisons))
...
+        a $(CODE null) class reference is invalid, as $(CODE null) has no contents.)
 
+        $(SPEC_RUNNABLE_EXAMPLE_FAIL
         ---
-        class C;
+        class C {}
         C c;
-        if (c < null)  // error
-            ...
+        if (c < null) {}  // compile-time error
```
dkorpel, line 679: Shouldn't this be put in a `void main() {}`? A top-level if-statement doesn't parse.

dkorpel (approved):

spec/expression.dd
ntrel: Examples that don't contain `main` are wrapped in a `main` function (and have `stdio` imported). So C here is a nested class.

dkorpel: Good to know

PR #3157, opened 2022-01-01
[spec/expression.dd] Improve InExpression

ntrel:
Make *InExpression* discuss a container, not just an associative array. Use key/value rather than *element*.
Mention result of `in` is a pointer *for AAs* (specifically).
Move sentence about precedence below result type description.
~~Add best practice note about overloading `in` only when there's sub-linear time complexity.~~
Add note about overloading `opBinaryRight`.

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -691,7 +691,7 @@ $(H2 $(LNAME2 in_expressions, In Expressions))
 $(GRAMMAR
 $(GNAME InExpression):
     $(GLINK ShiftExpression) $(D in) $(GLINK ShiftExpression)
-    $(GLINK ShiftExpression) $(D !in) $(GLINK ShiftExpression)
+    $(GLINK ShiftExpression) $(D ! in) $(GLINK ShiftExpression)
 )
 
 $(P A container such as an associative array
@@ -700,14 +700,17 @@ $(P A container such as an associative array
         -------------
         int foo[string];
         ...
-        if ("hello" in foo) {
+        if ("hello" in foo)
+        {
             // the string was found
         }
         -------------
 
-    $(P The return value of the $(I InExpression) is typically `bool` or a pointer.
+    $(P The result of an $(I InExpression) is a pointer for associative
+        arrays.
         The pointer is $(D null) if the container has no matching key.
-        If there is a match, the return value is a pointer to a value associated with the key.
+        If there is a match, the pointer points to a value associated
+        with the key.
     )
 
     $(P The $(D !in) expression is the logical negation of the $(D in)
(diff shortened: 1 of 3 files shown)


spec/expression.dd
```diff
@@ -691,30 +691,36 @@ $(H2 $(LNAME2 in_expressions, In Expressions))
...
     )
 
     $(P The $(D !in) expression is the logical negation of the $(D in)
         operation.
     )
 
+    $(P The $(D in) expression has the same precedence as the
+        relational expressions $(D <), $(D <)$(D =), etc.)
+
+    $(NOTE When $(DDSUBLINK spec/operatoroverloading, binary, overloading) `in`, normally only `opBinaryRight` would be defined.)
```
dkorpel, line 720: Explain why. "Because the operation is usually not defined by the element but by the container, which appears on the right hand side of `in` operator."

spec/expression.dd
```diff
@@ -691,30 +691,36 @@ $(H2 $(LNAME2 in_expressions, In Expressions))
...
+
+    $(BEST_PRACTICE Overloading `in` for a container should only be done when the operation is of sublinear time complexity.)
```
dkorpel, line 722: That comes up in the newsgroup sometimes but it's not official AFAIK. I would not add this in this PR but make a separate PR for this with rationale for this, and let it be reviewed by the leadership.

spec/expression.dd
```diff
@@ -805,13 +811,13 @@ $(GNAME CatExpression):
...
-        and the other is of that array's element type, that element
-        is converted to an array of length 1 of that element,
-        and then the concatenation is performed.
-    )
+    $(P A $(I CatExpression) concatenates a container's data with other data, producing
+        a new container of the same type as the result.
+        Typically the container is a dynamic array.)
+
+    $(P For a dynamic array, the other operand must either be another array or a
+        single value compatible with the element type of the array.
```
dkorpel, line 819:
```suggestion
        single value that implicitly converts to the element type of the array.
```

spec/expression.dd
```diff
@@ -805,13 +811,13 @@ $(GNAME CatExpression):
...
 
-    $(P A $(I CatExpression) concatenates arrays, producing
-        a dynamic array with the result. The arrays must be
-        arrays of the same element type. If one operand is an array
-        and the other is of that array's element type, that element
-        is converted to an array of length 1 of that element,
-        and then the concatenation is performed.
-    )
+    $(P A $(I CatExpression) concatenates a container's data with other data, producing
+        a new container of the same type as the result.
```
dkorpel, line 815: "the result has the same type as the result"?

spec/expression.dd
```diff
@@ -691,30 +691,36 @@ $(H2 $(LNAME2 in_expressions, In Expressions))
...
+        }
         -------------
 
-    $(P The $(D in) expression has the same precedence as the
-        relational expressions $(D <), $(D <)$(D =),
-        etc.
-        The return value of the $(I InExpression) is $(D null)
-        if the element is not in the array;
-        if it is in the array it is a pointer to the element.
+    $(P The return value of the $(I InExpression) is typically `bool` or a pointer.
```
dkorpel, line 708: I don't like "typically `bool`", I would keep this paragraph about `in` expressions on a built-in Associative Array.

dkorpel (changes requested):

ntrel:
I've split out the concatenation changes to #3232.

dkorpel (approved):

PR #3158, opened 2022-01-01
[expression.dd] Improve *CastExpression* docs

ntrel:
Add subheadings for *CastExpression*.
Make more examples runnable.
Add missing *CastQual* anchor.
Remove some unhelpful `GLINK`s when that link is already nearby.

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -1018,10 +1018,7 @@ $(H3 $(LNAME2 cast_expressions, Cast Expressions))
 $(GRAMMAR
 $(GNAME CastExpression):
     $(D cast $(LPAREN)) $(GLINK2 type, Type) $(D $(RPAREN)) $(GLINK UnaryExpression)
-    $(I CastQual)
-
-$(GNAME CastQual):
-    $(D cast $(LPAREN)) $(GLINK2 type, TypeCtors)$(OPT) $(D $(RPAREN)) $(GLINK UnaryExpression)
+    $(GLINK CastQual)
 )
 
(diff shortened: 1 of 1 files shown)


spec/expression.dd
```diff
@@ -1162,40 +1173,51 @@ $(GNAME CastExpression):
...
-    $(P Casting with no $(GLINK2 type, Type) or $(GLINK CastQual) removes
+    $(P Casting with no type or qualifiers removes
         any top level $(D const), $(D immutable), $(D shared) or $(D inout)
         type modifiers from the type
-        of the $(GLINK UnaryExpression).)
+        of the $(I UnaryExpression).)
 
         ---
         shared int x;
         assert(is(typeof(cast()x) == int));
```
dkorpel, line 1207: I think this could be made runnable with a `static assert`

MoonlightSentinel:
The various cast types should be nested in `Cast Expressions` instead of being independent children of `Unary Expressions`.

Also do you know why  `Pow Expressions` is listed as a separate section instead of being a child of `Unary Expressions`? Maybe there is an error in `Cast Expressions` that interrupts the current section.
(Doesn't need to be fixed in this PR as it already exists in `master`)

ntrel:
> The various cast types should be nested in Cast Expressions instead of being independent children of Unary Expressions.

Thanks, fixed.

> why Pow Expressions is listed as a separate section instead of being a child of Unary Expressions? 

Although the grammar has Pow expressions as part of Unary, they are actually binary so I think it's best to keep them as an independent heading. There's some tension between grammar and concepts in this page IMO, but it's OK.

PR #3160, opened 2022-01-05
[expression.dd] Improve IndexExpression & SliceExpression docs

ntrel:
* Make IndexExpression docs more generic.
* Mention pointer indexing & slicing.

* `$` is also declared for *ValueSeq* slices and can be for user types - make this paragraph independent of array slicing.
* Don't only talk about slicing arrays throughout, but slicing elements of the *PostfixExpression* instead. This is more generic.
* Mention first and last element expressions for arrays and pointers.
* Add *Slice Conversion to Static Array* subheading, move *ValueSeq* paragraph above it & add a *ValueSeq* link.

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -1235,20 +1235,27 @@ $(GNAME IndexExpression):
     $(GLINK PostfixExpression) $(D [) $(GLINK ArgumentList) $(D ])
 )
 
-$(P $(I PostfixExpression) is evaluated. If $(I PostfixExpression) is an
-expression of type static array or dynamic array, the symbol $(DOLLAR) is set to
-be the number of elements in the array. If $(I PostfixExpression) is a $(I
-ValueSeq), the symbol $(DOLLAR) is set to be the number of elements
-in the sequence. A new declaration scope is created for the evaluation of the
-$(GLINK ArgumentList) and $(DOLLAR) appears in that scope only.)
+    $(P $(I PostfixExpression) is evaluated.
+        If $(I PostfixExpression) is an expression of static or
+        dynamic array type, the result of the indexing is an lvalue
+        of the *i*th element in the array.
+        If $(I PostfixExpression) is a pointer `p`, the result is `*(p + i)`.
+        In each case, `i` is an integer evaluated from $(I ArgumentList).
+    )
 
-    $(P If $(I PostfixExpression) is a $(I ValueSeq),
-        then the $(GLINK ArgumentList) must consist of only one argument,
+    $(P If $(I PostfixExpression) is a $(DDSUBLINK spec/template, variadic-templates, $(I ValueSeq))
+        then the $(I ArgumentList) must consist of only one argument,
         and that must be statically evaluatable to an integral constant.
         That integral constant $(I n) then selects the $(I n)th
         expression in the $(I ValueSeq), which is the result
         of the $(I IndexExpression).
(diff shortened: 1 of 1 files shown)


spec/expression.dd
```diff
@@ -1266,43 +1266,51 @@ $(GNAME Slice):
...
+        If $(I PostfixExpression) is an expression of static array or
+        dynamic array type, the result of the slice is a dynamic array
+        of the element type of the $(I PostfixExpression).
+    )
+
+    $(P If $(I PostfixExpression) is a $(DDSUBLINK spec/template, variadic-templates, $(I ValueSeq)), then
+        the result of the slice is a new $(I ValueSeq) formed
+        from the upper and lower bounds, which must statically evaluate
+        to integral constants.
+        It is an error if those bounds are out of range.
```
maxhaton, line 1278:
```suggestion
        These bounds must be in range.
```

maxhaton: Reads a little bit nicer + not a huge fan of language specs mentioning "error" and similar.

spec/expression.dd
```diff
@@ -1266,43 +1266,51 @@ $(GNAME Slice):
...
+    $(P If the $(D [ ]) form is used, the slice is of all the elements in $(I PostfixExpression).
     )
 
     $(P A $(I SliceExpression) is not a modifiable lvalue.)
 
+$(H3 $(LNAME2 slice_to_static_array, Slice Conversion to Static Array))
+
     $(P If the slice bounds can be known at compile time, the slice expression
-    is implicitly convertible to an lvalue of static array. For example:)
+    may be implicitly convertible to an lvalue of static array. For example:)
```
dkorpel, line 1303: Why `may be`, are there cases where it isn't implicitly convertible?

ntrel: @dkorpel The *PostfixExpression* might be a user-defined type where `opSlice` doesn't return a dynamic array.

ntrel:
This pull now covers `IndexExpression` as well as it's closely related. I also updated the first paragraph of the `SliceExpression` docs and added a paragraph about slicing a pointer.

spec/expression.dd
ntrel: 'must be' is not as clear that this is checked at compile time. (I didn't change the original sentence, just moved it).

spec/expression.dd
```diff
@@ -1235,22 +1235,29 @@ $(GNAME IndexExpression):
...
-$(GLINK ArgumentList) and $(DOLLAR) appears in that scope only.)
-
-    $(P If $(I PostfixExpression) is a $(I ValueSeq),
-        then the $(GLINK ArgumentList) must consist of only one argument,
+    $(P $(I PostfixExpression) is evaluated.
+        If $(I PostfixExpression) is an expression of static or
+        dynamic array type, the result of the indexing is an lvalue
+        of the *i*th element in the array.
+        If $(I PostfixExpression) is a pointer `p`, the result is `*(p + i)`.
+        In each case, `i` is an integer evaluated from $(I ArgumentList).
```
dkorpel, line 1243: You use `i` twice and only define what it refers to afterwards. I think that's a bit awkward, I'd rather see `i` defined first.

ntrel: I have now defined `i` in the same sentence as the first use, hopefully that reads better.

dkorpel (approved):

PR #3161, opened 2022-01-05
[spec] Improve array docs

ntrel:
Add *Array Literals* section.
Rename *Array Usage* to *Array Assignment* & improve example, link to *Slicing*. Comment out error lines for clarity.
Add example for *Indexing*.
Remove unnecessary variables in *Pointer Arithmetic* example.

[expression.dd]
Simplify literal -> static array example.

[hash-map.dd]
Make first example runnable.
Add AA *Literals* section.

### DIFF
--- a/spec/arrays.dd
+++ b/spec/arrays.dd
@@ -121,15 +121,24 @@ int[]* e;     // pointer to dynamic array of ints
 ---------
 )
 
-$(H2 $(LNAME2 usage, Array Usage))
+$(H2 $(LNAME2 literals, Array Literals))
 
-        $(P There are two broad kinds of operations to do on an array -
-        those affecting the handle to the array,
-        and those affecting the contents of the array.
-        )
+---
+auto a1 = [1,2,3];  // type is int[], with elements 1, 2 and 3
+auto a2 = [1u,2,3]; // type is uint[], with elements 1u, 2u, and 3u
+int[2] = [1, 2];
+---
+    $(P `[]` is an empty array literal.)
 
-        $(P The handle to an array is specified by naming the array, as
-        in p, s or a:
+    $(P See $(DDSUBLINK expression, array_literals, Array Literals).)
+
+$(LEGACY_LNAME2 usage)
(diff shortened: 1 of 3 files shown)


maxhaton (approved):

spec/arrays.dd
```diff
@@ -694,7 +724,7 @@ $(H2 $(LNAME2 bounds, Array Bounds Checking))
 
         $(P It is an error to index an array with an index that is less than
         0 or greater than or equal to the array length. If an index is
-        out of bounds, a RangeError exception is
+        out of bounds, a `RangeError` exception is
```
dkorpel, line 727: This is not something that needs to be fixed in this PR necessarily, but a `RangeError` is not an `Exception`, it's an `Error` / `Throwable`. Also, currently there's a subclass of `RangeError`, `ArrayIndexError`, that gets thrown.

spec/expression.dd
```diff
@@ -1580,16 +1580,17 @@ $(GNAME ArrayLiteral):
             bar(arr[1 .. 3]);
             assert(arr == [1, 4, 5]);
 
-          //baz(arr[1 .. 3]); // cannot match length
+            //foo(arr[0..3]); // cannot match length
```
dkorpel, line 1583: `0 .. 3` should have spaces around `..` according to the style guide

spec/hash-map.dd
```diff
@@ -29,6 +31,17 @@ $(HEADERNAV_TOC)
         of the keys inserted into the array. In particular, in a $(D foreach) loop the
         order in which the elements are iterated is typically unspecified.)
 
+$(H2 $(LNAME2 literals, Literals))
+
+$(SPEC_RUNNABLE_EXAMPLE_RUN
+---
+auto aa = [21u:"he", 38:"ho", 2:"hi"];
```
dkorpel, line 38: Style guide doesn't specify this, but I think there should be a space after the `:`

spec/arrays.dd
ntrel: Thanks. Fixed, along with the style points.

PR #3163, opened 2022-01-07
[arrays.dd] Improve pointer docs

ntrel:
~~Depends on #3162 for anchor link.~~

Document dereference and address of operators with examples.
Mention pointing to multiple elements and how this is not `@safe`.
Add *Pointer Arithmetic* subheading for *Indexing* & add example. This replaces the old section, actually compiles and is simpler.

### DIFF
--- a/spec/arrays.dd
+++ b/spec/arrays.dd
@@ -235,11 +235,12 @@ $(SPEC_RUNNABLE_EXAMPLE_RUN
 int[] a = [1,2,3];
 int* p = a.ptr;
 
-*(p + 2) = 4; // same as `p[2] = 4`
+p[2] = 4;
 assert(a[2] == 4);
+writeln(p[3]); // undefined behaviour
 
 assert(p == &a[0]);
-p++;
+p++; // point to a[1]
 assert(*p == 2);
 ---
 )
--- a/spec/arrays.dd
+++ b/spec/arrays.dd
@@ -18,13 +18,24 @@ $(H2 $(LNAME2 array-kinds, Kinds))
 
 $(H3 $(LNAME2 pointers, Pointers))
 
+        $(P A pointer to type $(D T) has a value which is a reference (address) to another
+        object of type $(D T). It is commonly called a $(I pointer to T) and it's type is
+        `T*`. To access the object value, use the `*` dereference operator:
+        )
+
+$(SPEC_RUNNABLE_EXAMPLE_RUN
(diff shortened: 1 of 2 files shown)


spec/arrays.dd
```diff
@@ -18,13 +18,24 @@ $(H2 $(LNAME2 array-kinds, Kinds))
 
 $(H3 $(LNAME2 pointers, Pointers))
 
+        $(P A pointer to type $(D T) has a value which is a reference (address) to another
+        object of type $(D T). It is commonly called a $(I pointer to T) and it's type is
```
dkorpel, line 22:
```suggestion
        object of type $(D T). It is commonly called a $(I pointer to T) and its type is
```

dkorpel (approved):

spec/arrays.dd
```diff
@@ -192,6 +224,28 @@ assert(b[1] == 2);
...
+    arithmetic expressions to produce another pointer:)
+
+$(SPEC_RUNNABLE_EXAMPLE_RUN
+---
+int[] a = [1,2,3];
+int* p = a.ptr;
+
+*(p + 2) = 4; // same as `p[2] = 4`
+assert(a[2] == 4);
+
```
maxhaton, line 240: Mention `ptrdiff_t` somewhere?

spec/arrays.dd
```diff
@@ -192,6 +224,28 @@ assert(b[1] == 2);
...
+    $(P A pointer can also be indexed, but no bounds checks are done.
+    Unlike arrays, a pointer value can also be used in certain
+    arithmetic expressions to produce another pointer:)
+
+$(SPEC_RUNNABLE_EXAMPLE_RUN
+---
+int[] a = [1,2,3];
+int* p = a.ptr;
+
+*(p + 2) = 4; // same as `p[2] = 4`
```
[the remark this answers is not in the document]
ntrel: I have now changed this to just use `p[2]`, not `p + 2`. The latter is harder to explain, and is already explained in the *AddExpression* docs, which are linked to below this example. I have kept `p++` (but added a comment) as that is simpler. I have added an *AddExpression* example  with `p + 2`.

spec/arrays.dd
ntrel: I have added an *AddExpression* example showing that. I don't think we need it in the array docs, pointer indexing is enough IMO.

PR #3167, opened 2022-01-08
[spec] Make more examples runnable

ntrel:
Also tweak those examples:
Move const attribute after parameters.
Fix calling array delegate elements.
Rename 2 eponymous template member functions so they can be called explicitly (this is before the IFTI section). Needed to compile.
Replace some trivial writefln calls with writeln.

### DIFF
--- a/spec/template.dd
+++ b/spec/template.dd
@@ -362,8 +362,6 @@ $(GNAME TemplateThisParameter):
 
         $(SPEC_RUNNABLE_EXAMPLE_COMPILE
         ---
-        import std.stdio;
-
         struct S
         {
             void foo(this T)(int i) const
--- a/spec/template.dd
+++ b/spec/template.dd
@@ -796,10 +796,8 @@ $(GNAME TemplateSequenceParameter):
             }
         }
 
-        template write(Args...)
+        template write(Args...) // Args is a TypeSeq
         {
-            pragma(msg, Args); // Args is a TypeSeq
-
             void f(Args args) // args is a ValueSeq
             {
                 writeln("args are ", args);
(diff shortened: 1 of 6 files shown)


ljmf00:
> Also: Move const attribute after parameters. Rename gc -> GC. Fix calling array delegate elements. Rename 2 eponymous template member functions so they can be called explicitly (this is before the IFTI section). Add _Type Sequence Deduction_ subheading. Replace some trivial writefln calls with writeln.

Thanks for this! As per contributions guidelines, try to avoid making refactoring/unrelated changes in the same PR.

ntrel:
@ljmf00 I have removed the added subheading and gc -> GC typo changes. All the other changes are either needed for the examples to compile or are obvious minor tweaks to the now runnable examples.

spec/template.dd
```diff
@@ -360,14 +360,15 @@ $(GNAME TemplateThisParameter):
         infer the mutability of the `this` reference. For example, if
         `this` is `const`, then the function is marked `const`.)
 
+        $(SPEC_RUNNABLE_EXAMPLE_COMPILE
         ---
         import std.stdio;
```
dkorpel, line 365: This import is now not necessary anymore

spec/template.dd
```diff
@@ -777,30 +784,33 @@ $(GNAME TemplateSequenceParameter):
...
+            void f(Args args) // args is a ValueSeq
             {
                 writeln("args are ", args);
             }
         }
 
         void main()
         {
-            print!(1,'a',6.8).print();                    // prints: args are 1a6.8
-            write!(int, char, double).write(1, 'a', 6.8); // prints: args are 1a6.8
```
dkorpel, line 801: This is interesting, an eponymous function template cannot be called explicitly, it tries to call `write!(int, char, double)().write(1, 'a', 6.8);` and says it's missing arguments. I wonder if that's a bug.

dkorpel (approved):
Great stuff, the changes to the code exactly show why they should be runnable.

spec/template.dd
ntrel: Good point, removed.

spec/template.dd
ntrel: I think it's because of implicit parentheses for function calls.

PR #3171, opened 2022-01-11
[spec] Improve *AliasDeclaration* docs

ntrel:
Define aliases in terms of another symbol, not just another type.
Add subheadings.
Add note about capitalizing non-basic type aliases.
Move paragraph about type aliases being indistinguishable (from other aliases) up to type aliases section.
Move function type aliases example down to function type aliases section.
Make 2 examples runnable.
Simplify aliasing an overload set example - classes A and D are not needed.
Make *AliasAssign* sections child headings of *AliasDeclaration*.
Make *AliasAssign* examples get compiled; add static assert with `staticMap` and `T.sizeof`.

### DIFF
--- a/spec/declaration.dd
+++ b/spec/declaration.dd
@@ -281,12 +281,14 @@ $(GNAME AliasAssignment):
     $(GLINK_LEX Identifier) $(GLINK2 template, TemplateParameters)$(OPT) $(D =) $(GLINK StorageClasses)$(OPT) $(GLINK2 type, BasicType) $(GLINK2 function, Parameters) $(GLINK2 function, MemberFunctionAttributes)$(OPT)
 )
 
-    $(P $(I AliasDeclaration)s create a symbol that is an alias for another type,
-        and can be used anywhere that other type may appear.
+    $(P $(I AliasDeclaration)s create a symbol name that refers to another symbol.
+        That symbol name can be used anywhere that the aliased symbol may appear.
     )
 
+$(H3 $(LNAME2 alias-type, Type Aliases))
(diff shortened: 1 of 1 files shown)


spec/declaration.dd
```diff
@@ -296,12 +298,29 @@ alias myint = abc.Foo.bar;
...
+        Type aliases can sometimes look indistinguishable from
+        other symbol aliases:
+        )
+
+--------------------
+alias abc = foo.bar; // is it a type or a symbol?
 --------------------
 
+        $(P
+        The distinction is made in the semantic analysis pass.
```
maxhaton, line 317: I Don't think this should be mentioned in the spec.

spec/declaration.dd
```diff
@@ -545,15 +563,22 @@ $(GNAME AliasReassignment):
 
         $(P An alias declaration inside a template can be reassigned a new value.)
 
+        $(SPEC_RUNNABLE_EXAMPLE_COMPILE
         ---
-        template staticMap(alias F, T...)
+        alias AliasSeq(a...) = a;
```
maxhaton, line 568:
```suggestion
        import std.meta : AliasSeq;
```
?

maxhaton: Having the declaration inline isn't such a bad thing but best not to encourage people to redeclare it everywhere

spec/declaration.dd
ntrel: This was an existing sentence, I've now removed it.

maxhaton (approved):

PR #3172, opened 2022-01-11
Fix deadlink in heading "Memory Model"

quassy:
The heading was linking to https://dlang.org/spec/memory-model which does not exist. I guess it should have been an anchor link instead.

### DIFF
--- a/spec/intro.dd
+++ b/spec/intro.dd
@@ -66,7 +66,7 @@ $(OL
 )
 
 
-$(H2 $(LINK2 memory-model, Memory Model))
+$(H2 $(LNAME2 memory-model, Memory Model))
 
     $(P The $(I byte) is the fundamental unit of storage. Each byte has 8 bits and is stored at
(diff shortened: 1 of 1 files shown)


ntrel (approved):

ntrel:
The build error looks unrelated. (You could try `git commit -a` with no changes then force pushing to trigger the tests again).

quassy:
@ntrel On second read I think I misunderstood and you wanted me to force commit the same single commit 😅, but the checks seem to pass now and you should just be able to sqash merge this change to keep the history clean.

ntrel:
> @ntrel On second read I think I misunderstood and you wanted me to force commit the same single commit 😅, but the checks seem to pass now and you should just be able to sqash merge this change to keep the history clean.

Great. I don't have merge rights though. (Sorry I meant `git commit --amend`).

maxhaton (approved):

maxhaton:
Thanks

PR #3173, opened 2022-01-13
[spec/template.dd] Improve docs

ntrel:
Add *Template Declarations* heading:
Add example for template type parameter.
Fix: Value parameters are not limited to int, float, string.
Fix: Alias parameters can take local symbols.
Move paragraph about overloaded templates from *Explicit Template Instantiation* to *Template Declarations*.
Mention eponymous templates & often using short syntax templates instead.

Move sentence about value specialization to *Value Parameters*.
Add *Type Sequence Deduction* subheading of *Sequence Parameters*.

Tweak *Eponymous Templates* wording.
Rename *Aggregate Templates* to *Aggregate Type Templates*.

Move 'IFTI vs static if' section from *Eponymous Templates* to *IFTI* & add *Restrictions* subheading.
Move *Template Constructors* below *Function Templates* section.

### DIFF
--- a/spec/template.dd
+++ b/spec/template.dd
@@ -934,7 +934,8 @@ $(H2 $(LNAME2 implicit_template_properties, Eponymous Templates))
         }
         ------
 
-        $(P Using functions and more types than the template:)
+        $(P The following example has more than one eponymous member and uses
+        $(RELATIVE_LINK2 ifti, Implicit Function Template Instantiation):)
 
         ------
         template foo(S, T)
@@ -948,24 +949,6 @@ $(H2 $(LNAME2 implicit_template_properties, Eponymous Templates))
         {
             foo(1, 2, "test"); // foo!(int, int).foo(1, 2, "test")
             foo(1, 2); // foo!(int, int).foo(1, 2)
-        }
-        ------
-
-        $(P When the template parameters must be deduced, the eponymous members
-        can't rely on a $(LINK2 version.html#StaticIfCondition, `static if`)
-        condition since the deduction relies on how the in members are used:)
-
-        ------
-        template foo(T)
-        {
-            static if (is(T)) // T is not yet known...
-                void foo(T t) {} // T is deduced from the member usage
-        }
-
-        void main()
-        {
-            foo(0); // Error: cannot deduce function from argument types
-            foo!int(0); // Ok since no deduction necessary
         }
         ------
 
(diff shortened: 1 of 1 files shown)


spec/template.dd
```diff
@@ -53,7 +53,14 @@ $(GNAME TemplateParameter):
         $(I TemplateParameter) in case one is not supplied.
     )
 
-$(H2 $(LNAME2 explicit_tmp_instantiation, Explicit Template Instantiation))
+    $(NOTE `template` keyword declarations are less common in idiomatic D code.
```
maxhaton, line 56: Is this note necessary? Also it should probably be "less common than [something]" if this point is to be made.

spec/template.dd
```diff
@@ -36,11 +38,9 @@ $(GNAME TemplateParameter):
     )
 
     $(P Template parameters can be types, values, symbols, or sequences.
-        Types can be any type.
-        Value parameters must be of an integral type, floating point
-        type, or string type and
-        specializations for them must resolve to an integral constant,
-        floating point constant, null, or a string literal.
+        Type parameters can be any type.
+        Value parameters can have any type which can be statically
```
maxhaton, line 42: Isn't it the case that the *type* of a value parameter can be any valid D type, but whether an argument can be passed to it is the point being made about it being computable at compile?

spec/template.dd
ntrel: The 'than' is on the next line. The point is for people reading the page as a learning resource rather than a reference (the page seems intended for that as it gradually introduces things). They should not get the idea that verbose template blocks with explicit instantiation is typical in D.

spec/template.dd
ntrel: The *argument* can be of any compile-time constructable type. I essentially copied the first sentence of:
https://dlang.org/spec/template.html#template_value_parameter

I would prefer not to repeat the kinds of types which satisfy that requirement, as they are listed in the value parameter section. (There is a link to that just above this paragraph).

spec/template.dd
ntrel: @maxhaton I've reworded this a bit to mention single member templates, see 640802e

spec/template.dd
ntrel: @maxhaton I've now tweaked this to say *Value parameters can take any value which can be statically initialized at compile time*, so it no longer talks about types.

ntrel: Another tweak - *can take any expression...*. Also I restored the sentence about the specialization expression needing to be CT evaluatable, but I put it in the *TemplateValueParameter* section instead as that seems more appropriate - 4a0300c.

ntrel:
I have made a few changes:
* Tweak TemplateDeclaration docs & mention eponymous templates.
* Split out the *Template Instantiation* changes to #3177.

maxhaton (approved):

PR #3177, opened 2022-01-19
[spec/template.dd] Improve template instantiation docs

ntrel:
Add *Template Instantiation* heading:
Add description, mention IFTI.
Make *Explicit Instantiation* a subheading.
Add *Common Instantiation* subheading.
Make *Instantiation Scope* a subheading & fix example formatting.

### DIFF
--- a/spec/template.dd
+++ b/spec/template.dd
@@ -74,7 +74,18 @@ $(GNAME TemplateParameter):
         template declarations instead.)
 
 
-$(H2 $(LNAME2 explicit_tmp_instantiation, Explicit Template Instantiation))
+$(H2 $(LNAME2 template_instantiation, Template Instantiation))
+
+    $(P A template must be instantiated before use. This means passing one or
+    more arguments to the template which are substituted into the template
+    body, which becomes a new scoped entity.)
+
+    $(P $(RELATIVE_LINK2 function-templates, Function templates) can
+    often be implicitly instantiated by inferring the template
+    arguments from a function call. Other templates must be
+    instantiated explicitly.)
+
+$(H3 $(LNAME2 explicit_tmp_instantiation, Explicit Template Instantiation))
 
     $(P Templates are explicitly instantiated with:
     )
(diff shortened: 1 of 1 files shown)


spec/template.dd
```diff
@@ -53,7 +53,19 @@ $(GNAME TemplateParameter):
         $(I TemplateParameter) in case one is not supplied.
     )
 
-$(H2 $(LNAME2 explicit_tmp_instantiation, Explicit Template Instantiation))
+
+$(H2 $(LNAME2 template_instantiation, Template Instantiation))
+
+    $(P A template must be instantiated before use. This means passing one or
```
maxhaton, line 59: templates instantiations can have no arguments

ntrel: Sorry, I'm not sure what you mean. Instantiating does take arguments.

maxhaton: `temp!()()`

ntrel: I didn't write passing arguments to an instantiation though, I said *passing arguments to the template*. I don't understand what's wrong.

ntrel: Oh I see! I changed it to `This usually means passing one or more`...

dkorpel: A specification should be precise, I would avoid words like 'usually' or 'often'. How about "This means passing an argument list".

spec/template.dd
```diff
@@ -74,7 +74,18 @@ $(GNAME TemplateParameter):
...
-$(H2 $(LNAME2 explicit_tmp_instantiation, Explicit Template Instantiation))
+$(H2 $(LNAME2 template_instantiation, Template Instantiation))
+
+    $(P A template must be instantiated before use. This usually means
+    passing one or more arguments to the template which are substituted
+    into the template body, which becomes a new scoped entity.)
+
+    $(P $(RELATIVE_LINK2 function-templates, Function templates) can
+    often be implicitly instantiated by inferring the template
+    arguments from a function call. Other templates must be
```
dkorpel, line 85: Avoid 'often'.
```suggestion
    be implicitly instantiated if the compiler can infer the template
    arguments from a function call. Otherwise the template must be
```

spec/template.dd
ntrel: @dkorpel I've now added that phrasing.

dkorpel (approved):

dkorpel (approved):

PR #3180, opened 2022-01-20
[spec/statement.dd] Improve examples for break & continue

ntrel:


dkorpel (commented):
I like that the examples are runnable, but they are also a bit more complex. There's extra math operations and `writeln`s, but it's not really used since there's no expected output shown, so I think they make it a bit more confusing to read.

ntrel:
Superceded by #3576.

PR #3181, opened 2022-01-20
Remove underground rekordz as the link now seems to point to a drug c…

maxhaton:
…ompany


CyberShadow (approved):

CyberShadow:
The book timestamps need to be removed in CI, they are interfering with cache deduplication.

PR #3182, opened 2022-01-21
ImportC: add documentation for __import

WalterBright:
The implementation: https://github.com/dlang/dmd/pull/13539

### DIFF
--- a/spec/importc.dd
+++ b/spec/importc.dd
@@ -226,10 +226,12 @@ $(H2 $(LNAME2 extensions, Extensions))
     $(P Any declarations in scope can be accessed, not just
     declarations that lexically precede a reference.)
 
+
     $(H3 $(LNAME2 ctfe, Compile Time Function Execution))
 
     $(P Evaluating constant expressions includes executing functions in the
     same manner as D's CTFE can.)
+
 
     $(H3 $(LNAME2 inlining, Function Inlining))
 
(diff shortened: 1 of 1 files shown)


spec/importc.dd
```diff
@@ -245,6 +247,77 @@ $(H2 $(LNAME2 extensions, Extensions))
...
+    importing module, and preprocessing symbols in the importing file are not
+    available to the imported module.)
+
+    $(P A D module can be imported, in the same manner as that
+    of a $(GLINK2 module, ImportDeclaration).)
+
+    $(P Imports can be circular.)
+
+$(CCODE
+__import core.stdc.stdarg; // get D declaration of va_list
```
maxhaton, line 300: I think this example should be something like hello world instead since `va_list`s can't actually be manipulated (with all capabilities required by the standard) without a shim.

maxhaton: The rest is good.

csharpdf:
I'd appreciate this.

spec/importc.dd
WalterBright: I'm not understanding what you mean. User shims aren't required to do stdarg stuff, that is taken care of by the default importation of __builtins.c.

maxhaton: `va_arg` is a macro, that presumably won't work for a while?

WalterBright: va_arg is a template.

maxhaton: `va_arg` as in https://en.cppreference.com/w/c/variadic/va_arg is a macro.

I will be opening a PR to add basic preprocessor support soon

maxhaton (approved):

PR #3184, opened 2022-01-22
[spec/function.dd] Improve *Virtual Functions* section

ntrel:
Move simple overriding example from 'Function Inheritance and Overriding' to use as first example. This introduces `override` before the next more complex example about virtual vs non-virtual methods, which also uses `override`.
Explain that `override` is required and that it can catch bugs.
Use list for kinds of non-virtual functions.
Simplify examples a bit.
Move paragraph about Objective-C to end of section.
Add 'Covariance' subheading.
Add 'Calling Base Class Methods' subheading & tweak wording.
Change 'Function Inheritance and Overriding' subheading to 'Overload Sets and Overriding'. Add missing `override` attribute in an example and use `SPEC_RUNNABLE_EXAMPLE_FAIL`.
Add 'Default Values', 'Inherited Attributes', 'Restrictions' subheadings.

'Function Parameters' section:
Add subheading for `this` reference and Obj-C selector & move docs here.

attributes.dd:
Remove `override` paragraphs and link to 'Virtual Functions'.

### DIFF
--- a/spec/attribute.dd
+++ b/spec/attribute.dd
@@ -549,30 +549,7 @@ $(H2 $(LNAME2 return, $(D return) Attribute))
 
 $(H2 $(LNAME2 override, $(D override) Attribute))
 
-        $(P The $(D override) attribute applies to virtual functions.
-        It means that the function must override a function with the
-        same name and parameters in a base class. The override attribute
-        is useful for catching errors when a base class's member function
-        gets its parameters changed, and all derived classes need to have
-        their overriding functions updated.
-        )
-
----------------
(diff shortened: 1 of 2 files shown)


dkorpel (approved):
I didn't think Covariance needed its own heading since it was only two sentences.

Then I noticed the spec doesn't provide a complete definition of 'covariant' somewhere, instead there's some hints sprinkled all over the function section like "Safe functions are covariant with trusted or system functions", so I think it could use its own section, which can be expanded in a later PR.

PR #3190, opened 2022-01-26
add bitfields documentation

WalterBright:
Implementation: https://github.com/dlang/dmd/pull/13568


WalterBright:
@CyberShadow can we please have a way to find what the errors are with DAutoTest?

thewilsonator:
With DautoTest (unless you have LeTeX error in which case you're screwed) scroll _all_ the way to the bottom of of the Build Log (click view).

In this case your problem is 
```
Searching for tabs
./spec/struct.dd:134:	b.x = 4;
```

WalterBright:
Oh, I clicked on "base view". No errors there. I wonder what that is.

Anyhow, thanks! Hope it works now.

MrSmith33:
The text mentions `unit` several times, but never says what it is.

WalterBright:
@MrSmith33 it does now!

WalterBright:
The dmd bitfields PR has been merged, under the -preview=bitfields switch. This one needs to be, too.

spec/struct.dd
```diff
@@ -95,6 +95,57 @@ $(H2 $(LNAME2 struct_layout, Struct Layout))
     $(LI Avoid using empty structs as parameters or arguments to variadic functions.)
     ))
 
+$(H2 $(LNAME2 bitfields, Bit Field Declarations))
+
```
maxhaton, line 99:
```suggestion
$(P Bit fields are available under the preview flag $(D -preview=bitfields))
```

maxhaton:
> The dmd bitfields PR has been merged, under the -preview=bitfields switch. This one needs to be, too.

OK - As soon as it mentions the preview flag I will merge it

Moth-Tolias:
does the best practices section at line 44 still hold?

spec/declaration.dd
```diff
@@ -55,6 +55,12 @@ $(GNAME DeclaratorIdentifierList):
 $(GNAME DeclaratorIdentifier):$(LEGACY_LNAME2 VarDeclaratorIdentifier)
     $(GLINK_LEX Identifier)
     $(GLINK_LEX Identifier) $(GLINK2 template, TemplateParameters)$(OPT) $(D =) $(GLINK Initializer)
+    $(GLINK BitfieldDeclarator)
+
+$(GNAME BitfieldDeclarator):
+    $(D :) $(GLINK2 expression, AssignExpression)
+    $(GLINK_LEX Identifier) $(D :) $(GLINK2 expression, AssignExpression)
+    $(GLINK_LEX Identifier) $(D :) $(GLINK2 expression, AssignExpression) $(D =) $(GLINK Initializer)
```
tim-dlang, line 63: Using AssignExpression after `:` introduces an ambiguity for the following example:
`
int a : b = c;
`

The part `b = c` could be an AssignExpression or `b` could be an AssignExpression and `c` an Initializer.

The parser seems to use a ConditionalExpression instead: https://github.com/dlang/dmd/blob/df3b0559c21b65ebf62f508f100fc4e25ffd4885/src/dmd/parse.d#L4614

tim-dlang:
DeclaratorInitializer should also get a rule with BitfieldDeclarator in the grammar. DeclaratorIdentifier is only used for declarators after the first comma, while DeclaratorInitializer is used for the first declarator.

tim-dlang:
Another potential ambiguity in the grammar:

```
void f()
{
    a: b;
}
```

The statement could be a bitfield with `a` as type and `b` as width, but it could also be a labeled statement with `a` as label and `b` as a statement. Bitfields are probably only allowed in classes and structs, but the grammar would also allow them in functions.

spec/declaration.dd
MrSmith33: Q: Initial implementation [forbids](https://github.com/dlang/dmd/pull/13568/files#diff-7b6e6126b5e5b6f47c0389d7533a9336a434356c4b2c515fc43bb44328a7123aR4) bitfield initializers. Does this mean that they will be supported in the future?

spec/struct.dd
```diff
@@ -95,6 +95,57 @@ $(H2 $(LNAME2 struct_layout, Struct Layout))
...
+    $(P A bit field is declared with a $(GLINK2 declaration, BitfieldDeclarator). The bit field width
+    is specified by the $(GLINK2 expression, AssignExpression). It is evaluated at compile time, and
+    must result in an integer expression that ranges from 0 to the number of bits in type of the
```
ntrel, line 102:
```suggestion
    must result in an integer expression that ranges from 0 to the number of bits in the type of the
```

spec/struct.dd
```diff
@@ -95,6 +95,57 @@ $(H2 $(LNAME2 struct_layout, Struct Layout))
...
+    bit field.)
+
+    $(P The type of the bit field must be an integral type, signed or unsigned. The values assigned to
+    bit fields must fit into the specified width. The bit field is placed into a unit of memory at least
+    as large as needed to hold a value of the type of the bit field.)
+
+    $(P Anonymous bit fields do not have a $(GLINK_LEX Identifier) for them.)
```
ntrel, line 109:
```suggestion
    $(P Anonymous bit fields do not have an $(GLINK_LEX Identifier) for them.)
```

spec/declaration.dd
```diff
@@ -55,6 +55,12 @@ $(GNAME DeclaratorIdentifierList):
 $(GNAME DeclaratorIdentifier):$(LEGACY_LNAME2 VarDeclaratorIdentifier)
     $(GLINK_LEX Identifier)
     $(GLINK_LEX Identifier) $(GLINK2 template, TemplateParameters)$(OPT) $(D =) $(GLINK Initializer)
+    $(GLINK BitfieldDeclarator)
+
+$(GNAME BitfieldDeclarator):
+    $(D :) $(GLINK2 expression, AssignExpression)
+    $(GLINK_LEX Identifier) $(D :) $(GLINK2 expression, AssignExpression)
```
ntrel, line 62:
```suggestion
    $(GLINK_LEX Identifier)$(OPT) $(D :) $(GLINK2 expression, AssignExpression)
```

ntrel (approved):

AndrejMitrovic:
@WalterBright you may want to update this part: https://github.com/dlang/dlang.org/blob/ae9d2d63077db4daff8407280d4e6ece1547e942/spec/struct.dd#L90-L94

spec/declaration.dd
ntrel, line 62:
```suggestion
    $(GLINK_LEX Identifier)$(OPT) $(D :) $(GLINK2 expression, AssignExpression)
```

spec/declaration.dd
ntrel: Update for @tim-dlang's observation:
```suggestion
    $(GLINK_LEX Identifier) $(D :) $(GLINK2 expression, ConditionalExpression) $(D =) $(GLINK Initializer)
```

ntrel, line 63: Update for @tim-dlang's observation (fixed):
```suggestion
    $(GLINK_LEX Identifier) $(D :) $(GLINK2 expression, ConditionalExpression)
    $(GLINK_LEX Identifier) $(D :) $(GLINK2 expression, ConditionalExpression) $(D =) $(GLINK Initializer)
```

spec/declaration.dd
ntrel: Note the above trivial change will conflict with the important grammar change below so maybe ignore it.

spec/declaration.dd
ntrel: See https://github.com/dlang/dlang.org/pull/3190/files/ccdacb6866cd19d3617ec13d4792525e496c85b0#r884139913

ntrel:
@maxhaton If bitfields are here to stay it would be good to merge this after adding any of the diff suggestions as you see fit. After that I can make a pull to fix https://github.com/dlang/dlang.org/pull/3190#issuecomment-1140274338

adamdruppe:
We really shouldn't document this useless trash feature that never passed the DIP review to be added to the language.

spec/declaration.dd
WalterBright: They're both equivalent.

spec/declaration.dd
WalterBright: It would be a reasonable enhancement.

WalterBright: @tim-dlang you're right about the ConditionalExpression.

WalterBright: yes

WalterBright:
> you may want to update this part

Done.

WalterBright:
Ready to merge.

dkorpel (changes requested):
Like Adam said, this feature hasn't been added to the language. This specification text makes it sound like it is. Perhaps it can be documented on a separate page for preview switches.

WalterBright:
@dkorpel see https://github.com/dlang/dlang.org/pull/3190/files#diff-dee70e840ada3a820b3a373d649812b1157fd584ed0837e4a937ea3418f093e3R212

dkorpel:
A preview switch doesn't imply it's not part of the language, there are preview switches for current language features.

WalterBright:
Since bitfields do not influence anything else in the language semantics, there's no longer a purpose to putting bitfields behind a preview switch.

PR #3195, opened 2022-01-28
Fix issue 22711 - Effect of template UDAs on instance members is undo…

pbackus:
…cumented


maxhaton (approved):
Good spot. Thanks Paul

PR #3197, opened 2022-01-29
Correct hasMember trait example

andrejp88:
The `hasMember` trait made an odd claim — that adding `import std.stdio;` inside a struct S makes `write` a member of that struct. If you run the example, you can see that `write` isn't a member after all.

I left the `import std.stdio;` at the top to show that UFCS doesn't mean any function is a member.

### DIFF
--- a/spec/traits.dd
+++ b/spec/traits.dd
@@ -677,8 +677,6 @@ import std.stdio;
 struct S
 {
     int m;
-
-    import std.stdio; // imports write
 }
 
 void main()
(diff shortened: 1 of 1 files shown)


spec/traits.dd
```diff
@@ -688,7 +686,7 @@ void main()
     writeln(__traits(hasMember, S, "m")); // true
     writeln(__traits(hasMember, s, "m")); // true
     writeln(__traits(hasMember, S, "y")); // false
-    writeln(__traits(hasMember, S, "write")); // true
+    writeln(__traits(hasMember, S, "write")); // false
```
maxhaton, line 689:
```suggestion
    writeln(__traits(hasMember, S, "write")); // false, but callable like a member via UFCS
```

maxhaton (approved):

PR #3199, opened 2022-01-31
[spec/statement.dd] Improve 'Foreach over Associative Arrays'

ntrel:
Add example.
Fix: The variable types don't have to match key & value types exactly, they just need to be compatible.
Make existing example runnable & move above two sentences.
~~Use UNDEFINED_BEHAVIOR macro for unspecified iteration order.~~

### DIFF
--- a/spec/statement.dd
+++ b/spec/statement.dd
@@ -617,25 +617,43 @@ $(H3 $(LNAME2 foreach_over_associative_arrays, Foreach over Associative Arrays))
         can be one or two variables declared. If one, then the variable
         is said to be the $(I value) set to the elements of the array,
         one by one. The type of the
-        variable must match the type of the array contents. If there are
+        variable must be compatible with the array element type.)
+
+$(SPEC_RUNNABLE_EXAMPLE_RUN
+--------------
+// value type is int
+int[string] userAges = ["john":30, "sue":32];
+
+foreach (ref age; userAges)
+{
+    age++;
+}
+assert(userAges == ["john":31, "sue":33]);
+--------------
+)
+
+        $(P If there are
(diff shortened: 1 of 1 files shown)


spec/statement.dd
```diff
@@ -617,25 +617,43 @@ $(H3 $(LNAME2 foreach_over_associative_arrays, Foreach over Associative Arrays))
...
+
+foreach (string s, double d; aa)
 {
-    writefln("a['%s'] = %g", s, d);
+    writefln("aa['%s'] = %g", s, d);
 }
 --------------
+)
+
+        $(UNDEFINED_BEHAVIOR The order in which the elements of the
```
dkorpel, line 654: I think this is 'unspecified behavior', undefined behavior must be prevented by the compiler in `@safe` code.

spec/statement.dd
```diff
@@ -617,25 +617,43 @@ $(H3 $(LNAME2 foreach_over_associative_arrays, Foreach over Associative Arrays))
...
+}
+assert(userAges == ["john":31, "sue":33]);
+--------------
+)
+
+        $(P If there are
         two variables declared, the first is said to be the $(I index)
         and the second is said to be the $(I value). The $(I index)
-        must be of the same type as the indexing type of the associative
+        must be compatible with the indexing type of the associative
```
dkorpel, line 638: What counts as 'compatible', implicitly convertible?

spec/statement.dd
ntrel: OK, fixed.

spec/statement.dd
ntrel: Yes, updated.

dkorpel (approved):

PR #3202, opened 2022-02-01
[spec/statement.dd] Improve 'Foreach over Arrays'

ntrel:
The variable type can be compatible with the element type, not match exactly.
Mention modifying the element with `ref`.
Move paragraph about index type below example; use type inference.

### DIFF
--- a/spec/statement.dd
+++ b/spec/statement.dd
@@ -536,7 +536,8 @@ foreach (i, char c; a)
 }
 --------------
 )
-        $(P The $(I index) type must be `size_t` for a dynamic array.
+        $(P For a dynamic array, the $(I index) type must be compatible
+        with `size_t`.
         Static arrays may use any integral type that spans the length
         of the array.)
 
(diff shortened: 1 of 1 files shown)


spec/statement.dd
```diff
@@ -511,32 +511,38 @@ $(H3 $(LNAME2 foreach_over_arrays, Foreach over Arrays))
...
 char[] a = ['h', 'i'];
 
-foreach (size_t i, char c; a)
+foreach (i, char c; a)
 {
     writefln("a[%d] = '%c'", i, c);
 }
 --------------
 )
+        $(P The $(I index) type must be `size_t` for a dynamic array.
```
dkorpel, line 539: This is not true, it must be an integral type that a `size_t` converts to. It can also be a signed `long` for example.

dkorpel (commented):
Looks good, except for one inaccuracy.

spec/statement.dd
ntrel: Thanks, fixed.

dkorpel (approved):

PR #3203, opened 2022-02-01
[spec/statement.dd] Improve foreach docs

ntrel:
Explain the variables declared in `ForeachTypeList` better.

Move paragraph about inferring the type of the `ForeachType` declaration from the `ref` subheading to main foreach section.
Move note about `auto` being redundant from array subheading, along with the example.
Modify the example, make runnable. There's no real need to show `auto` causing an error.

~~Mark resizing/reassigning a container during foreach as undefined behaviour. Tweak example.~~

### DIFF
--- a/spec/statement.dd
+++ b/spec/statement.dd
@@ -445,7 +445,7 @@ for (int i = 0; i < 10; i++)
 
 $(H2 $(LEGACY_LNAME2 ForeachStatement, foreach-statement, Foreach Statement))
 
-$(P A `foreach` statement loops over the contents of an aggregate.)
+$(P A `foreach` statement iterates a series of values.)
 
 $(GRAMMAR
 $(GNAME AggregateForeach):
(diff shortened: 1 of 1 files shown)


spec/statement.dd
```diff
@@ -984,30 +992,28 @@ $(CONSOLE
...
-        $(P If not specified, the $(I Type)s in the $(I ForeachType) can be
-        inferred from
-        the type of the $(I ForeachAggregate).
-        )
+        $(P $(D ref) cannot be applied to an array index variable.)
 
 $(H3 $(LNAME2 foreach_restrictions, Foreach Restrictions))
 
-        $(P The aggregate itself must not be resized, reallocated, free'd,
+        $(UNDEFINED_BEHAVIOR The aggregate itself must not be resized, reallocated, free'd,
```
dkorpel, line 999: I'm not sure this is undefined behavior, that implies it must be prevented in `@safe` code by the compiler.

ntrel: OK, updated.

ntrel: I've extended the examples and moved this to #3205.

dkorpel (approved):

PR #3204, opened 2022-02-01
[spec/statement.dd] Improve 'Foreach over Structs & Classes with Ranges'

ntrel:
Link to `std.range`.
Add example using linked list.

Add 'Multiple Element Values' subheading.
Fix tuple -> sequence typo.
Fix tuple front example to not infinite loop & make runnable. Show single foreach variable is still legal.
Link to `Tuple`.

### DIFF
--- a/spec/statement.dd
+++ b/spec/statement.dd
@@ -873,16 +873,16 @@ $(H4 $(LNAME2 front-seq, Multiple Element Values))
 
         $(SPEC_RUNNABLE_EXAMPLE_RUN
         ---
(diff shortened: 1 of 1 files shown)


dkorpel (approved):
Eventually I would like to see an example that doesn't depend on Phobos tuples, since I found it pretty hard to figure out how to get multiple `foreach` variables on my own types.

ntrel:
@dkorpel Thanks for reviewing (and the other PRs). I've updated this to define `Tuple` in the example instead of using `std.typecons`.

maxhaton (approved):

PR #3207, opened 2022-02-05
[articles/ctarguments.dd] Various improvements

ntrel:
Use CONSOLE macro for output.
Use lowercase identifiers for value sequences.
Change 'Assignment' heading to 'Element assignment' and make it a subheading of 'Indexing and slicing'.
Change 'Loops' to 'foreach'.

Change 'Type sequence instantiation' to not be a subheading of type sequences because it's about specific lvalue sequences, which the next heading continues for value sequences in general.
Move `Tuple` sentence to 'Value sequences' heading (it applies to other value sequences as well besides just type sequence instances).
Remove unnecessary 'Aggregate field sequences' subheading.
Mention value sequences can't be returned from functions.

Add symbol sequence example.

### DIFF
--- a/articles/ctarguments.dd
+++ b/articles/ctarguments.dd
@@ -36,22 +36,24 @@ $(HEADERNAV_TOC)
     ---
     template Variadic(T...)
     {
-        static assert (T.length > 1);
+        static assert(T.length > 1);
         pragma(msg, T[0]);
         pragma(msg, T[1]);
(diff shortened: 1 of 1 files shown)


ntrel:
BTW I'm also working on updating the template sequence parameter docs, almost ready.

maxhaton (approved):

PR #3209, opened 2022-02-09
[articles/mixin.dd] Add concatenation, statements, types, expressions

ntrel:
Change `genStruct` eponymous template to enum template.
Add 5 sections: argument concatenation, declarations, statements, types, expressions.
Add *Comparison with C preprocessor* heading.

Also tweak spec examples:
Split mixin statement spec examples into 2. Change for -> foreach and show mixin code reading a variable declared in current scope.
Move mixin variable declaration example to *MixinDeclaration* from statement.dd.

### DIFF
--- a/articles/mixin.dd
+++ b/articles/mixin.dd
@@ -2,31 +2,96 @@ Ddoc
 
 $(D_S String Mixins,
 
+$(HEADERNAV_TOC)
+
         $(P String Mixins (not to be confused with
         $(DDLINK spec/template-mixin, Template Mixins, template mixins))
         enable string constants to be compiled as regular D code
         and inserted into the program.
+        Mixins can be used to generate:)
+
+        $(UL
+        $(LI Declarations)
+        $(LI Statements)
+        $(LI Types)
+        $(LI Expressions)
+        )
+        $(P
         Combining this with compile time manipulation of strings
         enables the creation of domain-specific languages.
-        )
-
-        $(P For example, here we can create a template that generates
-        a struct with the named members:
+        For example, we can generate a declaration `struct Foo { int bar; }`
+        using string constants for the identifiers:
         )
 
 ---
-template GenStruct(string Name, string M1)
+enum name = "Foo";
+enum m1 = "bar";
+mixin("struct " ~ name ~
+    "{ int " ~ m1 ~ "; }");
+
+enum f = Foo(2);
+static assert(f.bar == 2);
+---
+        $(P This can be encapsulated in an enum template:)
+---
+enum string genStruct(string name, string m1) =
+    "struct " ~ name ~ "{ int " ~ m1 ~ "; }";
+
+mixin(genStruct!("Foo", "bar"));
+---
+
+$(H2 $(LNAME2 multiple-args, Concatenation of Multiple Arguments))
+
+        $(P `mixin()` can also concatenate multiple arguments, converting any
+        non-strings to strings:)
+---
+enum name = "Foo";
+enum m1 = "bar";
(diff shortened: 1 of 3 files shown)


articles/mixin.dd
```diff
@@ -2,31 +2,96 @@ Ddoc
 
 $(D_S String Mixins,
 
+$(HEADERNAV_TOC)
+
         $(P String Mixins (not to be confused with
         $(DDLINK spec/template-mixin, Template Mixins, template mixins))
         enable string constants to be compiled as regular D code
         and inserted into the program.
+        Mixins can be used to generate:)
```
maxhaton, line 11: Perhaps work "or in place of" in here?

ntrel:
@maxhaton Thanks, I changed the wording as suggested.

I also:
* Added a section on 'Domain Specific Languages'.
* Moved the 'Generating Declarations' heading to the top.

articles/mixin.dd
```diff
@@ -2,31 +2,130 @@ Ddoc
...
+        ---
+        mixin(convertJS(q{
+            var msg = "Hello world!"
+            console.log(msg)
+            msg = 5
+            console.log(msg)
+            }));
+        ---
+        $(P `convertJS` would generate a string that would be compiled
+        to the following D code:)
```
dkorpel, line 113:
```suggestion
        $(P `convertJS` would take a string of JavaScript code and return a string with
        the following D code:)
```

articles/mixin.dd
```diff
@@ -2,31 +2,130 @@ Ddoc
...
+{
+    return mixin("x +", 1) * 7;  // same as `return (x + 1) * 7`
+}
 ---
 
+$(H2 $(LNAME2 dsl, Domain Specific Languages))
+
+        $(P
+        Combining compile-time function execution with `mixin()`
+        enables the use of domain-specific languages:)
```
dkorpel, line 103: Not a fan of this example. JavaScript is a general purpose programming language, a DSL is something like regex or SQL. Implementing `convertJS` would be a huge undertaking, and it doesn't reflect a real use case (or at least I don't see it). Why would you write this, instead of translating the JS offline and putting the result in your .d file?

I suggest using [pegged](https://github.com/PhilippeSigaud/Pegged) as an example.

dkorpel, line 103: Not a fan of this example. JavaScript is a general purpose programming language, a DSL is something like regex or SQL. Also, this doesn't reflect a real use case (or at least I don't see it). Why would you write this, instead of just writing D code?

I suggest using [pegged](https://github.com/PhilippeSigaud/Pegged) as an example.

dkorpel (changes requested):

articles/mixin.dd
maxhaton: I don't see why any of that really matters. If the reader can't discern "real use cases" from a demonstrative example then they're not going to get anything from any example.

Besides, JavaScript did literally start as the ultimate domain specific language a long time ago.

dkorpel: mixins are a confusing feature, and their working / purpose are [easily misunderstood](https://forum.dlang.org/post/hxoclgxkqatinlcnyzhr@forum.dlang.org). Consider the target audience for these articles: they're programmers who are new to D and might not have done any meta-programming beforehand. When you tell "`mixin("int y;")` is equivalent to `int y;`" the first question is "what does `mixin` do then?". 

The specification can be very dry in its explanations, but the article should help programmers understand what actual problems the feature solves. There's a tendency of new D users to complicate things with `mixin` when regular functions could be used, which is not helped by poor official examples.

PR #3210, opened 2022-02-09
[index.dd] Tweak character pair frequency example

ntrel:
Fix title: A static array of characters is not a tuple.
Split complex expression `aa[arr[i .. $][0 .. 2]]++;` into 2.

### DIFF
--- a/index.dd
+++ b/index.dd
@@ -204,20 +204,24 @@ void main()
 }
 ----
 )
-$(EXTRA_EXAMPLE_RUNNABLE Count frequencies of all 2-tuples,
+$(EXTRA_EXAMPLE_RUNNABLE Count frequencies of character pairs,
 ----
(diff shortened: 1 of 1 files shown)


index.dd
```diff
@@ -204,20 +204,24 @@ void main()
 }
 ----
 )
-$(EXTRA_EXAMPLE_RUNNABLE Count frequencies of all 2-tuples,
+$(EXTRA_EXAMPLE_RUNNABLE Count frequencies of character pairs,
 ----
 void main()
 {
     import std.stdio : writefln;
+    // associative array mapping char pairs to int
```
maxhaton, line 212:
```suggestion
    // An associative array mapping pairs of characters to int
```

dkorpel (approved):

PR #3215, opened 2022-02-11
Move ArrayInitializer grammar to expression.dd & improve docs

ntrel:
Rename ArrayInitializer to ArrayLiteral.
* ArrayInitializer is not referenced anywhere else.
* Existing ArrayLiteral grammar was wrong.

Tweak wording of array literal description.

**Update:** 
* Document `[`*AssignExpression* `:` *NonVoidInitializer*`]` syntax.
* Remove slicing from array literal section, just link to it (the example for each is the same).

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -1723,7 +1723,7 @@ $(GNAME ArrayMemberInitialization):
         int[2] sa = [1, 2];
         -------------
 
-    $(NOTE Slicing a dynamic array with statically known bounds also
+    $(NOTE Slicing a dynamic array with a statically known slice length also
         $(RELATIVE_LINK2 slice_to_static_array, allows conversion) to a static array.)
 
     $(P If any $(I ArrayMemberInitialization) is a
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -1716,37 +1716,15 @@ $(GNAME ArrayMemberInitialization):
         ---
 
     $(P By default, an array literal is typed as a dynamic array, but the element
-        count is known at compile time. So all array literals can be
-        implicitly converted to static array types. Slicing a dynamic array with
-        statically known bounds also allows conversion to a static array.)
+        count is known at compile time. So an array literal can be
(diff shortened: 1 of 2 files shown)


spec/expression.dd
```diff
@@ -1695,37 +1695,15 @@ $(GNAME ArrayMemberInitialization):
...
-            assert(arr == [1, 4, 5]);
-
-            //foo(arr[0 .. 3]); // cannot match length
-        }
+        int[2] sa = [1, 2];
         -------------
-        )
+
+    $(NOTE Slicing a dynamic array with statically known bounds also
+        $(RELATIVE_LINK2 slice_to_static_array, allows conversion) to a static array.)
```
MoonlightSentinel, line 1706: Implicit conversion is possible when the resulting *size* is known at compile time.  DMD recognizes e.g. this pattern:

```d
void foo(int[] array, int i)
{
    int[2] part = array[i .. i + 2];
    assert(part == [ 2, 3 ]);
}

void main()
{
    int[] arr = [ 1, 2, 3, 4 ];
    foo(arr, 1);
}
```

ntrel: BTW, I've added an example of this in #3231.

ntrel:
Rebased. Minor addition under Initializers: `See also $(GLINK VoidInitializer)`.

spec/expression.dd
```diff
@@ -1707,41 +1716,20 @@ $(GNAME ArrayLiteral):
         ---
 
     $(P By default, an array literal is typed as a dynamic array, but the element
-        count is known at compile time. So all array literals can be
-        implicitly converted to static array types. Slicing a dynamic array with
-        statically known bounds also allows conversion to a static array.)
+        count is known at compile time. So an array literal can be
```
dkorpel, line 1719:
```suggestion
        count is known at compile time. Therefore, an array literal can be
```

dkorpel (changes requested):

PR #3217, opened 2022-02-13
[spec/struct.dd] Improve struct initialization docs

ntrel:
Move `StructInitializer` grammar to struct.dd from declaration.dd. It's not a declaration and the docs for it are in struct.dd.
Add *Initialization* heading and reparent 5 headings.
Minor tweaks.

### DIFF
--- a/spec/struct.dd
+++ b/spec/struct.dd
@@ -140,7 +140,9 @@ S* p;     // ok, knowledge of members is not necessary
         $(LINK2 https://en.wikipedia.org/wiki/Opaque_pointer, PIMPL idiom).)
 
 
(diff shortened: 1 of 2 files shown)


spec/struct.dd
```diff
@@ -207,6 +221,8 @@ $(H2 $(LNAME2 default_union_init, Default Initialization of Unions))
         ---
         )
 
+        $(P It is an error to supply initializers for members other than the first one.)
```
dkorpel, line 224: This might change once we have named parameters, though the DIP doesn't mention unions.

dkorpel (approved):

PR #3218, opened 2022-02-14
Fix issue 15437 - documentation for typeof(someTemplate) == void

pbackus:
See also [issue 7947][1].

[1]: https://issues.dlang.org/show_bug.cgi?id=7947

### DIFF
--- a/spec/type.dd
+++ b/spec/type.dd
@@ -513,6 +513,15 @@ $(GNAME Typeof):
         typeof(S.foo) n;  // n is declared to be an int
         --------------------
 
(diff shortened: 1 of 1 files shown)


thewilsonator (approved):

ntrel:
So have we given up on making it an error?

> I briefly talked to both Andrei and Walter today at dconf about this. They both agreed, that calling typeof on a template should be an error (similiar to how calling typeof on a type is an error).

https://issues.dlang.org/show_bug.cgi?id=7947#c4

PR #3221, opened 2022-02-15
GNU version refers to hurd and not to GDC

etienne02:


spec/version.dd
```diff
@@ -224,7 +224,7 @@ $(H3 $(LEGACY_LNAME2 PredefinedVersions, predefined-versions, Predefined Version
         $(LONGTABLE_2COLS 0.6, Predefined Version Identifiers,
         $(THEAD Version Identifier, Description),
         $(TROW $(ARGS $(D DigitalMars)) , $(ARGS DMD (Digital Mars D) is the compiler))
-        $(TROW $(ARGS $(D GNU)) , $(ARGS GDC (GNU D Compiler) is the compiler))
+        $(TROW $(ARGS $(D GDC)) , $(ARGS GDC (GNU D Compiler) is the compiler))
```
maxhaton, line 227: GCC does not set version=GDC. It sets GNU as far as I've ever known

maxhaton (changes requested):

ibuclaw (changes requested):
This has been the case for nearly 20 years, and is not going to change.

maxhaton:
> This has been the case for nearly 20 years, and is not going to change.

@ibuclaw I think this PR was mistakenly fixing rather than aiming to change the status quo 

@etienne02 thanks for the effort, but as mentioned above this PR is bad.

PR #3225, opened 2022-02-18
[spec/template.dd] Improve TemplateSequenceParameter docs

ntrel:
Mention any argument passed to an alias parameter can be passed to a sequence parameter.
Show `AliasSeq` definition.
Link to CT sequences article.
`typeof` can be used on a *ValueSeq* to obtain a *TypeSeq*.
Mention a sequence is automatically expanded when referenced in a declaration or expression.
Split sequence `print` example into 2 - one with ValueSeq arguments, one with TypeSeq arguments.
A *TypeSeq* can similarly be used to declare lvalue variables.
Add *Sequence Operations* subheading & add example.
Explain how to construct a sequence from an existing one.
Mention AliasAssign for iterative construction.
Mention foreach on a sequence.

type.dd:
`typeof` can be used on a *ValueSeq* to obtain a *TypeSeq*.

### DIFF
--- a/spec/template.dd
+++ b/spec/template.dd
@@ -902,8 +902,8 @@ $(H4 $(LNAME2 seq-ops, Sequence Operations))
         $(DDSUBLINK spec/expression, index_expressions, indexing) an
         $(I AliasSeq) with `Seq[n]`. Indexes must be known at compile-time.
         The result is an lvalue when the element is a variable.)
-    $(LI Sub-sequences can be obtained using
-        $(DDSUBLINK spec/expression, slice_expressions, slicing) syntax.)
+    $(LI $(DDSUBLINK spec/expression, slice_expressions, Slicing)
+        produces a new sequence with a subset of the elements of the original sequence.)
     )
 
     $(SPEC_RUNNABLE_EXAMPLE_RUN
--- a/spec/template.dd
+++ b/spec/template.dd
@@ -932,7 +932,7 @@ $(H4 $(LNAME2 seq-ops, Sequence Operations))
     $(UL
     $(LI Construct a new sequence using the original sequence (or a slice of it) and additional elements before or after it.)
     $(LI Use $(DDSUBLINK spec/declaration, AliasAssign, Alias Assignment)
-        to build a new sequence iteratively.)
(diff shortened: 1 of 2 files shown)


spec/template.dd
```diff
@@ -818,60 +818,124 @@ $(GNAME TemplateSequenceParameter):
...
+$(H4 $(LNAME2 seq-ops, Sequence Operations))
+
+    $(UL
+    $(LI The number of elements in an $(I AliasSeq) can be retrieved with
+        the $(D .length) property.)
+    $(LI The $(I n)th element can be retrieved by
+        $(DDSUBLINK spec/expression, index_expressions, indexing) an
+        $(I AliasSeq) with `Seq[n]`. Indexes must be known at compile-time.
+        The result is an lvalue when the element is a variable.)
+    $(LI Sub-sequences can be obtained using
```
dkorpel, line 905: "Sub-sequence" is not defined as far as I can see, I would not give it a name and say that slicing a sequence produces a new sequence with a subset of the elements of the original sequence.

dkorpel (changes requested):

spec/template.dd
ntrel: OK, I've now used that wording.

spec/template.dd
```diff
@@ -818,60 +818,124 @@ $(GNAME TemplateSequenceParameter):
...
+    // slice first 3 elements
+    alias trio = nums[0 .. $-1];
+    // expand into an array literal
+    static assert([trio] == [1, 2, 3]);
+    ---
     )
 
-    $(P $(I AliasSeq)-s are static compile-time entities, there is no way
+    $(P $(I AliasSeq)s are static compile-time entities, there is no way
         to dynamically change, add, or remove elements either at compile-time or run-time.
```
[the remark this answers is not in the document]
ntrel: *AliasAssign* dynamically changes what an alias is aliasing. I updated the wording so it mentions iteratively building sequences plural. Each sequence assigned to an alias is not itself modified AIUI, rather each assignment creates a new sequence.

dkorpel (approved):

PR #3232, opened 2022-02-21
[spec] Improve concatenation docs

ntrel:
Split out from #3157.

Improve array concatenation example.
Add heading for *Array Appending*, separate relevant paragraphs.

Make *CatExpression* discuss a container, not just an array.
Mention concatenating a single value.

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -864,13 +864,12 @@ $(GNAME CatExpression):
     $(GLINK AddExpression) $(D ~) $(GLINK MulExpression)
 )
 
-    $(P A $(I CatExpression) concatenates arrays, producing
-        a dynamic array with the result. The arrays must be
-        arrays of the same element type. If one operand is an array
-        and the other is of that array's element type, that element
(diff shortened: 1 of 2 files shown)


spec/expression.dd
```diff
@@ -864,13 +864,12 @@ $(GNAME CatExpression):
...
 )
 
-    $(P A $(I CatExpression) concatenates arrays, producing
-        a dynamic array with the result. The arrays must be
-        arrays of the same element type. If one operand is an array
-        and the other is of that array's element type, that element
-        is converted to an array of length 1 of that element,
-        and then the concatenation is performed.
-    )
+    $(P A $(I CatExpression) concatenates a container's data with other data, producing
```
maxhaton, line 867: Is this actually what typical containers do? Perhaps it should be a BEST_PRACTICE

ntrel: What container doesn't it apply to? The new container doesn't necessarily have to be the same type as the LHS.

dkorpel (approved):

PR #3235, opened 2022-02-28
[spec/expression.dd] Improve IsExpression docs

ntrel:
Add Basic, Identifier and Parameter List form subheadings instead of an ordered list. Better for navigation and grouping relevant things together. Also use H5 for each kind of `is` expression, for anchors.

Move is(Bar == int) example above `== keyword` forms.
Add some missing docs for `== __vector`.
Add example for is(Type == keyword) and link to spec/traits.
Fix: `is` declaring an identifier also works in StaticAssert.

Make is(Type Identifier) example runnable. Also show that the identifier remains in scope even outside `static if` - see https://issues.dlang.org/show_bug.cgi?id=18609.
Split is ( Type Identifier : TypeSpecialization ) example in two.
Move is(Bar T == int) example above `== keyword` forms.
Explain `Identifier == keyword` form *result* for super, return, __parameters.

Split Parameter List form example into 3 and add titles. Define `struct Tuple(T...)` instead of importing std.tuple for clarity. Use 2 static asserts instead of `assert(0)`.

css:
Add missing h5 style, fixes text size too small.

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -2270,9 +2270,14 @@ $(GNAME TypeSpecialization):
 )
 
     $(P $(I IsExpression)s are evaluated at compile time and are
-        used for checking for valid types, comparing types for equivalence,
-        determining if one type can be implicitly converted to another,
-        and deducing the subtypes of a type.
+        used for:)
+    $(UL
+    $(LI checking for valid types)
+    $(LI comparing types for equivalence)
+    $(LI determining if one type can be implicitly converted to another)
+    $(LI deducing the subtypes of a type)
+    )
+    $(P
         The result of an $(I IsExpression) is a boolean of value `true`
         if the condition is satisfied. If the condition is not satisfied,
         the result is a boolean of value `false`.
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -2283,30 +2283,23 @@ $(GNAME TypeSpecialization):
         If it is not semantically correct, the condition is not satisfied.
     )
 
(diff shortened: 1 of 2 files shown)


spec/expression.dd
```diff
@@ -2335,20 +2329,33 @@ void foo()
         writeln("not satisfied");
 }
 -------------
-        )
 
-        $(LI $(D is $(LPAREN)) $(I Type) $(D ==) $(I TypeSpecialization) $(D $(RPAREN))$(BR)
+        $(H5 $(LNAME2 is-type-equal, $(D is $(LPAREN)) $(I Type) $(D ==) $(I TypeSpecialization) $(D $(RPAREN))))
```
maxhaton, line 2333: These seem (for me at least) to get listed without a number (the trailing paragraph is then numbered)

ntrel: Yes, I think the css code for that is in the `BODY_PREFIX` macro in `spec/spec.ddoc`. I'm not sure how to update it for h5.

spec/expression.dd
```diff
@@ -2357,28 +2369,37 @@ void foo()
...
 
-        then the condition is satisfied if $(I Type) is one of those. $(DDSUBLINK
+        then the condition is satisfied if $(I Type) is one of those.
+        )
+---
+static assert(is(Object == class));
+
+int i;
+static assert(!is(i == const));
+---
```
dkorpel, line 2379: Not runnable?

dkorpel (approved):

PR #3244, opened 2022-03-06
spec/lex.dd: Remove HexString from grammar

ibuclaw:
Hex strings have been obsolete since 2.086.

### DIFF
--- a/spec/lex.dd
+++ b/spec/lex.dd
@@ -291,7 +291,6 @@ $(GNAME StringLiteral):
     $(GLINK AlternateWysiwygString)
     $(GLINK DoubleQuotedString)
 
-    $(GLINK HexString)
     $(GLINK DelimitedString)
     $(GLINK TokenString)
 )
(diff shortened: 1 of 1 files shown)


maxhaton (approved):
LGTM, haven't checked if the final website is sound yet

ntrel (changes requested):
I suggest at least adding a link in the `\x` [string escape docs](https://github.com/dlang/dlang.org/blob/1d178c651e7f3a26ab61469ac99e69a1970aacb3/spec/lex.dd#L573):
```
See also: $(REF hexString, std,conv) 
```

ntrel:
> I suggest at least adding a link in the `\x` [string escape docs](https://github.com/dlang/dlang.org/blob/1d178c651e7f3a26ab61469ac99e69a1970aacb3/spec/lex.dd#L573):
> 
> ```
> See also: $(REF hexString, std,conv) 
> ```

Actually I can do it in another pull, I'm working on the escape sequence docs now anyway ;-)

ntrel:
Now done in #3245.

PR #3251, opened 2022-03-10
Add __traits(parameters) to language spec

pbackus:
Implementation PR: https://github.com/dlang/dmd/pull/13071

### DIFF
--- a/spec/traits.dd
+++ b/spec/traits.dd
@@ -80,6 +80,7 @@ $(GNAME TraitsKeyword):
     $(RELATIVE_LINK2 compiles, $(D compiles))
     $(RELATIVE_LINK2 toType, $(D toType))
     $(RELATIVE_LINK2 initSymbol, $(D initSymbol))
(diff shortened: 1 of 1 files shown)


adamdruppe (approved):

maxhaton (approved):

ntrel (commented):
Can you move this under *Function Parameter Traits* rather than the *Symbol Traits* heading please?

spec/traits.dd
```diff
@@ -1772,6 +1773,71 @@ $(H3 $(GNAME initSymbol))
...
+        {
+            return add(__traits(parameters));
+            // equivalent to;
+            //return add(x, y);
+        }
+
+        int nestedExample(int x)
+        {
+            // outer function's parameters
+            static assert(typeof(__traits(parameters)).length == 1);
```
ntrel, line 1805: Nit: I think we don't need typeof to access the sequence length. Also below.

PR #3255, opened 2022-03-12
Remove special case of __traits(parameters) in foreach

MoonlightSentinel:
The trait now consistently returns the function parameters

See dlang/dmd#13798

### DIFF
--- a/spec/traits.dd
+++ b/spec/traits.dd
@@ -1037,11 +1037,6 @@ $(H3 $(GNAME parameters))
         $(P If the function is nested, the parameters returned are those of the
         inner function, not the outer one.)
 
-        $(P When used inside a $(DDSUBLINK spec/statement, foreach-statement,
-        `foreach` statement) that calls an $(DDSUBLINK spec/statement,
-        foreach_over_struct_and_classes, `opApply` overload), the parameters
-        returned are those of the delegate passed to `opApply`.)
-
         ---
         int add(int x, int y)
         {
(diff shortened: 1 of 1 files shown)


spec/traits.dd
```diff
@@ -1083,13 +1078,12 @@ $(H3 $(GNAME parameters))
...
             foreach(idx; 0..5)
             {
-                // foreach does not call opApply
                 static assert(is(typeof(__traits(parameters)) == AliasSeq!(C, int)));
             }
             foreach(idx, elem; c)
             {
-                // foreach calls opApply
-                static assert(is(typeof(__traits(parameters)) == AliasSeq!(size_t, C)));
+                // Sees past the delegate passed to opApply
```
maxhaton, line 1085:
```suggestion
                // __traits(parameters) sees past the delegate passed to opApply
```

pbackus:
Would it make more sense to remove the `foreach` example entirely, since it's no longer a special case?

maxhaton:
> Would it make more sense to remove the `foreach` example entirely, since it's no longer a special case?



> Would it make more sense to remove the `foreach` example entirely, since it's no longer a special case?

I vote for keeping it because it confirms that it doesn't to this to the experienced user, and indicates that opApply exists to the uninitiated

MoonlightSentinel:
> I vote for keeping it because it confirms that it doesn't to this to the experienced user

Agreed. The spec explicitly states that the foreach body is passed as a delegate to `opApply`, so it's better to be explicit IMO.

maxhaton (approved):

PR #3256, opened 2022-03-20
spec/template.dd: correct misleading IFTI example

pbackus:
Contrary to what was previously stated in the text, the example did not
fail because the template parameter T was specialized. Instead, it
failed because the T used in the template parameter list referred to a
different type than the T used in the function parameter list, and the
former could not be deduced from the latter.

To make the example clearer, the second T has been changed to U, and the
text has been updated accordingly.


ljmf00 (approved):
Adding `U` makes it much more clear :)

Geod24 (approved):

PR #3258, opened 2022-03-20
spec/iasm.dd: Move GCC syntax to the bottom of the page

ibuclaw:
The generated contents table erroneously includes a dead "Other" subcategory because of the "$(COMMENT)" section that follows for DMD-style inline asm.

image


maxhaton (approved):
LGTM

Geod24 (approved):

PR #3259, opened 2022-03-22
Fix issue 22027 - `inout` doesn't imply return

dkorpel:
DMD PR: https://github.com/dlang/dmd/pull/12689


dkorpel:
DMD PR merged 
@RazvanN7

PR #3260, opened 2022-03-22
Add UFCS to style guide

dkorpel:
@atilaneves 
This came up in https://github.com/dlang/phobos/pull/8410#discussion_r826115306

I wrote down what I think is the norm in dmd, Phobos, druntime, and the spec. Style is very subjective, so this initial description deliberately does not go into too much detail to prevent nailing it down too hard.

### DIFF
--- a/dstyle.dd
+++ b/dstyle.dd
@@ -243,7 +243,7 @@ $(H3 $(LNAME2 properties, Properties))
 $(H3 $(LNAME2 property-syntax, Property syntax))
 
 $(P Do not use $(LINK2 spec/function.html#pseudo-member, UFCS) or
-$(LINK2 spec/function.html#optional-parenthesis, Optional Parentheses) outside of its intended use cases.
+$(LINK2 spec/function.html#optional-parenthesis, Optional Parentheses) outside of their intended use cases.
 Omitting parentheses is useful in generic code that does not care whether a member is a field or
 a function, for example the range primitive `front`.
 It can also make a chain of range functions more compact.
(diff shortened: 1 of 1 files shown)


dstyle.dd
```diff
@@ -240,6 +240,31 @@ $(H3 $(LNAME2 properties, Properties))
         properties should be nouns, just like if they were member variables.
         Getter properties should $(I not) alter state.)
 
+$(H3 $(LNAME2 property-syntax, Property syntax))
+
+$(P Do not use $(LINK2 spec/function.html#pseudo-member, UFCS) outside of its intended use cases.
```
atilaneves, line 245: I'm not sure why this is conflating UFCS with optional parens.

dkorpel: I didn't see the spec had a section on optional parens, I thought that also fell under UFCS

dstyle.dd
```diff
@@ -242,7 +242,8 @@ $(H3 $(LNAME2 properties, Properties))
 
 $(H3 $(LNAME2 property-syntax, Property syntax))
 
-$(P Do not use $(LINK2 spec/function.html#pseudo-member, UFCS) outside of its intended use cases.
+$(P Do not use $(LINK2 spec/function.html#pseudo-member, UFCS) or
+$(LINK2 spec/function.html#optional-parenthesis, Optional Parentheses) outside of its intended use cases.
```
atilaneves, line 246: "outside of their"

atilaneves (approved):

PR #3261, opened 2022-03-22
rotate gpg key

MartinNowak:
- updated keyring to add new gpg key
- d-keyring.gpg was signed with new key as the old one already expired
  (might cause a validation warning when updating the d-keyring)


MartinNowak:
> * d-keyring.gpg was signed with new key as the old one already expired

Maybe you could sign that instead @ibuclaw?

ibuclaw (approved):

PR #3262, opened 2022-03-22
Update Ref Return Scope Parameters to new design

dkorpel:
This removes the old design where it matters whether the function returns by `ref`, and replaces it with the new design where it matters if `return scope` appears together.

Reference: 
[Issue 22541 - DIP1000: Resolve ambiguity of ref-return-scope parameters](https://issues.dlang.org/show_bug.cgi?id=22541)
[Issue 22790 - ref-return-scope is always ref-return, scope, unless return-scope appear in that order](https://issues.dlang.org/show_bug.cgi?id=22790)

This is not final, like [Paul Backus proposed](https://forum.dlang.org/post/qqfqdsdhqwrrckturpal@forum.dlang.org) and [Walter mentioned](https://issues.dlang.org/show_bug.cgi?id=22541#c4):

> Where I'd like to go is only allow "return ref" or "return scope".

But this brings it up to date with where we are now.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -1942,7 +1942,7 @@ $(H3 $(LNAME2 ref-return-scope-parameters, Ref Return Scope Parameters))
         When a parameter is passed by `ref` and has both the `return` and `scope` attributes,
         it gets $(LINK2 #return-scope-parameters, `return scope`) semantics if and only if the `return` and `scope`
         keywords appear adjacent to each other, in that order.
-        In other cases, the parameter gets $(LINK2 #return-ref-parameters, `return ref`) semantics
+        In all other cases, the parameter has $(LINK2 #return-ref-parameters, `return ref`) semantics
         and regular $(LINK2 #scope-parameters, `scope`) semantics:)
 
 ---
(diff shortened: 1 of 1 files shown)


spec/function.dd
```diff
@@ -1937,78 +1937,91 @@ class C
 
 $(H3 $(LNAME2 ref-return-scope-parameters, Ref Return Scope Parameters))
 
-        $(P Parameters marked as `ref return scope` come in two forms:)
+        $(P It is not possible to have both `return ref` and `return scope` semantics
+        for the same parameter.
+        When a parameter is passed by `ref` and has both the `return` and `scope` attributes,
```
maxhaton, line 1942:
```suggestion
        When a parameter is passed by `ref` and has both the `return` and `scope` storage classes,
```

spec/function.dd
```diff
@@ -1937,78 +1937,91 @@ class C
...
+        it gets $(LINK2 #return-scope-parameters, `return scope`) semantics if and only if the `return` and `scope`
+        keywords appear adjacent to each other, and in that order.
```
maxhaton, line 1944:
```suggestion
        keywords appear adjacent to each other, in that order.
```

spec/function.dd
```diff
@@ -1937,78 +1937,91 @@ class C
...
+        In other cases, the parameter gets $(LINK2 #return-ref-parameters, `return ref`) semantics
```
maxhaton, line 1945:
```suggestion
        In all other cases, the parameter has $(LINK2 #return-ref-parameters, `return ref`) semantics
```

RazvanN7 (approved):
The changes to the spec are ok in the sense that they match the implementation, but honestly having semantics depend on the order of return and ref is just bad design.

PR #3263, opened 2022-03-23
Spec changes for static array .tupleof

Yoplitein:
See dlang/dmd#13870


spec/arrays.dd
```diff
@@ -580,6 +580,7 @@ Returns an array literal with each element of the literal being the $(D .init) p
         $(TROW $(D .ptr), Returns a pointer to the first element of the array.)
         $(TROW $(D .dup), Create a dynamic array of the same size and copy the contents of the array into it. The copy will have any immutability or const stripped. If this conversion is invalid the call will not compile.)
         $(TROW $(D .idup), Create a dynamic array of the same size and copy the contents of the array into it. The copy is typed as being immutable. If this conversion is invalid the call will not compile.)
+        $(TROW $(D .tupleof), Returns an $(DDSUBLINK spec/template, variadic-templates, expression sequence) of all elements of the array$(COMMA) as with $(D .tupleof) for structs and $(DDSUBLINK spec/class, class_properties, classes). May only be used on values$(COMMA) not types.)
```
ntrel, line 583: BTW you can wrap the description with `$(ARGS description)` to avoid the comma macro.

I don't think we need to mention it doesn't apply to types here specifically, because the other 3 properties above also don't apply to types. Perhaps we can add a THEAD for instance only properties (though that's not necessary to do for your pull).

ntrel (approved):

dkorpel (approved):

PR #3272, opened 2022-04-06
[spec/class] Improve tupleof docs

ntrel:
Undo recent change to properties heading as .tupleof is a type property too (oops).
Update template link to more relevant anchor.
tupleof doesn't include static fields.
Show examples of .tupleof on class type.
Expand instance example.

### DIFF
--- a/spec/class.dd
+++ b/spec/class.dd
@@ -184,8 +184,8 @@ $(SPEC_RUNNABLE_EXAMPLE_RUN
 ---
 class Foo { int x; long y; }
 
-static assert(Foo.tupleof[0].stringof == "x");
-static assert(typeof(Foo.tupleof).stringof == `(int, long)`);
+static assert(__traits(identifier, Foo.tupleof[0]) == "x");
+static assert(is(typeof(Foo.tupleof)[1] == long));
 
 void main()
 {
--- a/spec/class.dd
+++ b/spec/class.dd
@@ -207,6 +207,8 @@ void main()
 ---
 )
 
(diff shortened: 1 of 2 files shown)


spec/class.dd
```diff
@@ -184,16 +184,25 @@ $(SPEC_RUNNABLE_EXAMPLE_RUN
 ---
 class Foo { int x; long y; }
 
+static assert(Foo.tupleof[0].stringof == "x");
+static assert(typeof(Foo.tupleof).stringof == `(int, long)`);
```
dkorpel, line 188: Assuming an exact result for `.stringof` is bad practice, since the string representation for a type or expression can vary per implementation.

dkorpel (changes requested):

spec/class.dd
```diff
@@ -171,11 +171,11 @@ void test(Foo foo)
...
 ------
 
-$(H2 $(LNAME2 class_properties, Class Instance Properties))
+$(H2 $(LNAME2 class_properties, Class Properties))
 
         $(P The $(D .tupleof) property is an
-        $(DDSUBLINK spec/template, variadic-templates, expression sequence)
-        of all the fields in the class, excluding the hidden fields and
+        $(DDSUBLINK spec/template, homogeneous_sequences, lvalue sequence)
+        of all the non-static fields in the class, excluding the hidden fields and
```
maxhaton, line 178: Definition for hidden fields?

spec/class.dd
ntrel: OK, updated.

maxhaton (approved):

PR #3276, opened 2022-04-11
[spec/attribute] Document `final` attribute

ntrel:
Methods of a final class are always final (not by default).
Explain `final` in Virtual Functions section and add anchor.

### DIFF
--- a/spec/attribute.dd
+++ b/spec/attribute.dd
@@ -18,7 +18,7 @@ $(GNAME Attribute):
     $(RELATIVE_LINK2 static, $(D static))
     $(RELATIVE_LINK2 linkage, $(D extern))
     $(RELATIVE_LINK2 abstract, $(D abstract))
-    $(D final)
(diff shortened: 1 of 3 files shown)


dkorpel (approved):

spec/attribute.dd
```diff
@@ -664,6 +665,15 @@ $(P
         Note that the class is still abstract and cannot be instantiated directly.
 )
 
+$(H2 $(LNAME2 final, `final` Attribute))
+
+$(UL
+$(LI A class can be declared $(DDSUBLINK spec/class, final, `final`) to prevent
+    subclassing.)
+$(LI A class method can be declared $(DDSUBLINK spec/function, final, `final`)
```
maxhaton, line 673: Note about interfaces with final methods?

ntrel: Now added.

PR #3277, opened 2022-04-11
[spec] Improve @disable docs

ntrel:
Refine link to disabled default constructor.
Add anchor for disabled copy constructor & example.


ntrel:
Fixed ddoc error.

PR #3279, opened 2022-04-17
Document ImportC support for typeof operator

Yoplitein:
See dlang/dmd#14000

### DIFF
--- a/spec/importc.dd
+++ b/spec/importc.dd
@@ -288,6 +288,18 @@ _Static_assert(sizeof(A) == 1, "A should be size 1");
     $(P Arrays can have `register` storage class, and may be enregistered by the compiler. C11 6.3.2.1-3)
 
 
+    $(H3 $(LNAME2 typeof, typeof Operator))
+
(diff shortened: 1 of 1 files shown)


spec/importc.dd
```diff
@@ -288,6 +288,11 @@ _Static_assert(sizeof(A) == 1, "A should be size 1");
     $(P Arrays can have `register` storage class, and may be enregistered by the compiler. C11 6.3.2.1-3)
 
 
+    $(H3 $(LNAME2 typeof, typeof Operator))
+
+    $(P The `typeof` operator may be used in place of a type specifier, upon expressions or type names.)
```
dkorpel, line 293: I don't get "upon expressions or type names"

Yoplitein: The intent was to document that C's typeof accepts types as well as expressions, unlike D's typeof. I'll make it less ambiguous with a grammar.

dkorpel (approved):

PR #3280, opened 2022-04-17
Document `scope` inference from `pure`

dkorpel:
Assumes the changes of: https://github.com/dlang/dmd/pull/13993

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -2015,6 +2015,48 @@ S test6(ref return scope int* p)
 }
 ---
 
+$(H3 $(LNAME2 pure-scope-inference, Inferred `scope` parameters in `pure` functions))
+
+    When a parameter is not marked or inferred `scope`, it may still be `@safe` to assign it a `scope` pointer in a function call.
+    The following conditions need to be met:
+
+    $(UL
+    $(LI The function is $(RELATIVE_LINK2 pure-functions, `pure`), hence the argument cannot be assigned to a global variable)
+    $(LI The function is $(RELATIVE_LINK2 nothrow-functions, `nothrow`), hence the argument cannot be assigned to a thrown `Exception` object)
+    $(LI None of the other parameters have mutable indirections, hence the argument cannot be assigned to a longer-lived variable)
+    )
+
+    Then, the parameter is still treated as `scope` or `return scope` depending on the return type of the function:
+    $(UL
+    $(LI If the function returns by `ref` or has a return type that contains pointers, the argument could be returned, so it is treated as `return scope`)
+    $(LI Otherwise, the argument cannot escape the function, so it is treated as `scope`)
+    )
+
+---
+@safe:
+
+int dereference(int* x) pure nothrow;
+int* identity(int* x) pure nothrow;
+int* identityThrow(int* x) pure;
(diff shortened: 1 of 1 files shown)


spec/function.dd
```diff
@@ -2015,6 +2015,48 @@ S test6(ref return scope int* p)
 }
 ---
 
+$(H4 $(LNAME2 pure-scope-inference, Inferred `scope` parameters in `pure` functions))
```
ntrel, line 2018: I think this should be H3 as it's not a subheading of 'return ref scope'.
```suggestion
$(H3 $(LNAME2 pure-scope-inference, Inferred `scope` parameters in `pure` functions))
```

spec/function.dd
```diff
@@ -2015,6 +2015,48 @@ S test6(ref return scope int* p)
...
+
+    When a parameter is not marked or inferred `scope`, it may still be `@safe` to assign it a `scope` pointer in a function call.
+    The following conditions need to be met:
+
+    $(UL
+    $(LI The function is $(RELATIVE_LINK2 pure-functions, `pure`), hence the argument cannot be assigned to a global variable)
+    $(LI The function is $(RELATIVE_LINK2 nothrow-functions, `nothrow`), hence the argument cannot be assigned to a thrown `Exception` object)
+    $(LI None of the other parameters have mutable indirections, hence the argument cannot be assigned to a longer-lived variable)
```
ntrel, line 2026: Just a query: what about that feature where if the function returns void then `return` applies to the first parameter? I don't know if that's still supported (if so could be added separately).

spec/function.dd
```diff
@@ -2015,6 +2015,48 @@ S test6(ref return scope int* p)
...
+    $(LI If the function returns by `ref` or has a return type that contains pointers, the argument could be returned, so it is treated as `return scope`)
+    $(LI Otherwise, the argument cannot escape the function, so it is treated as `scope`)
+    )
+
+---
+@safe:
+
+int* identity(int* x) pure nothrow;
+int* identityThrow(int* x) pure;
+int dereference(int* x) pure nothrow;
```
ntrel, line 2040: Would be nicer if `dereference` was declared as the first function because it's called first, and as the others are in order.

ntrel (approved):

spec/function.dd
dkorpel: `return` semantics when assigning to the first parameter is still supported, but `pure` -> `scope` inference doesn't consider it yet: it sees the first parameter is mutable, and won't infer (return) scope for the other parameters. This is a possible enhancement to `pure` -> `scope` inference.

spec/function.dd
dkorpel: Done

spec/function.dd
dkorpel: Good catch. I haven't seen this rendered yet because my local build of dlang.org is suddenly acting up with linking errors.

ntrel: Thanks. Just FYI you can click the Details link on the CyberShadow test, then click New for function.html to see the changes rendered.

dkorpel: I didn't know that, very useful.

RazvanN7 (approved):

PR #3284, opened 2022-04-18
Improve spec of `scope` attribute

dkorpel:


spec/attribute.dd
```diff
@@ -617,27 +617,117 @@ auto i = 6.8;   // declare i as a double
...
+        (Resource Acquisition Is Initialization) protocol.
         This means that the destructor for an object is automatically called when the
         reference to it goes out of scope. The destructor is called even
         if the scope is exited via a thrown exception, thus $(D scope)
         is used to guarantee cleanup.
 )
 $(P
-        If there is more than one $(D scope) variable going out of scope
+        When a class is constructed with `new` and assigned to a local `scope` variable,
+        it may be allocated on the stack and permitted in a `@nogc` context.
```
maxhaton, line 697: Mention destruction i.e. deterministic lifetime

spec/attribute.dd
```diff
@@ -617,27 +617,117 @@ auto i = 6.8;   // declare i as a double
...
-        RAII (Resource Acquisition Is Initialization) protocol.
+        The $(D scope) attribute is used for parameters or local variables with a type that contains low-level pointers.
+        It signifies that the variable's pointer values will not escape the function
+        (in case of a $(DDSUBLINK spec/function, scope-parameters, scope parameters)) or
+        $(DDSUBLINK spec/statement, scope-statement, scope statement) that the variable is declared in,
+        for example by returning it or assigning it to a variable outside of its scope.
+)
+
+$(P
+        When applied to a global variable or a variable with a type that has no pointers, `scope` does nothing and is ignored.
```
maxhaton, line 628:
```suggestion
        When applied to a global variable or a variable with a type that has no pointers, `scope` is ignored.
```

spec/attribute.dd
```diff
@@ -617,27 +617,117 @@ auto i = 6.8;   // declare i as a double
 $(H2 $(LNAME2 scope, $(D scope) Attribute))
 
 $(P
-        The $(D scope) attribute is used for local variables and implements the
-        RAII (Resource Acquisition Is Initialization) protocol.
+        The $(D scope) attribute is used for parameters or local variables with a type that contains low-level pointers.
```
maxhaton, line 620: Low-level?

spec/attribute.dd
```diff
@@ -617,27 +617,117 @@ auto i = 6.8;   // declare i as a double
...
+        It signifies that the variable's pointer values will not escape the function
```
maxhaton, line 621: This section is clumsily worded, I'll be more specific when I'm at a proper keyboard.

spec/attribute.dd
dkorpel: It's in the paragraph above

spec/attribute.dd
PetarKirov: How about:

```suggestion
        The $(D scope) attribute is used for parameters or local variables of types that contains indirections.
```

(See https://dlang.org/phobos/std_traits#.hasIndirections)

spec/attribute.dd
RazvanN7: I suggest using smaller sentences. Also, it should be clear that any variable may be attributed with `scope`, however, the checks are performed only for types that contain indirections. The way this paragraph is worded makes it sound that it is an error to use scope on types that do not contain indirections.

RazvanN7: For example: "The scope attribute is used for parameters of local variables. If the type of the variable does not contain any indirections, scope is ignored. Otherwise, the compiler will mechanically check that the variable does not escape the scope it is declared in. A variable may escape its scope by being returned or being assigned to a variable outside of its scope".

spec/attribute.dd
```diff
@@ -617,27 +617,117 @@ auto i = 6.8;   // declare i as a double
...
-        RAII (Resource Acquisition Is Initialization) protocol.
+        The $(D scope) attribute is used for parameters or local variables of types that contain indirections.
+        It signifies that the variable's pointer values will not escape the function
+        (in case of a $(DDSUBLINK spec/function, scope-parameters, scope parameters)) or
+        $(DDSUBLINK spec/statement, scope-statement, scope statement) that the variable is declared in,
+        for example by returning it or assigning it to a variable outside of its scope.
+)
+
+$(P
+        When applied to a global variable or a variable with a type that has no pointers, `scope` is ignored.
```
RazvanN7, line 628: Ok, now I see that you mention this. I would state this up-front.

dkorpel: Done

RazvanN7 (approved):

PR #3289, opened 2022-04-30
Use DuckDuckGo for search

LorenDB:
Many people, including myself, distrust Google and stay away from it as much as possible. To that end, I propose that search be routed through DuckDuckGo to make sure that everybody can search the site with peace of mind from a privacy standpoint.

### DIFF
--- a/dpl-docs/views/layout.dt
+++ b/dpl-docs/views/layout.dt
@@ -128,7 +128,7 @@ html(lang='en-US')
             a.expand-toggle(href="#{root_dir}search.html", title="Search")
               span Search
             #search-box
-              form(method='get', action='https://google.com/search', autocomplete='off')
+              form(method='get', action='https://duckduckgo.com/', autocomplete='off')
                 input#domains(type='hidden', name='domains', value='dlang.org')
                 |             
                 input#sourceid(type='hidden', name='sourceid', value='google-search')


SixthDot:
I'd just use their static address instead of trusting the DNS service

maxhaton:
If the results are as good then ok I guess. If the results are worse then it's a no.

LorenDB:
https://www.google.com/search?domains=dlang.org&q=std.algorithm&sitesearch=dlang.org
https://duckduckgo.com/?q=std.algorithm+site%3Adlang.org&t=opera&ia=web

ljmf00 (approved):

dkorpel:
> https://www.google.com/search?domains=dlang.org&q=std.algorithm&sitesearch=dlang.org 
https://duckduckgo.com/?q=std.algorithm+site%3Adlang.org

This is a very unambiguous query, let's try 'map':

https://www.google.com/search?domains=dlang.org&q=map&sitesearch=dlang.org
https://duckduckgo.com/?q=map+site%3Adlang.org

DuckDuckGo only lists std.iteration.map as fourth result, under the spec for "Interfaces" which is the third result. I'm not sure the search quality is really on par.

ljmf00:
> > https://www.google.com/search?domains=dlang.org&q=std.algorithm&sitesearch=dlang.org
> > https://duckduckgo.com/?q=std.algorithm+site%3Adlang.org
> 
> This is a very unambiguous query, let's try 'map':
> 
> https://www.google.com/search?domains=dlang.org&q=map&sitesearch=dlang.org https://duckduckgo.com/?q=map+site%3Adlang.org
> 
> DuckDuckGo only lists std.iteration.map as fourth result, under the spec for "Interfaces" which is the third result. I'm not sure the search quality is really on par.

It seems that `dlang.org` is using Google Webmaster but not Bing Webmaster tools? Can someone confirm that `dlang.org` is being explicitly observed by Webmaster tools?

There might be another obvious reasons for DuckDuckGo not organizing the articles by relevance correctly. Our `<meta>` tags are too generic. Our `description` and `title` fields are the same for every page, making it harder for search engines to do SEO correctly. You can also provide locale and other interesting extra metadata that some search engines look for, e.g., OpenGraph protocol, https://ogp.me/. We should also generate relevant `keywords`. Our `keywords` field is really bad, and, again, shouldn't be generic to the website, but specific to the article.

See the example for this page: https://dlang.org/library/std/algorithm/iteration/map.html and compare it with other references from other languages like Rust, Go,...

Another important thing for SEO on documentation is to have a fairly standard ToC (table of contents), which we have, but is harder to read/parse by a search engine. It is preferable to have a `ol`/`li` list of available content, and it can be hidden if needed.

Bolpat:
> DuckDuckGo only lists std.iteration.map as fourth result, under the spec for "Interfaces" which is the third result. I'm not sure the search quality is really on par.

Not anymore. It lists `std.iteration.map as` first and the associative arrays (known as `unordered_map` in C++) spec as second result. Quite good IMO. Ordering of search results can change over time.

One big advantage with DuckDuckGo is non-personalized results. Google shows different results to different people searching the same thing at the same moment. Of course, *over time* results can (and should) change because websites do change. With DuckDuckGo, I, you, or anyone else, can input something and observe the quality of the results. It can even be automated (search input: `site:dlang.org map`; where does `std.iteration.map` appear?) With Google, one might get good or bad results because of results being personalized.

WebFreak001:
seems like the search index is fine now - should we just merge this?

dkorpel:
@mdparker What do you think of this?

mdparker:
@dkorpel At first blush, I don't have any objection to it other than personally preferring Google (I *like* having personalized results). I never use DuckDuckGo, so can't make an informed opinion of it. Is there a demand for the change other than one person?

Bolpat:
@mdparker, at least, add me. As far as I can tell, this change is neutral for most people and mildly valuable for those with privacy concerns. As stupid as it sounds, we could just merge this and see if people complain.

dkorpel:
> Is there a demand for the change other than one person?

I don't know, we could poll it in the news group

WebFreak001:
I think it's a good idea to change to DuckDuckGo for the short term - long term I would love if we could host at least our phobos and spec search entirely ourself, so we can keep the index up-to-date, avoid 404s, give smarter previews, not depend on external services and APIs and also bundle this with the offline DMD / Phobos documentation that's bundled with the compiler. I think there are a lot of positive aspects we can have here.

For how such an offline search example could look like, check out the search on my dub docs rewrite at the top right: https://docs.webfreak.org/

While you need JS for the search to work, we could also have a very simple search server for no-js / online searches available and could also collect information about which pages are searched for with which keywords.

ibuclaw:
Why not just go straight to cutting the middle man?  All good documentation generators support builtin searching (*ahem* ddox).

adamdruppe:
dpldocs.info/File

adamdruppe:
oops i clicked the wrong button but like we have plenty of phobos/druntime api searches

the main benefit of the site search i think is that it includes other things too

PR #3290, opened 2022-05-02
Add __traits(classInstanceAlignment)

kinke:
For https://github.com/dlang/dmd/pull/14063.

### DIFF
--- a/spec/traits.dd
+++ b/spec/traits.dd
@@ -73,6 +73,7 @@ $(GNAME TraitsKeyword):
     $(RELATIVE_LINK2 parent, $(D parent))
     $(RELATIVE_LINK2 child, $(D child))
     $(RELATIVE_LINK2 classInstanceSize, $(D classInstanceSize))
+    $(RELATIVE_LINK2 classInstanceAlignment, $(D classInstanceAlignment))
     $(RELATIVE_LINK2 getVirtualIndex, $(D getVirtualIndex))
     $(RELATIVE_LINK2 allMembers, $(D allMembers))
     $(RELATIVE_LINK2 derivedMembers, $(D derivedMembers))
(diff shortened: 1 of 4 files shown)


kinke:
(The Phobos library trait is buggy: https://g.godbolt.org/z/c4MWov45f)

spec/traits.dd
```diff
@@ -580,6 +581,17 @@ $(H3 $(GNAME classInstanceSize))
...
+$(H3 $(GNAME classInstanceAlignment))
+
+        $(P Takes a single argument, which must evaluate to either
+        a class type or an expression of class type.
+        The result
+        is of type $(CODE size_t), and the value is the alignment
+        of a runtime instance of the class type.
+        It is based on the static type of a class, not the
+        polymorphic type.
+        )
```
ljmf00, line 593: I would suggest 80 char text wrapping.

```suggestion
        $(P Takes a single argument, which must evaluate to either a class type
        or an expression of class type.
        The result is of type $(CODE size_t), and the value is the alignment of
        a runtime instance of the class type. It is based on the static type of
        a class, not the polymorphic type.
        )
```

kinke: [The block is a copy-pasta from classInstanceSize above.]

ljmf00: Right, we can refactor those in a subsequent PR.

ljmf00 (approved):
LGTM, although I don't have approval rights

spec/traits.dd
```diff
@@ -580,6 +581,17 @@ $(H3 $(GNAME classInstanceSize))
         polymorphic type.
         )
 
+$(H3 $(GNAME classInstanceAlignment))
+
+        $(P Takes a single argument, which must evaluate to either
+        a class type or an expression of class type.
+        The result
+        is of type $(CODE size_t), and the value is the alignment
+        of a runtime instance of the class type.
```
maxhaton, line 590:
```suggestion
        of an instance of the class type.
```
Runtime is probably unnecessary here

spec/traits.dd
```diff
@@ -580,6 +581,17 @@ $(H3 $(GNAME classInstanceSize))
...
+        It is based on the static type of a class, not the
```
maxhaton, line 591: Is dynamic type a concept we use in the spec elsewhere, it's quite clunky

maxhaton (approved):

PR #3293, opened 2022-05-10
clarify return-ref combined with scope parameters

WalterBright:
Discussion in comments to https://github.com/dlang/dmd/pull/12817


dkorpel (approved):
This is an improvement, but it still does not give a definition of `return ref`, and it doesn't explain why returning by ref gets this magical exemption for `scope` but returning by value doesn't.

```D
@safe:

ref int* retRefScopeC(return ref scope int* p)
{
    return p; // allowed
}

int* retRefScopeD(return ref scope int* p)
{
    return p; // not allowed?!
}

void main() 
{
    scope int* p;

    int* a = retRefScopeC(p); // allowed
    int* b = retRefScopeD(p); // not allowed?!
    assert(a == b); // exact same pointer!
}
```

WalterBright:
The spec actually isn't intended to give explanations, it's purely descriptive.

I'm not sure what is missing from the description of return-ref. I did add a description of how return-ref scope behaves.

PR #3299, opened 2022-05-17
[spec/function] Improve UFCS docs

ntrel:
Improve description.
Document that other local symbols besides imports do not match a UFCS call.
Split example & move member function paragraph.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -3782,12 +3782,13 @@ $(H2 $(LNAME2 function-attribute-inference, Function Attribute Inference))
 
 $(H2 $(LNAME2 pseudo-member, Uniform Function Call Syntax (UFCS)))
 
-        $(P A free function can be called like a member function when:)
+        $(P A free function can be called like a member function when both:)
         $(UL
-        $(LI The member function does not exist for the expression on the left hand side)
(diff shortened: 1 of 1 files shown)


spec/function.dd
```diff
@@ -3782,18 +3782,20 @@ $(H2 $(LNAME2 function-attribute-inference, Function Attribute Inference))
 
 $(H2 $(LNAME2 pseudo-member, Uniform Function Call Syntax (UFCS)))
 
-        $(P A free function can be called with a syntax equivalent to that of
-        a member function of its first parameter type. The free function is
-        called a $(I UFCS function).
+        $(P A free function can be called like a member function when:)
+        $(UL
+        $(LI The member function does not exist for the expression on the left hand side)
+        $(LI The free function's first parameter type matches the left hand expression)
```
dkorpel, line 3788: It's not clear that this list is a conjunction, i.e. these points must both hold, not just one of them. It's also unclear what "left hand expression" refers to.

dkorpel (changes requested):

spec/function.dd
ntrel: Updated.

dkorpel (approved):

PR #3300, opened 2022-05-17
Change exe to installer in button

maxhaton:


rikkimax:
After a recent post on the N.G. this is one improvement to the downloads page we came up with on Discord. That could improve the new user experience.

Separating out (i.e. tabbed) the installers and archives is another.

dkorpel (approved):

PR #3301, opened 2022-05-19
minor changes to undefined behavior notes

WalterBright:
to clarify

### DIFF
--- a/spec/declaration.dd
+++ b/spec/declaration.dd
@@ -590,8 +590,9 @@ $(GNAME VoidInitializer):
         $(P Normally, variables are initialized either with an explicit
         $(GLINK Initializer) or are set to the default value for the
         type of the variable. If the $(I Initializer) is $(D void),
-        however, the variable is not initialized. If its value is
-        used before it is set, undefined program behavior will result.
+        however, the variable is not initialized.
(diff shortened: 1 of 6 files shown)


spec/declaration.dd
```diff
@@ -590,8 +590,9 @@ $(GNAME VoidInitializer):
         $(P Normally, variables are initialized either with an explicit
         $(GLINK Initializer) or are set to the default value for the
         type of the variable. If the $(I Initializer) is $(D void),
-        however, the variable is not initialized. If its value is
-        used before it is set, undefined program behavior will result.
+        however, the variable is not initialized.
+        Void initializers for objects that contain indirections is not
+        allowed in @safe code.
```
dkorpel, line 595:
```suggestion
        allowed in `@safe` code.
```

dkorpel (approved):
Undefined behavior is not allowed in `@safe` code by definition, and there's already a list of things disallowed in `@safe` code: https://dlang.org/spec/function.html#safe-functions

So mentioning in `UNDEFINED_BEHAVIOR` paragraphs that it's disallowed in `@safe` code is redundant, though I don't think it's a problem either.

PR #3302, opened 2022-05-19
Try to move GSOC files into their own folder

maxhaton:
First step to trying to clear up the dlang.org tree.

It's a mess.


thewilsonator:
`dmd/src/dmd` next pls.

maxhaton:
> `dmd/src/dmd` next pls.

https://youtu.be/CZ44w8DMxJw 
That's fightin' talk boy

CyberShadow:
@maxhaton @thewilsonator  This broke the pages:

https://dlang.org/gsoc/gsoc2012.html

You need to update the macro which sets the site root, so that the resources (CSS) are loaded from the correct location.

(Not that I expect that these pages are going to be in high demand, but we should either keep them non-broken or delete them outright.)

maxhaton:
> @maxhaton @thewilsonator This broke the pages:
> 
> https://dlang.org/gsoc/gsoc2012.html
> 
> You need to update the macro which sets the site root, so that the resources (CSS) are loaded from the correct location.
> 
> (Not that I expect that these pages are going to be in high demand, but we should either keep them non-broken or delete them outright.)

Ok. Ddoc is awful...

I'll rectify this, could you make DAutoTester try to catch this? Maybe it already does but failed to go red?

CyberShadow:
Yeah, it catches some things but not this case apparently.

Search for "Warning:" in the build log. I see it has bitrotten and reports all pages as false positives today though.

PR #3303, opened 2022-05-19
Shove some things into a folder called community.

maxhaton:
These are the things that are obviously suitable for it, there is likely more elsewhere.

This one might break something.


CyberShadow:
When moving things around, please add redirects from the old to the new locations in `.htaccess` to avoid breaking external links to these resources.

CyberShadow:
> These are the things that are obviously suitable for it, there is likely more elsewhere.

Do you want to put together a plan for a final structure of the website?

maxhaton:
> > These are the things that are obviously suitable for it, there is likely more elsewhere.
> 
> Do you want to put together a plan for a final structure of the website?

I just want to make the build process less horrible and directory structure also less horrible.

Currently I am doing *something*, the final result may well be something else. I don't have the attention span to sort the makefile out so I'm just doing what I can first

thewilsonator:
@maxhaton ping on redirects. Lets get this in.

CyberShadow:
> Lets get this in.

https://dlang.org/gsoc/gsoc2012.html should probably be fixed first before more similar changes.

PR #3305, opened 2022-05-21
[spec/class.dd] Move abstract class docs to class.dd & add examples

ntrel:
They fit better in class.dd and the final class docs are there (as well as synchronized classes). 
Link to class.dd from attribute.dd instead.

The first commit just moves the paragraphs.
The second commit tweaks the docs and adds examples.

Move abstract class attribute after abstract methods rather than in the middle of them.

### DIFF
--- a/spec/class.dd
+++ b/spec/class.dd
@@ -1214,17 +1214,35 @@ $(P
 )
 
 $(P
-        Classes become abstract if any of its virtual member functions
+        A class is abstract if any of its virtual member functions
         are declared abstract or if they are defined within an
         abstract attribute.
         Note that an abstract class may also contain non-virtual member functions.
-)
-
-$(P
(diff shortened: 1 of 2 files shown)


dkorpel (approved):

spec/class.dd
```diff
@@ -1203,6 +1203,76 @@ $(P When a scope class reference goes out of scope, the destructor (if any) for
...
+        Only virtual member functions may be declared abstract; non-virtual
+        member functions and free-standing functions cannot be declared
+        abstract.
+)
+
+$(P
+        A class is abstract if any of its virtual member functions
+        are declared abstract or if they are defined within an
+        abstract attribute.
+        Note that an abstract class may also contain non-virtual member functions.
```
dkorpel, line 1220: I suppose you merged the two paragraphs because the following example applies to both of them.

PR #3306, opened 2022-05-21
[spec/struct] Improve introduction section

ntrel:
Show how to instantiate a struct and copy it.
Document `new` on a struct.
List kinds of member declarations:
Link to methods, static fields/methods, operator overloading & alias this.

### DIFF
--- a/spec/struct.dd
+++ b/spec/struct.dd
@@ -34,6 +34,50 @@ $(GNAME AggregateBody):
     $(D {) $(GLINK2 module, DeclDefs)$(OPT) $(D })
 )
 
+    $(P The following example declares a struct type with a single integer field:)
+
+$(SPEC_RUNNABLE_EXAMPLE_RUN
+---
+struct S
+{
+    int i;
+}
+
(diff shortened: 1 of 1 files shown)


spec/struct.dd
```diff
@@ -34,6 +34,50 @@ $(GNAME AggregateBody):
...
+
+    S b = a; // copy a
+    a.i++;
+    assert(a.i == 4);
+    assert(b.i == 3);
+}
+---
+)
+
+    $(P A struct is allocated on the stack by default. To allocate on
```
dkorpel, line 58: Only for local variables

dkorpel (changes requested):

spec/struct.dd
maxhaton: You can also say "struct instance". All local variables are stored on the stack but a class is a pointer to an instance

ntrel: Updated.

dkorpel (approved):

PR #3307, opened 2022-05-22
add Rationale to assert() specification

WalterBright:
This comes up now and then, it's nice to be able to point to the long discussions on it.


spec/expression.dd
```diff
@@ -2179,6 +2179,12 @@ $(GNAME AssertArguments):
     )
     )
 
+    $(RATIONALE
+    The rationale for and proper use of `assert` comes up now and then.
+    Here is a long discussion of it:
```
mdparker, line 2184: At a minimum, "comes" should be "come" (rationale AND use). Pedantically, "usage" is more appropriate than "use" here (*how* something is used vs. the fact that it is used). 

I'd suggest a rephrasing it all as:

"The rationale for and proper usage of assert are sometimes misunderstood. A relevant discussion can be found in the following forum thread:"

mdparker (changes requested):
I'd prefer my suggested rewrite for the less casual language, but the two minor changes work, too.

jmh530:
I would much rather a single simple explanation, with examples, and then the link to further discussion.

aG0aep6G:
This is an alarmingly low-effort and low-quality PR. A discussion thread with hundreds of messages of different authors displaying different opinions does not make a good rationale.

spec/expression.dd
WalterBright: done

PR #3310, opened 2022-05-22
[spec/class] Improve list of class components

ntrel:
This should not be exhaustive (we already have a TOC) but should mention notable things.

Add:
* nested classes
* operator overloading
* item for other declarations - link to [DeclDef](https://dlang.org/spec/module.html#DeclDef)
* links for interfaces, static fields

Remove:
* anonymous metaclass for Objective-C (not D so not that relevant, it is mentioned in member functions section)
* types - covered by DeclDef
* synchronized attribute - this is not a *component* of a class (it's in the TOC)
* static constructors and destructors - these are in the TOC but they don't
  seem to be particularly relevant to classes. In another
  pull I could merge these sections from class.dd into the module.dd
  docs: https://dlang.org/spec/module.html#staticorder, as they are
  children of DeclDef.
* unittests - covered by DeclDef, not that relevant to classes.

Move alias this out of methods.

### DIFF
--- a/spec/class.dd
+++ b/spec/class.dd
@@ -55,7 +55,7 @@ $(UL
         $(LI dynamic fields)
         $(LI $(DDSUBLINK spec/attribute, static, static) fields)
         $(LI $(RELATIVE_LINK2 nested, nested classes))
-        $(LI $(RELATIVE_LINK2 member-functions, member functions)
+        $(LI $(RELATIVE_LINK2 member-functions, member functions))
         $(UL
             $(LI static member functions)
             $(LI $(DDSUBLINK spec/function, virtual-functions, Virtual Functions))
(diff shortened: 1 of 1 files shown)


spec/class.dd
```diff
@@ -46,30 +46,26 @@ $(GNAME Interface):
     $(GLINK2 type, BasicType)
 )
 
-$(P Classes consist of:)
+$(P A class consists of:)
 
+$(COMMENT list of notable components, not general things or attributes)
```
dkorpel, line 51:
```suggestion
$(COMMENT list of notable components, not exhaustive)
```
Or something else that doesn't speak of "general things"

dkorpel (changes requested):

dkorpel (approved):

PR #3312, opened 2022-05-23
[spec] Improve return docs

ntrel:
function.dd:
Add anchors for contract preconditions, postconditions (see below).
Move sentence about non-void return type requiring at least one terminating statement from statement.dd to Return Values section. Also mention noreturn statements.
Add link to ref functions.

statement.dd:
Link to function return values.
Uncomment returning a expression when return type is void. This was commented out in 549bdb109 because the expression could be non-void before [this issue](https://issues.dlang.org/show_bug.cgi?id=5399) was fixed (mentioned in commit message). Document that the expression must be void in this case, and that this is useful in generic programming.
Link to `out` contract docs rather than contracts.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -362,7 +362,8 @@ $(H2 $(LNAME2 function-return-values, Function Return Values))
         is required if the function specifies a return type that is not void,
         unless the function contains inline assembler code.)
 
-        $(P Function return values not marked as $(D ref) are considered to be rvalues.
+        $(P Function return values not marked as $(RELATIVE_LINK2 ref-functions, `ref`)
+        are considered to be rvalues.
         This means they cannot be passed by reference to other functions.
         )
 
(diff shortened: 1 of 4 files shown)


spec/function.dd
```diff
@@ -357,7 +357,13 @@ $(GNAME OutStatement):
 
 $(H2 $(LNAME2 function-return-values, Function Return Values))
 
-        $(P Function return values not marked as $(D ref) are considered to be rvalues.
+        $(P At least one $(DDSUBLINK spec/statement, return-statement, return statement),
+        throw statement, noreturn statement or assert(0) expression
```
maxhaton, line 361: Link to noreturn statement docs?

dkorpel (approved):

spec/function.dd
ntrel: I've now added links and expanded on 'noreturn statement' which isn't really a thing.

PR #3318, opened 2022-06-15
Remove outdated clause from ReturnStatement spec

pbackus:
This sentence was added to account for a change in the behavior of
return statements in DMD 2.037. That change has long since been
reverted, and the sentence is now false.

See https://issues.dlang.org/show_bug.cgi?id=3478

### DIFF
--- a/spec/statement.dd
+++ b/spec/statement.dd
@@ -1505,8 +1505,6 @@ $(P
         An *Expression* of type void is allowed if the function specifies
         a void return type. The *Expression* will be evaluated,
         but nothing will be returned. This is useful in generic programming.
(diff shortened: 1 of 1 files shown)


maxhaton (approved):

ntrel:
@pbackus this sentence says if the void expression has no side effects it's an error. Why would we allow an expression that does nothing, that's bug prone surely?

pbackus:
@ntrel I wasn't able to track down when and why this was changed in the compiler, but the current behavior has been in place since at least DMD 2.060 (the oldest version on run.dlang.io), which was released in 2012.

PR #3319, opened 2022-06-16
Fix 23186 - wchar/dchar do not have their endianess defined

dkorpel:
@rikkimax

### DIFF
--- a/spec/type.dd
+++ b/spec/type.dd
@@ -123,6 +123,8 @@ $(H2 $(LEGACY_LNAME2 Basic Data Types, basic-data-types, Basic Data Types))
     $(TROW $(D dchar), $(D '\U0000FFFF'), unsigned 32 bit (UTF-32 code unit))
     )
 
+    $(P Endianness of basic types is part of the $(DDSUBLINK spec/abi, endianness, ABI))
+
     $(IMPLEMENTATION_DEFINED The real floating point type has at least the range and precision
     of the `double` type. On x86 CPUs it is often implemented as the 80 bit Extended Real
     type supported by the x86 FPU.
--- a/spec/abi.dd
+++ b/spec/abi.dd
@@ -47,6 +47,9 @@ $(H2 $(LNAME2 basic_types, Basic Types))
         $(TROW double, 64 bit IEEE 754 floating point value)
         $(TROW real, implementation defined floating point value$(COMMA) for x86 it is
          80 bit IEEE 754 extended real)
+        $(TROW char, 8 bit unsigned value)
+        $(TROW wchar, 16 bit unsigned value)
+        $(TROW dchar, 32 bit unsigned value)
         )
 
 $(H2 $(LNAME2 delegates, Delegates))


ljmf00 (approved):

spec/type.dd
```diff
@@ -119,8 +119,8 @@ $(H2 $(LEGACY_LNAME2 Basic Data Types, basic-data-types, Basic Data Types))
     $(TROW $(D cdouble), $(D double.nan+double.nan*1.0i), complex double)
     $(TROW $(D creal), $(D real.nan+real.nan*1.0i), complex real)
     $(TROW $(D char), $(D '\xFF'), unsigned 8 bit (UTF-8 code unit))
-    $(TROW $(D wchar), $(D '\uFFFF'), unsigned 16 bit (UTF-16 code unit))
-    $(TROW $(D dchar), $(D '\U0000FFFF'), unsigned 32 bit (UTF-32 code unit))
+    $(TROW $(D wchar), $(D '\uFFFF'), unsigned 16 bit (UTF-16 code unit) matching the target endianness)
```
maxhaton, line 122: > target 
Does that cover CTFE?

dkorpel: CTFE does not expose endianness.

ibuclaw: Why not say it for all?  Do `ushort`'s match the target endianness?  What about `double`'s?

I'm being pedantic, but I think leaving this detail for abi.dd is enough.

ibuclaw (changes requested):

rikkimax:
> > I'm being pedantic, but I think leaving this detail for abi.dd is enough.
> 
> I think so too, but @rikkimax says "this isn't an ABI thing, it's about encodings" https://issues.dlang.org/show_bug.cgi?id=23186#c2

With Unicode there are three different encodings associated with it, for each code point size. Big/Little/Unknown endian.

Basically, you can't read in a file, read a pipe or socket.  and cast it to the respective string type.

But now that you mention it @ibuclaw we should be defining float/double in terms of binary vs decimal. Because they are not interchangeable and both can be 32/64bit.

ibuclaw:
> > > I'm being pedantic, but I think leaving this detail for abi.dd is enough.
> > 
> > I think so too, but @rikkimax says "this isn't an ABI thing, it's about encodings" https://issues.dlang.org/show_bug.cgi?id=23186#c2
> 
> With Unicode there are three different encodings associated with it, for each code point size. Big/Little/Unknown endian.
> 
> Basically, you can't read in a file, read a pipe or socket.  and cast it to the respective string type.
> 

But these are determined by the BOM at the beginning of the file/stream, not the ABI of the target if I understand you then. So is there really any point in giving endianess a mention then? 

> But now that you mention it @ibuclaw we should be defining float/double in terms of binary vs decimal. Because they are not interchangeable and both can be 32/64bit.

They will always be binary as we don't have any specialization of floating point types.

rikkimax:
> But these are determined by the BOM at the beginning of the file/stream, not the ABI of the target if I understand you then. So is there really any point in giving endianess a mention then?

Yes, because at least we have documented that you can cast when the endian matches, otherwise you have to do the conversion.

It also means that string literals have their endianness defined. Since you can't copy the saved output of a string literal to file ext., and read it on another and assume it will be correct.

> > But now that you mention it @ibuclaw we should be defining float/double in terms of binary vs decimal. Because they are not interchangeable and both can be 32/64bit.
> 
> They will always be binary as we don't have any specialization of floating point types.

Which could be a bit of a problem moving forward if C does indeed get decimal floats ;)

ibuclaw:
> > But these are determined by the BOM at the beginning of the file/stream, not the ABI of the target if I understand you then. So is there really any point in giving endianess a mention then?
> 
> Yes, because at least we have documented that you can cast when the endian matches, otherwise you have to do the conversion.
> 
> It also means that string literals have their endianness defined. Since you can't copy the saved output of a string literal to file ext., and read it on another and assume it will be correct.
> 

String literals are 1-byte, so there are no distinctions between endianess there.  Wide literals follow native endianess, which matches the underlying integer types.

Conversion to/from native is a library matter.

> > > But now that you mention it @ibuclaw we should be defining float/double in terms of binary vs decimal. Because they are not interchangeable and both can be 32/64bit.
> > 
> > They will always be binary as we don't have any specialization of floating point types.
> 
> Which could be a bit of a problem moving forward if C does indeed get decimal floats ;)

They technically [do have them](https://gcc.gnu.org/onlinedocs/gcc/Decimal-Float.html#Decimal-Float), as well as [fixed point](https://gcc.gnu.org/onlinedocs/gcc/Fixed-Point.html#Fixed-Point), for decades now.

rikkimax:
> Conversion to/from native is a library matter.

Indeed, the best we can do is to acknowledge when it needs to happen.

> They technically [do have them](https://gcc.gnu.org/onlinedocs/gcc/Decimal-Float.html#Decimal-Float), as well as [fixed point](https://gcc.gnu.org/onlinedocs/gcc/Fixed-Point.html#Fixed-Point), for decades now.

At least with compiler-specific extensions, you can kinda ignore them when it comes to describing D's support for C.

If the things that look like they are going into C23, go in, we won't be able to describe D's support for C the same way we do now. Too much stuff we won't be able to represent 1:1 that we should be able to.

PR #3322, opened 2022-06-17
Update org list on main page

mdparker:
Some of the companies in the "Industry Proven" list aren't around or aren't using D anymore. This updates the list with our biggest sponsor and two other companies that have been using and contributing a long time.

### DIFF
--- a/book/contextMacros.ddoc
+++ b/book/contextMacros.ddoc
@@ -1,2 +1,1 @@
-COZUM_HTML = frontispiece.cozum.html
-DVER = 2.088
+DVER = 2.097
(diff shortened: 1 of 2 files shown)


book/contextMacros.ddoc
```diff
@@ -1 +1,2 @@
-DVER = 2.097
+COZUM_HTML = frontispiece.cozum.html
+DVER = 2.088
```
thewilsonator, line 2: Why move the version backwards?

mdparker: I have no idea where that came from. I didn't touch that file. I'll checkout the head version.

thewilsonator (approved):

PR #3323, opened 2022-06-17
Specs: Change D Main Function to allow contracts and shortened methods

etienne02:
Note: Dmd allow to put contracts on main function. If this should be rejected, I’ll update the spec accordingly

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -3083,15 +3083,20 @@ $(H2 $(LNAME2 main, $(D main()) Function))
 
         $(GRAMMAR
         $(GNAME MainFunction):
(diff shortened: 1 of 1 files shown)


spec/function.dd
```diff
@@ -3083,16 +3083,21 @@ $(H2 $(LNAME2 main, $(D main()) Function))
...
             $(D int)
             $(GLINK2 type, noreturn)
             $(RELATIVE_LINK2 auto-functions, $(D auto))
+
+        $(GNAME MainFunctionBody):
+            $(GLINK ShortenedFunctionBody)
+            $(GLINK SpecifiedFunctionBody)
         )
 
+
```
thewilsonator, line 3100: remove extra newline

thewilsonator (approved):

PR #3324, opened 2022-06-17
Remove contextMacros.ddoc

dkorpel:
The file is automatically generated, but tracked by git, so when you build the website, it adds to your diff.

Trailing white space removed by my editor.


thewilsonator (approved):

thewilsonator:
you could also add this to `.gitignore`

maxhaton (approved):

dkorpel:
> you could also add this to `.gitignore`

That was my first idea, but there's no use keeping the file around in the first place, it's only created so it can be passed to dmd

PR #3344, opened 2022-07-19
Fix download links

ryuukk:


thewilsonator (approved):

download.dd
```diff
@@ -201,8 +201,8 @@ Macros:
 
     BETA=$(COMMENT $0)
     _=BETA=$0
-    B_DMDV2=2.100.1
-    B_SUFFIX=beta.1
+    B_DMDV2=2.100.0
+    B_SUFFIX=beta.0
```
ibuclaw, line 205: Aren't these numbers wrong? At the very least try a partial revert of the merge_stable commit rather than make it up.

ibuclaw (changes requested):

jmh530:
This only addresses the beta release.

download.dd
ibuclaw: 🏥🎂👶

PR #3347, opened 2022-07-20
Fix Issue 18887 - inout badly described

RazvanN7:
Moved the inout section from functions to type qualifiers.


dkorpel:
Did you change the content?

RazvanN7:
Nope, simply copy-pasta. The only thing I added was a link in the inout function section to the new inout type qualifier section.

dkorpel (approved):

jmh530:
@RazvanN7 I get that it is only copy-pasta, but it would be nice to have an example of the usage of `inout const`.

spec/function.dd
```diff
@@ -698,139 +698,7 @@ $(H2 $(LNAME2 auto-ref-functions, Auto Ref Functions))
...
-            // rewritten to immutable(int)[]
-            immutable(int)[] z = bar(ia);
-        }
-        ---
-        )
-
-    $(P $(B Note:) Shared types cannot
-        be matched with `inout`.
-    )
+    $(P For extensive information read the $(LINK2 const3.hmtl#inout, $(D inout section of the type qualifiers documentation page)).)
```
ntrel, line 701: Typo hmtl. For LaTeX anchor I'd write this as:
```
$(DDSUBLINK2 spec/const3, inout, $(D inout) type qualifier).
```

PR #3349, opened 2022-07-22
Fix source links for druntime

schveiguy:
Since druntime has moved to dmd, all the source links are broken.


adamdruppe (approved):

maxhaton (approved):

thewilsonator (approved):

schveiguy:
Ugh, just noticed the "improve this page" links all wrong as well.

schveiguy:
That one seems more involved, not sure how to do it.

PR #3350, opened 2022-07-23
Fix Issue 19869 - `FunctionLiteral` allows incorrect forms

ntrel:
_SpecifiedFunctionBody_ allows `do` and contracts.
https://dlang.org/spec/function.html#SpecifiedFunctionBody

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -1865,7 +1865,7 @@ $(GNAME ParameterWithMemberAttributes):
 
 $(GNAME FunctionLiteralBody2):
     $(D =>) $(GLINK AssignExpression)
-    $(GLINK2 statement, BlockStatement)
+    $(GLINK2 function, SpecifiedFunctionBody)
 )
 
     $(P $(I FunctionLiteral)s (also known as $(LNAME2 lambdas, $(I Lambdas))) enable embedding anonymous functions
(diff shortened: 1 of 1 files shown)


spec/expression.dd
```diff
@@ -1854,7 +1854,7 @@ $(GNAME FunctionLiteral):
     $(D function) $(D ref)$(OPT) $(GLINK2 type, Type)$(OPT) $(GLINK ParameterWithAttributes)$(OPT) $(GLINK FunctionLiteralBody2)
     $(D delegate) $(D ref)$(OPT) $(GLINK2 type, Type)$(OPT) $(GLINK ParameterWithMemberAttributes)$(OPT) $(GLINK FunctionLiteralBody2)
     $(D ref)$(OPT) $(GLINK ParameterWithMemberAttributes) $(GLINK FunctionLiteralBody2)
-    $(GLINK2 function, SpecifiedFunctionBody)
+    $(GLINK2 statement, BlockStatement)
```
dkorpel, line 1857: This one looks right

spec/expression.dd
```diff
@@ -1865,14 +1865,14 @@ $(GNAME ParameterWithMemberAttributes):
 
 $(GNAME FunctionLiteralBody2):
     $(D =>) $(GLINK AssignExpression)
-    $(GLINK2 function, SpecifiedFunctionBody)
+    $(GLINK2 statement, BlockStatement)
```
dkorpel, line 1868: But this one doesn't look right, this is allowed:
```D
auto b = function int(int x) do {return true;};
```

dkorpel (changes requested):

spec/expression.dd
ntrel: Sorry, restored this line.

dkorpel (approved):

RazvanN7 (approved):

PR #3354, opened 2022-07-24
Fix Issue 15379 - "final" attribute on function parameter

ntrel:
`final` is not allowed here in D2.
```d
void foo(final int a){} // error!
```
```
finalpar.d(1): Error: variable `finalpar.foo.a` cannot be `final`, perhaps you meant `const`?
```

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -73,6 +73,10 @@ $(GNAME VariadicArgumentsAttribute):
     $(D shared)
 )
 
+$(NOTE In D2, declaring a parameter `final` is a semantic error, but not a parse error.)
+
+$(P See also: $(RELATIVE_LINK2 param-storage, parameter storage classes).)
+
 $(H3 Function attributes)
 
 $(GRAMMAR


dkorpel:
It's not a parse error though, this compiles:
```D
version(none) void foo(final int a) {}
```

So the grammar should allow it.

ljmf00:
> It's not a parse error though, this compiles:
> 
> ```d
> version(none) void foo(final int a) {}
> ```
> 
> So the grammar should allow it.

Why we should allow something that apparently doesn't have a useful meaning? I think moving the error to the parser, if we can have the same kind of error message, and fixing the grammar is a good path. Otherwise, if this has some meaning for templates, we should perhaps update the spec?

dkorpel:
> Why we should allow something that apparently doesn't have a useful meaning? 

I meant to say "the grammar should allow it _or the implementation should be updated to match the new grammar_".

> I think moving the error to the parser, if we can have the same kind of error message, and fixing the grammar is a good path.

There are plenty of errors that could be caught by the parser but are currently caught during semantic analysis, keeping the grammar simpler.
```D
version(none) 
void f() scope {
    break;
    default:
    switch (0) {}
    "useless";
}
```

I'm not against changing the parser here per se, but moving errors to parse time isn't a "good path" by default.

ntrel:
> There are plenty of errors that could be caught by the parser but are currently caught during semantic analysis, keeping the grammar simpler.

In this case the grammar is not simpler. I think the reason it parses was for the error advising to use const. That error is probably not needed anymore, or could perhaps be done in the parser. If not, the spec should note that final is not valid here.

ntrel:
> I think the reason it parses was for the error advising to use const

Perhaps it was so D1 and D2 code can co-exist in the same module? So perhaps it should still parse if this is still a goal.

ntrel:
As this seems intentional to not error on parsing, I have added a note to the docs instead of removing `final`.

@dkorpel Can you remove the waiting for dmd label now please?

dkorpel (approved):

RazvanN7 (approved):

PR #3355, opened 2022-07-25
[spec/struct] Add 'Union Literals' & 'Anonymous Structs & Unions'

ntrel:
The docs for a struct literal with a union field apply for non-anonymous unions too.
Fix: Non-overlapping union field data is not default initialized, it is
zeroed.
Fixes Issues 21188, 18855.

### DIFF
--- a/spec/struct.dd
+++ b/spec/struct.dd
@@ -6,7 +6,8 @@ $(HEADERNAV_TOC)
 
 $(H2 $(LNAME2 intro, Introduction))
 
-    $(P Whereas classes are reference types, structs are value types.
+    $(P Whereas $(DDLINK spec/class, Classes, classes) are reference types,
+    structs are value types.
     Structs and unions are simple aggregations of data and their
     associated operations on that data.
     )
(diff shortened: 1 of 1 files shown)


dkorpel (approved):
I think the text could be more precise in what an anonymous struct/union does instead of what it's used for. It doesn't cover arbitrarily complex / redundant usages: `struct S { struct { struct { union { struct { int x; }}}}}` Still, it's a good start.

PetarKirov:
@ntrel please update the commit message, so that all of the issues will be automatically closed by @dlang-bot. See: https://github.com/dlang/dlang-bot#referencing-multiple-issues

RazvanN7:
Hmm, the commit message does contain both issues, I wonder why the bot doesn't catch that.

RazvanN7 (approved):

PetarKirov:
I think the commit message should start with the issues that are fixed. Try this:

```
Fix issues 18855 and 21188 - [spec/struct] Add 'Union Literals' & 'Anonymous Structs & Unions' 

Fix Issue 21188 - Anonymous structs - not described.
Fix Issue 18855 - Behavior of Anonymous Union is Undocumented.

Fix: Non-overlapping union field data is not default initialized, it is
zeroed.
```

See: https://regex101.com/r/C7q7eT/1

ntrel:
@PetarKirov @RazvanN7 Updated commit message as suggested. No changes to diff.

ntrel:
@PetarKirov Do you know which file in `dlang-bot` has the regex to detect fixed issues? I'd like to have a look.

PetarKirov:
> Do you know which file in dlang-bot has the regex to detect fixed issues? I'd like to have a look.

I think this is what you're looking for: https://github.com/dlang/dlang-bot/blob/750bb49423033c0bc52d2515fde2b29238668265/source/dlangbot/bugzilla.d#L21

RazvanN7 (approved):

PR #3357, opened 2022-07-29
Fix Issue 15286 -  is(typeof(symbol))

RazvanN7:


spec/expression.dd
```diff
@@ -2370,6 +2370,12 @@ $(GNAME TypeSpecialization):
         compared against.
     )
 
+    $(P $(I IsExpression)s may be used in conjunction with $(D typeof) to check
+    whether an expression is semantically correct. For example, $(D is(typeof(foo)))
```
dkorpel, line 2374: "semantically correct" is currently a poorly defined term. Is `cast(ubyte*)0xDEADBEEF` semantically correct inside a `@safe` function? The spec for `__traits(compiles, X)` also uses "semantically correct" but I recall there being subtle differences between that and `is(typeof(X))`.

I suppose it's out of scope for this issue though.

dkorpel (approved):

spec/expression.dd
ntrel: Something like "to check whether an expression type checks correctly" would be more accurate. Errors not related to type checking are not done here AIUI. Could also suggest __traits(compiles) for full semantic checks.

RazvanN7: Semantically correct is essentially the same thing type checking from a compilers perspective. However, type-checking is a compiler developer specific term, whereas "semantically correct" is much more friendly to the user.

ntrel: @RazvanN7 
> Semantically correct is essentially the same thing type checking from a compilers perspective

That is not true when the context affects whether it compiles:
```d
int i;
void f() pure
{
    pragma(msg, is(typeof(i++))); // true
    pragma(msg, __traits(compiles, i++)); // false
}
```
So the new docs are misleading that `is(typeof(expr))` tests whether `expr` is *semantically correct*. Some semantic checks are skipped.

Also Don Clugston said in 2012:
> It checks if the expression has a type. That's all.

https://forum.dlang.org/post/k6o04g$259s$1@digitalmars.com

PR #3359, opened 2022-07-30
Fix AA word count example

ntrel:
Add Examples heading otherwise they're listed under Properties.
Make example get compiled.
Add missing import.

### DIFF
--- a/spec/hash-map.dd
+++ b/spec/hash-map.dd
@@ -468,7 +468,8 @@ $(P Properties for associative arrays are:)
         else evaluates the $(D create) delegate and adds it to the associative array))
     )
 
-$(HR)
+$(H2 $(LNAME2 examples, Examples))
+
 $(H3 $(LNAME2 aa_example, Associative Array Example: word count))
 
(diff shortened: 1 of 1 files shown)


ntrel:
I think the example would be clearer if it worked on `stdin` rather than accepting multiple filenames and opening each. It's supposed to demonstrate AAs rather than how to implement a command-line program. Happy to do that if anyone else agrees.

ntrel:
@CyberShadow Is there a limit on compilable example length? It seems to be getting cut off in the middle of a string literal.
```
                if (args.length > 2)
                {
                    writefln("
---
__stdin.d(64): Error: unterminated string constant starting at __stdin.d(64)
```
The example runs on clicking run in my local generated html file.

CyberShadow:
The problem has to be the string of `-------------------------------------`. There's probably a bug somewhere, it should be looking only at the start of the line and optional whitespace.

ntrel:
Tests now green.

dkorpel (approved):

PR #3361, opened 2022-07-30
Make some AA examples runnable

ntrel:


spec/hash-map.dd
```diff
@@ -392,13 +400,14 @@ $(H2 $(LNAME2 runtime_initialization, Runtime Initialization of Immutable AAs))
             aa = assumeUnique(temp);
         }
 
-        unittest
+        void main()
```
thewilsonator, line 403: do we not have a `SPEC_RUNNABLE_UNITTEST_EXAMPLE_RUN`?

thewilsonator (approved):

spec/hash-map.dd
ntrel: grepping for `UNIT` or `TEST` only found `DDOX_UNITTEST_HEADER`.

RazvanN7 (approved):

PR #3363, opened 2022-08-02
Fix function.dd typo

jmh530:
Should be `html` instead of `hmtl`


ntrel:
There's https://github.com/dlang/dlang.org/pull/3362 already which uses the DDSUBLINK macro instead (maybe better for LaTeX).

RazvanN7:
Yes, this has already been fixed. Thanks @jmh530 !

PR #3365, opened 2022-08-07
[spec/ddoc] Improve macro expansion section

ntrel:
Fixes Issue 15476 - DDOC_UNDEFINED_MACRO is undocumented.

Format more inline code.
Mention macros can take arguments *before* example using them.
Tweak generated HTML example and use 'such as' - this is not the exact output.
Improve wording of macro expansions.
Use bullet points.
Mention DDOC_UNDEFINED_MACRO.
Use RELATIVE_LINK2 macro.
Add 'Macro Arguments' and 'Macro Definitions' subheadings.
Add note about literal commas in an argument.

### DIFF
--- a/spec/ddoc.dd
+++ b/spec/ddoc.dd
@@ -515,9 +515,9 @@ $(P
 $(H3 $(LNAME2 inline_code, Inline Code))
 
 $(P
-    Inline code can be written between backtick characters ($(BACKTICK)), similarly
+    Inline code can be written between backtick characters (`$(BACKTICK)`), similarly
     to the syntax used on GitHub, Reddit, Stack Overflow, and other websites. Both
-    the opening and closing $(BACKTICK) character must appear on the same line to trigger this
+    the opening and closing `$(BACKTICK)` character must appear on the same line to trigger this
     behavior.
 )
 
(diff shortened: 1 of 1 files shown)


thewilsonator (approved):

spec/ddoc.dd
```diff
@@ -968,33 +970,65 @@ $(P
...
-    in it.
-    If a $(DOLLAR)(NAME) is desired to exist in the output without being
-    macro expanded, the $(DOLLAR) should be $(LINK2 #punctuation_escapes,
+    )
+    $(LI
+    If the macro name is undefined, the replacement text will
+    be $(D $)`(DDOC_UNDEFINED_MACRO(NAME))`. This defaults to empty.
+    )
+    $(LI
+    If $(D $)`(NAME)` is desired to exist in the output without being
```
RazvanN7, line 988: `desired` sounds a bit non-technical to me. Maybe replace with `is required in the output [...]`.

ntrel:
```suggestion
    If $(D $)`(NAME)` is required to exist in the output without being
```

RazvanN7 (approved):

PR #3368, opened 2022-08-08
spec/arrays.dd: remove invalid code in example

emanuele6:
I think this code was left here by mistake: it is invalid because it is missing an identifier name before `=` and I don't think it belongs in that example.

### DIFF
--- a/spec/arrays.dd
+++ b/spec/arrays.dd
@@ -153,9 +153,9 @@ int[]* e;     // pointer to dynamic array of ints
 $(H2 $(LNAME2 literals, Array Literals))
 
 ---
-auto a1 = [1,2,3];  // type is int[], with elements 1, 2 and 3
+auto a1 = [1,2,3];  // type is int[], with elements 1, 2, and 3
 auto a2 = [1u,2,3]; // type is uint[], with elements 1u, 2u, and 3u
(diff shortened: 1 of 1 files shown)


spec/arrays.dd
```diff
@@ -155,7 +155,6 @@ $(H2 $(LNAME2 literals, Array Literals))
 ---
 auto a1 = [1,2,3];  // type is int[], with elements 1, 2 and 3
 auto a2 = [1u,2,3]; // type is uint[], with elements 1u, 2u, and 3u
-int[2] = [1, 2];
```
thewilsonator, line 158:
```diff
-int[2] = [1, 2];
+int[2] = [1, 2];  // type is `int[2]`  with elements 1, 2 and always holds two elements
```

emanuele6: i guess, but you need to add an identifier name like `a3` in that case (as i said).

thewilsonator: Whoops that's what i meant to do.

emanuele6: I would remove the "always holds two elements" part since that has not really much to do with the "array literal".

ntrel (approved):

RazvanN7 (approved):

PR #3369, opened 2022-08-10
Fix Issue 23276 - DOC: &gt; instead of > in dmd-windows.html

RazvanN7:


thewilsonator (approved):

adamdruppe:
this kind of thing is why i was against the addition of markdown

too much magic syntax

Geod24:
This kind of things is why I am in favor of markdown. You'd express this with a `console` code block in markdown, no escaping needed.

adamdruppe:
The original `&gt;` thing is another wtf (what i was actually first going to rail against in here), but that's a design flaw ddoc and markdown share anyway.

PR #3370, opened 2022-08-12
Follow-up for PR #3357

RazvanN7:


RazvanN7:
cc @ntrel . Thanks for the clarifications.

spec/expression.dd
```diff
@@ -2371,9 +2371,9 @@ $(GNAME TypeSpecialization):
     )
 
     $(P $(I IsExpression)s may be used in conjunction with $(D typeof) to check
-    whether an expression is semantically correct. For example, $(D is(typeof(foo)))
-    will return $(D true) if $(D foo) is a semantically correct expression with a valid
-    type.
+    whether an expression type checks correctly. For example, $(D is(typeof(foo)))
+    will return $(D true) if $(D foo) is a syntactically correct expression and if
```
dkorpel, line 2375: This suggests you could give it a syntactically incorrect expression and it will return `false`. In reality, you won't get past the parsing stage when you put something syntactically incorrect inside `typeof()`.

dkorpel (changes requested):

spec/expression.dd
RazvanN7: Maybe I should just drop the "syntactically" part altogether and just state that it returns true if foo has a valid type?

dkorpel: That works

RazvanN7:
@dkorpel done.

dkorpel (approved):

PR #3373, opened 2022-08-15
Fix Issue 21279 - cast expression between integer types is not defined

ntrel:


dkorpel (approved):

spec/expression.dd
```diff
@@ -1204,6 +1204,41 @@ $(H4 $(LNAME2 cast_static_array, Static Arrays))
...
+
+        $(SPEC_RUNNABLE_EXAMPLE_RUN
+        ---
+        short c = -1;
+        auto d = cast(ushort) c;
+        assert(d == ushort.max);
+
+        ubyte e = 255;
+        auto f = cast(byte) e;
+        assert(f == -1);
```
schveiguy, line 1238: This is somewhat of a bad example, because ubyte doesn't need a cast for byte.

schveiguy (commented):
The two paragraphs seem to be overlapping in what they are specifying.

Both statements talk about signed values. I'm not exactly sure what the need for the second paragraph is.

spec/expression.dd
ntrel: Yes, `byte f = e;` works implicitly. But I'm hoping there will be a warning or error for that eventually. So I think it's good practice to cast when the values might not fit.

ntrel:
> Both statements talk about signed values. I'm not exactly sure what the need for the second paragraph is.

The first paragraph is about truncating to a smaller type. The first example shows this can flip the sign even when the source and target are both signed. The second paragraph is about differing signedness of source and target.

spec/expression.dd
schveiguy: The point I'm making is that there is no special rule for casting, as this implies.

schveiguy: Now that I'm looking for it, I also don't see anything in the spec on implicit conversions between types for assignment! Though it hints at it in the example here: https://dlang.org/spec/type.html#usual-arithmetic-conversions

ntrel: @schveiguy I have changed these to implicit conversions. I think it's OK to keep them under the cast docs as the first floating point example is also implicit conversions, not casts.

PR #3374, opened 2022-08-15
[spec/type] Improve Type Conversions section

ntrel:
Mention unsafe casts.
List more implicit conversions.
Add Class Conversions subheading.

### DIFF
--- a/spec/type.dd
+++ b/spec/type.dd
@@ -193,7 +193,7 @@ f = Foo.E;       // OK
 
     $(UL
     $(LI All types implicitly convert to $(RELATIVE_LINK2 noreturn, `noreturn`).)
-    $(LI Arrays implicitly convert to $(DDSUBLINK spec/arrays, void_arrays, `void[]`).)
+    $(LI Static and dynamic arrays implicitly convert to $(DDSUBLINK spec/arrays, void_arrays, `void[]`).)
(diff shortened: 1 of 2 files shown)


dkorpel (approved):

spec/type.dd
```diff
@@ -188,6 +191,15 @@ f = Foo.E;       // OK
 -------------------
 )
 
+    $(UL
+    $(LI All types implicitly convert to $(RELATIVE_LINK2 noreturn, `noreturn`).)
+    $(LI Arrays implicitly convert to $(DDSUBLINK spec/arrays, void_arrays, `void[]`).)
```
dkorpel, line 196: Perhaps specify Static and Dynamic arrays, but not Associative arrays

ntrel:
```suggestion
    $(LI Static and dynamic arrays implicitly convert to $(DDSUBLINK spec/arrays, void_arrays, `void[]`).)
```

ntrel:
@CyberShadow Just in case you didn't know, `dautotest` seems broken, affecting other recent dlang.org pulls too.

CyberShadow:
Yeah, everything is kind of broken due to issues.dlang.org being down.

If you'd like to spend some time on it, an improvement would be to make this error non-fatal, and skip the relevant parts of the website build if issues.dlang.org is unreachable.

PR #3377, opened 2022-08-16
[spec/expression] Improve docs for `this` and `super`

ntrel:
Fix - `this` is not a pointer for structs, it's a reference.
Mention `typeof(this)` works anywhere inside an aggregate type, not just member functions.
Fix - assigning to `this` does work for structs.
`super` is only allowed inside a class.
Mention using `super` from Calling Base Class Methods.

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -1623,7 +1623,7 @@ $(H3 $(LNAME2 super, super))
     $(P $(D super) is identical to $(D this), except that it is
         cast to $(D this)'s base class.
         It is an error if there is no base class.
-        It is an error to use $(D super) outside a class.
+        It is an error to use the $(D super) reference outside a non-static class member function.
         (Only class $(D Object) has no base class.)
         If a member function is called with an explicit reference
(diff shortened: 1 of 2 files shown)


spec/expression.dd
```diff
@@ -1615,14 +1616,14 @@ $(H3 $(LNAME2 this, this))
...
-    $(P Assignment to $(D this) is not allowed.)
+    $(P Assignment to $(D this) is not allowed for classes.)
 
 $(H3 $(LNAME2 super, super))
 
     $(P $(D super) is identical to $(D this), except that it is
         cast to $(D this)'s base class.
         It is an error if there is no base class.
-        It is an error to use $(D super) within a struct member function.
+        It is an error to use $(D super) outside a class.
```
dkorpel, line 1626: When you do this:
```D
class C
{
    struct S
    {
        void m() 
        {
            auto x = super;
        }
    }
}
```
The error is "`super` is only allowed in non-static class member functions". Maybe copy that wording in the spec.

dkorpel (approved):

PR #3379, opened 2022-08-16
[spec/type] Add Value Range Propagation section

ntrel:
@schveiguy:
> Also I cannot find "Value Range Propagation" anywhere in the spec.

### DIFF
--- a/spec/type.dd
+++ b/spec/type.dd
@@ -167,7 +167,7 @@ $(H3 $(LEGACY_LNAME2 Pointer Conversions, pointer-conversions, Pointer Conversio
 $(H3 $(LEGACY_LNAME2 Implicit Conversions, implicit-conversions, Implicit Conversions))
 
     $(P Implicit conversions are used to automatically convert
-    types as required.
+    types as required. The rules for integers are detailed in the next sections.
     )
 
     $(P An enum can be implicitly converted to its base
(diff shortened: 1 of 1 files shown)


spec/type.dd
```diff
@@ -344,6 +344,34 @@ ulong  u4 = long(-1); // ok, -1 can be represented in a ulong
     point types.
     )
 
+$(H3 $(LNAME2 vrp, Value Range Propagation))
+
+    $(P Besides type-based implicit conversions, D allows certain integer
+        expressions to implicitly convert to a narrower type after
+        integer promotion. This works by analysing the minimum and
+        maximum possible range of values for each expression.
```
maxhaton, line 352:
```suggestion
        integer promotion. This works by determining the possible range of values of an expression
```

spec/type.dd
```diff
@@ -344,6 +344,34 @@ ulong  u4 = long(-1); // ok, -1 can be represented in a ulong
...
+        If that range of values would fit inside a narrower type, implicit
```
maxhaton, line 353: "fit" will need a definition

spec/type.dd
maxhaton: possibly link to things like this (https://en.wikipedia.org/wiki/Value_range_analysis) although the form that the compiler performs is more limited.

spec/type.dd
ntrel:
```suggestion
        If that range of values matches or is a subset of a narrower type's value range, implicit
```

spec/type.dd
ntrel: Now added that link, and updated the docs.

PR #3381, opened 2022-08-19
[spec/type] Improve numeric type conversion examples

ntrel:
Expands on #3378.
Move integer promotion examples to that section and show `dchar` promotion.
Add rationale for integer promotion.
Add rationale for numeric type conversion rules.
Rename int, uint variables for signed/unsigned example and show long/ulong.
Add floating point example.

### DIFF
--- a/spec/type.dd
+++ b/spec/type.dd
@@ -274,6 +274,8 @@ $(H3 $(LEGACY_LNAME2 Integer Promotions, integer-promotions, Integer Promotions)
     column.
     )
 
+    $(P Integer promotion applies to each operand of a binary expression:)
+
     $(SPEC_RUNNABLE_EXAMPLE_COMPILE
     ---
     byte a;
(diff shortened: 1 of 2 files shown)


spec/type.dd
```diff
@@ -274,6 +274,33 @@ $(H3 $(LEGACY_LNAME2 Integer Promotions, integer-promotions, Integer Promotions)
...
+    int e = d * d; // OK
+    static assert(is(typeof(int() * d) == int));
+
+    dchar f;
+    static assert(is(typeof(f - f) == uint));
+    ---
+    )
+
+    $(RATIONALE
+    * 32-bit integer operations are often faster than smaller integer types
```
dkorpel, line 299: Are there even instructions to do sub 32-bit arithmetic operations?

ntrel: Well gcc uses the `al` 8 or `ax` 16-bit registers for `add`:
https://d.godbolt.org/z/77zxcooon

Here's the link I saw before writing that:
https://stackoverflow.com/questions/5069489/performance-of-built-in-types-char-vs-short-vs-int-vs-float-vs-double

Happy to remove this line if you prefer.

maxhaton: https://stackoverflow.com/questions/45660139/how-exactly-do-partial-registers-on-haswell-skylake-perform-writing-al-seems-to

Using the partial/sub registers can lead to some confusing performance traps.

dkorpel (approved):

PR #3384, opened 2022-08-20
[spec/struct] Improve union init & anonymous docs

ntrel:
Add simple Dynamic Initialization of Unions section.
Add paragraph defining an anonymous struct/union.

### DIFF
--- a/spec/struct.dd
+++ b/spec/struct.dd
@@ -379,6 +379,24 @@ S t = s; // sets t.a to 3, S.opCall(S) is not called
 ----
 )
 
+$(H3 $(LNAME2 dynamic_union_init, Dynamic Initialization of Unions))
+
+        $(P The $(RELATIVE_LINK2 static_union_init, static initializer syntax)
+        can also be used to initialize non-static variables.
+        The initializer need not be evaluable at compile time.)
+
+$(SPEC_RUNNABLE_EXAMPLE_COMPILE
+---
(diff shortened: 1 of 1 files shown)


spec/struct.dd
```diff
@@ -434,6 +452,12 @@ assert(u.b == [2, 0]);
 
 $(H2 $(LNAME2 anonymous, Anonymous Structs and Unions))
 
+    $(P An anonymous struct or union can be declared as a member of a
+    parent class, struct or union by omitting an identifier.
```
dkorpel, line 456:
```suggestion
    parent class, struct or union by omitting the identifier after `struct` or `union`.
```

spec/struct.dd
```diff
@@ -434,6 +452,12 @@ assert(u.b == [2, 0]);
...
+    Such a declaration simultaneously declares an anonymous type and a
+    hidden field of that type in the parent type. The fields are
```
dkorpel, line 458: It doesn't really declare a type, it just affects the `.offsetof` of the members. 'Hidden field' usually refers to context pointers or vtables, but that's not what's going on here. This description makes it appear more complex than it is, I would say that an anonymous `struct` and `union` specifies whether the contained member variables are being laid out sequentially or overlapped.

ntrel: Thanks, reworded.

dkorpel (approved):

PR #3387, opened 2022-08-22
[spec/expression] Improve index and slice docs

ntrel:
Fix mention of pointer index.
Add summary descriptions for index and slice expressions.
Mention overloading of index and slice operations.

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -849,7 +849,7 @@ assert(a[2] == 4);
 )
 
     $(P $(GLINK IndexExpression) can also be used with a pointer and has
-    the same behaviour as adding an integer then dereferencing the result.)
+    the same behaviour as adding an integer, then dereferencing the result.)
 
     $(P If the second operand is a pointer, and the first is an integral type,
         and the operator is $(D +),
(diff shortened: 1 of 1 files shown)


spec/expression.dd
```diff
@@ -849,7 +849,7 @@ assert(a[2] == 4);
 )
 
     $(P $(GLINK IndexExpression) can also be used with a pointer and has
-    the same behaviour as adding an integer.)
+    the same behaviour as adding an integer then dereferencing the result.)
```
dkorpel, line 852:
```suggestion
    the same behaviour as adding an integer and then dereferencing the result.)
```

```suggestion
    the same behaviour as adding an integer, then dereferencing the result.)
```

spec/expression.dd
```diff
@@ -1349,6 +1351,10 @@ $(GNAME IndexExpression):
         $(I ArgumentList) and `$` appears in that scope only.
     )
 
+    $(P The index operator can be $(DDSUBLINK spec/operatoroverloading, array, overloaded).
+        Using multiple indexes in *ArgumentList* is only supported for operator
```
dkorpel, line 1355:
```suggestion
        Using multiple indices in *ArgumentList* is only supported for operator
```

dkorpel (approved):

PR #3390, opened 2022-08-26
add Cheat Sheet for RefScopeReturn

WalterBright:
This adds a table for what various combinations of ref-scope-return mean. Once one satisfies oneself that it is correct, it becomes easier to just check the implementation and code against the table, rather than trying to puzzle out each case over and over.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -2524,6 +2524,271 @@ $(H3 $(LNAME2 this-reference, `this` Reference))
         )
 
 
+
+$(H2 $(LNAME2 refscopereturn, Ref Scope Return Cases))
+
+
+$(H3 $(LNAME2 rsr_definitions, Definitions))
+
+    $(DL
+    $(DT I) $(DD type that contains no indirections)
+    $(DT P) $(DD type that contains indirections)
+    $(DT X) $(DD type that may or may not contain indirections)
+    $(DT p) $(DD parameter of type P)
+    $(DT i) $(DD parameter of type I)
+    $(DT ref) $(DD `ref` or `out` parameter)
+
+    $(DT returned) $(DD returned via the `return` statement)
+    $(DT escaped) $(DD stored in a global or other memory not in the function$(RSQUO)s stack frame)
+    )
+
+$(H3 $(LNAME2 rsr_classification, Classification))
+
+$(P A parameter must be in one of the following states:)
+
+    $(TABLE2 Classification,
+    $(THEAD Term, Description)
+    $(TROW None,
+        `p` may be returned or escaped)
+
+    $(TROW ReturnScope,
+        `p` may be returned but not escaped)
+
+    $(TROW Scope,
+        `p` may be neither returned nor escaped)
+
+    $(TROW Ref,
+        `p` may be returned or escaped$(LF)
+        `ref` may not be returned nor escaped)
+
+    $(TROW ReturnRef,
+        `p` may be returned or escaped$(LF)
+        `ref` may be returned but not escaped)
+
+    $(TROW RefScope,
+        `p` may be neither returned nor escaped$(LF)
+        `ref` may not be returned nor escaped)
+
+    $(TROW ReturnRef-Scope,
+        `p` may be neither returned nor escaped$(LF)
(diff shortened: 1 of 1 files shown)


spec/function.dd
```diff
@@ -2524,6 +2524,288 @@ $(H3 $(LNAME2 this-reference, `this` Reference))
...
+---
+class C {
+    X foo()
+}
+---
+
+    $(P is treated as:)
+
+---
+X foo(P)
```
dkorpel, line 2749:
```suggestion
X foo(C)
```

spec/function.dd
```diff
@@ -2524,6 +2524,288 @@ $(H3 $(LNAME2 this-reference, `this` Reference))
         )
 
 
+
+$(H2 $(LNAME2 refscopereturn, Ref Scope Return Cases))
```
dkorpel, line 2528: It would be nice to give this a proper place and introduction instead of just dropping it in the middle of function.dd, which is already pretty big.

spec/function.dd
```diff
@@ -2524,6 +2524,288 @@ $(H3 $(LNAME2 this-reference, `this` Reference))
...
+    $(TROW ReturnRef-ReturnScope,
+        `p` may be returned but not escaped$(LF)
+        `ref` may be returned but not escaped$(LF)
+        This isn't expressible with the current syntax
+        and so is not allowed.)
+    )
+
+$(H3 $(LNAME2 rsr_mapping, Mapping Syntax Onto Classification))
+
+    $(P The juxtaposition of `return` immediately preceding `scope` means ReturnScope.
```
dkorpel, line 2589: This is already covered by another paragraph on this page

spec/function.dd
```diff
@@ -2524,6 +2524,288 @@ $(H3 $(LNAME2 this-reference, `this` Reference))
...
+
+    $(OL
+    $(LI `return ref` and `scope`)
+    $(LI `ref` and `return scope`)
+    )
+
+    $(P The ambiguity is resolved by:)
+
+    $(OL
+    $(LI If `return scope` appears, in that order, adjacent to each other, then
```
dkorpel, line 2764: This is the third time this appears on this page

WalterBright:
@CyberShadow any clue why DAutoTest is failing?

spec/function.dd
WalterBright: P is deliberate. It is treated as a pointer.

spec/function.dd
WalterBright: The whole spec could us a refactoring. The structure of it is very old. But this is properly a separate issue.

spec/function.dd
WalterBright: I only see two. Anyhow, removed one.

spec/function.dd
WalterBright: deleted the other

WalterBright:
@CyberShadow never mind! (I found it.)

spec/function.dd
```diff
@@ -2524,6 +2524,271 @@ $(H3 $(LNAME2 this-reference, `this` Reference))
...
+        Ref,)
+
+    $(TROW `X foo(ref scope P p)`,
+        RefScope,)
+
+    $(TROW `P foo(ref return scope P p)`,
+        Ref-ReturnScope,)
+
+    $(TROW `P foo(return ref scope P p)`,
+        ReturnRef-Scope,)
```
ibuclaw, line 2626: I'd expect these to be the same, or the keywords (or syntax) we use for dip1000 aren't expressive enough, or having ref as a keyword was a mistake.

WalterBright: See the included:
```
The juxtaposition of `return` immediately preceding `scope` means ReturnScope.
Otherwise, `return` and `ref` in any position means ReturnRef.
```

ibuclaw (approved):

spec/function.dd
dkorpel: Here's the first appearance:
https://github.com/dlang/dlang.org/blob/e60a7f08b67377d289cd871090b140ae11459fcf/spec/function.dd#L1844-L1846

PR #3353, opened 2022-07-24
Revert download and changelog version changes

ibuclaw:
[Revert "Temp fix homepage download links](https://github.com/dlang/dlang.org/commit/64b902c76a8f1f75c68a41db365e7eecc0a5405c) (#3351)

[Revert "Fix issue 23237 - temp force 2.100.0 download](https://github.com/dlang/dlang.org/commit/92f05854a43ee059e6e83726f0a450a652108e6d) (#3348)

[Revert "Switch back to 2.100.0, 2.100.1 is not released yet"](https://github.com/dlang/dlang.org/commit/701fb2a266a6748aae8a3cc575ef1f85a10acde7) (#3344)

Reverts the mess people have been making of the mechanical parts of the site pages.

To be merged when I get the SMS from the RM.

PR #3398, opened 2022-08-30
update download and changelog for v2.100.2-beta.1

MartinNowak:


dkorpel (approved):

ibuclaw (approved):
Note, there's also #3353 for when the .2 release goes out proper.

ibuclaw:
I don't think going through the beta and rc process is necessary here, but oh well.

PR #3403, opened 2022-09-02
[spec/arrays] Improve resizing section

ntrel:
Add *Growing an Array* subheading.
Mention reallocation.
Use list for when a copy is done, mention no spare capacity.
Add asserts to example.
Move `capacity` description to later example.
Add *capacity and reserve* subheading.

### DIFF
--- a/spec/arrays.dd
+++ b/spec/arrays.dd
@@ -754,8 +754,9 @@ array.length = i;
 
 $(H3 $(LNAME2 capacity-reserve, `capacity` and `reserve`))
 
-        $(P The $(D capacity) property determines how many elements can be appended
-        to the array without reallocating.)
+        $(P The $(D capacity) property gives the maximum length the array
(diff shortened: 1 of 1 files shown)


spec/arrays.dd
```diff
@@ -744,6 +752,10 @@ array.length = i;
         input from the console - it's unlikely to be longer than 80.
         )
 
+$(H3 $(LNAME2 capacity-reserve, `capacity` and `reserve`))
+
+        $(P The $(D capacity) property determines how many elements can be appended
```
dkorpel, line 757: This doesn't tell what `capacity` actually is: the `length` the array can grow to before needing to reallocate.

dkorpel: Maybe mention that the specific amount of elements is `capacity - length`

ntrel: Now done.

dkorpel (approved):

PR #3404, opened 2022-09-02
[spec/arrays] Improve assignment docs

ntrel:
Introduce .ptr property before example using it.
Remove wrong line showing `s = a` as an error.
~~Show assigning a shorter array literal to a static array is OK.~~ It is a runtime error.
Mention slicing produces a dynamic array.
Improve _Array Setting_ example.

Update: 
* Document assigning a dynamic array to a static array.
* Extend Copying example with dynamic array

### DIFF
--- a/spec/arrays.dd
+++ b/spec/arrays.dd
@@ -199,7 +199,8 @@ assert(a == []);
     using pointer $(RELATIVE_LINK2 slicing, slicing), so that the number of elements
     to copy is then known.)
 
-    $(P A static array can be assigned from a dynamic array. The lengths must match:)
+    $(P A static array can be assigned from a dynamic array - the data is copied.
+    The lengths must match:)
 
 $(SPEC_RUNNABLE_EXAMPLE_RUN
 ---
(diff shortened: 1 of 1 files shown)


spec/arrays.dd
```diff
@@ -181,8 +184,11 @@ p = a.ptr; // p points to the first element of the array a.
 
 // error, since the length of the array pointed to by p is unknown
 //s = p;
-//s = a; // error, as a's length is not known at compile-time
+
+s = [1,2]; // OK
+assert(s == [1,2,0]);
 s = [1,2,3];
+//s = [1,2,3,4]; // error, too many elements
```
dkorpel, line 191: Please put spaces after commas

dkorpel (approved):

spec/arrays.dd
ntrel: Turns out assigning a shorter literal is a runtime error. I have now documented that and assigning a dynamic array variable.

RazvanN7 (approved):

PR #3418, opened 2022-09-25
[spec] Document NewExpression forms

ntrel:
Add example.
Fix: scope new is only special for class types.


dkorpel (approved):

dkorpel (commented):
This leaves out the empty argument list form though (`new int();`)

PR #3422, opened 2022-10-01
[spec] List symbols which can be aliased

ntrel:
Make template example runnable.
Add import link.
Add static import for function alias example.

### DIFF
--- a/spec/declaration.dd
+++ b/spec/declaration.dd
@@ -243,7 +243,16 @@ $(GNAME AliasAssignment):
 
     $(P $(I AliasDeclaration)s create a symbol name that refers to another symbol.
         That symbol name can be used anywhere that the aliased symbol may appear.
+        The following symbols can be aliased:
     )
+    * $(RELATIVE_LINK2 alias-type, Types)
+    * $(RELATIVE_LINK2 alias-variable, Variables)
+    * Constants
+    * Modules
+    * Packages
+    * Function Literals
+    * Templates
+    * Template Instantiations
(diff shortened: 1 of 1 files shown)


spec/declaration.dd
```diff
@@ -243,7 +243,16 @@ $(GNAME AliasAssignment):
...
     $(P $(I AliasDeclaration)s create a symbol name that refers to another symbol.
         That symbol name can be used anywhere that the aliased symbol may appear.
+        The following symbols can be aliased:
     )
+    * $(RELATIVE_LINK2 alias-type, Types)
+    * $(RELATIVE_LINK2 alias-variable, Variables)
+    * Constants
+    * Modules
+    * Packages
+    * Function Literals
```
maxhaton, line 253: regular functions too

spec/declaration.dd
```diff
@@ -325,11 +336,13 @@ version (linux)
 --------------------
 
         $(P
-        Aliasing can be used to `import` a symbol from an import into the
-        current scope:
+        Aliasing can be used to $(DDSUBLINK spec/module, import-declaration, `import`)
+        a symbol from an import into the current scope:
```
maxhaton, line 340:
```suggestion
        a symbol from an imported module or package into the current scope:
```

spec/declaration.dd
ntrel: Thanks, now added.

dkorpel (approved):

PR #3334, opened 2022-07-09
[project] Merge druntime into the dmd repository

ibuclaw:
Fixes build script paths to work with new merged dmd-druntime repository

See https://github.com/dlang/projects/issues/92

PR #3428, opened 2022-10-03
posix.mak: Fix make clean recipe

ibuclaw:
Since the dmd/druntime merger in #3334, `make clean` has been broken in the dlang.org repository.


CyberShadow (approved):
Thanks!

dkorpel (approved):

PR #3429, opened 2022-10-03
Use downloads.dlang.org for all main download links

ibuclaw:
This is using our new downloads bucket, the old s3 bucket is there for archive only (no longer receiving any updates).


CyberShadow (approved):
I didn't even know we were exposing AWS URLs to people.

Edit: oh right, this was the https:// bodge-fix.

ibuclaw:
> Edit: oh right, this was the https:// bodge-fix.

Yes, the CNAME to the old S3 bucket was http-only.

PR #3437, opened 2022-10-17
Extract and test RUNNABLE_EXAMPLE tests including Phobos

ntrel:
Extract & run the tests (like for dlang.org SPEC_RUNNABLE_EXAMPLE_*) to check the examples actually compile.
RUNNABLE_EXAMPLE seems to have been introduced to phobos in 2018 to std.exception. Recently I added it to std.regex, and I have a pull to add it to std.experimental.allocator: https://github.com/dlang/phobos/pull/8604.

In theory it could be used in druntime too, but as of now it isn't.

---

Depends on https://github.com/dlang/phobos/pull/8611.

Remove RUNNABLE_EXAMPLE from 1 module.dd example where it imported a module that doesn't exist.
Show path of file when printing number of examples found.

### DIFF
--- a/tools/dspec_tester.d
+++ b/tools/dspec_tester.d
@@ -121,6 +121,7 @@ int main(string[] args)
         SpecType("$(SPEC_RUNNABLE_EXAMPLE_COMPILE", CompileConfig.TestMode.compile),
         SpecType("$(SPEC_RUNNABLE_EXAMPLE_RUN", CompileConfig.TestMode.run),
         SpecType("$(SPEC_RUNNABLE_EXAMPLE_FAIL", CompileConfig.TestMode.fail),
+        SpecType("$(RUNNABLE_EXAMPLE", CompileConfig.TestMode.run),
     ];
     foreach (file; specDir.dirEntries("*.dd", SpanMode.depth).parallel(1))
     {
--- a/tools/dspec_tester.d
+++ b/tools/dspec_tester.d
@@ -16,7 +16,6 @@ Author: Sebastian Wilzbach
 */
 
 import std.algorithm;
-import std.path;
 import std.stdio;
 import std.range;
 import std.regex;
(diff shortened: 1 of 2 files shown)


tools/dspec_tester.d
```diff
@@ -121,15 +125,27 @@ int main(string[] args)
         SpecType("$(SPEC_RUNNABLE_EXAMPLE_COMPILE", CompileConfig.TestMode.compile),
         SpecType("$(SPEC_RUNNABLE_EXAMPLE_RUN", CompileConfig.TestMode.run),
         SpecType("$(SPEC_RUNNABLE_EXAMPLE_FAIL", CompileConfig.TestMode.fail),
+        SpecType("$(RUNNABLE_EXAMPLE", CompileConfig.TestMode.run),
     ];
-    foreach (file; specDir.dirEntries("*.dd", SpanMode.depth).parallel(1))
+    auto fnames = chain(specDir.dirEntries("_*.dd", SpanMode.depth),
```
ntrel, line 130: The underscore was just for local testing, will remove tomorrow.

tools/dspec_tester.d
```diff
@@ -121,15 +125,27 @@ int main(string[] args)
...
+    auto files = chain(specDir.dirEntries("*.dd", SpanMode.depth),
+        stdDir.dirEntries("*.d", SpanMode.depth));
+    foreach (file; files.parallel(1))
     {
+        // auto-import current module if in phobos
+        string modImport = file.name.find(stdDir);
+        if (modImport.length)
+        {
+            modImport = "std" ~ modImport[stdDir.length..$];
+            modImport.popBackN(2); // ".d"
```
dkorpel, line 139:
```suggestion
            modImport = modImport.stripExtension();
```

dkorpel (approved):

tools/dspec_tester.d
ntrel: I added this and tweaked 2 imports.

ntrel:
@dkorpel the circleci check is still failing, seemingly unrelated. This is good to go.

dkorpel (approved):

PR #3442, opened 2022-10-19
Add scope array literals to spec

dkorpel:
Requested here: https://github.com/dlang/dmd/pull/14562#issuecomment-1282905802


maxhaton:
> What about when the element type is not plain old data as mentioned in the changelog: https://github.com/dlang/dmd/pull/14562/files#diff-a773cf7a39af2594f4e2e4f8aa41047d2b2f10a8558a5f0cb1c7d4c86f6d0304R23

As-if rule may be useful

pbackus:
It is worth noting that this change (and others like it) may **silently** introduce memory corruption into any `@trusted` code that relied on the old behavior. For example, the following code is memory safe when compiled with DMD 2.100, but will be unsound when compiled with DMD 2.101:

```d
@trusted int[] example()
{
    scope arr = [1, 2, 3];
    return arr;
}
```

PR #3443, opened 2022-10-21
Fix Issue 23467 - `tools/dspec_tester.d` calls `readText` multiple times per file

ntrel:
~Currently depends on #3437 (see last commit for now).~
This avoids reading and allocating the contents of the same file 4 times (would've been 3 times before #3437).


ntrel:
@wilzbach @dkorpel

thewilsonator (approved):

PR #3444, opened 2022-10-23
chore(spec/garbage): lookups on associative arrays doesn't use GC

ljmf00:

### DIFF
--- a/spec/garbage.dd
+++ b/spec/garbage.dd
@@ -374,7 +374,7 @@ $(H2 D $(LNAME2 op_involving_gc, Operations That Involve the Garbage Collector))
         $(LI Array concatenation)
         $(LI Array literals (except when used to initialize static data))
         $(LI Associative array literals)
-        $(LI Any insertion, removal, or lookups in an associative array)
+        $(LI Any insertion or removal in an associative array)
         $(LI Extracting keys or values from an associative array)
                 $(LI Taking the address of (i.e. making a delegate to) a nested function that
          accesses variables in an outer scope)


thewilsonator (approved):

Geod24 (approved):

Geod24:
> [1m/dev/shm/dtest/work/repo/dmd/compiler/src/dmd/dtemplate.d(3023): [1;31mError: [mfound `[0;36m$[m` instead of statement

Hum what ?

ntrel (changes requested):
This PR is wrong, lookups can throw which allocates an exception:
```d
auto aa = [1:2];
@nogc void main()
{
	int i = aa[2];
}
```
```
nogcaa.d(4): Error: indexing an associative array in `@nogc` function `D main` may cause a GC allocation
```
The wording needs to be reinstated and changed to "failed lookup".

ljmf00:
> This PR is wrong, lookups can throw which allocates an exception:
> 
> ```d
> auto aa = [1:2];
> @nogc void main()
> {
> 	int i = aa[2];
> }
> ```
> 
> ```
> nogcaa.d(4): Error: indexing an associative array in `@nogc` function `D main` may cause a GC allocation
> ```
> 
> The wording needs to be reinstated and changed to "failed lookup".

See https://github.com/dlang/dmd/pull/14585.

ljmf00-wekaio:
> This PR is wrong, lookups can throw which allocates an exception:

Also to clarify, it is not an `Exception` but rather an `Error`. Errors are thrown using a pre-allocated thread-local storage instance.

RazvanN7:
Even if an error would have been allocated with the GC, since errors should not be captured and are considered exceptional cases, I think that, generally, throwing errors should be @nogc.

RazvanN7 (approved):

PR #3445, opened 2022-10-23
[spec/arrays] Improve capacity docs

ntrel:
Remove sentence about appending to a static array. 
Add note about static array capacity being for generic code. 
Mention capacity is zero if not pointing to GC (array) memory. 
Extend example and explain it afterwards.
Explain when array length is reduced, capacity will be zero. 
Mention performance impact of capacity.

Fix Issue 23432 - document when array capacity is zero and capacity performance.

### DIFF
--- a/spec/arrays.dd
+++ b/spec/arrays.dd
@@ -576,8 +576,8 @@ Returns an array literal with each element of the literal being the $(D .init) p
         the number of bytes per array element.)
         $(TROW $(D .length), Returns the number of elements in the array.
         This is a fixed quantity for static arrays. It is of type $(D size_t).)
-        $(TROW $(D .capacity), Returns the number of elements that can be appended to the array without reallocating.
-        $(D .capacity) is always $(D 0) for static arrays because their size cannot be modified.)
+        $(TROW $(D .capacity), Always $(D 0) for static arrays because their size cannot be modified.
+            (This exists for generic code that handles both static and dynamic arrays.))
         $(TROW $(D .ptr), Returns a pointer to the first element of the array.)
         $(TROW $(D .dup), Create a dynamic array of the same size and copy the contents of the array into it. The copy will have any immutability or const stripped. If this conversion is invalid the call will not compile.)
         $(TROW $(D .idup), Create a dynamic array of the same size and copy the contents of the array into it. The copy is typed as being immutable. If this conversion is invalid the call will not compile.)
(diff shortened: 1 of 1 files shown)


ntrel:
@schveiguy @acehreli You might be interested in reviewing this if you have time.

spec/arrays.dd
```diff
@@ -576,8 +576,8 @@ Returns an array literal with each element of the literal being the $(D .init) p
         the number of bytes per array element.)
         $(TROW $(D .length), Returns the number of elements in the array.
         This is a fixed quantity for static arrays. It is of type $(D size_t).)
-        $(TROW $(D .capacity), Returns the number of elements that can be appended to the array without reallocating.
-        $(D .capacity) is always $(D 0) for static arrays because their size cannot be modified.)
+        $(TROW $(D .capacity), Always $(D 0) for static arrays because their size cannot be modified.
```
schveiguy, line 579: well....
```d
struct S
{
    int[4] x;
}
void main()
{
    auto s = [S()];
    import std.stdio;
    writeln(s[0].x.capacity); // 7
}
```

I think the capacity should be removed as a property for static arrays. It's not a property of static arrays, but of dynamic arrays. The reason it works is because D implicitly casts the static array to a dynamic array.

spec/arrays.dd
```diff
@@ -759,24 +760,63 @@ array.length = i;
...
+        a.reserve(2);
+        a = a[];
+        assert(a.capacity > 0);
+        a = a[0..2];
+        assert(a.capacity == 0);
+        ---
+        )
+
+        $(NOTE Accessing `.capacity` requires the runtime to
+        acquire a global lock and perform a table lookup.)
```
schveiguy, line 815: This is not true. There is a thread-local cache for capacity/appending. The lock is only required to fill in the cache, and then just the cache is used.

However, it still can have a performance impact, because you are talking about an opaque function that cannot be inlined, and the cache lookup isn't really that cheap either.

schveiguy (dismissed):
Please see the notes. I think identifying when the capacity is 0 is fine.

spec/arrays.dd
ntrel: Interesting, thanks. I have removed the static array property.

spec/arrays.dd
ntrel: OK, I've changed this to `may require`.

ntrel:
Added some more docs in e2b834e.

dkorpel (approved):

PR #3454, opened 2022-11-08
Make clear the search is using google

WebFreak001:
So people don't get surprised when they are suddenly on a different webpage when searching for something.


myOmikron:
I was very surprised!

dkorpel (approved):

Geod24 (approved):

PR #3455, opened 2022-11-14
merge stable

ibuclaw:


thewilsonator (approved):

ibuclaw:
Build failure is due to CI erroneously cloning druntime @CyberShadow.  This is preventing dlang.org site from being updated + release announcement -> Just merge? Or does DAutoTest need updating to stop doing this?

CyberShadow (changes requested):
Uh oh, the DMD docs fell off

posix.mak
```diff
@@ -431,10 +431,9 @@ ${GENERATED}/modlist-prerelease.ddoc : tools/modlist.d ${STABLE_DMD} $(DRUNTIME_
 		$(addprefix --dump , object std etc core dmd rt core.internal.array core.internal.util) >$@
```
ibuclaw, line 431: This looks like the problem @CyberShadow (`modlist-${LATEST}.ddoc` recipe has a different dmd/src path).

CyberShadow: Yes, well spotted!

ibuclaw: Actually, everything in the make files that uses `DMD_LATEST_DIR` needs to be aligned up with the use of `DMD_DIR`.

CyberShadow: Good idea, updated all the ones I could find.

ibuclaw: Looks much better now! :tada:

CyberShadow (approved):
Yep, looks good!

PR #3466, opened 2022-12-11
add spec/windows.dd

WalterBright:
The idea is to provide much better information about how Windows executables and dlls work and are constructed.

This is just a skeleton outline of the future page.

### DIFF
--- a/spec/windows.dd
+++ b/spec/windows.dd
@@ -1,0 +1,46 @@
+Ddoc
+
+$(SPEC_S Windows Programming,
+
+$(HEADERNAV_TOC)
+
+$(P This covers Windows programming with 32 bit OMF, 32 bit Mscoff,
+64 bit Mscoff, console programs, GUI programs, and DLLs.
+)
+
+$(H2 $(LNAME2 mscoff, Windows 32 and 64 bit MSCOFF Programs))
+
+    $(P 32 bit and 64 bit MSCOFF programs use the Microsoft Visual C/C++ compiler as the Associated
+    C Compiler, generate object files in the MSCOFF format, and use the Microsoft linker
+    to link them.
+    )
+
+$(H3 $(LNAME2 mscoff-console, Console Programs))
+
+$(H3 $(LNAME2 mscoff-windows, Windows GUI Programs))
+
+$(H3 $(LNAME2 mscoff-dlls, DLLs))
+
+$(H2 $(LNAME2 omf, Windows 32 bit OMF Programs))
+
+    $(P 32 bit OMF programs use the
+    $(LINK2 https://www.digitalmars.com/download/freecompiler.html, Digital Mars C/C++ compiler)
+    as the Associated C Compiler, generate object files in the OMF format, and use the
+    $(LINK2 https://www.digitalmars.com/ctg/optlink.html, Optlink linker)
+    to link them.
+    )
+
+$(H3 $(LNAME2 omf-console, Console Programs))
+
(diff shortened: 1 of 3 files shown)


spec/windows.dd
```diff
@@ -0,0 +1,46 @@
...
+$(P This covers Windows programming with 32 bit OMF, 32 bit Mscoff,
+64 bit Mscoff, console programs, GUI programs, and DLLs.
+)
+
+$(H2 $(LNAME2 mscoff, Windows 32 and 64 bit MSCOFF Programs))
+
+    $(P 32 bit and 64 bit MSCOFF programs use the Microsoft Visual C/C++ compiler as the Associated
+    C Compiler, generate object files in the MSCOFF format, and use the Microsoft linker
+    to link them.
+    )
```
rikkimax, line 16: That'll need to cover mingw and lld (which you can't enable if you have VS).

spec/windows.dd
```diff
@@ -0,0 +1,46 @@
...
+    as the Associated C Compiler, generate object files in the OMF format, and use the
+    $(LINK2 https://www.digitalmars.com/ctg/optlink.html, Optlink linker)
+    to link them.
+    )
+
+$(H3 $(LNAME2 omf-console, Console Programs))
+
+$(H3 $(LNAME2 omf-windows, Windows GUI Programs))
+
+$(H3 $(LNAME2 omf-dlls, DLLs))
```
rikkimax, line 37: That'll need to cover linker scripts which (for some reason) are required to generat a dll (at least last time I tried -shared wasn't enough).

rikkimax:
There should also be some talk on DllMain and how the compiler won't automatically add it for you and the implementation required to for use with druntime.

spec/windows.dd
WalterBright: I don't know the ecosystem around mingw/lld. Someone who does will have to write it.

spec/windows.dd
WalterBright: I know. This page is currently just an outline.

spec/windows.dd
rikkimax: > 32 bit and 64 bit MSCOFF programs use the Microsoft Visual C/C++ compiler as the Associated C Compiler, generate object files in the MSCOFF format, and use the Microsoft linker if available to link them. If not available mingw import libraries and LLD linker will be used instead.

Shouldn't need much, like what I wrote above.

spec/windows.dd
rikkimax: All good, just noting it from experience :)

spec/windows.dd
mdparker: That should specify that it's the MinGW import libraries and LLD that ship with dmd. Anyone attempting to use the import libraries in the MinGW distro may encounter difficulties, so it's important to clarify that.

ibuclaw:
What part of this is to do with the D language specification?

adamdruppe:
yeah i agree, this has nothing to do with the D language. might be a nice addition under dmd's documentation though.

WalterBright:
> What part of this is to do with the D language specification?

The same reason we have an ABI section.

WalterBright:
And a C portability guide, and a C++ portability guide, etc. If you want do split all this non-spec stuff off into a separate subdirectory, I'm all for it. Please do it! Meanwhile, my focus is on solving the problem of mass confusion about how things work under the hood.

WalterBright:
ping @dkorpel @RazvanN7 @maxhaton

RazvanN7 (approved):

PR #3467, opened 2022-12-11
[operatoroverloading.dd] Improve examples

ntrel:
Add opCmp example.
Make opCall example runnable.
Add opOpAssign example.
Make opIndex() example runnable.
Add example for `s.opIndex(s.opSlice(...))`, tweak paragraph.

### DIFF
--- a/spec/operatoroverloading.dd
+++ b/spec/operatoroverloading.dd
@@ -975,8 +975,8 @@ void main()
 
         $(P To overload array slicing of the form $(D a[)$(SLICE)$(D ]),
         two steps are needed.  First, the expressions of the form $(SLICE) are
-        translated via $(D opSlice!0) into user-defined objects that encapsulate
-        the endpoints $(I i) and $(I j). Then these user-defined objects are
+        translated via $(D opSlice!0) into objects that encapsulate
+        the endpoints $(I i) and $(I j). Then these objects are
         passed to $(D opIndex) to perform the actual slicing.)
 
 $(SPEC_RUNNABLE_EXAMPLE_RUN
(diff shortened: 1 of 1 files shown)


spec/operatoroverloading.dd
```diff
@@ -514,6 +515,23 @@ struct S
         version of $(D opEquals).  Otherwise, inequalities like $(D a <= b)
         will behave inconsistently with equalities like $(D a == b).)
 
+$(SPEC_RUNNABLE_EXAMPLE_RUN
+---
+struct S
+{
+    int i, j;
+    int opCmp(ref const S s) const { return i - s.i; } // ignore j
```
dkorpel, line 523: I don't think it's a good example, since it doesn't account for overflow. I suggest: `return (i > s.i) - (i < s.i);`

dkorpel (changes requested):

spec/operatoroverloading.dd
ntrel: Now fixed

spec/operatoroverloading.dd
```diff
@@ -514,6 +515,28 @@ struct S
...
+S c = {3, 0};
+assert(a < c);
+assert(a <= b);
+assert(!(a < b)); // opCmp ignores j
+assert(a != b);   // generated opEquals tests both i and j members
+---
+)
+
+$(BEST_PRACTICE Using `(i > s.i) - (i < s.i)` instead of `i - s.i` to
+compare integers avoids overflow/underflow.)
```
dkorpel, line 537:
```suggestion
compare integers avoids overflow.)
```

underflow happens with small floating points, going below `int.min` is officially still called overflow

RazvanN7 (approved):

PR #3471, opened 2022-12-14
add example to ModuleInfo documentation

adamdruppe:
Follow-up to the 100% correct information written by Walter, reviewed by me, and merged by Razvan in https://github.com/dlang/dlang.org/pull/3464 to demonstrate to the reader how things are supposed to work.


adamdruppe:
Force pushed with more information, to show how these modules must work with any combination of `-betterC` thanks to the ModuleInfo generation rules. I think this is complete now, ping @RazvanN7 @maxhaton

WalterBright (changes requested):
The prose needs to explain what is happening, and the tone of the existing prose should be done in the usual dry tone of specifications.

PR #3472, opened 2022-12-14
note that druntime must match compiler, not vice versa

adamdruppe:
See the function `dmd.toobj.genModuleInfo` for what specifically happens here. The compiler assumes a layout, druntime must match it instead of defining it.

### DIFF
--- a/spec/abi.dd
+++ b/spec/abi.dd
@@ -987,7 +987,7 @@ $(H2 $(LNAME2 ModuleInfo, ModuleInfo Instance))
         $(NOTE this feature may be removed.))
         )
 
-        $(P `ModuleInfo` is defined in Druntime's $(DRUNTIMESRC object.d).)
+        $(P `ModuleInfo` is defined in Druntime's $(DRUNTIMESRC object.d), which must match the compiler's output in both the values of flags and layout of fields.)
 
         $(P Modules compiled with $(DDSUBLINK dmd, switch-betterC, $(TT -betterC))
         do not have a `ModuleInfo` instance generated, because such modules must work


dkorpel (approved):

Geod24 (commented):
You are 100% right - but this is an implementation detail.
I don't think it belongs here. This is like saying "The `extern(C)` declarations must match the equivalent header files". It can be assumed to be in the general body of knowledge of the reader.

adamdruppe:
So consider this: suppose someone wrote a D compiler that actually did match the moduleinfo layout defined in druntime and modified druntime slightly, just adding some padding bytes or something, it is the same info tho. User has one library built with this compiler and wants to link it against a library built by dmd, which also generates moduleinfo, but does not check druntime to see the layout.

There's gonna be a runtime crash because one of the compilers is non-compliant to the spec. Which one is it? As the spec is written right now, I'd argue it is dmd, since the spec says it is *defined* in object.d.

With this PR merged, they'd be incompatible, but neither would be non-compliant (it'd make it implementation-defined behavior by the compiler+runtime combination, specifying they must agree, instead of being defined by the runtime's implementation that the compiler must adhere to).

(And before yesterday, ModuleInfo wasn't defined by the language spec at all. So if I made a D implementation that never generated it and implemented static ctors through some other means, would my D compiler be compliant? Before, yes, it was *all* implementation details. But now the spec spells out that ModuleInfo exists. So it isn't an implementation detail anymore.)

Probably ideal would be to actually define the abi in the spec directly, then say all conforming implementations must use that same definition, so they'd be all compatible. But right now, compilers and druntime implementations are tightly coupled, so you can't mix and match compilers for different modules, so just documenting that reality and saying the exact ABI is implementation defined by the whole is the easiest thing.

PR #3475, opened 2022-12-19
Remove version(integer) from docs

MrcSnm:
These got removed a while ago. Didn't do for the debug()  because I'm not sure about if it got removed from debug too.


dkorpel (approved):

dkorpel:
> I'm not sure about if it got removed from debug too.

It did

MrcSnm:
> > I'm not sure about if it got removed from debug too.
> 
> It did

Removed from debug too

PR #3479, opened 2022-12-22
Fix Issue 12298 - Templates can be used in mixin even when not declared as mixin template

RazvanN7:
I don't see any reason why this should be banned. Also, it seems that this is used extensively in dmd's testsuite. Come to think of it, I don't see why we need mixin template declarations, when we could just use normal templates.

### DIFF
--- a/spec/template-mixin.dd
+++ b/spec/template-mixin.dd
@@ -203,6 +203,26 @@ void main()
     duff_loop();  // executes foo() 10 times
 }
 ------
+
+        $(P A template declaration that is preceded by the $(D mixin) keyword
+        may only be instantiated only by being mixin-ed. A template declaration
+        that is not preceded by the $(D mixin) keyword may be both instantiated
+        normally and mixin-ed:)
+
+$(SPEC_RUNNABLE_EXAMPLE_RUN
+------
+mixin template Foo() { int foo; }
+template Bar() { int bar; }
+
+mixin Foo;  // can only be used by being mixin-ed
+mixin Bar;  // also can be mixed in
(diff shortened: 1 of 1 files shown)


spec/template-mixin.dd
```diff
@@ -202,6 +202,26 @@ void main()
     mixin duffs_device!(i, j, delegate { writeln("foo"); });
     duff_loop();  // executes foo() 10 times
 }
+------
+
+        $(P A template declaration that is preceded by the $(D mixin) keyword
+        may only be instantiated only by being mixin-ed. A template declaration
```
dkorpel, line 208:
```suggestion
        may only be instantiated by being mixin-ed. A template declaration
```

PetarKirov:
@WalterBright @atilaneves where do you stand on the distinction between `template`s and `mixin template`s? Are `mixin templates` a redundant feature in the language? IIRC `scope class` was similarly deprecated.

adamdruppe:
`scope class` was deprecated due to shortcomings in `@safe` implementation, not redundancy. It should actually be revisited now that dip1000 is less broken.

Geod24:
Yeah, `mixin template` is just redundant and should probably go.

adamdruppe:
What do you gain by removing it? mixin is already a keyword and it indicates intent.

(if anything, i'd go ahead and ban mixing in non-mixin templates, since they probably aren't written with the other scoping rules in mind and it changes out from under them!)

atilaneves:
Hmm... I'm not sure. @WalterBright ?

PR #2935, opened 2021-01-25
Correct download links to download years

Geod24:
Fix a couple issues with https://github.com/dlang/dlang.org/pull/2933
CC @schveiguy

PR #3484, opened 2023-01-02
Use release version-based paths on the downloads page

ibuclaw:
Doing #3481, #3253,  every year is just silly.


ibuclaw:
@Geod24 it seems that you're the one to blame #2935.  We're no longer on AWS anymore, and B2 doesn't support symlinks, so I doubt whatever you ran into is still the case.

CyberShadow:
Yeah, I think the problem was that it was redirecting to `http://` OSLT.

dkorpel (approved):

PR #3489, opened 2023-01-13
Fix ImportC square example

dkorpel:
See https://forum.dlang.org/post/uqllydlolryuzhkcghro@forum.dlang.org

### DIFF
--- a/spec/importc.dd
+++ b/spec/importc.dd
@@ -37,7 +37,7 @@ $(H2 $(LNAME2 examples, Quick Examples))
     hello world
     )
 
-    $(P C function in file $(TT square.c):)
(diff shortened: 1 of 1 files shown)


thewilsonator (approved):

adamdruppe (changes requested):
oh don't make the function name ugly, change the filename. I'd suggest something generic like "functions.c"

or fix the bad idea that is importC reusing D modules with naked top-level names.

dkorpel:
> oh don't make the function name ugly

I thought `sqr` would be a nice short name, it's how I call it in my code. It's sometimes used for square root though, so for clarity I'll take up your suggestion.

PR #3492, opened 2023-01-16
Fix possible data race with runnable example tester

ntrel:
https://issues.dlang.org/show_bug.cgi?id=15129 strikes again (like https://github.com/dlang/dmd/pull/14817).


dkorpel (commented):
`hasFailed = true;` on line 149 should become an atomic operation

dkorpel (approved):

ntrel:
Surely dmd should error when trying to assign to a shared variable if it's not safe, just like read-modify-write errors.

PR #3494, opened 2023-01-16
Fix Issue 23636 - No spec docs for shared qualifer

ntrel:


anon17:
Additions:

Mutable data that is meant to be shared among multiple threads should be declared with the `shared` qualifier. This prevents unsynchronized reading and writing to the data, which would otherwise cause data races. It's done by differentiating between shared and unshared data, so they can't implicitly substitute each other without a cast, and thus shared data can't be accidentally treated as unshared. The `shared` type attribute is transitive (like `const` and `immutable`).

Lastly, to declare global/static data to be implicitly shared across multiple threads without any compiler checks, see $(DDSUBLINK spec/attribute, gshared, `__gshared`). Such data has unshared type.

RazvanN7 (approved):

RazvanN7:
@anon17 feel free to submit a PR.

PR #3500, opened 2023-01-28
Fix &lt; and &gt; in C_HEADER_DESCRIPTION

ntrel:
E.g. at top of https://dlang.org/library/core/stdc/stdlib.html:
```
from the standard C header &lt;stdlib.h&gt;.
```

### DIFF
--- a/html.ddoc
+++ b/html.ddoc
@@ -175,7 +175,7 @@ _=
 _=Generic blurb to describe C header bindings
 
 C_HEADER_DESCRIPTION = This module contains bindings to selected types and
-functions from the standard C header $(HTTP $1, $(D $(LESS)$2$(GREATER))). Note that this
+functions from the standard C header $(HTTP $1, $(D <$2>)). Note that this
 is not automatically generated, and may omit some types/functions from the
 original C header.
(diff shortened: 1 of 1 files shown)


adamdruppe:
The change here used to be wrong in ddoc, I don't know if that's changed recently but make sure it generates correctly in regular ddoc too.

(or deprecate ddoc and only use good generators but since this is the official website thing it gotta work there for now)

ntrel:
Something weird is happening, the original actually looks OK the way I generate the HTML. **My change is wrong**. In the online page linked in the description, the link text has this HTML (at least in Firefox):
```
<code class="lang-d"><span class="pun">&amp;</span><span class="pln">lt</span><span class="pun">;</span><span class="pln">stdlib<wbr></span><span class="pun">.</span><span class="pln">h</span><span class="pun">&amp;</span><span class="pln">gt</span><span class="pun">;</span></code>
```
So the entity characters are getting broken up into `span`'d punctuation characters.

ntrel:
Filed: https://issues.dlang.org/show_bug.cgi?id=23659

PR #3501, opened 2023-01-28
Fix Issue 22418 - Error in documentation on strings

andraam21:


thewilsonator (commented):
Please leave the indentation as is was

RazvanN7:
Please add a commit that contains "Fix Issue 22418" in the commit message so that dlang-bot auto-closes the issue when the PR is merged.

andraam21 (commented):
Fix Issue 22418

spec/arrays.dd
```diff
@@ -1008,10 +1008,11 @@ $(H3 $(LNAME2 strings, Strings))
 $(SPEC_RUNNABLE_EXAMPLE_FAIL
 ---------
 char[] str1 = "abc";                // error, "abc" is not mutable
```
RazvanN7, line 1010: The bug report asks for a redesign of this example because currently the errors do not match. For example, the line above yields:

```
cannot implicitly convert expression `"abc"` of type `string` to `char[]`
```

spec/arrays.dd
```diff
@@ -1008,10 +1008,11 @@ $(H3 $(LNAME2 strings, Strings))
...
-char[] str2 = "abc".dup;            // ok, make mutable copy
+char[] str2 = "abc".dup;            // ok, makes mutable copy
 immutable(char)[] str3 = "abc";     // ok
-immutable(char)[] str4 = str1;      // error, str4 is not mutable
-immutable(char)[] str5 = str1.idup; // ok, make immutable copy
+immutable(char)[] str4 = str2;      // error, str4 is not mutable
```
RazvanN7, line 1013: This is not the error that the compiler outputs.

RazvanN7 (changes requested):
The PR does not fix the issue. The OP asks for a clarification on what happens here and in its current form the PR does not address that. To be accepted, this needs to:

- the comments must match the errors that the compiler outputs
- the difference between different combinations of immutable(char)[] or immutable(char[]) must be highlighted.

spec/arrays.dd
```diff
@@ -988,6 +988,16 @@ int[Color.max + 1] value =
...
+)
+
+        $(P All elements of a static array can be initialized to a specific value with:)
+
+$(SPEC_RUNNABLE_EXAMPLE_RUN
+---------
+int[4] a = 42; // set all elements of a to 42
+
+assert(a == [42, 42, 42, 42]);
+---------
```
RazvanN7, line 1000: This seems to be unrelated to this PR. Maybe you somehow screwed the rebase?

spec/arrays.dd
```diff
@@ -1021,14 +1032,37 @@ immutable(char)[] str5 = str1.idup; // ok, make immutable copy
...
+string str4 = str2;      // error, cannot implicitly convert expression `str2` of type `char[]` to `string`
+string str5 = str3;      // ok, makes a mutable str5 with immutable aray contents
+string str6 = str2.idup; // ok, makes immutable copy
 ---------
 )
 
+        $(P
+        Immutable is transitive, data referred to by an immutable is
+        also immutable:
+        )
```
RazvanN7, line 1047: This is to vague. Rephrase to: "The type `immutable(char)[]` represents an array of `immutable char`s. However, the reference to the string is mutable. If the the reference to the string needs to be `immutable` as well it can be declared `immutable char[]` or `immutable string`".

And then you can put your examples, there is no need for the second paragraph.

spec/arrays.dd
```diff
@@ -988,16 +988,6 @@ int[Color.max + 1] value =
...
-)
-
-        $(P All elements of a static array can be initialized to a specific value with:)
-
-$(SPEC_RUNNABLE_EXAMPLE_RUN
----------
-int[4] a = 42; // set all elements of a to 42
-
-assert(a == [42, 42, 42, 42]);
----------
```
RazvanN7, line 1000: this should not be deleted.

spec/arrays.dd
```diff
@@ -1031,13 +1022,29 @@ immutable(char)[] str5 = str1.idup; // ok, make immutable copy
...
 string str3 = "abc";     // ok
-string str4 = str1;      // error, str4 is not mutable
-string str5 = str1.idup; // ok, make immutable copy
+string str4 = str2;      // error, cannot implicitly convert expression `str2` of type `char[]` to `string`
+string str5 = str3;      // ok, makes a mutable str5 with immutable aray contents
+string str6 = str2.idup; // ok, makes immutable copy
 ---------
 )
+        $(P
+        The type immutable(char)[] represents an array of immutable chars. However, the reference to the string is
```
RazvanN7, line 1034: please highlight the compiler keyword : immutable(char)[], immutable string etc.

RazvanN7 (approved):
Squash the commits and we are ready to go

PR #3502, opened 2023-01-28
Fix Issue  6583 - cast() operation not fully specified

Robert21721:
Added some information and examples about pointers and the casting process
 https://issues.dlang.org/show_bug.cgi?id=6583

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -1145,15 +1145,31 @@ $(H4 $(LNAME2 cast_class, Class References))
     )
 
 $(H4 $(LNAME2 cast_pointers, Pointers))
-    $(P Casting a pointer)
+    $(P Casting a pointer modifies the type that will be obtained as a result of dereferencing,
+        along with the number of bytes on which the pointer arithmetic is performed.
+    )
+
+    $(P In the following example, variable val has 4 bytes and then its address is casted to char* in the ch variable.
+        Now, if ch is dereferenced, the printed value will be the character 'a' 
+        which corresponds to the integer value 97 (the least significant byte of val). If ch is incremented,
+        the printed value will be the character 'b' (the second least significant byte of val).
+    )
 
         $(SPEC_RUNNABLE_EXAMPLE_RUN
             -------------
-            int a = 353;
-            char *c = cast(char*) (&a);
-            writeln(*c);
+            int val = 25185; // 00000000 00000000 01100010 01100001
+            char *ch = cast(char*)(&val);
+
+            writeln(*ch);    // a
+            writeln(cast(int)(*ch)); // 97
+            writeln(*(ch + 1));  // b
+            writeln(cast(int)(*(ch + 1)));   // 98
             -------------
         )
+
+    $(P Here is an example of the reverse process. It will result in undefined behaviour due to the fact that 
+        the memory space following the c variables location has unknown content.
+    )
 
         $(SPEC_RUNNABLE_EXAMPLE_RUN
             -------------
@@ -1162,6 +1178,8 @@ $(H4 $(LNAME2 cast_pointers, Pointers))
             writeln(*p);
             -------------
         )
+
+    $(P The same thing can be done using a dynamically allocated array.)
 
         $(SPEC_RUNNABLE_EXAMPLE_RUN
             -------------
@@ -3022,4 +3040,3 @@ Macros:
         USUAL_ARITHMETIC_CONVERSIONS=$(DDSUBLINK spec/type, usual-arithmetic-conversions, Usual Arithmetic Conversions)
         INTEGER_PROMOTIONS=$(DDSUBLINK spec/type, integer-promotions, Integer Promotions)
         _=
-
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -1150,7 +1150,7 @@ $(H4 $(LNAME2 cast_pointers, Pointers))
     )
 
     $(P In the following example, variable val has 4 bytes and then its address is casted to char* in the ch variable.
-        Now, if ch is dereferenced, the printed value will be the character 'a' 
+        Now, if ch is dereferenced, the printed value will be the character 'a'
         which corresponds to the integer value 97 (the least significant byte of val). If ch is incremented,
         the printed value will be the character 'b' (the second least significant byte of val).
     )
@@ -1167,7 +1167,7 @@ $(H4 $(LNAME2 cast_pointers, Pointers))
             -------------
         )
 
-    $(P Here is an example of the reverse process. It will result in undefined behaviour due to the fact that 
+    $(P Here is an example of the reverse process. It will result in undefined behaviour due to the fact that
         the memory space following the c variables location has unknown content.
     )
 
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -1144,6 +1144,43 @@ $(H4 $(LNAME2 cast_class, Class References))
         (i.e. a reinterpret cast).
     )
 
+$(H4 $(LNAME2 cast_pointers, Pointers))
+    $(P Casting a pointer)
+
+        $(SPEC_RUNNABLE_EXAMPLE_RUN
+            -------------
+            int a = 353;
+            char *c = cast(char*) (&a);
+            writeln(*c);
+            -------------
+        )
+
+        $(SPEC_RUNNABLE_EXAMPLE_RUN
+            -------------
+            char c = 'a';
+            int *p = cast(int*) (&c);
+            writeln(*p);
+            -------------
+        )
+
+        $(SPEC_RUNNABLE_EXAMPLE_RUN
+            -------------
+            import core.stdc.stdlib;
+
+            int *p = cast(int*) malloc(5 * int.sizeof);
+            for (int i = 0; i < 5; i++) {
+                p[i] = i + 'a';
+            }
+            // p = {97, 98, 99, 100, 101}
+
+            char* c = cast(char*) p;     // c = {97, 0, 0, 0, 98, 0, 0, 0, 99 ...}
+            for (int i = 0; i < 5 * int.sizeof; i++) {
+                writeln(c[i]);
+            }
+            -------------
+        )
+
+
 $(H4 $(LNAME2 cast_array, Arrays))
 
     $(P Casting a dynamic array to another dynamic array is done only if the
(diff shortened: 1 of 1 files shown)


spec/expression.dd
```diff
@@ -1144,6 +1144,61 @@ $(H4 $(LNAME2 cast_class, Class References))
...
+
+        $(SPEC_RUNNABLE_EXAMPLE_RUN
+            -------------
+            import core.stdc.stdlib;
+
+            int *p = cast(int*) malloc(5 * int.sizeof);
+            for (int i = 0; i < 5; i++) {
+                p[i] = i + 'a';
+            }
+            // p = {97, 98, 99, 100, 101}
```
SixthDot, line 1192: use square brackets please, so that this looks like a D array literal

spec/expression.dd
```diff
@@ -1144,6 +1144,61 @@ $(H4 $(LNAME2 cast_class, Class References))
...
+
+            char* c = cast(char*) p;     // c = {97, 0, 0, 0, 98, 0, 0, 0, 99 ...}
```
SixthDot, line 1194: same

SixthDot (changes requested):

SixthDot:
you need to reword the first commit message from "fixed ...." to "fix ...." so that the bot get triggered.

SixthDot (approved):
sorry for the last suggestion, that has not worked, the PR looks formally OK to me however.

RazvanN7:
@Robert21721 changing the title of the PR is not going to trigger the bot. You need to add a commit (a.k.a git commit -m "Fix Issue 6583 - ...").

spec/expression.dd
```diff
@@ -1144,6 +1144,61 @@ $(H4 $(LNAME2 cast_class, Class References))
...
+$(H4 $(LNAME2 cast_pointers, Pointers))
+    $(P Casting a pointer modifies the type that will be obtained as a result of dereferencing,
+        along with the number of bytes on which the pointer arithmetic is performed.
+    )
+
+    $(P In the following example, variable val has 4 bytes and then its address is casted to char* in the ch variable.
+        Now, if ch is dereferenced, the printed value will be the character 'a'
+        which corresponds to the integer value 97 (the least significant byte of val). If ch is incremented,
+        the printed value will be the character 'b' (the second least significant byte of val).
+    )
```
RazvanN7, line 1156: I think this paragraph is a bit redundant. You don't need to explain every line of code, the example is fairly self explanatory.

spec/expression.dd
```diff
@@ -1144,6 +1144,61 @@ $(H4 $(LNAME2 cast_class, Class References))
...
+            writeln(*ch);    // a
+            writeln(cast(int)(*ch)); // 97
+            writeln(*(ch + 1));  // b
+            writeln(cast(int)(*(ch + 1)));   // 98
+            -------------
+        )
+
+    $(P Here is an example of the reverse process. It will result in undefined behaviour due to the fact that
+        the memory space following the c variables location has unknown content.
+    )
```
RazvanN7, line 1172: The above paragraph should simply state: "The following code results in undefined behavior due to accessing memory outside of the bounds of variable `c`".

spec/expression.dd
```diff
@@ -1144,6 +1144,61 @@ $(H4 $(LNAME2 cast_class, Class References))
...
+
+        $(SPEC_RUNNABLE_EXAMPLE_RUN
+            -------------
+            char c = 'a';
+            int *p = cast(int*) (&c);
+            writeln(*p);
+            -------------
+        )
+
+    $(P The same thing can be done using a dynamically allocated array.)
```
RazvanN7, line 1182: The wording is not suitable for technical prose. Avoid using expressions such as "the same thing".

In this case you could word it as: "The same outcome can be obtained by using a dynamically allocated array".

However, the example does not present a case of undefined behavior. No out of bounds access is performed, therefore this example is not "the same" as the previous one.

RazvanN7 (changes requested):
This does not fully address the issue at hand. The following need to be treated for this PR to be accepted:

- casting a pointer type to a basic type
- casting a signed to an unsigned and vice-versa


As suggested in the issue, maybe these could be treated in related to what C does.

spec/expression.dd
```diff
@@ -1100,6 +1100,66 @@ $(GNAME CastExpression):
         (foo) - p;      // subtract p from foo
         -------------
 
+$(H4 $(LNAME2 cast_basic_data_types, Basic Data Types))
+    $(P Cast between different types of data will not be performed automatically.
```
RazvanN7, line 1104: This is too vague. One might get the impression that an `uint` must be cast to a `long`, which is not the case. This needs to be explained thoroughly. You can state that where [implicit conversions](https://dlang.org/spec/type.html#implicit-conversions) cannot be applied, casts may be used to force the type system to accept the reinterpretation. The int/byte example below is not necessary as it is thoroughly treated in the above link.

Then  you just need to explain the different types of casts that exist.

spec/expression.dd
```diff
@@ -1100,6 +1100,66 @@ $(GNAME CastExpression):
...
+            ushort e = cast(ushort) a;   // 11111011 10000101
+
+            writeln(b);
+            writeln(c);
+            writeln(d);
+            writeln(e);
+            -------------
+        )
+
+    $(P In this case, the sign will be extended.)
```
RazvanN7, line 1131: Please add a generalizing statement. I.E: "when casting an integral type from a narrower type to a wider type ... bla bla ... sign extension ... bla bla ".

Do not add meaningless paragraphs like this one.

spec/expression.dd
```diff
@@ -1100,6 +1100,66 @@ $(GNAME CastExpression):
...
+            ushort c = cast(ushort) a;  // 00000000 10000101
+            short d = cast(short)b; // 11111111 10000101
+
+            writeln(c);
+            writeln(d);
+            -------------
+        )
+
+    $(P It is also possible to cast pointers to basic data types but this operation
+        does not provide control over the obtained result.
```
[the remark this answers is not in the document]
RazvanN7: Shouldn't this paragraph be merged with the "Pointers" section?

spec/expression.dd
```diff
@@ -1144,6 +1204,55 @@ $(H4 $(LNAME2 cast_class, Class References))
         (i.e. a reinterpret cast).
     )
 
+$(H4 $(LNAME2 cast_pointers, Pointers))
+    $(P Casting a pointer modifies the type that will be obtained as a result of dereferencing,
+        along with the number of bytes on which pointer arithmetic is performed.
+    )
```
RazvanN7, line 1210: Please be concise. You are talking about casting a pointer type to another pointer type. From your wording that is not obvious, sincer you can cast a pointer to another pointer or to a basic type, or to a class/struct (and the other way around).

spec/expression.dd
```diff
@@ -1144,6 +1204,55 @@ $(H4 $(LNAME2 cast_class, Class References))
...
+            char *ch = cast(char*)(&val);
+
+            writeln(*ch);    // a
+            writeln(cast(int)(*ch)); // 97
+            writeln(*(ch + 1));  // b
+            writeln(cast(int)(*(ch + 1)));   // 98
+            -------------
+        )
+
+    $(P The same outcome can be obtained by using a dynamically allocated array.)
```
RazvanN7, line 1224: This is too vague. What outcome are you referring to?

spec/expression.dd
```diff
@@ -1144,6 +1204,55 @@ $(H4 $(LNAME2 cast_class, Class References))
...
+
+            char* c = cast(char*) p;     // c = [97, 0, 0, 0, 98, 0, 0, 0, 99 ...]
+            for (int i = 0; i < 5 * int.sizeof; i++) {
+                writeln(c[i]);
+            }
+            -------------
+        )
+
+    $(P The following code results in undefined behavior due to accessing memory
+        outside of the bounds of variable c.
```
RazvanN7, line 1244: This should be a more general statement. For example: "When casting a pointer of type A to a pointer of type B and type B is wider than type A, if the portion of memory that exceeds the size of A is accessed, then that will result in undefined behavior.". This last sentence can still be improved, I meant it just as an example.

spec/expression.dd
```diff
@@ -1100,6 +1100,60 @@ $(GNAME CastExpression):
         (foo) - p;      // subtract p from foo
         -------------
 
+$(H4 $(LNAME2 cast_basic_data_types, Basic Data Types))
+    $(P There are some cases where the $(DDSUBLINK spec/type, implicit-conversions, implicit conversion)
+        cannot be applied, so cast may be used to force the type system to accept the reinterpretation.
```
RazvanN7, line 1105: Rephrase to: "For situations where $(DDSUBLINK spec/type, implicit-conversions, implicit conversions)" on basic types cannot be performed, the type system may be forced to accept the reinterpretation of a memory region by using a cast.

spec/expression.dd
```diff
@@ -1100,6 +1100,60 @@ $(GNAME CastExpression):
...
+    )
+
+    $(P This situation may occur, for example, when trying to assign a value stored on four bytes
+        to a variable stored on one byte. Without a cast, the following error will be displayed:
+        "cannot implicitly convert expression a of type int to byte".
```
RazvanN7, line 1110: Rephrase to: "An example of such a scenario is represented by trying to store a wider type into a narrower one:"

spec/expression.dd
```diff
@@ -1100,6 +1100,60 @@ $(GNAME CastExpression):
...
+    )
+
+    -------------
+    int a;
+    byte b = a; // error
```
RazvanN7, line 1115:
```suggestion
    byte b = a; // cannot implicitly convert expression a of type int to byte
```

spec/expression.dd
```diff
@@ -1100,6 +1100,60 @@ $(GNAME CastExpression):
...
+    -------------
+
+    $(P When the destination type is smaller than the source type,
```
RazvanN7, line 1118: "When casting a source type that is wider than the destination type"

spec/expression.dd
```diff
@@ -1100,6 +1100,60 @@ $(GNAME CastExpression):
...
+            ushort e = cast(ushort) a;   // 11111011 10000101
+
+            writeln(b);
+            writeln(c);
+            writeln(d);
+            writeln(e);
+            -------------
+        )
+
+    $(P Casting from a narrower type to a wider type could be done automatically,
```
RazvanN7, line 1137: What does "could be done automatically mean" ? Is it done or is it not done?

I assume you want to say: "For integral types casting from a narrower type to a wider type is done by performing sign extension."

RazvanN7 (changes requested):
Almost there.

RazvanN7 (approved):
Squash the commits and we are ready to go.

PR #3503, opened 2023-01-29
Add core.sys to the sidebar

ntrel:
Fix issue 5872.


ntrel:
Some html core.sys docs are missing, some don't have any content. DAutoTest failing. Probably best to close this.

PR #3505, opened 2023-01-30
Fix Issue 21132 - Duplicate keys in an AA literal

dumitrache-adrian92:
I specified the behavior of an associative array literal if pairs with the same key are declared and offered an example.

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -1907,6 +1907,16 @@ $(GNAME ValueExpression):
         are inserted as arguments in place of the sequence.
     )
 
+    $(P Associative array initializers may contain duplicate keys,
+        however, in that case, the last $(I KeyValuePair) lexicographically
+        encountered is stored.
+    )
+
+        ---
+        auto aa = [21: "he", 38: "ho", 2: "hi", 2:"bye"];
+        assert(aa[2] == "bye")
+        ---
+
 $(H3 $(LNAME2 function_literals, Function Literals))
 
 $(GRAMMAR


spec/expression.dd
```diff
@@ -1906,6 +1906,16 @@ $(GNAME ValueExpression):
...
         are inserted as arguments in place of the sequence.
     )
+    
+    $(P If there are duplicate keys in the $(I KeyValuePairs), then
+        the value returned for that key is going to be that of the
+        last $(I KeyValuePair) declared.
+    )
+    
+        ---
+        auto aa = [21u: "he", 38: "ho", 2: "hi", 2:"bye"];
```
RazvanN7, line 1916:
```suggestion
        auto aa = [21: "he", 38: "ho", 2: "hi", 2:"bye"];
```

We need the reader to focus on the important aspect. Having 21 as unsigned in this example is just noise.

spec/expression.dd
```diff
@@ -1906,6 +1906,16 @@ $(GNAME ValueExpression):
...
+        // aa[2] will be equal to "bye"
```
RazvanN7, line 1917: use assert instead of comment: `assert(aa[2] == "bye")`.

spec/expression.dd
```diff
@@ -1906,6 +1906,16 @@ $(GNAME ValueExpression):
         a $(I ValueSeq), then the elements of the $(I ValueSeq)
         are inserted as arguments in place of the sequence.
     )
+    
+    $(P If there are duplicate keys in the $(I KeyValuePairs), then
+        the value returned for that key is going to be that of the
+        last $(I KeyValuePair) declared.
```
RazvanN7, line 1912: I think it's worthwhile to mention that the duplicates are not actually stored, rather that the value for a specific key is updated to the last assigned value. Also, it's important to note that duplicates can be present in the AA initializer, but only the last lexicographically declared variable is stored.

RazvanN7 (changes requested):
Mind the comments

spec/expression.dd
```diff
@@ -1907,6 +1907,18 @@ $(GNAME ValueExpression):
         are inserted as arguments in place of the sequence.
     )
 
+    $(P You can declare duplicate keys in the $(I KeyValuePairs). The
+        value returned for that key is going to be that of the last
+        lexicographically $(I KeyValuePair) declared. This is because
+        duplicates are not actually stored, the value for that key is
+        just updated to the last assigned value.
+    )
```
RazvanN7, line 1915: The message is fine, but the wording is not quite adequate for technical prose. Avoid using "you" and try to use as much as possible the impersonal 3rd person:

"Associative array initializers may contain duplicate keys, however,  in that case, the last lexicographically $(I KeyValuePair) encountered is stored".

Much more concise and technical.

dumitrache-adrian92: Thanks for the advice!

RazvanN7 (approved):

RazvanN7:
@dumitrache-adrian92 congrats on your first PR. FYI: we are running a contest, whoever makes the most points from PRs gets a prize every 3 months. You can check your points here: https://bot.dlang.io/contributor_stats_cycle . The rules of the game are explicitly stated here: https://dlang.org/blog/2021/09/16/bugzilla-reward-system/

PR #3506, opened 2023-01-31
expound on what export means

WalterBright:


adamdruppe (approved):

adamdruppe:
this is accurate right now, a definition we probably want to revisit at some point but since it does what it says here we should merge.

dkorpel (approved):

rikkimax:
> Currently DllExport and internal symbol may be the same symbol. This will cause linker warnings on Windows [LNK4217](https://learn.microsoft.com/en-us/cpp/error-messages/tool-errors/linker-tools-warning-lnk4217) but can be silenced safely with ``/IGNORE:4217``.

I know this shouldn't be in the spec, but until we've done the export redesign it should go someplace.

PR #3512, opened 2023-02-01
Fix Issue 21178 - It is not explained what is "unknown"

Mateiuss:
Added an example of using an opaque enum (and also an error in using it).

### DIFF
--- a/spec/enum.dd
+++ b/spec/enum.dd
@@ -126,8 +126,7 @@ enum E : C
         $(P An empty enum body signifies an opaque enum - the enum members are unknown.)
         ---
         enum X;          // opaque enum
-        writeln(X.init); // error: enum x is opaque and has no default initializer
-                         // See 17.1.1 to understand what .init means
+        writeln(X.init); // error: enum X is opaque and has no default initializer
         ---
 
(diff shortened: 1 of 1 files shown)


spec/enum.dd
```diff
@@ -123,8 +123,12 @@ enum E : C
 ---
 )
 
-        $(P An empty enum body (For example $(CODE enum E;)) signifies an opaque
-        enum - the enum members are unknown.)
+        $(P An empty enum body signifies an opaque enum - the enum members are unknown.)
+        ---
+        enum X;          // opaque enum
+        writeln(X.init); // error: enum x is opaque and has no default initializer
+                         // See 17.1.1 to understand what .init means
```
RazvanN7, line 130: There's no need for this line.

RazvanN7:
Also, could you please have some information in the commit message? "Fix Issue 21178" does not offer any information.
At the very least you could put the title of the issue in the commit message.

spec/enum.dd
Mateiuss: I made a new commit where I erased that line. Also I added a proper message to the commit.

RazvanN7 (approved):

PR #3516, opened 2023-02-02
Fix issue 16707

AndreeaDrehuta:
Fixed an issue that seemed to be resolved and changed the names to match

### DIFF
--- a/spec/template.dd
+++ b/spec/template.dd
@@ -913,7 +913,7 @@ $(H4 $(LNAME2 homogeneous_sequences, Homogeneous Sequences))
 
         template print(Types...) // Types must be a TypeSeq
         {
-            void write(Types args) // args is a ValueSeq
+            void write_type(Types args) // args is a ValueSeq
             {
                 writeln("args are ", args);
             }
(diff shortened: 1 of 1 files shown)


spec/template.dd
```diff
@@ -892,15 +892,15 @@ $(H4 $(LNAME2 homogeneous_sequences, Homogeneous Sequences))
 
         template print(args...) // args must be a ValueSeq
         {
-            void f()
+            void write()
```
RazvanN7, line 895: just call this function `print`

spec/template.dd
```diff
@@ -892,15 +892,15 @@ $(H4 $(LNAME2 homogeneous_sequences, Homogeneous Sequences))
...
             {
                 writeln("args are ", args);
             }
         }
 
         void main()
         {
-            print!(1, 'a', 6.8).f(); // prints: args are 1a6.8
+            print!(1, 'a', 6.8).write(); // prints: args are 1a6.8
```
RazvanN7, line 903: and then you can simply do: "print!(1, 'a', 6.8)()"

spec/template.dd
```diff
@@ -913,15 +913,15 @@ $(H4 $(LNAME2 homogeneous_sequences, Homogeneous Sequences))
 
         template print(Types...) // Types must be a TypeSeq
         {
-            void f(Types args) // args is a ValueSeq
+            void write_type(Types args) // args is a ValueSeq
```
RazvanN7, line 916: same commentary as above.

RazvanN7 (approved):

PR #3518, opened 2023-02-02
Remove -dip1000 note from scope function section

GrimMaple:
-dip1000 was enabled by default, so this note is no longer relevant


RazvanN7:
> -dip1000 was enabled by default, so this note is no longer relevant

That is not quite accurate. -dip1000 is not enabled by default, however, you do receive deprecations for non-dip1000 compliant code. With -dip1000 an error is issued. I don't know if this difference needs to be highlighted in the spec.

dkorpel (approved):
Even though it raises only deprecation warnings by default, scope analysis is still being done without `-dip1000`, so the removal is accurate.

ntrel:
I couldn't find which release makes -dip1000 errors a deprecation by default. The example compiler seems to be 2.102.

ntrel:
BTW -dip1000 is still needed e.g. to return the address of a `ref return` parameter.

PR #3520, opened 2023-02-03
Issue 20997 - Missing example of scope guard executing after return statement

BVRazvan:
Issue link: https://issues.dlang.org/show_bug.cgi?id=20997

I added an example to clarify the order of execution when scope() statement interferes with return statement.

### DIFF
--- a/spec/statement.dd
+++ b/spec/statement.dd
@@ -1983,6 +1983,37 @@ $(CONSOLE
         return; nor may it be entered with a goto. A $(D scope(failure))
         statement may not exit with a return.
 
+$(SPEC_RUNNABLE_EXAMPLE_RUN
+--------------
+import std.stdio;
+
+int foo()
+{
+    scope(exit) writeln("Inside foo()");
+    return bar();
+}
+
+int bar()
+{
+    writeln("Inside bar()");
+    return 0;
+}
+
+int main()
+{
+    foo();
+    return 0;
+}
+--------------
+)
+
+        writes:
+
+$(CONSOLE
+Inside bar()
+Inside foo()
+)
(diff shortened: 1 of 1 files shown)


spec/statement.dd
```diff
@@ -1902,6 +1902,40 @@ or when it exits due to exception unwinding. $(D scope(failure)) executes
...
+
+        writes:
+
+$(CONSOLE
+Inside bar()
+Inside foo()
+)
+
+        A return lies inside the scope of a function; therefore, a $(D scope())
+        statement will be executed subsequently.
```
RazvanN7, line 1937: There's no need for this sentence.

spec/statement.dd
```diff
@@ -1902,6 +1902,40 @@ or when it exits due to exception unwinding. $(D scope(failure)) executes
...
+}
+--------------
+)
+
+        writes:
+
+$(CONSOLE
+Inside bar()
+Inside foo()
+)
```
RazvanN7, line 1934: Please move this example so that it is grouped with the others.

RazvanN7 (changes requested):
Please include the issue description in the commit message.

spec/statement.dd
BVRazvan: Changed it, please check.

spec/statement.dd
BVRazvan: Moved it, please check.

RazvanN7 (approved):

RazvanN7:
Please squash commits. The git history seems to be messed up.

RazvanN7 (approved):

PR #3523, opened 2023-02-04
Issue 16603 - Add explanation for the double-ending comment problem

AncaElenaNegoita:
Also modify the code to respect the lex structure.

### DIFF
--- a/spec/lex.dd
+++ b/spec/lex.dd
@@ -130,8 +130,15 @@ $(GNAME Comment):
     $(GLINK LineComment)
     $(GLINK NestingBlockComment)
 
+$(P In lex language, in order to concatenate, + is used instead of $(OPT) for a block comment.)
+$(P The given example with 2  $(D */)  $(D */) one following the other one will generate an error because the
+syntax is incorrect, there is an invalid sequence: the comment stops at the first  $(D */), and the other one
+is left outside. There can't be used all the Unicode characters because errors are generated depending on
+the chosen character. In this case, because a comment is represented by a text between $(D /**/), then when a third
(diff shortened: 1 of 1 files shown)


spec/lex.dd
```diff
@@ -130,8 +130,15 @@ $(GNAME Comment):
     $(GLINK LineComment)
     $(GLINK NestingBlockComment)
 
+$(P In lex language, in order to concatenate, + is used instead of $(OPT) for a block comment.)
```
RazvanN7, line 133: The reference to lex is irrelevant here. I don't really understand what you are trying to convey with this sentence.

spec/lex.dd
```diff
@@ -130,8 +130,15 @@ $(GNAME Comment):
...
+$(P The given example with 2  $(D */)  $(D */) one following the other one will generate an error because the
+syntax is incorrect, there is an invalid sequence: the comment stops at the first  $(D */), and the other one
+is left outside. There can't be used all the Unicode characters because errors are generated depending on
+the chosen character. In this case, because a comment is represented by a text between $(D /**/), then when a third
+one appears, the code doesn't know what it means and will be outside of the comment section. )
```
RazvanN7, line 138: This paragraph contains grammar mistakes and is overly complicated. Please try to simplify.

RazvanN7 (changes requested):
I don't think we should be changing the grammar here. The grammar is correct. Also, the first line on block comments is: "Block comments can span multiple lines, but do not nest."

I would argue that this issue should be closed as WONTFIX.

PR #3524, opened 2023-02-10
add documentation for getModuleClasses

WalterBright:
Implementation: https://github.com/dlang/dmd/pull/14699/

### DIFF
--- a/spec/traits.dd
+++ b/spec/traits.dd
@@ -60,6 +60,7 @@ $(GNAME TraitsKeyword):
     $(RELATIVE_LINK2 getLinkage, $(D getLinkage))
     $(RELATIVE_LINK2 getLocation, $(D getLocation))
     $(RELATIVE_LINK2 getMember, $(D getMember))
+    $(RELATIVE_LINK2 getModuleClasses, $(D getModuleClasses))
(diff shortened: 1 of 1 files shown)


spec/traits.dd
```diff
@@ -1330,6 +1330,31 @@ void main()
 ---
 )
 
+$(H3 $(GNAME getModuleClasses))
+
+        $(P Optionally takes one argument which is a module name or a fully qualified import.
```
adamdruppe, line 1335: What happens if you do `import a = std.stdio;`? Is `a` considered fully qualified at that point?

What if you use it on an imported symbol instead of an imported module?

It also isn't really a module name per se, but rather a module symbol, since it isn't quoted.

spec/traits.dd
```diff
@@ -1330,6 +1330,31 @@ void main()
...
+        If there is no argument, the current module is used.
+        It returns a tuple of strings representing the fully qualified names of classes defined in
```
adamdruppe, line 1337: What use case does this have?

dkorpel:
Closing, as the addition was reverted (https://github.com/dlang/dmd/pull/14892)

PR #3533, opened 2023-02-18
[spec] Show static assert `is` identifier example

ntrel:
The docs mention this but it's not clear what it means without an example.
Specify `is` identifier must be used in the first argument to introduce an identifier.

BTW, as a compiler enhancement, shouldn't the identifier still be in scope *after* the static assert statement?

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -2811,8 +2811,18 @@ $(H4 $(LNAME2 is-identifier, Identifier Forms))
         )
 
 $(SPEC_RUNNABLE_EXAMPLE_COMPILE
+---
+struct S
+{
+    int i;
+}
+static assert(is(typeof(S.i) T) && T.stringof == "int");
+---
+)
+$(SPEC_RUNNABLE_EXAMPLE_COMPILE
 -------------
 alias Bar = short;
+
 void foo()
 {
     static if (is(Bar T))
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -2799,7 +2799,8 @@ $(H4 $(LNAME2 is-identifier, Identifier Forms))
     $(P *Identifier* is declared to be an alias of the resulting
         type if the condition is satisfied. The *Identifier* forms
         can only be used if the $(I IsExpression) appears in a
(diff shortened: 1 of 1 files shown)


spec/expression.dd
```diff
@@ -2810,9 +2811,19 @@ $(H4 $(LNAME2 is-identifier, Identifier Forms))
         is declared to be an alias of $(I Type).
         )
 
+$(SPEC_RUNNABLE_EXAMPLE_COMPILE
+---
+struct S
+{
+    int i;
+}
+static assert(is(typeof(S.i) T) && T.stringof == "int");
```
dkorpel, line 2820:
```suggestion
static assert(is(typeof(S.i) T) && is(T == int));
```

Don't rely on the output of `.stringof`, it's implementation defined

dkorpel (changes requested):

spec/expression.dd
SixthDot: it's a bug in __traits(identifier)  in first place

```
template temp(T)
{
    pragma(msg, __traits(identifier, T));
}
alias Tint = temp!int; 
```

> Error: argument `int` has no identifier

dkorpel: What does `__traits(identifier)` have to do with this code example?

SixthDot: > What does `__traits(identifier)` have to do with this code example?

> Don't rely on the output of .stringof, it's implementation defined

dkorpel: You don't need a string here. My suggestion uses another `is()` expression.

SixthDot: that was not your initial argument, you initial argument was 

> Don't rely on the output of .stringof, it's implementation defined

dkorpel: Just because there's no utility for getting a string for built-in types doesn't mean `.stringof` isn't implementation defined anymore, and it doesn't validate this usage, especially considering you don't even need a string in the first place.

ntrel: @dkorpel changed to use `T.sizeof`. I wanted to show that T is valid after any `is` expression.

dkorpel (approved):

PR #3534, opened 2023-02-20
[spec] Improve `is(Type == keyword)` and Identifier docs

ntrel:
Describe `== TypeCtor` more precisely and add example. 
Remove `is(i == const)` example as even if `const int i;` was used it would still be false because `i` isn't a type.
Add note to prefer isModule and isPackage __traits (ideally `== module` and `== package` would be deprecated as they only pass when *Type* is not a type, `is()` is for checking types).
Improve wording.
Add example of type implicitly converting to match pattern.
Update `Identifier ==` example and add pattern example.
For `Identifier == Keyword` form, link to non-_Identifier_ form to avoid repeating the same info.

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -2795,7 +2795,7 @@ static assert(is(ModuleInfo == struct));
                 $(D immutable)
                 $(D inout)
                 $(D shared)
-        then the condition is satisfied if *Type* is declared with that *TypeCtor*:
+        then the condition is satisfied if *Type* is of that *TypeCtor*:
         )
 $(SPEC_RUNNABLE_EXAMPLE_COMPILE
 ---
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -2781,15 +2781,12 @@ $(SPEC_RUNNABLE_EXAMPLE_COMPILE
 ---
 static assert(is(Object == class));
 static assert(is(ModuleInfo == struct));
-static assert(is(object == module));
 ---
 )
-        $(NOTE The `module` and `package` forms are only satisfied when *Type* is not a type,
-        unlike the other forms. Prefer using $(DDSUBLINK spec/traits, isModule, isModule)
-        and $(DDSUBLINK spec/traits, isPackage, isPackage) `__traits` instead.
-
-        $(P $(DDSUBLINK
-        spec/module, package-module, Package modules) are considered to be both
+        $(P The `module` and `package` forms are only satisfied when *Type* is *not* a type,
+        unlike the other forms. The $(DDSUBLINK spec/traits, isModule, isModule)
+        and $(DDSUBLINK spec/traits, isPackage, isPackage) `__traits` should be used instead.
+        $(DDSUBLINK spec/module, package-module, Package modules) are considered to be both
         packages and modules.
         )
         $(P
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -2790,6 +2790,28 @@ static assert(is(ModuleInfo == struct));
         packages and modules.
         )
         $(P
+        *TypeSpecialization* can also be one of these keywords:
+        )
+        $(TABLE
+        $(THEAD keyword, condition)
+        $(TROW `super`, `true` if *Type* is a class or interface)
+        $(TROW `return`,
+            $(ARGS `true` if *Type* is a function, delegate or function pointer))
+        $(TROW `__parameters`,
+            $(ARGS `true` if *Type* is a function, delegate or function pointer))
+        )
+        $(SPEC_RUNNABLE_EXAMPLE_COMPILE
+        ---
+        class C {}
+        static assert(is(C == super));
+
+        void foo(int i);
+        static assert(!is(foo == return));
+        static assert(is(typeof(foo) == return));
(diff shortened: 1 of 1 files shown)


spec/expression.dd
```diff
@@ -2878,55 +2901,57 @@ void foo()
...
+        U u;
+    }
+    static assert(is(U == long));
+    ---
+    )
+        $(P
+        If *TypeSpecialization* is a keyword, the condition is satisfied
+        in the same manner as the
+        $(RELATIVE_LINK2 is-type-equal, `is(Type == Keyword)` form).
+        *TypeSpecialization* can also be one of these keywords:
```
ntrel, line 2943: I think these 3 keywords also work for the form without Identifier. `== return` does anyway.

RazvanN7 (commented):
Mind the nits.

spec/expression.dd
```diff
@@ -2773,27 +2773,41 @@ void foo()
...
-
-int i;
-static assert(!is(i == const));
+static assert(is(ModuleInfo == struct));
 ---
-        $(P $(DDSUBLINK
-        spec/module, package-module, Package modules) are considered to be both
+)
+        $(P The `module` and `package` forms are only satisfied when *Type* is *not* a type,
+        unlike the other forms. The $(DDSUBLINK spec/traits, isModule, isModule)
```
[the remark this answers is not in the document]
ntrel: Sorry I don't understand what situations you mean. I thought they were equivalent.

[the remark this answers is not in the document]
ntrel: I suppose there is a difference when the Identifier form is used, but that can be worked around easily with isModule. I would keep my wording, 'for such situations' doesn't seem to follow from the previous sentence.

spec/expression.dd
```diff
@@ -2773,27 +2773,41 @@ void foo()
                 $(D __vector)
                 $(D function)
                 $(D delegate)
-                $(D const)
-                $(D immutable)
-                $(D inout)
-                $(D shared)
                 $(D module)
                 $(D package)
```
thewilsonator, line 2777: what about `return` I had no idea that was a thing.

spec/expression.dd
```diff
@@ -2748,7 +2748,8 @@ void foo()
         $(H5 $(LNAME2 is-type-equal, $(D is $(LPAREN)) $(I Type) $(D ==) $(I TypeSpecialization) $(D $(RPAREN))))
```
thewilsonator, line 2748: I had always assumed that in `is(foo == keyword)` that foo was a symbol, not a type. 
```d
int x;
pragma(msg, is(x == float)); // false, also x is not a type.
```

spec/expression.dd
ntrel: Yes that is documented under the Identifier form, but I it seems to work without an Identifier too. Haven't tested `== super` and `== __parameters` yet.

spec/expression.dd
ntrel: AFAIK only `== module` and `== package` are satisfied when the argument is not a type. That's why they break the concept of IsExpression testing if a type is valid.

@RazvanN7 Perhaps they would be a good candidate for the [Language Simplification project](https://github.com/orgs/dlang/projects/11/views/1).

spec/expression.dd
thewilsonator: `is(foo == super)` and`is(foo ==  __parameters)` work very differently though, they bind `foo` to the superclass and parameters respectively IIRC

spec/expression.dd
thewilsonator: They really aren't, they have only one meaning, that meaning is obvious, and that form does not make sense to repurpose.

spec/expression.dd
ntrel: > they bind foo to the superclass and parameters respectively IIRC

No, they declare an optional Identifier. I have tested them now, they work without Identifier like `== return` does:
```d
class C {}
pragma(msg, is(C == super)); // true
C c;
pragma(msg, is(c == super)); // false

void foo(int i);
pragma(msg, is(foo ==  __parameters)); // false
pragma(msg, is(typeof(foo) ==  __parameters)); // true
```
I will move these 3 up from the Identifier form.

spec/expression.dd
ntrel: Then at minimum the grammar saying Type should be changed just for the module and package forms, because they do not accept a type, unlike all the forms which require only a type.

RazvanN7 (approved):

spec/expression.dd
ntrel: I suppose the grammar is technically OK, because Type includes QualifiedIdentifier which can accept ModuleFullyQualifiedName or PackageName, even though it's not actually a type.

ntrel: The source code actually states:
> if (e.tok2 == TOK.package_ || e.tok2 == TOK.module_) // These is() expressions are special because they can work on modules, not just types.

https://github.com/dlang/dmd/commit/9952fbb6c29#diff-f0a120d764c7fab59bd3210b19fe685d2f106e4a1b9a33266804ae99dd3bbd4eR5500

Special cases need strong justification to justify the cost of the complexity on the user.

PR #3537, opened 2023-02-22
[spec] Improve __traits(isSame)

ntrel:
Use 'the result is' not 'returns'.
Mention an alias may resolve to the same symbol.
Use static asserts instead of writeln.
Compare `object` instead of `std` so no import needed. 
Add example comparing expressions.
Split up large lambda example.

### DIFF
--- a/spec/traits.dd
+++ b/spec/traits.dd
@@ -1677,33 +1677,40 @@ void main()
 
 $(H3 $(GNAME isSame))
 
-        $(P Takes two arguments and returns bool $(D true) if they
-        are the same symbol, $(D false) if not.)
+        $(P Compares two arguments and evaluates to `bool`.)
+
+        $(P The result is `true` if the two arguments are the same symbol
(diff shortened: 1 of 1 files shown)


spec/traits.dd
```diff
@@ -1677,33 +1677,40 @@ void main()
...
 
-        $(P If the two arguments are expressions made up of literals
-        or enums that evaluate to the same value, true is returned.)
+        $(P The result is `true` if the two arguments are expressions
+        made up of literals or enums that evaluate to the same value.)
 
+$(SPEC_RUNNABLE_EXAMPLE_COMPILE
+---
+enum e = 3;
+static assert(__traits(isSame, (e), 3));
```
ntrel, line 1710: Note: Parens on `e` are needed to do expression value comparison - see https://issues.dlang.org/show_bug.cgi?id=20762.

dkorpel (approved):

PR #3552, opened 2023-03-09
Bind to external classes instance properties

MrcSnm:


spec/objc_interface.dd
```diff
@@ -43,6 +43,26 @@ $(HEADERNAV_TOC)
...
+        NSString label() @selector("label");
+        NSString label(NSString) @selector("setLabel:");
+    }
+    ---
+
+    $(P
+        Sometimes Objective-C classes will be documented having properties. Those
+        properties must be bound by using the the property name as a selector for its
+        getter and camel cased version of its name with `set` whenever setting it via D code.
+        This behavior doesn't seem to be documented on Apple's documents.
```
dkorpel, line 63: I suppose it's better than no documentation, but this kind of uncertain language does not fit the specification. Do you know any other D + Objective-C users that can review this information?

jacob-carlborg: The information is correct. It's documented here [1] under the "Use Accessor Methods to Get or Set Property Values" header. Although I think the wording in this PR could be improved, but English is not my native language.

[1] https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/EncapsulatingData/EncapsulatingData.html#//apple_ref/doc/uid/TP40011210-CH5-SW1

MrcSnm: I'll use the same word as the documentation on your linked resource then

dkorpel (approved):

PR #3558, opened 2023-03-17
[spec/attribute] Group attributes

ntrel:
Add groups Mutability, Shared Storage, Function.
Note: Safety attributes aren't included in Function Attributes because `@system` applies to variables (not documented in spec yet). 
Add link for `auto`.
Add Class Attributes group, move `override` here.


dkorpel (approved):

ntrel:
Rebased (no changes).

PR #3560, opened 2023-03-18
[spec/attribute] Tweak groups

ntrel:
Follows on from #3558.

Add subheading for synchronized under Shared Storage Attributes rather than just a link in the grammar. That way it appears in the TOC. 
Rename _Class Attributes_ to _OOP Attributes_ as some of them also apply to interfaces.

### DIFF
--- a/spec/attribute.dd
+++ b/spec/attribute.dd
@@ -31,7 +31,7 @@ $(GNAME Attribute):
     $(RELATIVE_LINK2 scope, $(D scope))
     $(RELATIVE_LINK2 shared, $(D shared))
     $(RELATIVE_LINK2 static, $(D static))
-    $(DDSUBLINK spec/class, synchronized-classes, `synchronized`)
+    $(RELATIVE_LINK2 synchronized, `synchronized`)
 
 $(GNAME FunctionAttributeKwd):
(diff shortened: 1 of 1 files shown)


spec/attribute.dd
```diff
@@ -905,7 +909,7 @@ void main() @nogc
 )
 
 
-$(H2 $(LNAME2 class-attributes, Class Attributes))
+$(H2 $(LNAME2 oop-attributes, OOP Attributes))
```
dkorpel, line 912: I don't like this rename, it breaks existing links to this part of the spec, and I think it's clear that interfaces fall under the category of classes.

ntrel: I can keep the anchor (it was only added just before I submitted this). The reason I changed heading to OOP is because synchronized is only a class or method attribute, so you might expect to find it under class attributes rather than shared storage attributes.

ntrel, line 912:
```suggestion
$(H2 $(LNAME2 class-attributes, OOP Attributes))
```

dkorpel: > it was only added just before I submitted this

Oh okay, nevermind then.

PR #3561, opened 2023-03-19
Don't wrap RUNNABLE_EXAMPLE_COMPILE code in `void main`

ntrel:
Show 'Compile' button instead of 'Run' for _EXAMPLE_COMPILE.
Also add tooltip for Run explaining code is wrapped in main if missing.
Remove `void main() {}` workarounds in some _COMPILE examples.
Fix _COMPILE examples to wrap code in a function when needed.


js/run.js
```diff
@@ -179,13 +179,17 @@ function parseOutput(res, o, oTitle)
...
+function wrapIntoMain(code, compile) {
     var currentPackage = $('body')[0].id;
 
     // dynamically wrap into main if needed
     if (code.indexOf("void main") >= 0 || code.indexOf("int main") >= 0) {
         return code;
     }
+    else if (compile) {
+        // avoid missing main link error
+        return code + "\nvoid main() {}";
```
ntrel, line 191: The Compile button does run the code too (perhaps I should add a tooltip to state that). That's useful for examples the user may want to execute but that the test suite doesn't need to (e.g. writeln).

Compile currently doesn't auto import stdio, though perhaps it should.

ntrel: I think compile shouldn't auto-import std.stdio as usually that's not needed.

CyberShadow:
What motivates this change?

I suppose the obvious disadvantage is that many examples now have a bit of boilerplate. What is the advantage?

dkorpel:
> What motivates this change?

https://github.com/dlang/dlang.org/pull/3557#discussion_r1140379645

ntrel:
> many examples now have a bit of boilerplate. What is the advantage?

* Testing a declaration declared inside a function is not the same as testing a declaration in module scope. Nested declarations might need a `static` attribute (to be equivalent) which is not needed at global scope, and this is easily forgotten.
* Some declarations are not valid inside a function, so wrapping them in main causes errors. This removes workarounds declaring a dummy main in some examples.

CyberShadow:
Thanks!

Then I guess I will leave this to @WalterBright, as the primary author of the specification.

spec/type.dd
```diff
@@ -476,36 +479,42 @@ $(H3 $(LNAME2 vrp, Value Range Propagation))
...
-    //ubyte k = i & 0x14A; // error, 0x14A > ubyte.max
-    ushort k = i & 0x14A; // OK
-
-    extern ubyte b;
-    k = i & b; // OK, 0 ... b.max
-    //b = b + b; // error, b.max + b.max > b.max
-    s = b + b; // OK, 0 ... b.max + b.max
+    void fun()
+    {
+        extern char c;
```
dkorpel, line 484: extern doesn't make sense on local variables

ntrel: Now changed to use parameters instead.

dkorpel (approved):

PR #2857, opened 2020-09-24
Update the grammar for named arguments

Hackerpilot:
https://github.com/dlang/DIPs/blob/master/DIPs/accepted/DIP1030.md

PR #3565, opened 2023-03-23
Add grammar for named function arguments

Bolpat:
Named arguments are implemented in https://github.com/dlang/dmd/pull/14575 and https://github.com/dlang/dmd/pull/14776.


dkorpel:
The grammar in the DIP has been superseded, there's now a separate NamedArgumentList so you can't have named arguments inside a mixin, pragma etc.

dkorpel:
I think https://github.com/dlang/dlang.org/pull/2857 has the correct grammar, just needs to be rebased.

Bolpat:
Didn’t see the PR. I’ll look into it soon.

PR #3568, opened 2023-03-24
Fix Issue 23807 - FunctionLiteral spec should not use Parameters

ntrel:
https://dlang.org/spec/expression.html#FunctionLiteral
https://dlang.org/spec/function.html#Parameters

A literal's parameter can just be a _FundamentalType_ or `...` but in all other cases an identifier is required. This specifies the current dmd behaviour. Some test code:
```d
void main()
{
    alias Int = int;
    alias a = (const i, Int, const...) { writeln(Int); }; // Int is a variable, not a type
    alias b = (short, const int k = 1) { writeln(k); }; // removing `int` for k is an error
    a(4,3);
    b(3);
    b(3,2);
}
```

Also use list for description and add item for omitted parameter types.

function.dd: Merge two variants of _ParameterList_ grammar using OPT.

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -2056,10 +2056,24 @@ $(GNAME FunctionLiteral):
     $(IDENTIFIER) $(D =>) $(GLINK AssignExpression)
 
 $(GNAME ParameterWithAttributes):
-    $(GLINK2 function, Parameters) $(GLINK2 function, FunctionAttributes)$(OPT)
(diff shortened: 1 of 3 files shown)


dkorpel:
I see semantic differences indeed, but are the two lists really parsed differently?

> removing `int` for k is an error

A semantic error, not a parsing error though

ntrel:
OK, moved other changes to #3568.

PR #3569, opened 2023-03-24
Make version drop-down arrow clickable

ntrel:
Otherwise only the text is clickable to show the list of options.

E.g. at the top right:
https://dlang.org/spec/expression.html
image

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1706,6 +1706,7 @@ div.openInEditorButton i{
    font-family: FontAwesome;
    content: "\f0d7";
    margin-left: -15px;
+   pointer-events: none; /* propagate click */
 }
 /* Runnable-examples css -end */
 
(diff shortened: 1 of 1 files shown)


CyberShadow:
The PR title and commit message say "Make version drop-down arrow clickable". But this just removes the arrow altogether, doesn't it?

To make it clickable, you can add `pointer-events: none;` to make clicks go through it and to the `select` element.

ntrel:
> The PR title and commit message say "Make version drop-down arrow clickable". But this just removes the arrow altogether, doesn't it?

It actually removed the style so the default caret got shown. I then changed the style so 'master' didn't get cut off.

> To make it clickable, you can add pointer-events: none; to make clicks go through it and to the select element.

OK thanks, reverted my other changes. Not sure the best way to fix width getting cut off.

dkorpel (approved):

PR #3571, opened 2023-03-31
IsExpression: comment out alias template instance with parameters

ntrel:
Needed for https://github.com/dlang/dmd/pull/15053. After merging that, it will be an error.


dkorpel:
Also needed for https://github.com/dlang/dmd/pull/15085, so merging now

dkorpel (approved):

PR #3572, opened 2023-03-31
[AssertExpression] Improve docs

ntrel:
Change link to use precondition spec in function.dd (contracts spec is a stub). 
Fix broken anchor in contracts.dd.
Add class invariant link.
Move compile-time evaluated condition paragraph down to its own subheading (avoid distracting from the runtime evaluation docs which are quite complicated).
Add short example.
Move message argument sentence down to its own subheading. 
Reword paragraph about compiler switches. Remove 'continuing execution' list item - that seems to be covered by the first sentence (whether the condition is evaluated or not).
Move best practice section up.
For CT evaluation, avoid mentioning 'release mode' which was undefined.
Add link to static assert.

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -2383,31 +2383,26 @@ $(GNAME AssertArguments):
     $(GLINK AssignExpression) $(D ,) $(GLINK AssignExpression) $(D ,)$(OPT)
 )
 
-    $(P The first $(I AssignExpression) must evaluate to true. If it does not, an $(I Assert Failure)
+    $(P The first $(I AssignExpression) must evaluate to `true`. If it does not, an $(I Assert Failure)
     has occurred and the program enters an $(I Invalid State).
     )
 
-    $(P If the first $(I AssignExpression) consists entirely of compile time constants,
-    and evaluates to false, it is a special case; it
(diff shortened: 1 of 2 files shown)


dkorpel (approved):

spec/expression.dd
```diff
@@ -2417,21 +2412,51 @@ $(GNAME AssertArguments):
...
     )
-    If the optional second $(I AssignExpression) is provided, the implementation may
-    evaluate it and print the resulting message upon assert failure:
 
+    $(P See also: $(DDSUBLINK spec/version, static-assert, `static assert`).)
+
+$(H4 $(LNAME2 assert-message, Assert Message))
+
+    $(P The second $(I AssignExpression), if present, must be implicitly
+        convertible to type $(D const(char)[]).
```
Lyratelle, line 2456: assert can now have multiple parameters, each of which is converted to string and concatenated. So this sentence should start with "any following AssignExpression, if present, ...

ntrel: @Lyratelle No that only works with `static assert`. Otherwise `dmd` would have to link in a runtime function to e.g. convert int to string.

PR #3577, opened 2023-04-06
make AA "Not yet Implemented" note better

WebFreak001:
wasn't able to build, since the makefile seems to be broken nowadays, so just trusting ye old CI here.

See https://forum.dlang.org/thread/kqfkzghjoyruzgbsvmje@forum.dlang.org

### DIFF
--- a/spec/hash-map.dd
+++ b/spec/hash-map.dd
@@ -361,9 +361,13 @@ $(SPEC_RUNNABLE_EXAMPLE_RUN
 
 $(H2 $(LNAME2 static_initialization, Static Initialization of AAs))
 
-        NOTE: Not yet implemented.
+        NOTE: Static initialization, e.g. assigning an AA value to a global or
+        static variable, isn't implemented yet. AAs can be constructed and used
+        at runtime and at compile time as usual. Pre-computed AA values embedded
+        into the executable are however not yet possible.
 
         ---------
+        // in global scope
         immutable long[string] aa = [
           "foo": 5,
           "bar": 10,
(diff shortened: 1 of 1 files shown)


spec/hash-map.dd
```diff
@@ -361,9 +361,13 @@ $(SPEC_RUNNABLE_EXAMPLE_RUN
 
 $(H2 $(LNAME2 static_initialization, Static Initialization of AAs))
 
-        NOTE: Not yet implemented.
+        NOTE: Static initialization, e.g. assigning a global variable with some
```
SixthDot, line 364: I think the wording is not great, I propose that instead

"Initialization using an AA literal, computed during compilation, is not implemented, instead initialization must be done at run-time, for example in a static constructor."

Can be amended ;)

but otherwise I think that another problem is that this should be a specification of the array literal, not of the AA, i.e the restriction should be in https://dlang.org/spec/expression.html#associative_array_literals.

WebFreak001: it's also other things than just AA literals, any AA value generated at compile time.

dkorpel, line 364:
```suggestion
        NOTE: Static initialization, i.e. assigning a global variable with some
```
I think "assigning a global variable" is the only example of static initialization, or are there others?

dkorpel (approved):

spec/hash-map.dd
WebFreak001: `switch (foo) { case ["foo": "bar"]: x; }` would be one I think, but I don't think switches are even usable with AAs.

`static` variables would be another example, other than global variables. (even if they are basically the same)

WebFreak001: improved the wording, included `static` as example.

WebFreak001: improved the wording.

WebFreak001:
ping

PR #3588, opened 2023-04-23
clarify what isIntegral, isScalar, etc. __traits mean

WalterBright:
Try to clear away the fog over precisely what these traits mean.


dkorpel (approved):
I don't think deprecated complex types and unimplemented `cent` types need to be mentioned, but it doesn't hurt either.

PR #3605, opened 2023-05-12
Specify that nested static struct might also be POD

SixthDot:
because

```d
struct S
{
    static struct T
    {

    }
}

static assert(__traits(isPOD, S.T)); // OK
```

### DIFF
--- a/spec/struct.dd
+++ b/spec/struct.dd
@@ -211,7 +211,7 @@ $(H2 $(LNAME2 POD, Plain Old Data))
     $(P A struct or union is $(I Plain Old Data) (POD) if it meets the following criteria:)
 
     $(OL
-    $(LI it is not nested or static)
+    $(LI it is static, or not nested)
     $(LI it has no postblits, copy constructors, destructors, or assignment operators)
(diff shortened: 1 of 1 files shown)


RazvanN7 (approved):

spec/struct.dd
```diff
@@ -211,7 +211,7 @@ $(H2 $(LNAME2 POD, Plain Old Data))
     $(P A struct or union is $(I Plain Old Data) (POD) if it meets the following criteria:)
 
     $(OL
-    $(LI it is not nested)
+    $(LI it is not nested or static)
```
RazvanN7, line 214: This sounds weird. Maybe a more correct phrasing would be: "it is not nested or if it is nested it is declared as `static`".

dkorpel: Indeed, the current wording could be interpreted as either "not (nested or static)" or "(not nested) or static"

SixthDot: ok, updated.

dkorpel (approved):

PetarKirov (approved):

PetarKirov (approved):

PR #3607, opened 2023-05-17
add ref-return-scope table

WalterBright:
The table with all the permutations should help.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -4248,7 +4248,7 @@ Macros:
         TITLE=Functions
         ASSIGNEXPRESSION=$(GLINK2 expression, AssignExpression)
         CHECK=&#10004;
-        REFP=$(LINK2 #ref-params, `ref`)
-        RETURN_REF=$(LINK2 #return-ref-parameters, `return ref`)
-        SCOPE=$(LINK2 #scope-parameters, `scope`)
-        RETURN_SCOPE=$(LINK2 #return-scope-parameters, `return scope`)
+        REFP=$(LINK2 #ref-params, $(D ref))
+        RETURN_REF=$(LINK2 #return-ref-parameters, $(D return ref))
+        SCOPE=$(LINK2 #scope-parameters, $(D scope))
+        RETURN_SCOPE=$(LINK2 #return-scope-parameters, $(D return scope))


maxhaton:
A table is good but much better is to enforce a canonical order for each semantic expressible

WalterBright:
> Is there a good reason why these permutations even exist?

Legacy code compatibility.

> A table is good but much better is to enforce a canonical order for each semantic expressible

Possibly true, but who wants to update all their code?

pbackus:
iirc the eventual plan is to require that `return` be written directly before the attribute it modifies (either as `return ref` or `return scope`), but disallowing `ref return` is a breaking change, so for now we're stuck with this awkwardness.

I think the best way to explain this stuff is to start with just `ref` and `scope`, and then introduce `return` as a modifier that can apply to either `ref` or `scope` (but not both). Giving all of the possible combinations equal standing in a table (including combinations where the attributes do not interact at all, like `scope ref`) obscures the underlying semantic rules, and is more likely to make the reader's eyes glaze over than to teach them anything.

Happy to take a crack at writing some documentation for this myself if desired.

WalterBright:
> A table is good but much better is to enforce a [...]

> the eventual plan is to require [...]

This PR is documenting the way D behaves now. Initiatives to improve the language should be a separate discussion. The table expresses the meaning better than the prose did.

spec/function.dd
```diff
@@ -1905,9 +1905,32 @@ $(H3 $(LNAME2 ref-return-scope-parameters, Ref Return Scope Parameters))
...
         keywords appear adjacent to each other, in that order.
-        Specifying a `return ref` and `scope` parameter enables returning a reference to a scope pointer.
-        In all other cases, the parameter has $(LINK2 #return-ref-parameters, `return ref`) semantics
-        and regular $(LINK2 #scope-parameters, `scope`) semantics.)
+        This behavior is summarized as follows:)
+
+        $(TABLE2 Combinations of `ref` $(COMMA) `return` and `scope`,
+        $(THEAD Order, Interpretation)
+        $(TROW `ref`,              $(REFP))
+        $(TROW `return`,           `return`)
```
dkorpel, line 1913: `return` is not an interpretation
```suggestion
        $(TROW `return`,           $(RETURN_SCOPE))
```

spec/function.dd
```diff
@@ -4230,3 +4248,7 @@ Macros:
         TITLE=Functions
         ASSIGNEXPRESSION=$(GLINK2 expression, AssignExpression)
         CHECK=&#10004;
+        REFP=$(LINK2 #ref-params, `ref`)
+        RETURN_REF=$(LINK2 #return-ref-parameters, `return ref`)
+        SCOPE=$(LINK2 #scope-parameters, `scope`)
+        RETURN_SCOPE=$(LINK2 #return-scope-parameters, `return scope`)
```
PetarKirov, line 4254:
```suggestion
        REFP=$(LINK2 #ref-params, $(D ref))
        RETURN_REF=$(LINK2 #return-ref-parameters, $(D return ref))
        SCOPE=$(LINK2 #scope-parameters, $(D scope))
        RETURN_SCOPE=$(LINK2 #return-scope-parameters, $(D return scope))
```

spec/function.dd
WalterBright: It depends on the context. `See mtype.buildScopeRef`

spec/function.dd
```diff
@@ -1905,9 +1905,32 @@ $(H3 $(LNAME2 ref-return-scope-parameters, Ref Return Scope Parameters))
...
+        $(TROW `return scope`,     $(RETURN_SCOPE))
+        $(TROW `scope return`,     $(RETURN_SCOPE))
+        $(TROW `ref scope`,        $(REFP) and $(SCOPE))
+        $(TROW `scope ref`,        $(REFP) and $(SCOPE))
+        $(TROW `ref return scope`, $(REFP) and $(RETURN_SCOPE))
+        $(TROW `return scope ref`, $(REFP) and $(RETURN_SCOPE))
+        $(TROW `ref scope return`, $(RETURN_REF) and $(SCOPE))
+        $(TROW `return ref scope`, $(RETURN_REF) and $(SCOPE))
+        $(TROW `scope return ref`, $(RETURN_REF) and $(SCOPE))
+        $(TROW `scope ref return`, $(RETURN_REF) and $(SCOPE))
```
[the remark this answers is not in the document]
WalterBright: Combining things to reduce the lines in the table is less clear

spec/function.dd
WalterBright: The back tick does the same thing as $(D ).

adamdruppe: False.

dukc:
> iirc the eventual plan is to require that return be written directly before the attribute it modifies (either as return ref or return scope), but disallowing ref return is a breaking change, so for now we're stuck with this awkwardness.

Well once we have language editions this can be done. It's otherwise a great plan IMO, but has one problem. How would `return` be applied to the `this` reference of a struct or union? `void memFun() return ref {/+...+/}` does not parse.

PR #3610, opened 2023-05-18
Fix Issue 23922 - Fix dmd/compiler/samples broken links

RazvanN7:
cc @CyberShadow


dkorpel (approved):

CyberShadow (approved):

CyberShadow:
Hah, so `std.socket` was the macro's only user after all.

RazvanN7:
Yep.

RazvanN7:
Thanks anyway, TIL.

PR #3611, opened 2023-05-19
add article on ref, return, and scope annotations

WalterBright:
This article is intended to be more of an informal tutorial about how ref, return and scope annotations work rather than a formal specification.

This PR is the first part.

The second part will be a tutorial on the various ways pointers can escape.

The third part will be about common problems with it and how to resolve them.

### DIFF
--- a/articles/RefReturnScope.dd
+++ b/articles/RefReturnScope.dd
@@ -200,7 +200,7 @@ a scoped value, no matter how twisty the code is.)
 )
 
 Macros:
-        TITLE=Coralling Wild Pointers With $D(ref return scope)
+        TITLE=Coralling Wild Pointers With $(D ref return scope)
         ITEMR=$(LI $(RELATIVE_LINK2 $1, $+))
         ITEM=<hr>$(H3 <a name="$1">$+</a>)
         SUBNAV=$(SUBNAV_ARTICLES)
--- a/articles/RefReturnScope.dd
+++ b/articles/RefReturnScope.dd
@@ -200,7 +200,7 @@ a scoped value, no matter how twisty the code is.)
 )
 
 Macros:
-        TITLE=Coralling Wild Pointers With `ref return scope`
+        TITLE=Coralling Wild Pointers With $D(ref return scope)
         ITEMR=$(LI $(RELATIVE_LINK2 $1, $+))
         ITEM=<hr>$(H3 <a name="$1">$+</a>)
         SUBNAV=$(SUBNAV_ARTICLES)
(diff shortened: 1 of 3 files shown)


PetarKirov:
I've added the article to the makefile as it wasn't getting compiled.

articles/RefReturnScope.dd
```diff
@@ -0,0 +1,207 @@
...
+
+$(P Curses! Foiled again!)
+
+$(P The operational idea here is, while compiling @safe code, it is not be possible to escape
+a scoped value, no matter how twisty the code is.)
+
+)
+
+Macros:
+        TITLE=Coralling Wild Pointers With `ref return scope`
```
PetarKirov, line 203:
```suggestion
        TITLE=Coralling Wild Pointers With $D(ref return scope)
```

articles/RefReturnScope.dd
```diff
@@ -0,0 +1,207 @@
...
+    int i;
+    int* p = &i; // p is now inferred to be scope
+    auto q = fin(p);  // q now contains the address of i, and so scope is also inferred
+    return q;  // error: scope variable `q` may not be returned
+}
+---
+
+$(P Curses! Foiled again!)
+
+$(P The operational idea here is, while compiling @safe code, it is not be possible to escape
```
PetarKirov, line 197:
```suggestion
$(P The operational idea here is, while compiling `@safe` code, it is not be possible to escape
```

articles/RefReturnScope.dd
```diff
@@ -0,0 +1,207 @@
...
+
+$(P Curses! Foiled again!)
+
+$(P The operational idea here is, while compiling `@safe` code, it is not be possible to escape
+a scoped value, no matter how twisty the code is.)
+
+)
+
+Macros:
+        TITLE=Coralling Wild Pointers With $D(ref return scope)
```
PetarKirov, line 203:
```suggestion
        TITLE=Coralling Wild Pointers With $(D ref return scope)
```

PetarKirov (approved):

articles/RefReturnScope.dd
WalterBright: The backticks should do the same thing.

PetarKirov: I know they do in general in DDoc, but I looked at the rendered HTML and
`` `ref return scope` `` appeared verbatim in the page title. This is possibly a bug.

PetarKirov: I found an existing issue: https://issues.dlang.org/show_bug.cgi?id=18703

adamdruppe: This is an intentional part of the design: the backticks are not processed in macros so it is possible to define macros that give access to literal backticks.

CyberShadow:
I suggest checking the rendered HTML in a web browser before merging:
https://github.com/dlang/dlang.org/pull/3628#issuecomment-1576093537

PR #3612, opened 2023-05-19
move glossary to spec

WalterBright:
It turns out that glossary.html is unreachable via the tables of contents. This fixes that by moving it into the spec. A glossary is useless if nobody can find it.


CyberShadow:
> It turns out that glossary.html is unreachable via the tables of contents.

It's in the navigation ("Resources" menu), FWIW.

CyberShadow:
Could you please add a redirect in `.htaccess`, so that old external links don't get broken?

WalterBright:
@CyberShadow done

CyberShadow:
Argh, the redirect is wrong.

PR #3614, opened 2023-05-23
Move Template(Mixin)Declaration to Declaration in grammar

tim-dlang:
This allows to declare them inside functions.
For mixin templates this is supported by DMD since https://github.com/dlang/dmd/pull/14240
Declaring normal templates in functions is also supported by DMD.

### DIFF
--- a/spec/module.dd
+++ b/spec/module.dd
@@ -31,8 +31,6 @@ $(GNAME DeclDef):
     $(GLINK2 version, DebugSpecification)
     $(GLINK2 version, VersionSpecification)
     $(GLINK2 version, StaticAssert)
-    $(GLINK2 template, TemplateDeclaration)
-    $(GLINK2 template-mixin, TemplateMixinDeclaration)
     $(GLINK2 template-mixin, TemplateMixin)
     $(GLINK MixinDeclaration)
     $(GLINK EmptyDeclaration)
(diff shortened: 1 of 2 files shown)


spec/module.dd
```diff
@@ -31,8 +31,6 @@ $(GNAME DeclDef):
     $(GLINK2 version, DebugSpecification)
     $(GLINK2 version, VersionSpecification)
     $(GLINK2 version, StaticAssert)
```
ntrel, line 33: Looks like StaticAssert can be removed here as it's in Declaration.

spec/module.dd
```diff
@@ -31,8 +31,6 @@ $(GNAME DeclDef):
...
-    $(GLINK2 template, TemplateDeclaration)
-    $(GLINK2 template-mixin, TemplateMixinDeclaration)
     $(GLINK2 template-mixin, TemplateMixin)
     $(GLINK MixinDeclaration)
```
ntrel, line 35: TemplateMixin and MixinDeclaration can be moved to Declaration too.

ntrel (approved):
While reviewing I noticed a few things that can also be updated (though I don't think those are necessary for this pull to get merged).

spec/module.dd
tim-dlang: Moving MixinDeclaration to Declaration could create an ambiguity with MixinStatement.
But it should be possible to move TemplateMixin if it is also removed from NonEmptyStatementNoCaseNoDefault.

ntrel: OK, thanks.

dkorpel (approved):

spec/module.dd
ntrel: Now moved TemplateMixin: #3660

PR #3619, opened 2023-05-31
[spec/function] Improve `return` struct method attribute docs

ntrel:
Fixes Issue 17934 - [scope] scopeness entrypoint for unique/ref-counted missing.

See https://issues.dlang.org/show_bug.cgi?id=17934#c12 for why this feature fixes that issue.

cc @dukc

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -1725,24 +1725,6 @@ ref int gun(return ref int x) {
 }
 ---
 )
-        $(P Struct non-static methods marked with the `return` attribute ensure the returned
-        reference will not outlive the struct instance.
-        )
-
----
-struct S
-{
-    private int x;
-    ref int get() return { return x; }
-}
-
-ref int escape()
-{
-    S s;
-    return s.get(); // Error: escaping reference to local variable s
-}
----
-
         $(P Returning the address of a `ref` variable is also checked in `@safe` code.)
 
 $(SPEC_RUNNABLE_EXAMPLE_FAIL
@@ -1832,6 +1814,52 @@ void uranus()
 }
 ---
 )
+
+$(H4 $(LNAME2 struct-return-methods, Struct Return Methods))
+
+        $(P Struct non-static methods can be marked with the `return` attribute to ensure a returned
+        reference will not outlive the struct instance.
+        )
+
+$(SPEC_RUNNABLE_EXAMPLE_FAIL
+---
+struct S
+{
+    private int x;
+    ref int get() return { return x; }
+}
+
+ref int escape()
+{
+    S s;
+    return s.get(); // Error: escaping reference to local variable s
+}
+---
+)
+        $(P The hidden `this` ref-parameter then becomes `return ref`.)
+
+        $(P The `return` attribute can also be used to limit
+        the lifetime of the returned value, even when the method is not `ref`:
+        )
+
+---
+struct S
+{
+    private int i;
+    int* get() return @safe => &i;
+}
+
+void f() @safe
+{
+    int* p;
+    {
+        S s;
+        int *q = s.get(); // OK, q has shorter lifetime than s
+        p = s.get(); // error, p has longer lifetime
+        p = (new S).get(); // OK, heap allocated S
+    }
+}
+---
 
 $(H3 $(LNAME2 scope-parameters, Scope Parameters))
 


spec/function.dd
```diff
@@ -1899,6 +1899,38 @@ class C
...
+---
+@safe:
+
+struct S
+{
+    int* p;
+    // return applies to the `this` reference
+    int* getReturn() return => p;
+    int* getReturnScope() return scope => p;
+}
```
pbackus, line 1919: There is no such thing as a "`return` parameter". In this example, the `this` reference passed to `getReturn` is a `return ref` parameter, because the `this` parameter of a `struct` method is implicitly `ref`.

ntrel: OK, thanks. Updated.

spec/function.dd
```diff
@@ -1803,6 +1785,53 @@ void venus()
...
+---
+)
+        $(P The `return` attribute can also be used for wrapper structs to limit
+        the lifetime of the returned value:
+        )
+
+$(SPEC_RUNNABLE_EXAMPLE_COMPILE
+---
+@safe:
```
Bolpat, line 1815: This is easily missed. Better annotate every function with `@safe` explicitly. It’s just two of them.

spec/function.dd
```diff
@@ -1803,6 +1785,53 @@ void venus()
...
+
+$(SPEC_RUNNABLE_EXAMPLE_FAIL
+---
+struct S
+{
+    private int x;
+    ref int get() return { return x; }
+}
+
+ref int escape()
```
Bolpat, line 1802: Annotate `@safe`.

dkorpel:
I agree that `return-ref` can be used as 'scope entrypoint', but haven't managed to convince Walter: https://github.com/dlang/dmd/pull/12817

spec/function.dd
```diff
@@ -1803,6 +1785,53 @@ void venus()
 }
 ---
 
+$(H4 $(LNAME2 struct-return-methods, Struct Return Methods))
+
+        $(P Struct non-static methods marked with the `return` attribute ensure the returned
+        reference will not outlive the struct instance.
+        )
+
```
dukc, line 1793: I think this should explicitly clarify that `return` means making the implicitly `ref` this pointer `return ref`.

spec/function.dd
```diff
@@ -1803,6 +1785,53 @@ void venus()
...
+    private int* p;
+    int* get() return => p;
+}
+
+void f()
+{
+    int* p;
+    {
+        S s;
+        //p = s.get(); // deprecated
```
dukc, line 1828: For consistency with other code examples here, don't comment it out and call it an error. Besides, DIP25 is already the default as of 2.103.0, and as I understand it it makes this an error, not a deprecation.

dukc (commented):
Along with Quirins arguments, otherwise LGTM.

spec/function.dd
ntrel: > For consistency with other code examples here, don't comment it out

Now done. (I often prefer not to use RUNNABLE_EXAMPLE_FAIL as it doesn't check which particular lines errored).

> and call it an error. Besides, DIP25 is already the default as of 2.103.0

OK. BTW the example compiler is still 2.102.

> , and as I understand it it makes this an error, not a deprecation
> the spec doesn't seem to mention the need of the preview flag anymore with other dip1000 errors either

It was removed in https://github.com/dlang/dlang.org/pull/3518. -dip1000 is still needed sometimes (even with git master), see my comment there.

spec/function.dd
```diff
@@ -1807,25 +1808,23 @@ ref int escape()
...
-void f()
+void f() @safe
 {
     int* p;
     {
         S s;
-        //p = s.get(); // deprecated
+        p = s.get(); // error, p has longer lifetime than s
         p = (new S).get(); // OK, heap allocated S
     }
```
Bolpat, line 1829: This can be compacted and be made more expressive without the `{}` scope:
```d
struct S
{
    private int p;
    int* get() return @safe => &p;
}
void main() @safe
{
    int* p = S().get(); // error, `p` has longer lifetime than temporary S()
    int* q = (new S).get(); // OK, heap allocated
}
```
First, returning the address of the `int` field – opposed to the copy of a pointer-type field – makes it obvious why there is a `return` attribute.
Second `S().get()` removes the necessity for the scope.

pbackus:
It is worth noting that this technique does not actually solve the problem from the linked bug report, where the destructor makes a `@trusted` call to `free`, because the destructor can be called early while the returned pointer is still in scope.

```d
import core.stdc.stdlib;

struct List
{
    int* front() @safe return
    {
        return data;
    }

    ~this() @trusted scope
    {
        free(data);
        data = null;
    }

    @disable this(this);

    int* data;
}

List list() @trusted
{
    auto l = List(cast(int*) malloc(int.sizeof));
    *l.data = 0;
    return l;
}

void main() @safe
{
    auto l = list();
    scope elem = l.front;
    destroy(l);
    int value = *elem; // use after free
}
```

pbackus:
It seems to me like there are two things being documented here:

1. The fact that the `this` parameter of a `struct` method is implicitly `ref`, and therefore `return` by itself on a `struct` method means `return ref`.
2. The technique of using a `return ref` parameter to trick the compiler into restricting the lifetime of an unrelated pointer returned by a function.

I think for clarity's sake it would be better to document both of these things separately. And I am not sure the language spec is the most appropriate place to document (2), given that it's a technique/idiom, not a language feature (maybe [d-idioms][1] should have it?).

[1]: https://p0nce.github.io/d-idioms/

ntrel:
> the destructor can be called early while the returned pointer is still in scope.

Yes, to solve that `object.destroy` and `std.algorithm.move` should `static assert(false)` when a struct has `@disable`d members `destroy` and `move`, or something similar. I see that as a separate issue, but let me know if you want this not to close the bugzilla issue.

ntrel:
> I am not sure the language spec is the most appropriate place to document (2), given that it's a technique/idiom, not a language feature

I'm just trying to document that the return attribute applies to the returned value when the function does not return by ref. The idiom for a pointer wrapper is indeed more involved, which is why I left all the other stuff out.

spec/function.dd
ntrel: I have extended the example to make the scope more useful.

ntrel:
> document that the return attribute applies to the returned value when the function does not return by ref

Changed to return address of int field as suggested by Quirin.

spec/function.dd
ntrel: I've now removed RUNNABLE until -dip1000 is the default, so the example runner works as documented.

ntrel:
> let me know if you want this not to close the bugzilla issue

Removed auto-close.

dukc:
>Yes, to solve that object.destroy and std.algorithm.move should static assert(false) when a struct has @disabled members destroy and move, or something similar. I see that as a separate issue, but let me know if you want this not to close the bugzilla issue.

A destructor can never be disabled, since by definition it is always called at end of the scope. Preventing calling object.destory early wouldn't, unfortunately, solve the issue either:

```D
void main() @safe
{
    auto pList = new List();
    *pList = list();
    scope elem = pList.front;
    pList = null; // Nothing refers to the list struct anymore
    GC.collect(); // Thus it gets freed here and the destructor called
    int value = *elem; // Oh no
}
```

But I agree that's a separate issue that needn't stop this PR.

dukc:
I see you addressed my observations. I think you still should add the `@safe` annotations as Quirin asked, but otherwise this is good to go IMO.

dukc:
> And I am not sure the language spec is the most appropriate place to document (2)

I think it's fine. It's essentially saying that it's legal to restrict lifetime further than the actual lifetime of the returned struct, not just a way of helping the compiler to understand the limits of the actual lifetime.

spec/function.dd
```diff
@@ -1831,6 +1813,52 @@ void uranus()
...
+    private int i;
+    int* get() return @safe => &i;
+}
+
+void f() @safe
+{
+    int* p;
+    {
+        S s;
+        int *q = s.get(); // OK, q has shorter lifetime than s
```
Bolpat, line 1855: Nitpicking: `int* q` instead of `int *q`.

ntrel:
```suggestion
        int* q = s.get(); // OK, q has shorter lifetime than s
```

ntrel:
Merge please?

dkorpel (approved):

PR #3628, opened 2023-06-04
Add article on building Windows DLLs

WalterBright:
Part of the problems with creating Windows DLLs is caused by the near complete lack of documentation on it. This is a first step.

### DIFF
--- a/articles/dll-windows.dd
+++ b/articles/dll-windows.dd
@@ -1,0 +1,221 @@
+Ddoc
+
+$(D_S $(TITLE)
+
+$(HEADERNAV_TOC)
+
+Windows DLLs (aka shared libraries) are a method of sharing instances of executable
+code and data between process. Although they perform the same role as shared libraries
+in other systems like Linux, OSX, and FreeBSD, they are implemented quite differently.
+The information in this article is specific to Windows DLLs.
+
+
+$(H2 Build a DLL)
+
+$(H3 Code for the DLL)
+
+$(OL
+$(LI Create the file myll.d:
+
+---
+module mydll;
+
+import core.sys.windows.dll;
+import core.stdc.stdio;
+
+mixin SimpleDllMain;
+
+export void entry()
+{
+    printf("called mydll.entry()\n");
+}
+---
+)
+
+$(LI Compile and link the DLL:
+
+$(CONSOLE
+dmd -shared mydll
+)
+
+which will create the files $(TT mydll.lib) (the $(I import library)) and
+$(TT mydll.dll) (the $(I dll)).
+)
+
+
+$(LI Create the file mydll.di:
+
+---
+module mydll;
+
+export void entry();
+---
+)
+
+$(LI Create the file myexe.d:
+
+---
+module myexe;
+
+import mydll;
+
+int main()
+{
+    mydll.entry();
+    return 0;
+}
+---
+)
+
+$(LI Compile the $(TT myexe.d) file and link is with the $(TT mydll.lib)
+file to create the $(TT myexe.exe) file:
+$(CONSOLE
+dmd myexe.d mydll.lib
+)
+)
+
+$(LI Run myexe:
+
+$(CONSOLE
+C:> myexe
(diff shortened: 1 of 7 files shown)


WalterBright:
C requires __declspec(import) and __declspec(export), DllMain, etc. So does C++.

In any case, I am documenting what it does now.

WalterBright:
As usual, the 10,000 line logs for DAutoTest and circleci give no clue what is worng.

WalterBright:
```
Loading navigation
std.json.JSONException@std/json.d(1547): Found 'd' when expecting '}'. (Line 1:36578)
```
???

WalterBright:
Ok, I deleted the body of the article, to see if it passes now.

WalterBright:
I'm dead in the water on this.

dlang.org.ddoc
```diff
@@ -493,7 +493,9 @@ $(SUBNAV_TEMPLATE
         $(ROOT_DIR)articles/template-comparison.html, Template Comparison,
         $(ROOT_DIR)articles/d-array-article.html, D Slices,
         $(ROOT_DIR)articles/cppcontracts.html, D's Contract Programming,
-        $(ROOT_DIR)articles/dll-linux.html, Writing Shared Libraries
+        $(ROOT_DIR)articles/dll-linux.html, Writing Shared Libraries for Linux,
+        $(ROOT_DIR)articles/RefReturnScope.html, Coralling Wild Pointers With $(D ref return scope),
```
CyberShadow, line 497: Please try to remove special formatting from the navigation, I don't think we can use it in all circumstances.

CyberShadow:
FWIW, there is a fair amount of information on the wiki:
https://wiki.dlang.org/Win32_DLLs_in_D

dlang.org.ddoc
WalterBright: that did the trick, thanks!

WalterBright:
Ok, put the article text back in as the boilerplate has passed.

WalterBright:
@CyberShadow thanks, I put a reference link to it in the article.

CyberShadow:
Uh-oh.

image

CyberShadow:
More:

image

CyberShadow:
Speaking of things escaping their scope... @WalterBright We have too many function/parameter annotations, not even the website layout can contain them 😂

WalterBright:
@CyberShadow did you change the style.css file? Because I didn't! I also don't know where the strange url is coming from?

CyberShadow:
Yes, I pushed two commits to your branch with [attempts of] fixes to the above.

The strange URL is due to formatting (again) in the ref return scope article's `TITLE`. It's apparently used to construct a Bugzilla URL and link, so that users can easily file bugs against specific pages on our website. Ideally we would strip HTML tags or macros from titles before constructing that URL, but I don't think we can do that in DDoc.

WalterBright:
Well, since it is changed, I can't just push anymore as it will revert your additions. If you could just fix the $(TITLE) that would do the trick!

CyberShadow:
Ah, just pull or rebase first.

I can move it to a separate PR if you're having trouble with Git.

CyberShadow:
Patch what? Sorry.

CyberShadow:
How can I do that without changing the file and pushing it to GitHub?

WalterBright:
You did it for the other patches. I don't understand.

Or just approve this, I'll merge it, then I'll do another PR for the $(TITLE)

CyberShadow:
Oh, but I did fix the title. There's two commits, one for the title and the other for the stylesheet.

CyberShadow (approved):
LGTM ether way

WalterBright:
Oh, ok then. Thanks!

articles/dll-windows.dd
```diff
@@ -0,0 +1,221 @@
+Ddoc
+
+$(D_S $(TITLE)
```
ntrel, line 3: Fixed: #3632

PR #3630, opened 2023-06-04
Improve `return ref` examples

ntrel:
Make some examples runnable.
Mention `@safe` is needed to check returning address of `return ref` parameter - the (now default) DIP25 is not enough.
Note -dip1000 *is* still needed to return the address of a `return ref` parameter.
Add example for void function with `return ref` parameter - this needs -dip1000 so can't be made runnable yet.

Note the example compiler is currently 2.102 which needs -dip25, so the example with the mars function errors but pluto always errors anyway so _FAIL is OK.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -1751,7 +1751,7 @@ int* pluto(ref int i) @safe
 
 int* mars(return ref int i) @safe
 {
-    return &i;  // OK with -dip1000
+    return &i;  // OK with -preview=dip1000
 }
 ---
 )
@@ -1777,7 +1777,7 @@ void main() @safe
 {
     int i;
     S s;
-    s.f(i); // OK with -dip1000, lifetime of `s` is shorter than `i`
+    s.f(i); // OK with -preview=dip1000, lifetime of `s` is shorter than `i`
     *s.p = 2;
     assert(i == 2);
 }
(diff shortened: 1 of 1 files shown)


spec/function.dd
```diff
@@ -1739,26 +1740,49 @@ ref int escape()
...
+int* pluto(ref int i) @safe
 {
     return &i; // error: returning &i escapes a reference to parameter i
 }
 
-int* mars(return ref int i)
+int* mars(return ref int i) @safe
 {
-    return &i;  // ok
+    return &i;  // OK with -dip1000
```
dkorpel, line 1754:
```suggestion
    return &i;  // OK with -preview=dip1000
```

spec/function.dd
```diff
@@ -1739,26 +1740,49 @@ ref int escape()
...
+    {
+        p = &i;
+    }
+}
+
+void main() @safe
+{
+    int i;
+    S s;
+    s.f(i); // OK with -dip1000, lifetime of `s` is shorter than `i`
```
dkorpel, line 1780:
```suggestion
    s.f(i); // OK with -preview=dip1000, lifetime of `s` is shorter than `i`
```

dkorpel (approved):

PR #3632, opened 2023-06-05
[articles/dll-windows] Fix missing sentences

ntrel:
Otherwise article text starts after `Linux,`:
https://dlang.org/articles/dll-windows.html

Also fix 2 typos.


CyberShadow (approved):
Thanks!

PR #3636, opened 2023-06-10
Fix Issue 23946 - spec states "there can only be one destructor"

ntrel:
...which can be confusing because of mixin templates

### DIFF
--- a/spec/class.dd
+++ b/spec/class.dd
@@ -712,8 +712,9 @@ $(GNAME Destructor):
         }
         ------
 
-        $(P There can be only one destructor per class, the destructor
-        does not have any parameters,
+        $(P There can be only one destructor declared per class, although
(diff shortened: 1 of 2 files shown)


spec/class.dd
```diff
@@ -712,8 +712,9 @@ $(GNAME Destructor):
         }
         ------
 
-        $(P There can be only one destructor per class, the destructor
-        does not have any parameters,
+        $(P There can be only one destructor declared per class, although
+        other destructors $(DDSUBLINK spec/template-mixin, destructors, can be mixed in).
```
dkorpel, line 716: Is this intentional behavior?

adamdruppe: Yes, but it was never written. Destructors of member variables are all called alongside the declared one, and a mixin template is a special kind of member. It is basically an embedded struct with its names exposed to the top level scope.

dkorpel (approved):

PR #3641, opened 2023-06-20
add reference to DLL best practices

WalterBright:
Lots of useful recommendations I didn't know about.


adamdruppe:
Yeah, it is worth noting that the normal suggestion for a D DllMain to call runtime.initialize is actually breaking those rules, since it runs module static constructors too, which *may* be not allowed in the locked context, since you don't know what those ctors too.

dkorpel (approved):

PR #3642, opened 2023-06-21
 [spec/lex] Improve floating point docs

ntrel:
Condense *FloatLiteral*, *DecimalFloat*, _Suffix_ grammar.
Describe floating literal requirements.
Describe decimal floats with examples.
Add hexadecimal float example.
Move existing examples up.
Underscore is only allowed after a digit.
Use list for suffixes, show more examples.
Formatting tweaks.

### DIFF
--- a/spec/lex.dd
+++ b/spec/lex.dd
@@ -928,7 +928,8 @@ $(GNAME LeadingDecimal):
 0x1p-52                // double.epsilon
 ---
 
-        $(P Floating literals can have embedded $(D $(UNDERSCORE)) characters to improve readability, which are ignored.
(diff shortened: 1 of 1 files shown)


spec/lex.dd
```diff
@@ -947,7 +957,7 @@ $(GNAME LeadingDecimal):
         1.;  // OK, double
         ---
 
-        $(P NOTE: Floating literals followed by $(B i) to denote imaginary
+        $(NOTE Floating literals followed by $(B i) to denote imaginary
```
maxhaton, line 960: These are supposed to be deprecated, you could zap this while you're here.

dkorpel (approved):

PR #3646, opened 2023-06-22
[spec/lex] Optional IdentifierStart after string/numeric literal suffix

ntrel:
This is for https://github.com/dlang/dmd/pull/15339.

I have ignored the ImaginarySuffix FloatLiteral variants, as they are deprecated.


dkorpel:
In the other PR you said:

> More importantly, we can never add any new literal suffixes without breaking code if we don't have an error.

When you only put `IdentifierStartError` after existing suffixes, you still can't add new suffixes without breaking code.

ntrel:
Yes, this change will only catch existing suffixes running into an identifier. The more general errors could be done afterwards - now there's a way to handle them in the grammar.

PR #3647, opened 2023-06-24
Fix Issue 24012 - [spec/cpp_interface] _d_dynamicArray generated by -…

ntrel:
…HC not documented

Also add link mentioning static array should be passed by ref.


dkorpel:
I don't think this documentation belongs in the language spec, it's specific to dmd's dtoh utility.

dkorpel (approved):

PetarKirov (approved):

PR #3648, opened 2023-06-25
scope does not affect operator overloading

WalterBright:
This documents existing behavior. See https://issues.dlang.org/show_bug.cgi?id=23941 for discussion.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -1295,6 +1295,8 @@ $(H2 $(LNAME2 function-overloading, Function Overloading))
 
         $(P Literals do not match $(CODE ref) or $(CODE out) parameters.)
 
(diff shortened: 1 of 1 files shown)


spec/function.dd
```diff
@@ -1295,6 +1295,8 @@ $(H2 $(LNAME2 function-overloading, Function Overloading))
 
         $(P Literals do not match $(CODE ref) or $(CODE out) parameters.)
 
+        $(P $(CODE scope) parameter storage class does not affect operator overloading.)
```
ntrel, line 1298: Looking at the linked issue, shouldn't this say function overloading rather than operator overloading?

WalterBright: Oops! good catch

ntrel (approved):

dkorpel (approved):

PR #3652, opened 2023-07-03
[spec] Improve opCast & `opCast!bool` docs

ntrel:
Add runnable example.
Clarify when implicit cast to bool happens. This does *not* happen whenever a bool type is expected.
Mention result of `opCast` is implicitly converted to the `cast` type.

### DIFF
--- a/spec/operatoroverloading.dd
+++ b/spec/operatoroverloading.dd
@@ -281,6 +281,8 @@ assert(b);
 //b = s; // error
 ---
 )
+    $(P If the return type of `opCast` differs from the *type* parameter of
+        the `cast`, then the result is implicitly converted to *type*.)
 
 $(H3 $(LNAME2 boolean_operators, Boolean Operations))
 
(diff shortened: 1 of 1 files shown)


spec/operatoroverloading.dd
```diff
@@ -283,11 +300,16 @@ if (e)   =>  if (e.opCast!(bool))
 if (!e)  =>  if (!e.opCast!(bool))
 ---
 
-        $(P etc., whenever a bool result is expected. This only happens, however, for
-        instances of structs. Class references are converted to bool by checking to
+        $(P and similarly for `assert(e)` and
+        $(DDSUBLINK spec/expression, logical_expressions, logical operators) used
```
dkorpel, line 304: I think there are a few more, `static assert`, `while`, `for`, `do ... while`

ntrel: Thanks, changed to use 'other boolean conditional expressions'.

dkorpel (approved):

PR #3654, opened 2023-07-05
[spec/arrays] Improve vector operations docs

ntrel:
Change title to Vector Operations, array operations is not precise enough IMO and hard to google for.
Define which operations are supported. These were found from code in `expressionsem.d`:
```d
            // Look for valid array operations
            if (exp.memset != MemorySet.blockAssign &&
                exp.e1.op == EXP.slice &&
                (isUnaArrayOp(exp.e2.op) || isBinArrayOp(exp.e2.op)))
```
(Those predicates are in `arrayop.d`).

Rewrite example & make runnable.
Add a note warning about invalid expressions for vector op.

Fixes Issue 5636 - Array ops use lexicographic comparison instead of vector-style element-wise.

### DIFF
--- a/spec/arrays.dd
+++ b/spec/arrays.dd
@@ -463,9 +463,9 @@ a ~= b; // a becomes the concatenation of a and b
         setting dynamic array length) for details.
         )
 
-$(H2 $(LNAME2 array-operations, Array Operations))
+$(H2 $(LNAME2 array-operations, Vector Operations))
 
-        $(P Many array operations, also known as vector operations,
+        $(P Many array operations
         can be expressed at a high level rather than as a loop.
         For example, the loop:
         )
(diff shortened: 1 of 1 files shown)


spec/arrays.dd
```diff
@@ -490,16 +490,35 @@ a[] = b[] + 4;
...
+a[] += 10 - (b[] ^^ 2);
+assert(a == [9, 6, 1]);
 ---
+)
+
+        $(NOTE In particular, an expression using
+        $(GLINK2 expression, ConditionalExpression),
+        $(DDSUBLINK spec/expression, logical_expressions, logical expressions),
+        $(GLINK2 expression, CmpExpression),
+        concatenation `~` or a function call is *not* a vector op.)
```
ntrel, line 521: The reason for this is because those operations are defined on slices as well as the element type, so it would be ambiguous if vector ops allowed these.

ntrel: E.g. unfortunately `slice ? 0 : 1`, `slice && 1` as well as more justified expressions like `s1 < s2` are legal.

spec/arrays.dd
```diff
@@ -490,16 +490,35 @@ a[] = b[] + 4;
 
         $(P A vector operation is indicated by the slice operator appearing
         as the left-hand side of an assignment or an op-assignment expression.
-        The right-hand side can be an expression consisting either of an array
-        slice of the same length and type as the left-hand side or a scalar expression
-        of the element type of the left-hand side, in any combination.
-        )
+        The right-hand side can be certain combinations of:)
```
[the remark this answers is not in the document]
ntrel: This is defined under the list *The following operations are supported*.

spec/arrays.dd
```diff
@@ -525,6 +525,9 @@ assert(a == [9, 6, 1]);
         has zero elements in it.
         )
 
+        $(P If the element type defines matching overloaded operators,
+        those methods must be `pure nothrow @nogc`.)
```
ntrel, line 529: ATM you get a druntime error if e.g. `opBinary` violates those. `@system` seems to be fine.

dkorpel (approved):

PR #3655, opened 2023-07-07
add spec/legacy page

WalterBright:
Create the spec page where legacy code support will be documented.


WalterBright:
ping @dkorpel

dkorpel (approved):

PR #3659, opened 2023-07-11
add body legacy code

WalterBright:
Moved it from the deprecate page to the legacy page.

Corresponding code change: https://github.com/dlang/dmd/pull/15394


WalterBright:
@dkorpel ping!

dkorpel (approved):

PR #3668, opened 2023-07-25
[spec] Improve class destructor docs

ntrel:
~~`destroy` only nulls a class reference, it doesn't "inform the garbage collector that an object is no longer referred to". Mention `__delete` instead.~~

List when dtor called including `destroy`.
Make example runnable.
Use list for destructor characteristics.
Minor tweaks.
Change 'auto object' to 'scope class instance'
Whether the GC calls the destructor is implementation defined.
Use 'Important' heading for the order of destructor calls & use PANEL macro to split paragraph in two with grouping.
 Fix wording about destroy as it applies to scope instances too.

### DIFF
--- a/spec/class.dd
+++ b/spec/class.dd
@@ -702,10 +702,13 @@ $(GNAME Destructor):
     $(D ~ this ( )) $(GLINK2 function, MemberFunctionAttributes)$(OPT) $(GLINK2 function, FunctionBody)
 )
 
-        $(P The destructor function is called when the object
-        is deleted by the garbage collector, or when a
-        $(DDSUBLINK spec/attribute, scope-class-var, `scope` object) goes out of scope.
-        The syntax is:)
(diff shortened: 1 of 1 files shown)


dkorpel (approved):

ntrel:
@dkorpel I was wrong, `destroy` does call the class destructor. I've now updated this and made some additions.

ntrel:
Rendered:
https://dtest.dlang.io/artifact/website-50b8611567f12a501ae03a840328a058703fd611-edfb9028ee12d9a297b0a9d57db657a3/web/spec/class.html#destructors

dkorpel (approved):

dkorpel:
> I was wrong, `destroy` does call the class destructor.

It's always a source of confusion what `free` `delete` and `destroy` do exactly.

ntrel:
I'm not sure why circleci is failing.

PR #3670, opened 2023-07-27
Add PANEL macro to visually group certain paragraphs

ntrel:
I copied the CSS property values from already used values. The `border-radius` being 5px rather than 4px is copied from the `message-box` style. (I think the 4px value is for smaller items like buttons).

Use it for `AssertExpression` docs to group several 'implementation defined' paragraphs and an example together.
https://dlang.org/spec/expression.html#AssertExpression

image


CyberShadow (approved):
Nice

PR #3671, opened 2023-07-28
[spec/class] Improve super class docs

ntrel:
Move *Super Class* section above *Fields*, which has an example using inheritance.
Add simple example.
Add note about multiple inheritance/interfaces.

### DIFF
--- a/spec/class.dd
+++ b/spec/class.dd
@@ -50,7 +50,7 @@ $(P A class consists of:)
 
 $(COMMENT list of notable components, not exhaustive)
 $(UL
-        $(LI a super class)
+        $(LI a $(RELATIVE_LINK2 super_class, super class))
         $(LI $(DDLINK spec/interface, Interfaces, interfaces)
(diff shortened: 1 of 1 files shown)


spec/class.dd
```diff
@@ -100,6 +100,23 @@ $(H2 $(LNAME2 access_control, Access Control))
...
+        )
+
+---
+class A { }     // A inherits from Object
+class B : A { } // B inherits from A
+---
+
+        $(P Multiple class inheritance is not supported.
+        However, a class can inherit from multiple
+        $(DDLINK spec/interface, Interfaces, interfaces) in addition to its super class.)
```
adamdruppe, line 118: might be worth noting that the parent class must be first, then interfaes listed after, separated by commas

the grammar says that too but in more grammar-ese

ntrel: Thanks, updated.

dkorpel (approved):

PR #3680, opened 2023-08-04
Move the base of {Index,Slice}Expression into PostfixExpression

Dante-Broggi:
Them being independent expressions in the grammar made PostfixExpression non-trivially left recursive.

A more direct representation of the parser would be to combine indexing and slicing, but I assume the separation was deliberate for reader clarity.

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -905,7 +905,7 @@ assert(a[2] == 4);
 ---
 )
 
-    $(P $(GLINK IndexExpression) can also be used with a pointer and has
+    $(P $(GLINK IndexOperation) can also be used with a pointer and has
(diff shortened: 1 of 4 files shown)


spec/expression.dd
```diff
@@ -1592,47 +1592,47 @@ S s = S(1, 2);
     Uniform construction syntax for built-in scalar types))
 
 
-$(H2 $(LNAME2 index_expressions, Index Expressions))
+$(H3 $(LNAME2 index_operations, Index Operations))
```
ntrel, line 1595: This and the other LNAME2 changes would need to use LEGACY_LNAME2 to preserve the original anchor.

dkorpel (approved):

PR #3681, opened 2023-08-05
Allow underscores at more positions in grammar for numbers

tim-dlang:
DecimalDigitsNoSingleUS and HexDigitsNoSingleUS previously did not accept underscores at the start and end at the same time, but DMD allows this. For example the hex literal `0x_1_` and the floating point literals `0_1_.0`, `1e_1_` and `0x_1_p_1_` are accepted by DMD, but were not allowed by the grammar. BinaryDigitsNoSingleUS was already correct.

The grammar is also simplified by using optional parts instead of multiple rules.


tim-dlang:
The CircleCI failure seems to be unrelated and may be fixed by https://github.com/dlang/dmd/pull/15504.

dkorpel (approved):

PR #3687, opened 2023-08-21
[articles/template-comparison] Improve D examples

ntrel:
Fix `static if` example syntax.
Make 2 examples runnable.
Fix & simplify `is` example (the template parameter has to be an alias).

### DIFF
--- a/articles/template-comparison.dd
+++ b/articles/template-comparison.dd
@@ -206,7 +206,7 @@ template&lt;&gt; class factorial&lt;1&gt;
             based on Template Arguments)
             $(TD Yes:
 ---
-template void foo(T)(T i)
+void foo(T)(T i)
 {
   static if (can_fast_foo!(T))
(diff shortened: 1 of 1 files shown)


articles/template-comparison.dd
```diff
@@ -206,7 +206,7 @@ template&lt;&gt; class factorial&lt;1&gt;
             based on Template Arguments)
             $(TD Yes:
 ---
-template void foo(T)(T i)
+void foo(T)(T i)
 {
   static if (can_fast_foo!(T))
```
PetarKirov, line 211: The C++ example is broken, btw:

PetarKirov (approved):
Nice work!

articles/template-comparison.dd
PetarKirov: Nevermind, I later saw that you fixed that in here: https://github.com/dlang/dlang.org/pull/3688

PR #3702, opened 2023-10-07
Change spec for SwitchStatement to include changes introduced in dlang/dmd#15656

Mai-Lapyst:
This PR changes the specs for SwitchStatement to include that the condition of a switch statement now can also be a declaration. See dlang/dmd#15656 for more infos.

### DIFF
--- a/spec/statement.dd
+++ b/spec/statement.dd
@@ -1220,7 +1220,7 @@ $(H2 $(LEGACY_LNAME2 SwitchStatement, switch-statement, Switch Statement))
 
 $(GRAMMAR
 $(GNAME SwitchStatement):
-    $(D switch $(LPAREN)) $(EXPRESSION) $(D $(RPAREN)) $(PSSCOPE)
+    $(D switch $(LPAREN)) $(GLINK IfCondition) $(D $(RPAREN)) $(PSSCOPE)
 
 $(GNAME CaseStatement):
     $(D case) $(GLINK2 expression, ArgumentList) $(D :) $(PSSEMI_PSCURLYSCOPE_LIST)$(OPT)
@@ -1241,14 +1241,27 @@ $(GNAME StatementNoCaseNoDefault):
     $(GLINK ScopeBlockStatement)
 )
 
-        $(P $(I Expression) is evaluated.
-        If the type of $(I Expression) is an `enum`, it is
+        $(P If the $(I IfCondition) is an *Expression*, it is evaluated.
+        If the type of the *Expression* is an `enum`, it is
         (recursively) converted to its $(GLINK2 enum, EnumBaseType).
-        Then, if the type is an integral, $(I Expression) undergoes
+        Then, if the type is an integral, the *Expression* undergoes
         $(DDSUBLINK spec/type, integer-promotions, Integer Promotions).
-        Then, the type of $(I Expression) must be either an integral or
+        Then, the type of the *Expression* must be either an integral or
         a static or dynamic array of $(CODE char), $(CODE wchar), or $(CODE dchar).
         )
+
+        $(P If an $(D auto) $(I Identifier) is provided, it is declared and
+        initialized to the value and type of the *Expression*. Its scope
+        extends from when it is initialized to the end of the *ScopeStatement*.)
+
+        $(P If a $(I TypeCtors) $(I Identifier) is provided, it is declared
+        to be of the type specified by $(I TypeCtors) and is initialized with
+        the value of the *Expression*. Its scope extends from when it is
+        initialized to the end of the *ScopeStatement*.)
+
+        $(P If a $(I Declarator) is provided, it is declared and initialized
+        to the value of the *Expression*. Its scope extends from when it is
+        initialized to the end of the *ScopeStatement*.)
 
         $(P The resulting value is
         compared against each of the case expressions. If there is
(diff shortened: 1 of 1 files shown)


spec/statement.dd
```diff
@@ -1241,15 +1241,28 @@ $(GNAME StatementNoCaseNoDefault):
     $(GLINK ScopeBlockStatement)
 )
 
-        $(P $(I Expression) is evaluated.
-        If the type of $(I Expression) is an `enum`, it is
+        $(P If the $(I IfCondition) is an *Expression*, it is evaluated.
```
ntrel, line 1244:
```suggestion
        $(P The *Expression* from the $(I IfCondition) is evaluated.
```

spec/statement.dd
```diff
@@ -1241,15 +1241,28 @@ $(GNAME StatementNoCaseNoDefault):
...
+        extends from when it is initialized to the end of the *ScopeStatement*.)
+
+        $(P If a $(I TypeCtors) $(I Identifier) is provided, it is declared
+        to be of the type specified by $(I TypeCtors) and is initialized with
+        the value of the *Expression*. Its scope extends from when it is
+        initialized to the end of the *ScopeStatement*.)
+
+        $(P If a $(I Declarator) is provided, it is declared and initialized
+        to the value of the *Expression*. Its scope extends from when it is
+        initialized to the end of the *ScopeStatement*.)
```
ntrel, line 1264: I think this could be more concise:
```suggestion
        $(P If an $(I Identifier) `=` prefix is provided, a variable is declared with that
        name, initialized to the value and type of the *Expression*. Its scope
        extends from when it is initialized to the end of the *ScopeStatement*.)

        $(P If $(I TypeCtors) and/or a specific type is provided for *Identifier*, those
        are used for the variable declaration which is initialized by an implicit
        conversion from the value of the *Expression*.)
```

ntrel (approved):

spec/statement.dd
```diff
@@ -1220,7 +1220,7 @@ $(H2 $(LEGACY_LNAME2 SwitchStatement, switch-statement, Switch Statement))
 
 $(GRAMMAR
 $(GNAME SwitchStatement):
-    $(D switch $(LPAREN)) $(EXPRESSION) $(D $(RPAREN)) $(PSSCOPE)
+    $(D switch $(LPAREN)) $(GLINK IfCondition) $(D $(RPAREN)) $(PSSCOPE)
```
thewilsonator, line 1223: `IfCondition` probably technically correct , in that it does the right thing, but I think the most correct thing to put is an `AssignExpression`.

Mai-Lapyst: I dont think so, since `AssignExpression` dosnt covers `auto x = expr` to my knowlege, or any declaration for that matter.

pbackus: The function used in DMD, `parseAssignCondition`, is the exact same one used for `if` and `while` statements, which both use `IfCondition` in the grammar. So `IfCondition` is correct here (though perhaps it should be renamed to reflect its expanded usage).

spec/statement.dd
```diff
@@ -1241,15 +1241,23 @@ $(GNAME StatementNoCaseNoDefault):
...
 
+        $(P If an $(I Identifier) `=` prefix is provided, a variable is declared with that
+        name, initialized to the value and type of the *Expression*. Its scope
+        extends from when it is initialized to the end of the *ScopeStatement*.)
+
+        $(P If $(I TypeCtors) and/or a specific type is provided for *Identifier*, those
+        are used for the variable declaration which is initialized by an implicit
+        conversion from the value of the *Expression*.)
+
         $(P The resulting value is
```
pbackus, line 1261: Probably best to use the same wording here as in the `if` and `while` statement sections:

```
        $(P If an $(D auto) $(I Identifier) is provided, it is declared and
        initialized to the value and type of the *Expression*. Its scope
        extends from when it is initialized to the end of the *ScopeStatement*.)

        $(P If a $(I TypeCtors) $(I Identifier) is provided, it is declared
        to be of the type specified by $(I TypeCtors) and is initialized with
        the value of the *Expression*. Its scope extends from when it is
        initialized to the end of the *ScopeStatement*.)

        $(P If a $(I Declarator) is provided, it is declared and initialized
        to the value of the *Expression*. Its scope extends from when it is
        initialized to the end of the *ScopeStatement*.)
```

spec/statement.dd
```diff
@@ -1241,15 +1241,23 @@ $(GNAME StatementNoCaseNoDefault):
...
         )
 
+        $(P If an $(I Identifier) `=` prefix is provided, a variable is declared with that
+        name, initialized to the value and type of the *Expression*. Its scope
+        extends from when it is initialized to the end of the *ScopeStatement*.)
+
+        $(P If $(I TypeCtors) and/or a specific type is provided for *Identifier*, those
+        are used for the variable declaration which is initialized by an implicit
+        conversion from the value of the *Expression*.)
+
```
pbackus, line 1260: Probably best to use the same wording here as in the `if` and `while` statement sections:

```
        $(P If an $(D auto) $(I Identifier) is provided, it is declared and
        initialized to the value and type of the *Expression*. Its scope
        extends from when it is initialized to the end of the *ScopeStatement*.)

        $(P If a $(I TypeCtors) $(I Identifier) is provided, it is declared
        to be of the type specified by $(I TypeCtors) and is initialized with
        the value of the *Expression*. Its scope extends from when it is
        initialized to the end of the *ScopeStatement*.)

        $(P If a $(I Declarator) is provided, it is declared and initialized
        to the value of the *Expression*. Its scope extends from when it is
        initialized to the end of the *ScopeStatement*.)
```

ntrel: Although those docs are more precise, I was trying to reduce repetition. I think both can be obtained - I've made a separate pull to do that for _IfCondition_: #3705.

Sorry if my commit here delayed this pull.

dkorpel (approved):

PR #3703, opened 2023-10-10
[spec] Add missing PragmaDeclaration to DeclDef

etienne02:
According to dmd parser this section is in DeclDef, so I added it there.


etienne02:
I see that Pragma appears in AttributeSpecifier in DeclDef, but Pragma is not exactly PragmaDeclaration.
I don’t know if this valid because there is some duplication with this change (but I don’t know if AttributeSpecifier covers all cases of PragmaDeclaration), so maybe it’s the PragmaDeclaration that is a duplicate and should be removed instead.

etienne02:
After searches, it seems that PragmaDeclaration is the duplicate. Attribute declaration is already covered in AttributeSpecifier.

PR #3706, opened 2023-10-30
Improve function (pointer) type docs

ntrel:
Fix Issue 24210 - Function types are not documented.

Add examples.
Make existing examples runnable.
Use list for kinds of delegate.
Remove C pointer to function syntax as it's no longer supported.

### DIFF
--- a/spec/type.dd
+++ b/spec/type.dd
@@ -140,8 +140,8 @@ $(H2 $(LEGACY_LNAME2 Derived Data Types, derived-data-types, Derived Data Types)
     $(LI $(DDSUBLINK spec/arrays, static-arrays, Static Arrays))
     $(LI $(DDSUBLINK spec/arrays, dynamic-arrays, Dynamic Arrays))
     $(LI $(DDLINK spec/hash-map, Associative Array, Associative Arrays))
-    $(LI $(DDLINK spec/function, Functions, Functions))
-    $(LI $(RELATIVE_LINK2 delegates, Delegates))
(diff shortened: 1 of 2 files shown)


spec/type.dd
```diff
@@ -548,36 +548,62 @@ values `false` and `true`, respectively. Casting an expression to `bool` means
...
 
-$(P Delegates are declared similarly to function pointers:)
+$(P A function type has the form:)
+
+$(GRAMMAR
+$(GLINK TypeCtor)$(OPT) $(GLINK BasicType) $(GLINK2 function, Parameters) $(GLINK2 function, FunctionAttributes)$(OPT)
+)
+
+$(P A function type e.g. `int(int)` is only used for type tests.
+A function type $(DDSUBLINK spec/declaration, alias-function, can be aliased).)
```
SixthDot, line 561: I have a serious reserve on this change. It seems to bring the special syntax allowed only for AliasDecl. But in other context that syntax can create ambiguities. For example

```
S(T,U)
```

requires work during expsema, as it can be a ctor call or a function type

SixthDot: well, to go further, it's a bit like `T[2]` which can be either an array Type, an IndexExp, or an attempt to statically evaluate the element of a template parameter.

So finally, you should think more about that. Does this really work ?

ntrel: Function type syntax is only allowed in a type context, so that rules out a call expression.

SixthDot (approved):

dkorpel (approved):

PR #3710, opened 2023-11-02
Document deprecation of 'is' for static arrays

pbackus:
This was deprecated in DMD 2.072.0, but the corresponding spec PR was never merged.

See https://github.com/dlang/dmd/pull/6089

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -688,6 +688,10 @@ $(GNAME IdentityExpression):
     ---
     )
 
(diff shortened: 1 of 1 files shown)


adamdruppe:
Should merge this but also get a follow up from the dictatorship about what this deprecation is supposed to lead to.

pbackus:
I think probably the most reasonable way forward is a `-preview` switch that changes it to do a value comparison.

PR #3712, opened 2023-11-03
[spec] Simplify `VarDeclarations` grammar

ntrel:
C multiple identifier variable declarations with separate *TypeSuffixes* are no longer supported, so *VarDeclarations* no longer needs to use *Declarators* and the grammar can be simplified quite a bit.
*Declarator* is kept for the uses in other files, but could be inlined.

The old alias grammar was the only place using *Declarators*, so I updated it to be more precise using *Identifiers*, as using different *TypeSuffixes* for a second identifier is an error anyway. (AIUI each identifier is the same type).

I added legacy aliases for the old grammar terms.

### DIFF
--- a/spec/declaration.dd
+++ b/spec/declaration.dd
@@ -41,31 +41,24 @@ $(H3 $(LNAME2 variable-declarations, Variable Declarations))
 
 $(GRAMMAR
 $(GNAME VarDeclarations):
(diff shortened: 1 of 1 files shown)


ntrel:
Looks like ddoc failed on this unrelated dmd file:
```
Sources: 
 -> /home/circleci/dmd/compiler/src/dmd/access.d

Targets: 
 -> /home/circleci/dlang.org/web/phobos/dmd_access.html
```

dkorpel (approved):

PR #3715, opened 2023-11-04
[spec] Improve value sequence docs

ntrel:
Prefer using TypeSeq for variadic template functions. Mention returning `Tuple`.
Add subheading and example for lvalue sequences.
Minor tweaks.

### DIFF
--- a/spec/template.dd
+++ b/spec/template.dd
@@ -933,10 +933,46 @@ $(H4 $(LNAME2 homogeneous_sequences, Homogeneous Sequences))
         ---
         )
 
+    $(BEST_PRACTICE Prefer using a *TypeSeq* to declare variadic template
+        functions because then there is only one template instantiation for
+        each set of types.)
+
+    $(NOTE A value sequence cannot be returned from a function - instead, return a
+        $(REF Tuple, std,typecons).)
+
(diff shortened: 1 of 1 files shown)


spec/template.dd
```diff
@@ -933,10 +933,46 @@ $(H4 $(LNAME2 homogeneous_sequences, Homogeneous Sequences))
         ---
         )
 
+    $(BEST_PRACTICE Prefer using a *TypeSeq* to declare variadic template
```
dkorpel, line 936: Prefer a TypeSeq over what exactly?

spec/template.dd
```diff
@@ -933,10 +933,46 @@ $(H4 $(LNAME2 homogeneous_sequences, Homogeneous Sequences))
...
+        functions because then there is only one template instantiation for
+        each set of types.)
```
dkorpel, line 938:
```suggestion
        each list of types.)
```

A different order still gives a different template instantiation

spec/template.dd
PetarKirov: I also don't understand what the recommendation means.

ntrel: I've now removed this paragraph.

dkorpel (approved):

PR #3716, opened 2023-11-07
Document enum copying and assignment behavior

pbackus:
An enum derived from a struct type bypasses its base type's copy constructor, postblit, and identity opAssign overloads (if any). This behavior is a special case, and does not apply to other operator overloads or special member functions of the base type (e.g., destructors), so it should be documented explicitly.

### DIFF
--- a/spec/enum.dd
+++ b/spec/enum.dd
@@ -201,6 +201,75 @@ X.sizeof // is same as int.sizeof
         in order to compute the $(CODE .max) and $(CODE .min) properties.
         )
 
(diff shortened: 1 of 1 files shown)


pbackus:
CircleCI once again failing due to running out of memory; see https://github.com/dlang/dmd/pull/15504 for more info.

dkorpel (approved):
Is this desirable behavior? Perhaps it could be fixed in a new language edition.

pbackus:
@dkorpel I think it's acceptable if you're only using actual enum members, but it becomes problematic when base-type values are cast to the enum type. See https://issues.dlang.org/show_bug.cgi?id=24225

PR #3719, opened 2023-11-08
[spec] Fix missing legacy.html

ntrel:
Also:
`-wo` does nothing after https://github.com/dlang/dmd/pull/15612, comment it out.
Remove non-standard macros.
Fix empty paragraphs.
Fix missing/wrong Previous/Next items.


CyberShadow:
Thanks! There still seems to be a stray `)` (see doc tester).

dkorpel (approved):

PR #3721, opened 2023-11-09
[spec/legacy] Add old alias syntax and postblit

ntrel:
As of last month the `-wo` switch does nothing, but I think it's still valuable to explain which features shouldn't be used in new code. These can then become errors in future editions (and perhaps revive the `-wo` switch for warnings in an older edition).

Please let me know if there are any other obsolete features that should be listed here.

### DIFF
--- a/spec/declaration.dd
+++ b/spec/declaration.dd
@@ -244,6 +244,8 @@ auto c = new C();  // c is a handle to an instance of class C
 
 $(H2 $(LNAME2 alias, Alias Declarations))
 
+    $(NOTE New code should use the *AliasAssignments* form only.)
+
 $(GRAMMAR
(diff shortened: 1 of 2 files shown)


SixthDot (approved):
There's indeed no reason to use the old syntax anymore.

dkorpel (approved):

tim-dlang:
> Please let me know if there are any other obsolete features that should be listed here.

The documentation for hex strings has been removed in https://github.com/dlang/dlang.org/pull/3244, but hex strings were added back to DMD in https://github.com/dlang/dmd/pull/15393.
Currently they seem to be undocumented. I'm not sure if they should be on the legacy page or if the old documentation should just be brought back, but the grammar should be documented somewhere.

PR #3744, opened 2023-12-13
Document named arguments

dkorpel:


spec/expression.dd
```diff
@@ -2499,6 +2512,13 @@ $(H3 $(LNAME2 uniform_construction_syntax, Uniform construction syntax for built
         auto b = wchar();   // same as: wchar.init
         ---
 
+    $(P The argument may not be given a name:)
+
+        ---
+        version (none) auto a = short(x: 1); // Allowed, parses according to grammar
```
RazvanN7, line 2518: Is this `version(none)` line intended? Seems to be identical to the following line, except for the comment.

RazvanN7 (approved):

spec/expression.dd
dkorpel: It's meant to show that adding a named argument is a semantic error, not a parser error. Code inside `version (none)` is parsed but not semantically analyzed, so it is allowed there. I wonder if I could make that clearer, or if I should just remove that since it's an unimportant detail to draw attention to.

RazvanN7: I think this is an implementation detail that's not relevant to the user.

spec/struct.dd
```diff
@@ -473,31 +473,58 @@ $(H2 $(LEGACY_LNAME2 StructLiteral, struct-literal, Struct Literals))
...
+        )
+        $(P
+        If there is a union field in the struct, only one
         member of the union can be initialized inside a
         struct literal. This matches the behaviour for union literals.
         )
 
+        $(SPEC_RUNNABLE_EXAMPLE_FAIL
+        ---
+        struct S { int x = 1; int y = 2, int z = 3; }
```
ntrel, line 514: Fixed in #3746

spec/struct.dd
```diff
@@ -473,31 +473,58 @@ $(H2 $(LEGACY_LNAME2 StructLiteral, struct-literal, Struct Literals))
...
+
+        S s0 = S(y: 5, 6, x: 4); // `6` is assigned to field `z`, which comes after `y`
+        S s1 = S(y: 5, z: 6);    // Field x is not assigned, set to default initializer `1`
+        S s2 = S(y: 5, x: 4, 5); // Error: field `y` is assigned twice
+        S s3 = S(z: 2, 3);       // Error: no field beyond `z`
```
ntrel, line 519: This line actually segfaults:
https://issues.dlang.org/show_bug.cgi?id=24281

This is why I don't like SPEC_RUNNABLE_EXAMPLE_FAIL, because it hides unwanted failures!

PR #3746, opened 2023-12-13
[spec] Tweak named arguments docs

ntrel:
Follow up to #3744.

Minor tweaks.
Use numbered lists to describe argument matching.
Fix SPEC_RUNNABLE_EXAMPLE_FAIL example syntax (a good example of why using that macro is bug-prone).
Add asserts to examples.


dkorpel (approved):
Thanks!

PR #3752, opened 2024-01-01
fix Issue 24310 - importC: Document workaround for using C symbols wh…

WalterBright:
…ich are also D keywords


AndrejMitrovic (approved):
Awesome! You worded that way better than I could. :)

dkorpel (approved):

kinke:
[wrong referenced issue number]

dkorpel:
I re-opened 24310 and closed 24312

PR #3756, opened 2024-01-12
fix Issue 24331 - @nogc and GC.disable() are often confused

WalterBright:


thewilsonator (approved):

spec/function.dd
```diff
@@ -3752,6 +3752,14 @@ $(H2 $(LNAME2 nogc-functions, No-GC Functions))
             }
             ---
 
+	$(BEST_PRACTICE Since a function marked `@nogc` will not do any GC allocations,
+	that implies it will not cause any GC collections to run. However,
+	another thread may still allocate with the GC and trigger a collection.
+	The recommended way to prevent GC collections from being run is to call
+	$(LINK2 https://dlang.org/phobos/core_memory.html#GC.disable, core.memory.GC.disable())
+	instead. This will stop collections from being run in any thread until a corresponding
+	call to `core.memory.GC.enable()` is run. GC allocations can also still be performed
```
rikkimax, line 3761: > GC allocation can still be performed while `GC.disable()` is in effect.

Original sentence isn't reading right.

WalterBright: I don't know what you mean?

WalterBright: Oh I see now

tgehr (approved):
Looks good to me. I think Adam Wilson and deadalnix are in the mid term also trying to roll back the guarantee that there are no background collections though.

WalterBright:
Well, when/if the GC changes the documentation will have to change, too!

PR #3762, opened 2024-01-28
js: Update jQuery to 2.2.4 and add jQuery Migrate plugin

ibuclaw:
Dependabot is complaining.

https://github.com/dlang/dlang.org/security/dependabot

Versions 1.x and 2.x no longer receive any patches, and there are likely some uses of deprecated 1.x features in the dlang.org site.  Bump it to the most up-to-date version that the migration plugin will allow us to in order to flag problems. Afterwards, it should be possible to jump to the latest 3.x version.

Just guessing, no idea if this will work.


thewilsonator (approved):

maxhaton:
I didn't even know that it used any JS at all.

PR #3768, opened 2024-02-14
Add Interpolation Expression Sequences to the spec

schveiguy:
See dlang/dmd#15715

### DIFF
--- a/spec/istring.dd
+++ b/spec/istring.dd
@@ -53,7 +53,7 @@ $(GNAME InterpolationExpression):
     $(B $ (LPAREN)) $(GLINK AssignExpression) $(B $(RPAREN))
 )
 
-$(H2 $(LNAME2 expression_translation Expression Translation)
+$(H2 $(LNAME2 expression_translation, Expression Translation))
 
 $(P When the lexer encounters an Interpolation Expression Sequence, the token
 is translated into a sequence of expressions, which replaces the single token.
@@ -85,7 +85,7 @@ assert(items == tuple(
     InterpolationFooter()));                     // denotes the end of an IES
 ------
 
-$(H2 $(LNAME2 core_interpolation core.interpolation Types)
+$(H2 $(LNAME2 core_interpolation, core.interpolation Types))
 
 $(P Types defined in $(D core.interpolation) need not be imported to use IES.
 These are automatically imported when an IES is used. The types are defined so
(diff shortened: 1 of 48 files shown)


schveiguy:
OK, I think this is ready for review. This needs to be merged before the next release.

"Interpolation Expression Sequence" is too long for the menu bar... Any ideas for an abbreviated name? Maybe just "Interpolations"?

schveiguy:
Unsure if adding this in where lexical goes (which I think is where it really belongs) is better than tacking on the end? Walter suggested [here](https://github.com/dlang/dmd/pull/15715#pullrequestreview-1824895012) that the spec should be on its own page.

I dislike that I had to bump all chapters up by 1.

spec/istring.dd
```diff
@@ -0,0 +1,228 @@
+Ddoc
+
+$(SPEC_S Interpolation Expression Sequence,
+
+$(HEADERNAV_TOC)
+
+$(P Interpolation Expression Sequences (IES) are expressions that intersperse
+string literal data, and values. An interpolation expression is written like a
```
adamdruppe, line 8: i'd delete the comma lol

otherwise tl;dr but no gross errors jumped out at me

schveiguy:
ping, this should go in before the release candidate is branched.

thewilsonator (approved):

spec/istring.dd
```diff
@@ -0,0 +1,228 @@
...
+
+------
+// simple version of std.typecons.Tuple
+struct Tuple(T...) { T value; }
+Tuple!T tuple(T...)(T value) { return Tuple!T(value); }
+
+import core.interpolation;
+string name = "John Doe";
+auto items = tuple(i"Hello, $(name), how are you?");
+assert(items == tuple(
```
ntrel, line 127: Pretty sure the above lines can be simplified to:
```d
alias AliasSeq(A...) = A;
import core.interpolation;
auto items = i"Hello, $(name), how are you?";
assert(items == AliasSeq!(
```

schveiguy: I will try it.

schveiguy: Yes, this works, I will simplify.

schveiguy: Hm... on second thought, this is doing something that I wasn't intending -- it's creating a compile-time sequence of things, including an alias to `name`. What I really wanted was to show that the sequence is the same at runtime, e.g. what the list of values is. So I think I will leave it as-is.

spec/istring.dd
```diff
@@ -0,0 +1,228 @@
...
+escapes are recognized inside these literals.)
+
+$(H3 $(LNAME2 ies_token, Token IES Literals))
+
+$(GRAMMAR
+$(GNAME InterpolatedTokenLiteral):
+    $(D iq{) $(GLINK InterpolatedTokenStringTokens)$(OPT) $(D })
+
+$(GNAME InterpolatedTokenStringTokens):
+    $(GLINK InterpolatedTokenStringTokenNoBraces)
```
tim-dlang, line 88: InterpolatedTokenStringTokenNoBraces is not defined. Here is a PR: https://github.com/dlang/dlang.org/pull/3772

PR #3769, opened 2024-02-16
d-keyring.gpg: Run gpg --refresh-keys

ibuclaw:
Roll over the expired keys with new signing keys.


MartinNowak:
> Roll over the expired keys with new signing keys.

Hi @ibuclaw, this apparently forgot to sign the updated keyring and refresh the docs page.
Also see 607d422606647a80d07437cb49183f7337db6019
I assume you ran `gpg --keyring ./d-keyring.gpg --no-default-keyring --refresh-keys`, but not the signing and `gpg_keys.dd` part.

PR #3771, opened 2024-02-19
ImportC: function-like macros

WalterBright:
Document the addition of function-like macro support.


WalterBright:
Ah, back to guesswork about what DAutoTest didn't like.

WalterBright:
Doing a binary search to figure out why DAutoTest fails, since DAutoTest won't tell me.

WalterBright:
Ok, ready to go!

dkorpel (approved):

PR #3772, opened 2024-02-25
Fix grammar for InterpolatedTokenStringTokens

tim-dlang:
InterpolatedTokenStringTokenNoBraces is not defined. Instead use InterpolatedTokenStringToken, so the string can also end with braces.


schveiguy (approved):
thanks, don't know how I managed to mess that up, I thought I copied the token string grammar...

PR #3778, opened 2024-03-11
[trivial] [spec/class] Tweak `static this` docs

ntrel:
Use panel for rationale section.
Tweak formatting.
Show example line is an error.

BTW I think the docs for StaticConstructor and StaticDestructor should be moved to module.dd, I might do that in another pull.

### DIFF
--- a/spec/class.dd
+++ b/spec/class.dd
@@ -926,7 +926,7 @@ class Foo
 class Foo
 {
     static ~this() { ... }  // a static destructor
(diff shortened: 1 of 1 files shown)


ntrel:
I don't know why DAutoTest failed:

> /bin/ld: cannot find -lphobos2: No such file or directory

thewilsonator (approved):

PR #3781, opened 2024-03-13
entry draft for a possible new header for platform compiler support

Emmankoko:
This PR is to propose an addition for Platform compiler support header to `cpp_interface.dd` file being the interfacing to C++ reference page. Might be verbose but I think it's needed especially for beginners. I remember nearly giving up when I started fresh with C++ interop because I was running my C++ code with g++ on windows and being linked to my D code compiled with DMD. It was a next level headache being a beginner.
so it can just be a simple instruction to let users know that the following documentation works with this compiler1-compiler2 pair on platform A. Ditto for platform B and platform C.

@Geod24

### DIFF
--- a/spec/cpp_interface.dd
+++ b/spec/cpp_interface.dd
@@ -42,6 +42,8 @@ $(H2 $(LNAME2 general_idea, The General Idea))
     $(LI matching C++ virtual function table layout for single inheritance)
     )
 
+$(H2 LNAME2 platform-compiler support, Platform-Compiler Support)
+
 $(H2 $(LNAME2 global-functions, Global Functions))
 
     $(P C++ global functions, including those in namespaces, can be declared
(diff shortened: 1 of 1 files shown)


dkorpel:
What's the status of this, are you waiting for approval to write the content? It's hard to judge whether it fits in this spot when there's only a header text, but whatever you end up writing, I'm sure we can find a place for it; be it here, in an article, or on a wiki page.

Emmankoko:
opened for review.....

spec/cpp_interface.dd
```diff
@@ -42,6 +42,12 @@ $(H2 $(LNAME2 general_idea, The General Idea))
     $(LI matching C++ virtual function table layout for single inheritance)
     )
 
+$(H2 LNAME2 platform-compiler support, Platform-Compiler Support)
+
+    $(P When building on windows, you should preferrably link your D code to MSVC compiled binaries.
+    you can link to both gcc and clang compiled binaries on linux.
```
dkorpel, line 48:
```suggestion
    You can link to both gcc and clang compiled binaries on linux.
```

dkorpel (approved):

PR #3783, opened 2024-03-18
[spec/struct] Clarify static initialization

ntrel:
Add links to `.init`.
Fix struct `.init` link to *default* struct initialization. 
Specify what happens when a struct member initializer without a specified identifier follows one with an identifier.
Specify that a union member initializer without identifier refers to the first member.
Before these 2 behaviours had to be inferred from the examples.

### DIFF
--- a/spec/class.dd
+++ b/spec/class.dd
@@ -358,9 +358,9 @@ $(GNAME Constructor):
 )
 
         $(P Fields are by default initialized to the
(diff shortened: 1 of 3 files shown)


ntrel:
DAutoTest:
```
/bin/ld: cannot find -lphobos2: No such file or directory
```

dkorpel (approved):

PR #3785, opened 2024-03-18
[spec/function] Improve safe function spec

ntrel:
Fix wrong 'not allowed...: no' double negative.
Casting between pointer types is allowed under certain conditions. 
Use sub-list formatting.
Don't allow accessing `@system` variables (I think these aren't documented yet).
Make array bounds exception a note.


thewilsonator (approved):

thewilsonator:
Looks like DAutoTest failed to link phobos???

dkorpel (approved):

PR #3791, opened 2024-03-27
Mention ImportC in interfaceToC.dd

dkorpel:
This is one of the first pages new users land on when searching for "dlang c interface" or "dlang c binding", so it's important to point them to the right pragmatic direction instead of going straight into the manual / low-level way. 

https://forum.dlang.org/post/jenrwaimqbwnlzsqnozt@forum.dlang.org

Also shorten the first paragraph, I don't think the point about VMs is very interesting these days.

### DIFF
--- a/spec/interfaceToC.dd
+++ b/spec/interfaceToC.dd
@@ -5,15 +5,21 @@ $(SPEC_S Interfacing to C,
 $(HEADERNAV_TOC)
 
         $(P D is designed to fit comfortably with a C compiler for the target
-        system. D makes up for not having its own VM by relying on the
-        target environment's C runtime library. It would be senseless to
-        attempt to port to D or write D wrappers for the vast array of C APIs
-        available. How much easier it is to just call them directly.
+        system. It is able to call C functions directly without requiring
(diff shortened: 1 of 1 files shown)


thewilsonator (approved):

spec/interfaceToC.dd
```diff
@@ -86,9 +92,8 @@ extern (C)
 $(H2 $(LNAME2 storage_allocation, Storage Allocation))
 
         $(P C code explicitly manages memory with calls to
-        $(LINK2 http://www.digitalmars.com/rtl/stdlib.html#malloc, malloc()) and
-        $(LINK2 http://www.digitalmars.com/rtl/stdlib.html#free, free()).
-        D allocates memory using the D garbage collector,
+        $(LINK2 https://www.man7.org/linux/man-pages/man3/malloc.3.html, malloc() and free()).
+        D allocates memory using the $(LINK2 garbage.html, D garbage collector),
```
thewilsonator, line 96: I think the CI is complaining about an incomplete URL here

PR #3798, opened 2024-04-03
Update c++ interop reference page on lifetime handling interoperability

Emmankoko:
calling C++ constructors in D are now possible and so must be updated in documentation


spec/cpp_interface.dd
```diff
@@ -933,17 +933,124 @@ $(H2 $(LNAME2 packing-and-alignment, Packing and Alignment))
...
+
+$(CONSOLE
+$(GT) g++ -c foo.cpp
+$(GT) dmd bar.d foo.o -L-lstdc++ && ./bar
+calling C++ integer constructor
+calling C++ destructor
+)
+
+    $(P Note that you cannot call C++ Copy constructor using D classes since classes in D are reference types.
+    you need value semantics to be able to copy so you need to call them using D struct.)
```
ntrel, line 991:
```suggestion
    $(P Note that the C++ Copy constructor cannot be called using D classes since classes in D are reference types.
    Value semantics are needed to be able to copy, so use a D struct.)
```

spec/cpp_interface.dd
```diff
@@ -933,17 +933,124 @@ $(H2 $(LNAME2 packing-and-alignment, Packing and Alignment))
...
+$(GT) g++ -c foo.cpp
+$(GT) dmd bar.d foo.o -L-lstdc++ && ./bar
+calling C++ integer constructor
+calling C++ copy constructor
+calling C++ destructor
+calling C++ destructor
+)
+
+    $(P Notice you don't need to call destroy on a struct object to invoke the destructor
+    since it does stack allocation and its lifetimes ends after leaving the stack.)
```
ntrel, line 1053:
```suggestion
    $(P Notice there is no need to call destroy on a struct object to invoke the destructor
    since it does stack allocation (by default) and its lifetime ends with its declaration scope.)
```

ntrel (commented):
Thanks for working on this. Just to bear in mind, please try to avoid using 'you' in the spec.

spec/cpp_interface.dd
Emmankoko: I think that's better. Thanks for the correction!

spec/cpp_interface.dd
Emmankoko: Just force pushed the changes! Thanks again for the suggestions.

ntrel: Thanks!

dkorpel (approved):
Thanks for writing this!
I don't use C++ so I can't verify the info deeply, perhaps @Geod24 can take a quick look?

Emmankoko:
> Thanks for writing this!
> I don't use C++ so I can't verify the info deeply, perhaps @Geod24 can take a quick look?

Okay let me ping him

PR #3801, opened 2024-04-08
ddoc/source/assert_writeln_magic: Add a new "// no-writeln" syntax

CyberShadow:
This allows adding this comment to a line to disable any rewriting done by this module, other than removing said comment.

https://github.com/dlang/phobos/pull/8975#discussion_r1555582231


dkorpel:
This works, but any new contributor encountering

```D
assert(fun() == 2); // no-writeln
```

Will have no idea what that comment means, and may be tempted to remove it. I wish there is a cleaner solution.

PR #3830, opened 2024-05-12
Fix Bugzilla 24548 - [spec] Boolean condition conversion is not docum…

ntrel:
…ented

Add *IfStatement* subheading for boolean conditions. Link to this for other boolean conditions in the spec.
Also use subheading instead of panel for `if` condition variables.


maxhaton:
Lgtm

thewilsonator (approved):

PR #3832, opened 2024-05-16
[spec] Improve foreach on a tuple range docs

ntrel:
State that using only one variable will make it a value sequence.
Use finite range.


spec/statement.dd
```diff
@@ -949,41 +949,42 @@ $(H4 $(LNAME2 front-seq, Multiple Element Values))
         expands to a $(DDSUBLINK spec/template, homogeneous_sequences, value sequence)
         whose length matches the number of variables. Each variable is assigned
         to the corresponding value in the sequence.
+        If there is only one variable, it will be a value sequence.
```
dkorpel, line 952: This sentence was a bit confusing to me on the first read, but the code example does clarify it.

dkorpel (approved):

PR #3835, opened 2024-05-22
[spec/template] Tweak template value parameters

ntrel:
Use list of argument kinds, add booleans & characters. 
Use '*valid* template value arguments' for clarity. 
Tweak example to use enum instead of function.

### DIFF
--- a/spec/template.dd
+++ b/spec/template.dd
@@ -525,23 +525,29 @@ $(GNAME TemplateValueParameterDefault):
 )
 
     $(P A template value parameter can take an argument of any expression which can
-        be statically evaluated at compile time.
-        Template value arguments can be integer values, floating point values,
-        nulls, string values, array literals of template value arguments,
-        associative array literals of template value arguments,
-        or struct literals of template value arguments.)
+        be statically evaluated at compile time:)
+
+        * `bool` values
+        * Character values
+        * Integer values
+        * Floating point values
+        * `null`
+        * String values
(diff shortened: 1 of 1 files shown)


spec/template.dd
```diff
@@ -525,23 +525,29 @@ $(GNAME TemplateValueParameterDefault):
 )
 
     $(P A template value parameter can take an argument of any expression which can
-        be statically evaluated at compile time.
-        Template value arguments can be integer values, floating point values,
-        nulls, string values, array literals of template value arguments,
-        associative array literals of template value arguments,
-        or struct literals of template value arguments.)
+        be statically evaluated at compile time:)
```
thewilsonator, line 528: what about `void` "valued" expressions?

ntrel: OK, updated.

dkorpel (approved):

spec/template.dd
```diff
@@ -525,17 +525,14 @@ $(GNAME TemplateValueParameterDefault):
...
 
     $(P A template value parameter can take an argument of any expression which can
-        be statically evaluated at compile time:)
+        be statically evaluated at compile time as one of the following:)
 
-        * `bool` values
-        * Character values
-        * Integer values
-        * Floating point values
+        * A `bool`, character, integer or string value
```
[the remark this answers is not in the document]
ntrel:
```suggestion
        * A `bool`, character, integer, floating point or string value
```

PR #3838, opened 2024-05-28
Fix Issue 24571 - Grammar spec should not use "opt"

Bolpat:


dkorpel:
This includes a stray commit [update download and changelog for v2.109.0-rc.1](https://github.com/dlang/dlang.org/pull/3838/commits/d643ce8bfaded1951deb726c09c81dff091d71c0)

Bolpat:
@dkorpel, I noticed, but I don’t know why. Does it matter?

PR #3844, opened 2024-06-03
[spec] Document unsafe bool operations

ntrel:
The following are implementation defined for bool:

* Casting runtime array to `bool[]`
* Reinterpreting a byte as bool (e.g. by union field)
* `void` initialization

Note: Only 0 or 1 are safe values:
https://dlang.org/spec/function.html#safe-values

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -1350,6 +1350,9 @@ $(H4 $(LNAME2 cast_array, Arrays))
             ---
         )
 
+    $(IMPLEMENTATION_DEFINED Casting a non-literal array to `bool[]` when any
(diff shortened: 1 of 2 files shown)


ntrel:
> The following are implementation defined for bool:

As I understand it anyway. A value like `0x2` interpreted as a boolean can be both true and false:
https://issues.dlang.org/show_bug.cgi?id=20148#c3

So it would depend on the machine instruction used to test the bool.

ntrel:
Changed to undefined behaviour rather than implementation defined, as it is not allowed in `@safe` code.

dkorpel (approved):
I'm not sure if this needs to be undefined behavior or implementation defined, but we can relax it later.

thewilsonator (approved):

PR #3859, opened 2024-06-26
Specify `opApply` as an alias to a function template instance

Bolpat:
In particular, cf. [Issue 24633](https://issues.dlang.org/show_bug.cgi?id=24633).


ntrel:
DAutoTest:
```
Searching for tabs
./spec/statement.dd:732:		S s;
make: *** [posix.mak:822: test] Error 1
```

spec/statement.dd
```diff
@@ -698,88 +698,100 @@ foreach (string s, double d; aa)
...
 $(GNAME OpApplyParameters):
-    *OpApplyParameter*
-    *OpApplyParameter*, *OpApplyParameters*
+    $(GLINK ParameterDeclaration)
+    $(GLINK ParameterDeclaration), *OpApplyParameters*
 
 $(GNAME OpApplyParameter):
-    $(GLINK ForeachTypeAttributes)$(OPT) $(GLINK2 type, BasicType) $(GLINK2 declaration, Declarator)
+    $(GLINK ParameterStorageClass)$(OPT) $(GLINK2 type, BasicType) $(GLINK2 declaration, Declarator)
 )
```
ntrel, line 718: *OpApplyParameter* should be removed as it is no longer part of the *OpApplyDeclaration* grammar.

spec/statement.dd
```diff
@@ -698,88 +698,100 @@ foreach (string s, double d; aa)
...
-    a `ref` delegate parameter both forms are supported.
-    )
+    Above, `opApply` is matched both when `i` is and is not `ref`, so by using
+    a `ref` delegate parameter, both forms are supported.
 
-    $(P There can be multiple $(D opApply) and $(D opApplyReverse) functions -
-        one is selected
-        by matching each parameter of `dg` to each $(I ForeachType)
-        declared in the $(I ForeachStatement).)
+    Stating the type of the variable like in the first loop is optional,
```
ntrel, line 748:
```suggestion
    Stating the type of the variable as in the first loop is optional,
```

spec/statement.dd
```diff
@@ -698,88 +698,100 @@ foreach (string s, double d; aa)
...
         $(GLINK GotoStatement) whose matching label is outside the *ForeachStatement*.
     )
 
-    $(P For example, consider a class that is a container for two elements:)
+    $(P The $(D opApply) and $(D opApplyReverse) member functions can be overloaded.
+        Selection works similar to overload resolution by
+        comparing the number of `foreach` variables and number of parameters of `body` and,
+        if more than one overload remains unique overload,
+        matching the parameter types of `body` and each $(I ForeachType)
+        declared in the $(I ForeachStatement) when given.)
```
ntrel, line 776:
```suggestion
        comparing the number of `foreach` variables with the number of parameters of `body` and,
        if more than one overload remains,
        matching each parameter of `body` against each $(I ForeachType)
        declared in the $(I ForeachStatement).)
```
Note: I've used *each parameter* rather than *each parameter type* because `ref` affects overload resolution as well as the type.

spec/statement.dd
```diff
@@ -811,57 +823,159 @@ $(CONSOLE
...
-    The `scope` storage class on the $(D dg) parameter means that the delegate does
-    not escape the scope of the $(D opApply) function (an example would be assigning $(D dg) to a
-    global variable). If it cannot be statically guaranteed that $(D dg) does not escape, a closure may
-    be allocated for it on the heap instead of the stack.
 
     $(BEST_PRACTICE Annotate delegate parameters to `opApply` functions with `scope` when possible.)
+
+    $(PANEL
+        The `scope` storage class on the `body` parameter means that the delegate does
+        not escape the scope of the apply function (an example would be assigning`body` to a
```
ntrel, line 831:
```suggestion
        not escape the scope of the apply function (e.g. assigning `body` to a
```

spec/statement.dd
```diff
@@ -811,57 +823,159 @@ $(CONSOLE
...
+        global variable). If it cannot be statically guaranteed that `body` does not escape, a closure may
+        be allocated for it on the heap instead of the stack.
     )
 
-    $(P $(B Important:) If $(D opApply) catches any exceptions, ensure that those
-        exceptions did not originate from the delegate passed to $(D opApply). The user would expect
+    $(P $(B Important:) If apply functions catch any exceptions, ensure that those
```
ntrel, line 836:
```suggestion
    $(P $(B Important:) If an apply function catches any exceptions, ensure that those
```

spec/statement.dd
```diff
@@ -811,57 +823,159 @@ $(CONSOLE
...
         exceptions thrown from a `foreach` body to both terminate the loop, and propagate outside
         the `foreach` body.
      )
 
 $(H4 $(LNAME2 template-op-apply, Template `opApply`))
 
-    $(P $(D opApply) can also be a templated function,
-        which will infer the types of parameters based on the $(I ForeachStatement).
-        For example:)
+    $(P `opApply` and `opApplyReverse` can also be a function templates,
```
ntrel, line 844:
```suggestion
    $(P `opApply` and `opApplyReverse` can also be function templates,
```

spec/statement.dd
```diff
@@ -811,57 +823,159 @@ $(CONSOLE
...
+        which can optionally infer the types of parameters based on the $(I ForeachStatement).
+    )
+
+    $(P $(B Note:) An apply function template cannot infer `foreach` variable types.)
```
ntrel, line 848:
```suggestion
    $(NOTE An apply function template cannot infer `foreach` variable types.)
```

spec/statement.dd
```diff
@@ -811,57 +823,159 @@ $(CONSOLE
...
-            foreach (int a, int b; S()) { }  // calls first opApply function
-            foreach (int a, int b, float c; S()) { }  // calls second opApply function
+            foreach (int a, int b; S()) { } // int delegate(ref int, ref int) pure nothrow @nogc @safe
+            foreach (bool b, string s; S()) { } // int delegate(ref bool, ref string) pure nothrow @nogc @safe
+            //foreach (x; S()) { } // Error: cannot infer type for `foreach` variable `x`, perhaps set it explicitly
         }
         --------------
         )
 
+$(H4 $(LNAME2 template-instance-op-apply, `opApply` as an alias to a explicit function template instance))
```
ntrel, line 870:
```suggestion
$(H4 $(LNAME2 template-instance-op-apply, `opApply` as an alias to an explicit function template instance))
```

spec/statement.dd
```diff
@@ -811,57 +823,159 @@ $(CONSOLE
...
+
+    $(P `opApply` and `opApplyReverse` can be aliases to an appropriate function or function template.
+        However, special treatment is given to apply functions that are aliases of an appropriate function template instance.)
+
+    $(P In that case, the function template instance is used for overload selection and `foreach` variable type inference,
+        but the template is instantiated again with the actual delegate type of the `foreach` body.
+        This way, apply functions can infer attributes depending on the attributes of `body` delegate.)
```
ntrel, line 877:
```suggestion
        This way, apply functions can infer attributes depending on the attributes of the `body` delegate.)
```

spec/statement.dd
```diff
@@ -1583,7 +1697,7 @@ foreach (item; list)
         $(P Any intervening $(RELATIVE_LINK2 try-statement, `finally`) clauses are executed,
         and any intervening synchronization objects are released.)
 
-        $(P $(D Note:) If a `finally` clause executes a `throw` out of the finally
+        $(P $(B Note:) If a `finally` clause executes a `throw` out of the finally
```
ntrel, line 1700:
```suggestion
        $(NOTE If a `finally` clause executes a `throw` out of the finally
```

ntrel (changes requested):

ntrel:
I think it would be better to implement [23666](https://issues.dlang.org/show_bug.cgi?id=23666) than have to document the alias workaround. However, the latter does work.

ntrel:
I don't think this fixes [23116](https://issues.dlang.org/show_bug.cgi?id=23116) - it's just a workaround for it.

spec/statement.dd
Bolpat: I didn’t like it either. I’m working on removing it.

Bolpat:
@ntrel, I now went over the relevant section with a text-to-speech tool and found (hopefully) the last issues. If the checks succeed, please have a look and merge if you don’t find any problems. Content-wise, I’m satisfied with how it looks and reads.

spec/statement.dd
```diff
@@ -808,60 +814,180 @@ $(GNAME OpApplyParameter):
...
+        )
+
+    $(PANEL The `opApplyImpl` pattern is generally preferable to
+        overloading many apply functions with all possible combinations of attributes.)
+
+    $(P Multiple apply function aliases can exist, and selection and `foreach` variable type inference work.
+        They can alias instances of the same function template or different function templates.)
+
+    --------------
+    class Tree(T)
```
ntrel, line 926: This example should be runnable too.

ntrel:
> please have a look and merge if you don’t find any problems

Just to mention, I don't have merge rights ;-)

spec/statement.dd
```diff
@@ -808,60 +814,180 @@ $(GNAME OpApplyParameter):
...
-            foreach (int a, int b; S()) { }  // calls first opApply function
-            foreach (int a, int b, float c; S()) { }  // calls second opApply function
+            foreach (int a, int b; S()) { } // int delegate(ref int, ref int) pure nothrow @nogc @safe
+            foreach (bool b, string s; S()) { } // int delegate(ref bool, ref string) pure nothrow @nogc @safe
+            //foreach (x; S()) { } // Error: cannot infer type for `foreach` variable `x`, perhaps set it explicitly
         }
         --------------
         )
 
+$(H4 $(LNAME2 template-instance-op-apply, `opApply` as an Alias to an Explicit Function Template Instance))
```
ntrel, line 866: This might be a better subheading:
```suggestion
$(H4 $(LNAME2 template-instance-op-apply, `opApply` with Attribute Inference))
```

Bolpat: Agreed. This is a “what it intends to do” vs. “what it technically is” ― and usually, “what it intends” is better. When I wrote this section, I focused on getting the technicalities right.

The introductory sentence to the subsection does not read weird, so I can just take your suggestion. 👍🏻

spec/statement.dd
```diff
@@ -698,108 +698,114 @@ foreach (string s, double d; aa)
...
+        These functions must each take exactly one function parameter of delegate type.
     )
 
-$(GRAMMAR
-$(GNAME OpApplyDeclaration):
-    `int opApply` `(` `scope` `int delegate` `(` $(I OpApplyParameters) `)` `dg` `)` `;`
-
-$(GNAME OpApplyParameters):
-    *OpApplyParameter*
-    *OpApplyParameter*, *OpApplyParameters*
```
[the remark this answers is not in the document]
Bolpat: It’s not a real grammar entity and just hangs there. Formally speaking, a grammar has a root ([`Module`](https://dlang.org/spec/grammar.html#Module) for D) from which parsing starts. Every other non-terminal symbol is only useful when it’s referenced as the right-hand side of any rule. There is no rule with `OpApplyDeclaration` and therefore, it’s useless. As far as the D grammar is concerned, `opApply` isn’t special and just like any other [function declaration](https://dlang.org/spec/function.html#FuncDeclaration).

In principle, we could add an `INFORMATIVE_GRAMMAR` block (cf. #3860) such as:
<pre>
int opApply ( scope int delegate <i><a href="https://dlang.org/spec/function.html#Parameters">Parameters</a></i> body )
</pre>
Honestly, I see little value in it. The examples do a great job at demonstrating how it’s done.

Apart from that, it’s technically wrong. The delegate type could given in various ways, such as `DG` which is an alias to a delegate type, or `opApply` could be a template that takes `DG` as a parameter. Or `opApply` can be an alias itself.

spec/statement.dd
```diff
@@ -698,108 +698,114 @@ foreach (string s, double d; aa)
...
-
-    $(P where each $(I OpApplyParameter) of `dg` must match a $(GLINK ForeachType)
-        in a $(GLINK ForeachStatement),
-        otherwise the *ForeachStatement* will cause an error.)
-
-    $(P Any *ForeachTypeAttribute* cannot be `enum`.)
-
     $(PANEL
-    To support a `ref` iteration variable, the delegate must take a `ref` parameter:
+    To support a `ref` iteration, the delegate must take `ref` parameters:
```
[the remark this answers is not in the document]
Bolpat: Yes, it should be rephrased. I’m not happy with it either. I don’t like _must_ here as it’s not a real requirement (such as `opApply` **must** return a nonzero result of `body`), it’s a mere conditional: If you want XYZ, do ABC.

```suggestion
    To support iteration by reference, let the delegate take the respective parameters as `ref` parameters:
```

What do you think?

spec/statement.dd
```diff
@@ -808,60 +814,180 @@ $(GNAME OpApplyParameter):
...
+            int opApplyImpl(Body)(scope Body body) => body(42);
+        }
+        void main() @nogc nothrow pure @safe
+        {
+            // Error: `@nogc` function `D main` cannot call non-@nogc function `onlineapp.A.opApply`
+            // Error: `pure` function `D main` cannot call impure function `onlineapp.A.opApply`
+            // Error: `@safe` function `D main` cannot call `@system` function `onlineapp.A.opApply`
+            // Error: function `onlineapp.A.opApply` is not `nothrow`
+            // (You might not encounter all of these errors at once.)
+            static assert(!__traits(compiles, () @safe {
```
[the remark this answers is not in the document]
Bolpat: > `@safe` is inherited from `main`.

In fact, it’s not:
```d
void main() @safe
{
    // static assert fails:
    static assert(!__traits(compiles, {
        int x;
        int* p = &x; // something unsafe
    }));
}
```
Without any attribute, the static assert fails.
I noticed that when I tried out, which is why I put it in in the first place.
Because the DAutoTest doesn’t like failing static asserts, I don’t see much of an alternative to leaving it in.

spec/statement.dd
```diff
@@ -808,60 +814,180 @@ $(GNAME OpApplyParameter):
...
+4
 )
-    $(PANEL
-    The `scope` storage class on the $(D dg) parameter means that the delegate does
-    not escape the scope of the $(D opApply) function (an example would be assigning $(D dg) to a
-    global variable). If it cannot be statically guaranteed that $(D dg) does not escape, a closure may
-    be allocated for it on the heap instead of the stack.
 
-    $(BEST_PRACTICE Annotate delegate parameters to `opApply` functions with `scope` when possible.)
+    $(BEST_PRACTICE Annotate delegate parameters to apply functions with `scope` when possible.)
```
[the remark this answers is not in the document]
Bolpat: grafik
I know exactly why I changed it. The “9.” looks bad. I honestly have no issue with it going after the PANEL, it would look like that (faked with HTML DOM manipulation):
grafik

Bolpat:
In all honesty, this is already too mission-creeped. Either merge this, or undo the changes except the new section, and put the improvements in a new PR.

@thewilsonator, review?

spec/statement.dd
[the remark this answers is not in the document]
ntrel: It's informative as to what the symbol should resolve to.

PR #3860, opened 2024-06-27
Replace `GRAMMAR_INFORMATIVE` by `INFORMATIVE_GRAMMAR`

Bolpat:
Because `GRAMMAR_INFORMATIVE` starts with `GRAMMAR`, there are stray ones included in [the grammar spec](https://dlang.org/spec/grammar.html).

### DIFF
--- a/book/dlang.org.ddoc
+++ b/book/dlang.org.ddoc
@@ -184,7 +184,7 @@ GLOSSARY = $(HTTP dlang.org/spec/glossary.html#$0, $0)
 GLOSSARY2 = $(HTTP dlang.org/spec/glossary.html#$1, $2)
 GNAME=<a id="$0">$(SPANC gname, $0)</a>
 GRAMMAR=$(TC pre, bnf notranslate, $0)
-GRAMMAR_INFORMATIVE=$(GRAMMAR $0)
+INFORMATIVE_GRAMMAR=$(GRAMMAR $0)
 GRAMMAR_LEX=$(GRAMMAR $0)
 GREEN=$(SPANC green, $0)
 GSELF=$(I $0)
--- a/dlang.org.ddoc
+++ b/dlang.org.ddoc
@@ -185,7 +185,7 @@ GLOSSARY = $(HTTP dlang.org/spec/glossary.html#$0, $0)
 GLOSSARY2 = $(HTTP dlang.org/spec/glossary.html#$1, $2)
 GNAME=<a id="$0">$(SPANC gname, $0)</a>
 GRAMMAR=$(TC pre, bnf notranslate, $0)
-GRAMMAR_INFORMATIVE=$(GRAMMAR $0)
+INFORMATIVE_GRAMMAR=$(GRAMMAR $0)
 GRAMMAR_LEX=$(GRAMMAR $0)
 GREEN=$(SPANC green, $0)
 GSELF=$(I $0)
--- a/spec/class.dd
+++ b/spec/class.dd
@@ -1446,7 +1446,7 @@ $(GNAME AnonBaseClassList):
 
 which is equivalent to:
 
-$(GRAMMAR_INFORMATIVE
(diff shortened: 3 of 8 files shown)


Bolpat:
I just checked [the prospect Grammar page](https://dtest.dlang.io/artifact/website-8fff91ddb2a6198ba8c65314c5ddc5283261a8df-ddfdd7aa1698db0e4dc66a210cfd8e4c/web/spec/grammar.html) and it looks good.

Bolpat:
@ntrel or @thewilsonator, one of you might take a quick look at this. Thanks in advance 😄

thewilsonator:
Why is this necessary or desired?

dkorpel (approved):

dkorpel:
> Why is this necessary or desired?

To remove blocks like this from the main D grammar:

image

Bolpat:
@thewilsonator, there is [a little D program](https://github.com/dlang/dlang.org/blob/master/ddoc/source/preprocessor.d#L293) that collects all blocks starting with `$(GRAMMAR` and adds them to the _D Grammar_ page. The issue is that `$(GRAMMAR_INFORMATIVE` starts with `$(GRAMMAR` and is collected as well. I thought about changing the script to do a lookahead and make sure `GRAMMAR` isn’t followed by another word character, but honestly, I didn’t understand the code, and IMO, it would have required some non-trivial change, as the `find` operation is in a UFCS chain. So I changed the name of the DDoc macro instead.

---

### [Before](https://dtest.dlang.io/artifact/website-868574ac69dd837646b9400862818274aa7fe177-8fbe664334ab1bf82ebb3fa69675fdeb/web/spec/grammar.html), the `GRAMMAR_INFORMATIVE` made _D Grammar_ have stray explanation grammar blocks:
grafik

### [With this change](https://dtest.dlang.io/artifact/website-8fff91ddb2a6198ba8c65314c5ddc5283261a8df-bcb26f2e70c72e6a348e771d920e1c92/web/spec/grammar.html), using `INFORMATIVE_GRAMMAR`, those aren’t included anymore:
grafik

### The second block of old is from [here](https://dtest.dlang.io/artifact/website-8fff91ddb2a6198ba8c65314c5ddc5283261a8df-bcb26f2e70c72e6a348e771d920e1c92/web/spec/class.html#anonymous), where it is part of an explanation that only makes sense in context:
grafik

thewilsonator:
Surely you can do 
```diff
-            enum grammarKey = "$(GRAMMAR";
+            enum grammarKey = "$(GRAMMAR";
```
then.

Bolpat:
> Surely you can do
> 
> ```diff
> -            enum grammarKey = "$(GRAMMAR";
> +            enum grammarKey = "$(GRAMMAR";
> ```
> 
> then.

Is this a joke that I don’t get? Those two lines are exactly the same.

> Or is that intended to pick up `"$(GRAMMAR_SUMMARY"` too?

`GRAMMAR_SUMMARY` is only used in one place. It’s likely that it’s being picked up, but doesn’t actually do anything when it’s picked up.

thewilsonator:
> Is this a joke that I don’t get? 

Nope, that was supposed to be 
```diff
-            enum grammarKey = "$(GRAMMAR";
+            enum grammarKey = "$(GRAMMAR ";
```

my bad.

Bolpat:
> > Is this a joke that I don’t get?
> 
> Nope, that was supposed to be
> 
> ```diff
> -            enum grammarKey = "$(GRAMMAR";
> +            enum grammarKey = "$(GRAMMAR ";
> ```
> 
> my bad.

Most definitely, a space or newline could follow. Possibly also something else. I’ll keep it as-is for now. The Right Thing would be to check for the word ending, but that’s non-trivial.

Bolpat:
@ntrel or @thewilsonator, merge?

Bolpat:
@thewilsonator, merge?

PR #3861, opened 2024-06-27
Consistently use `extern(X)`

Bolpat:
Linkage is not a control-flow statement (such as `if` or `while`) and as with `assert`, `pragma` and many others, no space should be used between the keyword and the opening parenthesis.

### DIFF
--- a/htod.dd
+++ b/htod.dd
@@ -121,7 +121,7 @@ $(P Produces the file test.d:)
 /* Converted to D from test.h by htod */
 module test;
 //C     unsigned u;
-extern (C):
+extern(C):
 uint u;
 //C     #define MYINT int
 //C     void bar(int x, long y, long long z);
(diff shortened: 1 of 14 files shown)


ntrel:
Makes sense as it is an attribute (though I'm used to seeing a space for it). I think all attributes with a parameter don't use a space, except `align (` sometimes.
Related - `version` and `debug` sometimes don't have a space in the spec.

Bolpat:
> Makes sense as it is an attribute (though I'm used to seeing a space for it). I think all attributes with a parameter don't use a space, except align ( sometimes.

Honestly, I’m just testing the waters here. If this is being rejected, I’m not trying `align` either.

> Related - `version` and `debug` sometimes don't have a space in the spec.

Related, but opposite. `version` and `debug` are conditional compilation / code-generation tools like `static if` and `static foreach`. Those are suggested by the [D Style page](https://dlang.org/dstyle.html) to have a space.

PR #3862, opened 2024-06-28
Condition → Conditional; IfCondition → Condition

Bolpat:
Using `IfCondition` in the grammar is misleading. It can be used for `if`, but also for `while` and `switch`. On the other hand, `Condition`, which refers to conditional compilation constructs, isn’t the stuff in parentheses of a `static if`, `debug` or `version`, but the whole `static if (…)` / `debug (…)` / `version (…)`; therefore, I replaced it with `Conditional`.

### DIFF
--- a/spec/statement.dd
+++ b/spec/statement.dd
@@ -234,10 +234,10 @@ $(P If statements provide simple conditional execution of statements.)
 
 $(GRAMMAR
 $(GNAME IfStatement):
-    $(D if $(LPAREN)) $(GLINK IfCondition) $(D $(RPAREN)) $(GLINK ThenStatement)
-    $(D if $(LPAREN)) $(GLINK IfCondition) $(D $(RPAREN)) $(GLINK ThenStatement) $(D else) $(GLINK ElseStatement)
+    $(D if $(LPAREN)) $(GLINK Condition) $(D $(RPAREN)) $(GLINK ThenStatement)
+    $(D if $(LPAREN)) $(GLINK Condition) $(D $(RPAREN)) $(GLINK ThenStatement) $(D else) $(GLINK ElseStatement)
 
-$(GNAME IfCondition):
+$(GNAME Condition):
     $(EXPRESSION)
     $(D auto) $(GLINK_LEX Identifier) $(D =) $(EXPRESSION)
     $(D scope) $(GLINK_LEX Identifier) $(D =) $(EXPRESSION)
@@ -283,7 +283,7 @@ $(H3 $(LNAME2 boolean-conditions, Boolean Conversion))
 $(H3 $(LNAME2 condition-variables, Condition Variables))
 
         $(P
-        When an $(I Identifier) form of *IfCondition* is used, a
+        When a variable declaration form of *Condition* is used, a
         variable is declared with that name and initialized to the
         value of the *Expression*.
         )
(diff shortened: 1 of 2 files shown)


dkorpel:
`Condition` and `Conditional` are easy to mix up. The dmd codebase calls IfCondition an AssignCondition, I suggest renaming it to that for consistency. 

If there's still need to clarify `Condition` then, perhaps it can become `StaticCondition`.

Bolpat:
> `Condition` and `Conditional` are easy to mix up.

I don’t know how you’d mix those up. A condition is something that you evaluate to a boolean to test it.
An *`IfCondition`* should be a condition for an `if` statement specifically, and is immediately a misnomer if it can be used anywhere where no `if` is present. IMO, *`Condition`* is the perfect name, but I’m not married to *`Conditional`*. I chose it because it it wasn’t taken and immediately follows the [*`ConditionalDeclaration`*](https://dlang.org/spec/grammar.html#ConditionalDeclaration) and [*`ConditionalStatement`*](https://dlang.org/spec/grammar.html#ConditionalStatement). Note that there’s also a [`ConditionalExpression`](https://dlang.org/spec/grammar.html#ConditionalExpression) which is unrelated to conditional compilation. (It should probably named *`ConditionExpression`*.)

`ConditionalCompilationIntroducer` instead for the `static if`/`version`/`debug` thing?

> The dmd codebase calls IfCondition an AssignCondition, I suggest renaming it to that for consistency.

I’d really like to keep *`Condition`* for what’s in `if`/`while`/`switch` parentheses because it perfectly matches what it means to the average programmer.

Using `AssignCondition` makes sense for the compiler, I guess, as it spends (probably) much more effort on these:
```
    auto Identifier = Expression
    scope Identifier = Expression
    TypeCtors Identifier = Expression
    TypeCtors? BasicType Declarator = Expression
```

If you want *`AssignCondition`* in the grammar, that would be my suggestion:
```diff
-   IfCondition:
+   Condition:
        Expression
+       AssignCondition
+
+   AssignCondition:
        auto Identifier = Expression
        scope Identifier = Expression
        TypeCtors Identifier = Expression
        TypeCtorsopt BasicType Declarator = Expression
```

Using *`AssignCondition`* for the *`Expression`* case is weird.

> If there's still need to clarify `Condition` then, perhaps it can become `StaticCondition`.
 
But *`Condition`* is simply wrong. It refers not the the boolean expression inside parentheses, but the whole “introducer” to the conditional compilation construct, e.g. <code>static if ( [AssignExpression](https://dlang.org/spec/grammar.html#AssignExpression) )</code> is an example.

dkorpel:
> ConditionalCompilationIntroducer

That sounds like a Java name :slightly_smiling_face: 

the_world_seen_by_an_objectoriented_programmer

> But Condition is simply wrong.

`static if (xyz)` `debug(xyz)` and `version(xyz)` are all different conditions to compile the following block, the condition isn't merely `xyz`.

But that's a bit besides the point. I think it's not worth bikeshedding over the best names for grammar production rules. They are mostly local. It's best to just pick something and be consistent with it. This PR changes them for marginally better clarity at best, so unless the names are made consistent with dmd (which is a nice property as they often need to be compared), I'd err on the side of keeping them as is.

Bolpat:
>  I think it's not worth bikeshedding

Okay, the let’s not do it! 😁 

> This PR changes them for marginally better clarity at best, so unless the names are made consistent with dmd (which is a nice property as they often need to be compared), I'd err on the side of keeping them as is.

You’re letting the perfect be the enemy of good.

IIUC, we’re arguing over _how much_ of an improvement the change is. To me, this improvement makes the spec a little more approachable. Most of my contributions are small, but I hope they add up.

The spec and the compiler are conceptually separate. I understand why the DMD devs want consistency, I even made DMD PRs to make grammar changes consistent after that was pointed out to me, and I’d be happy to file a DMD PR to adapt the names in the DMD source code _after_ this PR is merged. Why I would waste time preparing a DMD PR before is beyond me. Just realize that this isn’t a standard all potential contributors to this repo can meet.

dkorpel:
> IIUC, we’re arguing over _how much_ of an improvement the change is.

My point is: you think it's better, I think it's worse, so 0% improvement, and a PR should have at least positive improvement to get merged. 2 people is a low sample size, but it's not worth getting more people involved and discussing this at length.

> Okay, the let’s not do it! 😁

Shall we just close this then?

PR #3863, opened 2024-06-28
Allow `ref` and `auto ref` in `IfCondition` declarations

Bolpat:
As [DIP 1046](https://github.com/dlang/DIPs/blob/master/DIPs/accepted/DIP1046.md) *(`ref` For Variable Declarations)* has been accepted, probably, `ref` (and `auto ref`) should be allowed for the condition in `if`, `while`, and `switch` statements:

```d
int** pptr = …;
if (ref int* ptr = *pptr) …;
```
Of course, the condition tests whether the variable (`ptr` in the example) is `true`-like, not whether the reference is a non-null reference. In the example, it tests `ptr !is null`, and **not** `&ptr !is null` (or, equivalently, `pptr !is null`).

```d
if (auto ref x = mystery()) …;
```
Especially in template contexts, `auto ref` could be used to make `x` be a `ref` if `mystery()` returns by reference, but a value otherwise.

### DIFF
--- a/spec/statement.dd
+++ b/spec/statement.dd
@@ -241,6 +241,8 @@ $(GNAME IfCondition):
     $(EXPRESSION)
     $(D auto) $(GLINK_LEX Identifier) $(D =) $(EXPRESSION)
     $(D scope) $(GLINK_LEX Identifier) $(D =) $(EXPRESSION)
+    $(D ref) $(GLINK_LEX Identifier) $(D =) $(EXPRESSION)
+    $(D auto ref) $(GLINK_LEX Identifier) $(D =) $(EXPRESSION)
     $(GLINK2 type, TypeCtors) $(GLINK_LEX Identifier) $(D =) $(EXPRESSION)
     $(GLINK2 type, TypeCtors)$(OPT) $(GLINK2 type, BasicType) $(GLINK2 declaration, Declarator) $(D =) $(EXPRESSION)
 
--- a/spec/statement.dd
+++ b/spec/statement.dd
@@ -239,12 +239,18 @@ $(GNAME IfStatement):
 
 $(GNAME IfCondition):
     $(EXPRESSION)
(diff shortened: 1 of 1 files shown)


dkorpel:
This doesn't account for combinations, these also parse currently:

```D
if (scope ref x = b)
if (const ref x = b)
if (scope const ref x = b)
if (auto scope ref const x = b)
```

Bolpat:
I didn’t know `auto` and `ref` allowed other tokens between them. But apparently, they do.

Bolpat:
@dkorpel, it it fine? Merge?

dkorpel:
The current formulation doesn't allow `if (const ref x = y)`

dkorpel (approved):
I misread it

Bolpat:
> I misread it

No issue. I misread a lot, actually.

ntrel:
I don't think `if (ref x = y)` actually compiles yet. https://github.com/dlang/dmd/pull/16428 is not merged, and I don't know if it covers that case.

Bolpat:
> I don't think `if (ref x = y)` actually compiles yet. [dlang/dmd#16428](https://github.com/dlang/dmd/pull/16428) is not merged, and I don't know if it covers that case.

The spec technically isn’t supposed to document DMD’s behavior, but what DMD’s behavior ought to be. And it ought to be because the respective DIP was accepted.

ntrel:
@Bolpat I think it's confusing. There should at least be a note that the reference compiler doesn't implement it.

> it ought to be because the respective DIP was accepted.

The DIP didn't mention `if (ref`?

Bolpat:
> The DIP didn't mention `if (ref`?

The DIP is written rather incomplete. It’s of course intended.

I just tried out and `if (ref` works. `if (auto ref` does not work, but probably ought to.

ntrel:
Ok, thanks. The dmd pull is merged now anyway.

PR #3878, opened 2024-07-12
[spec/memory-safe-d] Improve `return` attribute docs

ntrel:
Fix Bugzilla 24659 - Memory safe D page lacks information on return ref.

Reword `return scope` docs (before this was just listed as `return`). 
Move sentence about constructor `return` attribute after method attribute.

### DIFF
--- a/spec/memory-safe-d.dd
+++ b/spec/memory-safe-d.dd
@@ -57,7 +57,7 @@ $(H2 $(LNAME2 usage, Usage))
          passed to functions. Such pointers include: raw pointers, arrays, `this`, classes, `ref` parameters, delegate/lazy parameters,
          and aggregates containing a pointer.)
 
-         $(P The $(DDSUBLINK spec/function, return-ref-parameters, `return ref`) attributes on a parameter indicates:)
+         $(P The $(DDSUBLINK spec/function, return-ref-parameters, `return ref`) attributes on a parameter indicate:)
 
          * A pointer derived from the parameter's address may be returned from the function
            (e.g. returning the parameter by reference).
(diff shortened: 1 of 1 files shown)


rikkimax:
@dkorpel is this enough? It doesn't seem to be.

ntrel:
I added a link rather than repeat examples - this page doesn't have examples for return scope either.

dkorpel:
> is this enough

Honestly most of the memory safety page is redundant specification-wise, it's mostly informative.

spec/memory-safe-d.dd
```diff
@@ -57,19 +57,31 @@ $(H2 $(LNAME2 usage, Usage))
          passed to functions. Such pointers include: raw pointers, arrays, `this`, classes, `ref` parameters, delegate/lazy parameters,
          and aggregates containing a pointer.)
 
+         $(P The $(DDSUBLINK spec/function, return-ref-parameters, `return ref`) attributes on a parameter indicates:)
```
dkorpel, line 60:
```suggestion
         $(P The $(DDSUBLINK spec/function, return-ref-parameters, `return ref`) attribute on a parameter indicates:)
```

Or

```suggestion
         $(P The $(DDSUBLINK spec/function, return-ref-parameters, `return ref`) attributes on a parameter indicate:)
```

spec/memory-safe-d.dd
```diff
@@ -57,19 +57,31 @@ $(H2 $(LNAME2 usage, Usage))
...
+
+         * A pointer derived from the parameter's address may be returned from the function
+           (e.g. returning the parameter by reference).
+         * A pointer derived from the parameter's address may be stored in the first parameter of
+           a function returning `void`, *iff* the first parameter is `ref`.
```
dkorpel, line 65: Or the implicit `this` parameter. Perhaps instead of repeating the exact rules, say 'or returned by assignment to another parameter' with a link to the exact rules.

ntrel: Updated.

dkorpel (approved):

PR #3885, opened 2024-07-22
[spec/expression] Add array literal heap allocation section

ntrel:
Also tweak `@nogc` items.

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -2159,6 +2159,9 @@ $(H4 $(LNAME2 array-literal-heap, GC Allocation))
 
     * It initializes or assigns to a static array.
     * It initializes a $(DDSUBLINK spec/attribute, scope, `scope`) slice.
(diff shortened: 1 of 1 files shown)


spec/expression.dd
```diff
@@ -2153,6 +2155,29 @@ $(GNAME ArrayLiteral):
         }
         ---
 
+    $(P An array literal is not GC allocated if:)
+
+    * It initializes or assigns to a static array.
+    * It initializes a $(DDSUBLINK spec/attribute, scope, `scope`) slice.
+
```
dkorpel, line 2162: If it's used in a `foreach` where the element type isn't `ref`, it's also not GC allocated

```D
foreach (i; [6, 7]) {} // converted to static array
foreach (ref i; [8, 9]) {} // GC allocated because `ref` may escape
```

dkorpel (approved):

spec/expression.dd
ntrel: Thanks, I also added 2 other cases.

PR #3886, opened 2024-07-23
Rephrase IfCondition variables

Bolpat:


spec/statement.dd
```diff
@@ -296,18 +294,28 @@ $(H3 $(LNAME2 condition-variables, Condition Variables))
...
         variable is declared with that name and initialized to the
-        value of the *Expression*.
+        value of the $(I Expression).
         )
 
-        * If $(D auto) $(I Identifier) is provided, the type of the variable
-          is the same as *Expression*.
+        *   If both `auto` and `ref` are present among $(I IfConditionStorageClasses),
+            the variable is a reference
+            if and only if the $(I Expression) is an lvalue;
```
dkorpel, line 302: On second thought, given that the DIP doesn't describe this, and the implementation doesn't implement it, and Walter has described his dislike for `auto ref` in the past, I don't think the spec should define `auto ref` semantics for `if`, so I don't think this PR should elaborate on it.

I'll ask Walter what the intended behavior is here.

dkorpel: He said pulled spec changes shouldn't predate an implementation, and it should be discussed on the news group first, so I suggest removing this specification.

dkorpel (approved):
The `auto ref` semantics are the same between IfCondition and regular locals, so this can be factored out in a next PR.

spec/statement.dd
dkorpel: Implementation is coming (https://github.com/dlang/dmd/pull/16771)

PR #3887, opened 2024-07-25
Fix `FunctionLiteral` grammar

Bolpat:
In the function literal grammar, it says that after the `function`/`delegate` keyword and optionally `ref` or `auto ref`, there can follow a general [*`Type`*](https://dlang.org/spec/type.html#Type):
<pre>
<i>FunctionLiteral</i>:
    function <i>RefOrAutoRef</i>? <i>Type</i>? <i>ParameterWithAttributes</i>? <i>FunctionLiteralBody</i>
    delegate <i>RefOrAutoRef</i>? <i>Type</i>? <i>ParameterWithMemberAttributes</i>? <i>FunctionLiteralBody</i>
    …
</pre>

The compiler does not support general *`Type`*, which is <code>*TypeCtors*? *BasicType* *TypeSuffixes*?</code>, but only <code>*BasicType* *TypeSuffixes*?</code>.

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -2274,11 +2274,14 @@ $(H3 $(LNAME2 function_literals, Function Literals))
 
 $(GRAMMAR
 $(GNAME FunctionLiteral):
-    $(D function) $(GLINK RefOrAutoRef)$(OPT) $(GLINK2 type, Type)$(OPT) $(GLINK ParameterWithAttributes)$(OPT) $(GLINK FunctionLiteralBody)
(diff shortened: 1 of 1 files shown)


dkorpel (approved):
Seems like an unnecessary limitation, is there a bugzilla issue for this?

Bolpat:
> Seems like an unnecessary limitation, is there a bugzilla issue for this?

It’s not that unnecessary as normally, `ref` and storage-class type qualifiers commute with `ref`, i.e. normally, one can write `const ref int` or `ref const int` interchangeably. Only here, you couldn’t.

Another issue is that with primary type syntax, `ref` would be viable to introduce a general type, but not a basic type. So if the spec allows a general type, one could write, e.g.:
```d
function ref ref int function()[] () => *null;
```
instead of:
```d
function ref (ref int function())[] () => *null;
```
Which is a lot clearer.

This is how I found this, by the way.

PR #3915, opened 2024-09-23
Fix Bugzilla 14945 - unions are missing from the ABI page

ntrel:
Also list empty struct size and add example.

### DIFF
--- a/spec/abi.dd
+++ b/spec/abi.dd
@@ -68,9 +68,14 @@ $(H2 $(LNAME2 delegates, Delegates))
         function's stack frame (nested functions).
         )
 
-$(H2 $(LNAME2 structs, Structs))
+$(H2 $(LNAME2 structs, Structs and Unions))
 
-        $(P Conforms to the target's C ABI struct layout.)
+        $(P These conform to the target's C ABI struct layout, except:)
+
+        * An `extern(D)` struct or union with no fields has size 1, not 0.
+          See $(DDSUBLINK spec/struct, struct_layout, Struct Layout).
+        * Unions are $(RELATIVE_LINK2 name_mangling, name-mangled) as if they were structs.
+        * The rest of this document treats unions extrinsically the same as structs.
 
 $(H2 $(LNAME2 classes, Classes))
 
--- a/spec/struct.dd
+++ b/spec/struct.dd
@@ -176,7 +176,18 @@ $(H2 $(LNAME2 struct_layout, Struct Layout))
     the first field and the start of the object.
     )
 
-    $(P Structs with no fields of non-zero size (aka $(I Empty Structs)) have a size of one byte.)
+    $(P $(DDSUBLINK spec/attribute, linkage, `extern(D)`) structs with no fields of non-zero
+    size (aka $(I Empty Structs)) have a size of one byte.)
+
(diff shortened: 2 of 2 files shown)


spec/struct.dd
```diff
@@ -176,7 +176,18 @@ $(H2 $(LNAME2 struct_layout, Struct Layout))
     the first field and the start of the object.
     )
 
-    $(P Structs with no fields of non-zero size (aka $(I Empty Structs)) have a size of one byte.)
+    $(P $(DDSUBLINK spec/attribute, linkage, `extern(D)`) structs with no fields of non-zero
```
Bolpat, line 179: Avoid the double negation “no … with non-…”ł and instead use “all … with …”.
```suggestion
    $(P $(DDSUBLINK spec/attribute, linkage, `extern(D)`) structs with all fields of zero
```
Mathematically speaking, any _all_ statement about things is vacuously true if there are none. Most people who read this part of the spec will understand it, but if you’re in doubt, you could do this instead:
```suggestion
    $(P $(DDSUBLINK spec/attribute, linkage, `extern(D)`) structs with no fields or all fields of zero
```

spec/abi.dd
```diff
@@ -68,9 +68,14 @@ $(H2 $(LNAME2 delegates, Delegates))
         function's stack frame (nested functions).
         )
 
-$(H2 $(LNAME2 structs, Structs))
+$(H2 $(LNAME2 structs, Structs and Unions))
 
-        $(P Conforms to the target's C ABI struct layout.)
+        $(P These conform to the target's C ABI struct layout, except:)
+
+        * An `extern(D)` struct or union with no fields has size 1, not 0.
```
Bolpat, line 75: Also `extern(C++)` structs seem to have non-zero size judging by running a simple example. The C++ Standard (all versions) agrees with that.
```suggestion
        * An `extern(D)` or `extern(C++)` struct or union with no fields has size 1, not 0.
```

spec/struct.dd
```diff
@@ -176,7 +176,18 @@ $(H2 $(LNAME2 struct_layout, Struct Layout))
...
+    size (aka $(I Empty Structs)) have a size of one byte.)
+
+    $(SPEC_RUNNABLE_EXAMPLE_COMPILE
+    ---
+    extern(C) struct C {}
+    struct D {}
+
+    static assert(C.sizeof == 0);
+    static assert(D.sizeof == 1);
```
Bolpat, line 188: Might be worth adding an `extern(C++)` example.
```suggestion
    extern(C) struct C {}
    extern(C++) struct Cpp {}
    struct D {}

    static assert(C.sizeof == 0);
    static assert(Cpp.sizeof == 1);
    static assert(D.sizeof == 1);
```

spec/abi.dd
ntrel: This paragraph is about differences from C. The link that follows is for further information.

spec/struct.dd
ntrel: I already linked to `extern()` where C++ linkage is described.

ntrel:
Rebased, tests now passing.

dkorpel (approved):

PR #3916, opened 2024-09-25
Simplify template parameter specialization and default

Bolpat:
For template alias parameters, [specialization](https://dlang.org/spec/template.html#TemplateAliasParameterSpecialization) and [default](https://dlang.org/spec/template.html#TemplateAliasParameterDefault) are simply stated as optional, but for template [type](https://dlang.org/spec/template.html#TemplateTypeParameter) and [value](https://dlang.org/spec/template.html#TemplateValueParameter) parameters, the optionality of specialization and default are given in 4 different lines. That is inconsistent and unnecessary. Parsing the optional constructs is simple as each has its own specific introducer, `:` for a specialization and `=` for default.

### DIFF
--- a/spec/template.dd
+++ b/spec/template.dd
@@ -288,10 +288,7 @@ $(H3 $(LNAME2 template_type_parameters, Type Parameters))
 
 $(GRAMMAR
 $(GNAME TemplateTypeParameter):
-    $(GLINK_LEX Identifier)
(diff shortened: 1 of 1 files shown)


Bolpat:
@thewilsonator or @ntrel, this is fairly trivial

thewilsonator (approved):

PR #3917, opened 2024-09-26
Replace `BasicType` by `PrimaryType`

Bolpat:
Essentially, this change renames *`BasicType`* to *`PrimaryType`* and gives *`BasicType`* a new meaning.

The entity *`BasicType`* and the term *basic type* will be used for the non-recursive rules in the type grammar. The recursive rules are <code>*TypeCtor*(*Type*)</code> and <code>__vector(*Type*)</code>. Those would constitute *elaborate primary types* (with *elaborate* = *non-basic*). The non-recursive rules are e.g. fundamental types, identifiers, `mixin` types, `typeof` types, etc., stuff that doesn’t immediately require parsing a *`Type`* again.

This is in prospect of the Primary Type Syntax DIP which, if accepted, will make *`PrimaryType`* truly be the type equivalent to `PrimaryExpression` in the expression syntax. Note that any DIP that adds tuples to the grammar based on parentheses, e.g. one that makes `(int, int)` a type, will profit from this change because such a type would be an elaborate primary type, and parts of the grammar spec (or forum discussions) can then easily distinguish *primary* and *basic* types in places where parsing ambiguity must be avoided.

In particular, this change allows the grammar spec to require a non-trivial type construct that is not a qualified type in code. This makes sense in places where the qualifier would be ignored if given, e.g. base class lists and, possibly in the future, the argument of `__vector`.

I’m not that deep into the SIMD types, so I left it alone, but as far as I understand, it could be constrained syntactically so that providing an ill-formed vector type fails fast. Essentially, a vector type would probably be <code>__vector(*BasicType* *TypeSuffixes*<sub>opt</sub>)</code> where the *`TypeSuffixes`* are optional because if *`BasicType`* is something other than a *`FundamentalType`*, it could already be a static array.

The parser changes required by this change are implemented in [this PR](https://github.com/dlang/dmd/pull/16882).

## Changes

Most changes are *`BasicType`* to *`PrimaryType`*.

### class.dd

Remove the unnecessary *`SuperClassOrInterface`* and *`Interfaces`* / *`Interface`* grammar entities, those are all *`BasicType`* now. A base class should not be spelled out in code as e.g. `const(Object)` as it’s equivalent to `Object`. if a `mixin` or ´typeof` resolves to a qualified type, the qualifiers are still ignored. This is about parsing only.

### declaration.dd

Rephrased a paragraph mentioning “basic type” with no actual connection to *`BasicType`*.

### expression.dd

Also renames *`BasicTypeWithSuffixes`* to *`SuffixedPrimaryType`*

### simd.dd

Because I inlined the `Vector` grammar entity, some rephrasing was needed.

### template.dd

For *`TemplateValueParameter`*, I changed *`BasicType`* to *`Type`* because this is what the compiler implements and there’s no good reason to change the compiler. This change would also encompass the Primary Type Syntax DIP and (likely) any tuple DIP.

### type.dd

Rephrased a paragraph mentioning “basic type” with no actual connection to *`BasicType`*.


spec/expression.dd
```diff
@@ -1644,7 +1601,7 @@ $(GNAME PostfixExpression):
     $(GSELF PostfixExpression) $(D ++)
     $(GSELF PostfixExpression) $(D --)
     $(GSELF PostfixExpression) $(D $(LPAREN)) $(GLINK NamedArgumentList)$(OPT) $(D $(RPAREN))
-    $(GLINK2 type, TypeCtors)$(OPT) $(GLINK2 type, BasicType) $(D $(LPAREN)) $(GLINK NamedArgumentList)$(OPT) $(D $(RPAREN))
```
tim-dlang, line 1647: Why remove `$(GLINK2 type, TypeCtors)$(OPT)`? The following compiles with DMD:
```
auto x = const int(5);
```

spec/class.dd
```diff
@@ -32,18 +32,12 @@ $(GNAME ClassDeclaration):
...
 
-$(GNAME SuperClassOrInterface):
-    $(GLINK2 type, BasicType)
-
-$(GNAME Interfaces):
-    $(GLINK Interface)
-    $(GLINK Interface) $(D ,) $(GSELF Interfaces)
-
-$(GNAME Interface):
-    $(GLINK2 type, BasicType)
```
tim-dlang, line 46: SuperClassOrInterface and Interfaces are still referenced by AnonBaseClassList and BaseInterfaceList.

spec/type.dd
```diff
@@ -24,23 +24,20 @@ $(GNAME TypeCtor):
     $(D inout)
     $(D shared)
 
+$(GNAME PrimaryType):
+    $(GLINK BasicType)
+    $(D __vector) $(D $(LPAREN)) $(GLINK Type) $(D $(RPAREN))
+    $(GLINK TypeCtor)$(OPT) $(D $(LPAREN)) $(GLINK Type) $(D $(RPAREN))
```
tim-dlang, line 30: Should this be without `$(OPT)` for now? This change is from your DIP, but I thought this pull request is only a preparation.

Bolpat:
There have been some unintentional changes slipping through. I don’t know how or why… I’ll look into this tomorrow.

spec/type.dd
Bolpat: Fixed.

spec/class.dd
Bolpat: Removed. 👍🏻

spec/expression.dd
Bolpat: You’re right. I added it back in.

PR #3919, opened 2024-10-19
Spec: specify what a TemplateMixinDeclaration is

WebFreak001:
I haven't checked with the compiler source if this actually holds true, but when trying it out with the compiler I get

```
a.d(10): Error: template instance `foo!()` mixin templates are not regular templates
```

which makes me believe this is how it works. Please correct me if I'm wrong here.


dkorpel (approved):

WebFreak001:
so can we hit merge?

PR #3920, opened 2024-10-23
Update areas-of-d-usage.dd

cyrusmsk:
Updated game engines details in language description. Previous links are not presentable anymore - Dash is just not opening, and the gfm is archived. I think new examples should be updated for people who will be interested. Maybe bindings for Godot and Raylib also worth to mention here.


schveiguy:
This is still up, but it's really old and out of date (9 years old). https://github.com/Circular-Studios/Dash

schveiguy (approved):
I think the changes are fine, good to post to newer/maintained engines. I think there's also https://code.dlang.org/packages/pixelperfectengine which is maintained by @ZILtoid1991

dkorpel (approved):

PR #4154, opened 2024-12-29
add move constructor documentation

WalterBright:
As promised.

### DIFF
--- a/spec/struct.dd
+++ b/spec/struct.dd
@@ -1373,6 +1373,268 @@ $(H3 $(LNAME2 implicit-copy-constructors, Implicit Copy Constructors))
 
     $(P If the generated copy constructor fails to type check, it will receive the `@disable` attribute.)
 
+$(H2 $(LEGACY_LNAME2 StructMoveConstructor, struct-move-constructor, Struct Move Constructors))
+
+    $(P Move constructors are very much like copy constructors. The difference is that copy
+    constructors make a copy of the original, while move constructors move the contents of
+    the original, and the lifetime of the original ends.)
+
+    $(NOTE Do not use postblits in the same struct with move constructors.)
+
+    $(COMMENT $(P A `struct` that defines a move constructor
+    is not $(RELATIVE_LINK2 POD, POD).))
+    $(P If a move constructor is declared, also declare a copy constructor.)
+
+    $(P A constructor declaration is a move constructor declaration if it meets
+    the following requirements:)
+
+    $(UL
+    $(LI It takes exactly one parameter without a
+    $(DDSUBLINK spec/function, function-default-args, default argument),
+    followed by any number of parameters with default arguments.)
+
+    $(LI Its first parameter is not a
+    $(DDSUBLINK spec/function, ref-params, `ref` parameter).)
+
+    $(LI The type of its first parameter is the same type as
+    $(DDSUBLINK spec/type, typeof, `typeof(this)`), optionally with one or more
+    $(DDLINK spec/const3, Type Qualifiers, type qualifiers) applied to it.)
+
+    $(LI It is not a
+    $(DDSUBLINK spec/template, template_ctors, template constructor declaration).)
+    )
+
+    ---
+    struct A
+    {
+        this(ref return scope A rhs) {}                    // copy constructor
+        this(return scope A rhs) {}                        // move constructor
+        this(return scope const A rhs, int b = 7) {}       // move constructor with default parameter
+    }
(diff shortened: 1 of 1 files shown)


spec/struct.dd
```diff
@@ -1373,6 +1373,229 @@ $(H3 $(LNAME2 implicit-copy-constructors, Implicit Copy Constructors))
 
     $(P If the generated copy constructor fails to type check, it will receive the `@disable` attribute.)
 
+$(H2 $(LEGACY_LNAME2 StructMoveConstructor, struct-move-constructor, Struct Move Constructors))
+
+    $(P Move constructors are very much like copy constructors. The difference is that copy
+    constructors make a copy of the original, while move constructors move the contents of
+    the original, and the lifetime of the original ends.)
+
+    $(NOTE Do not use postblits in the same struct with move constructors.)
```
rikkimax, line 1382: If we have to have a warning like this, we should probably not support it.

WalterBright: I plan on investigating if postblits should be allowed if there are move constructors.

rikkimax: What I'm wondering here is perhaps we have been thinking about postblit a bit wrong.

Instead of it being the third hook, make it the common denominator of both copy/move constructors. So it is always called after they are.

WalterBright: The rationale behind postblit was as another way to do a move constructor, not an enhancement to move constructors.

spec/struct.dd
```diff
@@ -1373,6 +1373,237 @@ $(H3 $(LNAME2 implicit-copy-constructors, Implicit Copy Constructors))
...
+
+    $(COMMENT $(P A `struct` that defines a move constructor
+    is not $(RELATIVE_LINK2 POD, POD).))
```
ntrel, line 1385: Just to check, was this intended to appear in the output? (The COMMENT macro is ignored).

spec/struct.dd
```diff
@@ -1373,6 +1373,237 @@ $(H3 $(LNAME2 implicit-copy-constructors, Implicit Copy Constructors))
...
+    lvalue can be coerced into being an rvalue using
+    $(DDSUBLINK spec/expression, RvalueExpression, `__rvalue(Expression)`).)
+
+    $(P If a move constructor is defined, implicit calls to it will be inserted
+    in the following situations:)
+
+    $(OL
+    $(LI When a variable is explicitly initialized:)
+
+    $(SPEC_RUNNABLE_EXAMPLE_RUN
```
ntrel, line 1428: All the added runnable/compilable examples in this PR fail to compile on run.dlang.org due to the compiler version it has:

> Error: `struct A` may not define both a rvalue constructor and a copy constructor

spec/struct.dd
```diff
@@ -1373,6 +1373,237 @@ $(H3 $(LNAME2 implicit-copy-constructors, Implicit Copy Constructors))
...
+        this(return scope A another) {} // move constructor
+        this(ref return scope A rhs) { assert(0); }
+    }
+
+    void fun(A a) {}
+
+    void main()
+    {
+        A a;
+        fun(__rvalue(a)); // no constructor gets called
```
ntrel, line 1469: The comment is confusing - the paragraph says an implicit move constructor call will be inserted.

spec/struct.dd
```diff
@@ -1373,6 +1373,237 @@ $(H3 $(LNAME2 implicit-copy-constructors, Implicit Copy Constructors))
...
+        // `a` is no longer a valid object
+    }
+    ---
+    )
+)
+
+$(H3 $(LNAME2 disable-move, Disabled Moving))
+
+    $(P When a move constructor is defined for a `struct` (or marked `@disable`), the compiler no
+    longer implicitly generates default move constructors for that `struct`:
```
ntrel, line 1479: I suggest a link to the relevant section which the reader won't have seen yet.
```suggestion
    longer $(RELATIVE_LINK2 implicit-move-constructors, implicitly generates) default move constructors for that `struct`:
```

WalterBright: added

spec/struct.dd
WalterBright: It was awaiting the compiler PR.

spec/struct.dd
WalterBright: COMMENT macros are ignored on purpose.

WalterBright:
ready to pull

dkorpel (approved):

PR #4156, opened 2024-12-30
[spec/function] Remove MissingFunctionBody from FunctionBody

### LINKED ISSUE #4101: grammar allows missing function body for AutoFuncDeclaration
### Description

This does not compile:
```
void foo()() {
    nothrow bar() {} // allowed
    nothrow bar(); // should parse according to spec
}
```
The spec says that in an AutoFuncDeclaration a MissingFunctionBody is allowed, yet even before instantiating template foo it raises an error:
Error: function declaration without return type.

I suggest only allowing SpecifiedFunctionBody in an AutoFuncDeclaration since it would always be a semantic error anyways (you can't infer the return type without a function body).

https://dlang.org/spec/grammar.html#AutoFuncDeclaration
https://dlang.org/spec/grammar.html#FunctionBody

ntrel:
Fixes #4101.
I think this is more natural - as a prototype does not have a body, so why would it be grouped under something called _FunctionBody_.

Changing the definition resolves an inaccuracy in the spec here:

> A disabled default constructor may not have a FunctionBody.

https://dlang.org/spec/struct.html#disable_default_construction

Which becomes valid with this pull.

I also noticed that [AutoFuncDeclaration](https://dlang.org/spec/function.html#AutoFuncDeclaration) seems to be the only kind of function declaration that gives a parse error on a _MissingFunctionBody_ - `static this()` doesn't (I suppose the body definition could be added at link time).

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -9,6 +9,7 @@ $(H2 $(LNAME2 grammar, Function Declarations))
 $(GRAMMAR
 $(GNAME FuncDeclaration):
     $(GLINK2 declaration, StorageClasses)$(OPT) $(GLINK2 type, BasicType) $(GLINK FuncDeclarator) $(GLINK FunctionBody)
+    $(GLINK2 declaration, StorageClasses)$(OPT) $(GLINK2 type, BasicType) $(GLINK FuncDeclarator) $(GLINK MissingFunctionBody)
     $(GLINK AutoFuncDeclaration)
 
 $(GNAME AutoFuncDeclaration):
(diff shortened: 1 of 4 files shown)


spec/function.dd
```diff
@@ -3448,21 +3448,17 @@ $(H2 $(LNAME2 main, $(D main()) Function))
         $(D main()) must be declared as follows:
         )
 
-        $(GRAMMAR
-        $(GNAME MainFunction):
-            $(GLINK MainReturnDecl) $(D main$(LPAREN)$(RPAREN)) $(GLINK2 statement, MainFunctionBody)
```
ntrel, line 3453: Note: MainFunctionBody link should have been GLINK.

dkorpel (approved):

PR #4164, opened 2025-01-18
add RefCast to Expressions

WalterBright:
Implementation PR: https://github.com/dlang/dmd/pull/20728

### DIFF
--- a/spec/expression.dd
+++ b/spec/expression.dd
@@ -1199,7 +1199,8 @@ $(H3 $(LNAME2 cast_expressions, Cast Expressions))
 $(GRAMMAR
 $(GNAME CastExpression):
     $(D cast $(LPAREN)) $(GLINK2 type, Type) $(D $(RPAREN)) $(GLINK UnaryExpression)
-    $(GLINK CastQual)
+    $(GLINK RefCast)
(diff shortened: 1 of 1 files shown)


spec/expression.dd
```diff
@@ -1598,14 +1599,31 @@ $(H4 $(LNAME2 cast_struct, Structs))
...
 
+$(H4 $(LNAME2 refcast, Reference Cast))
+
+$(GRAMMAR
+$(GNAME RefCast):
+    $(D cast $(LPAREN)) $(D ref) $(GLINK2 type, Type) $(D $(RPAREN)) $(GLINK UnaryExpression)
+)
+
+    $(P A $(I RefCast) takes a $(I UnaryExpression) representing an lvalue and forcibly changes
+    its type to $(I Type). It is equivalent to the expression $(D *cast(Type*)&UnaryExpression).
```
ntrel, line 1610:
```suggestion
    its type to $(I Type), producing an lvalue. It is equivalent to the expression $(D *cast(Type*) &UnaryExpression).
```

ntrel (changes requested):
This needs to add *RefCast* to https://dlang.org/spec/function.html#safe-functions

PR #4165, opened 2025-01-25
Update traits.dd

chibisi:
Added template alias example to isSame trait

### DIFF
--- a/spec/traits.dd
+++ b/spec/traits.dd
@@ -1931,6 +1931,30 @@ alias daz = foo;
 static assert(__traits(isSame, foo, daz));
 ---
 )
+        $(P isSame will also match template symbols as well as types.)
+
+$(SPEC_RUNNABLE_EXAMPLE_COMPILE
+---
+struct Foo(T){
+    T x;
+}
+
+struct Bar(T){
+    T x;
+}
+
+struct Point(T){
+    T x;
+    T y;
+}
+
+enum isFooOrBar(alias FB) = __traits(isSame, FB, Foo) || __traits(isSame, FB, Bar);
+
+static assert(isFooOrBar!(Foo));//true
+static assert(isFooOrBar!(Bar));//true
+static assert(isFooOrBar!(Point));//false
+---
+)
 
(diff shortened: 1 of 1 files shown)


chibisi:
Hmm, let me make an alteration and try to submit again.

spec/traits.dd
```diff
@@ -1930,6 +1930,30 @@ static assert(__traits(isSame, object, object));
...
+struct Point(T){
+    T x;
+    T y;
+}
+
+enum isFooOrBar(alias FB) = __traits(isSame, FB, Foo) || __traits(isSame, FB, Bar);
+
+static assert(isFooOrBar!(Foo));//true
+static assert(isFooOrBar!(Bar));//true
+static assert(isFooOrBar!(Point));//false
```
rikkimax, line 1955: No need to write out the result, the static assert already does that.

```suggestion
static assert(isFooOrBar!(Foo));
static assert(isFooOrBar!(Bar));
static assert(!isFooOrBar!(Point));
```

spec/traits.dd
```diff
@@ -1930,6 +1930,30 @@ static assert(__traits(isSame, object, object));
 alias daz = foo;
 static assert(__traits(isSame, foo, daz));
 ---
+)
+        $(P isSame will also match template symbols as well as types.)
```
rikkimax, line 1934: I'm not 100% on this proposal, but it might be a little better.

```suggestion
        $(P isSame matches against non-instantiated templates.)
```

rikkimax (changes requested):

spec/traits.dd
rikkimax:
```d
__stdin.d(18): Error: static assert:  `isFooOrBar!(Point)` is false
static assert(isFooOrBar!(Point));//false
^
```

Test failing due to missing not.

rikkimax (approved):

rikkimax:
Well done!

chibisi:
> Well done!

Thanks!

PR #4167, opened 2025-02-10
Ban undefined behavior at CTFE

Bolpat:
D should ban undefined behavior (UB) during compile-time function execution (CTFE) to ensure that the results are consistent and predictable. This restriction helps maintain the integrity of CTFE and guarantees that any constant expressions will yield the same result every time they are evaluated, give or take implementation-defined features. By avoiding undefined behavior, the language lives up the goal of being safe and reliable. If UB were allowed at CTFE, deterministic builds become impossible as a feature.

Considering what’s already banned during CTFE, it makes no sense whatsoever that UB is not generally ruled out.

This is not a breaking change because UB being what it is, allows the compiler to issue a diagnostic in the meaning of the old spec. The new spec would require a diagnostic and any remaining UB at CTFE that’s not caught constitutes a compiler bug.

C++ bans UB during CTFE (“constant evaluation” in C++ terminology) for the reasons mentioned in the first paragraph (possibly among others).

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -3595,6 +3595,8 @@ void main()
         Casts between signed and unsigned types are permitted.)
 
     $(LI Reinterpretation of overlapped fields in a union is not permitted.)
+
+    $(LI Undefined behavior must not occur.)
     )
 
     $(P Pointers are permitted in CTFE, provided they are used safely:)
@@ -3681,7 +3683,7 @@ static assert(countTen(12) == 12);  // invalid, modifies y.
 
     $(IMPLEMENTATION_DEFINED
     Functions executed via CTFE can give different results
-    from run time when implementation-defined or undefined-behavior occurs.
+    from run time when implementation-defined occurs.
     )
 
 


thewilsonator (approved):
Do you have any specific examples of undefined behaviour that must not occur?

jmdavis:
> Do you have any specific examples of undefined behaviour that must not occur?

I think that it's in response to: https://github.com/dlang/dmd/issues/20827

Mathias asked whether it was in the spec that UB is banned during CTFE, and I guess that he didn't find it there, so he's now adding it to the spec. I don't know if that's what we ultimately want or not. CTFE does already ban a number of useful things, but actually enforcing no UB during CTFE would likely lead to yet more things being banned. I don't know what currently can happen during UTFE that could be considered UB, but since you can obviously have UB  during runtime, banning it during compile time means that there has to be stuff that you can do at runtime that you can't do at compile time, and if anything, any time that those restrictions pop up, it's annoying. We obviously can't ever do everything that we can do at runtime during CTFE, but in general, it would be nice if we could do more rather than less.

So, maybe this is a good idea, and maybe it isn't. In essence, it's an argument for requiring that code be `@safe` to be callable during CTFE.

Bolpat:
> > Do you have any specific examples of undefined behaviour that must not occur?
> 
> I think that it's in response to: [dlang/dmd#20827](https://github.com/dlang/dmd/issues/20827)

Exactly this. Why you would want to actually mutate a `const` in CTFE is beyond me. Another example would be casting a function pointer from impure to `pure`. I can’t imagine banning UB takes away anything useful.

jmdavis:
> Exactly this. Why you would want to actually mutate a const in CTFE is beyond me. Another example would be casting a function pointer from impure to pure. I can’t imagine banning UB takes away anything useful.

If the function pointer is used in a context where it follows the rules of `pure` even if it's not typed as `pure`, then that isn't a problem (though such casts are certainly risky in general, because you have to be very sure of what you're doing). And the main reason that banning UB risks taking away useful code is for the same reason that banning it at runtime would taking away useful code. In the general case, you can't always detect it, so in order to ban it, you have to ban more than the things that are strictly a problem. For instance, `@safe` code cannot have UB, forcing some code to be `@trusted` and be verified by the programmer, because the compiler is unable to determine that that `@trusted` code is actually memory safe and does not introduce UB, which is why it has to be `@trusted` and not `@safe`. So, in order to fully prevent UB, the compiler would have to not allow `@trusted` code, which would effectively make it impossible to do anything that the compiler couldn't absolutely prove was memory safe.

CTFE is a somewhat different beast than runtime, and it already bans certain things that you can do just fine at runtime, and depending on the situation, we may want to ban more, but every time that anything is banned during CTFE that you can do at runtime, CTFE is then able to do less. And even if we all agree that casting away `const` and mutating is a problem, that doesn't mean that banning all UB during CTFE won't result in banning stuff that's perfectly fine, because the compiler isn't smart enough to fully verify it. If the compiler were actually fully able to verify UB in all cases without banning stuff that is actually fine, then we wouldn't need `@trusted`. And from what I can see, banning UB during CTFE is basically saying that all code called during CTFE has to be `@safe`, which seems needlessly restrictive.

And personally, I'm perfectly fine with the code doing dumb stuff if you cast away `const` and mutate during CTFE, precisely because whenever you cast away `const`, you're effectively promising to not mutate the value and are willing to deal with the UB if you do. If you don't want to risk those issues, then don't cast away `const`. It's allowed at runtime to allow you to work with stuff that isn't `const` but doesn't actually mutate, and I don't really see why CTFE should be special in that regard. If Walter wants to further restrict CTFE by banning casting away `const` during CTFE like he said in the forum thread linked in that issue, then that's his choice, and we'll have to deal with it, but it _would_ mean that CTFE would further become more restrictive than runtime, which is less than ideal - all to save you from shooting yourself in the foot when you're doing something that you should only be doing when you're very careful and know what you're doing.

rikkimax:
> Why you would want to actually mutate a `const` in CTFE is beyond me.

Destructors ignore const.

So this has (unfortunately) already been established that you need to be able to do so.

Bolpat:
> > Why you would want to actually mutate a `const` in CTFE is beyond me.
> 
> Destructors ignore const.

They handle `const` in their way. Destructors are special because using an object after running a destructor on it is (probably) UB.

PR #4169, opened 2025-02-18
Mention .tupleof as a data-member only alternative to allMembers

MrcSnm:


spec/traits.dd
```diff
@@ -1872,6 +1872,9 @@ void main()
         $(P The order in which the strings appear in the result
         is not defined.)
 
+        $(P `WARNING:` If you're trying to iterate only data members,
+        use the builtin property .tupleof instead.)
```
0xEAB, line 1876: IMHO this only applies to structs. With classes, `.tupleof` does not take fields from the base class into account according to the spec.

PR #4170, opened 2025-02-18
Mention the existence of the .tupleof property

MrcSnm:


spec/property.dd
```diff
@@ -283,6 +283,30 @@ $(D object.TypeInfo_Class)).)
 $(P $(CODE .classinfo) applied to an interface gives the information for the
 interface, not the class it might be an instance of.)
 
+$(H2 $(LNAME2 tupleof, .tupleof Property))
+
+$(P $(CODE .tupleof) provides a tuple representation of a struct. That way,
+it is possible to iterate on each by using $(CODE foreach).)
```
0xEAB, line 289:
```suggestion
$(P $(CODE .tupleof) provides a symbol sequence of all the non-static fields in a struct or class.
The order of the fields in the tuple matches the order in which the fields are declared.

For classes, hidden fields and fields of the bass class are excluded.

A common use case is to iterate over the fields with a $(CODE foreach) loop.)
```

MrcSnm: Thanks for the input!

0xEAB: IMHO the mention of the `foreach` use-case is a bit specific for a language spec. (One could also access the sequence members using indexes.)

0xEAB: See also: <https://dlang.org/spec/class.html#tupleof>

ntrel:
This seems to repeat the information [here](https://dlang.org/spec/class.html#tupleof).

There are actually many properties not documented on the properties page, as they are instead linked to under 'see also' (just before the `.init` Property section). This includes links to class properties and struct properties. (Arguably the page could make this clearer).

thewilsonator (approved):

PR #4175, opened 2025-02-22
Fix #4174 - Invalid C enum in ImportC spec example

### LINKED ISSUE #4174: Invalid C enum in ImportC spec example
This enum:
`enum SQL_MAX_NUMERIC_LEN { 16 };`

Causes DMD to error with the following error:
Error: found `16` when expecting `}`

I have labeled this a regression because I am using the latest nightly and this documentation says that above construction should work: https://dlang.org/spec/importc.html#enums

dkorpel:


thewilsonator (approved):
Do we have `SPEC_RUNNABLE_EXAMPLE` for importC?

dkorpel:
I don't think so

PR #4179, opened 2025-03-05
Update gpg_keys.dd with my public key

dkorpel:
So that I can sign release binaries, while there still is no shared account for git/ssh/gpg keys.


thewilsonator (approved):

thewilsonator:
(You made this PR from a branch not from your remote, so I suspect the CI will fail, feel free to merge this, but please delete the branch after)

PR #4185, opened 2025-03-09
Fix issue 4184: The spec should warn about using copy constructors.

jmdavis:
Until https://github.com/dlang/dmd/issues/20970 is fixed, we need to be warning against the use of copy constructors rather than warning against the use of postblit constructors.

So, I fixed the warning and put both it and the description of what happens when mixing postblit constructors and copy constructors together on both the section for postblit constructors and the one for copy constructors.

I suspect that move constructors merit a similar warning, but I'm not sure what their current state is, so this just fixes the warning for postblit constructors and copy constructors.

### DIFF
--- a/spec/struct.dd
+++ b/spec/struct.dd
@@ -1122,6 +1122,29 @@ $(H3 $(LNAME2 field-init, Field initialization inside a constructor))
 
 $(H2 $(LEGACY_LNAME2 StructCopyConstructor, struct-copy-constructor, Struct Copy Constructors))
 
+    $(PANEL
+        $(P $(RED Warning:) The plan is for
+        $(RELATIVE_LINK2 struct-copy-constructor, copy constructors) to replace
+        $(RELATIVE_LINK2 struct-postblit, postblit constructors), leaving
+        postblit constructors only in legacy code. However, because the
+        compiler hooks in druntime which deal with dynamic arrays and
+        associative arrays have not yet all been updated to support copy
+        constructors properly
+        ($(HTTPS github.com/dlang/dmd/issues/20970, issue #20970)), any type
+        that might be used in a dynamic array or associative array should use a
+        postblit constructor rather than a copy constructor. The copy
+        constructor will not be called in all the cases where it should be for
+        the elements of a dynamic array or the keys or values in an associative
+        array. Postblit constructors do not have this problem.)
+
+        $(P For backward compatibility reasons, a `struct` that explicitly
+        defines both a copy constructor and a postblit will only use the
+        postblit for implicit copying. However, if the postblit is disabled,
+        the copy constructor will be used. If a struct defines a copy
+        constructor (user-defined or generated) and has fields that define
+        postblits, a deprecation will be issued, informing that the postblit
+        will have priority over the copy constructor.))
+
     $(P Copy constructors are used to initialize a `struct` instance from
     another instance of the same type. A `struct` that defines a copy constructor
     is not $(RELATIVE_LINK2 POD, POD).)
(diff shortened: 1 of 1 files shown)


spec/struct.dd
```diff
@@ -1106,6 +1106,26 @@ $(H3 $(LNAME2 field-init, Field initialization inside a constructor))
...
+    $(HTTPS github.com/dlang/dmd/issues/20970, issue #20970), any type that might
+    be used in a dynamic array or associative array should use a postblit
+    constructor rather than a copy constructor. The copy constructor will not
+    be called in all the cases where it should be for the elements of a dynamic
+    array or the keys or values in an associative array. Postblit constructors
+    do not have this problem.)
+
+    For backward compatibility reasons, a `struct` that explicitly defines both
+    a copy constructor and a postblit will only use the postblit for implicit
+    copying. However, if the postblit is disabled, the copy constructor will be
```
0xEAB, line 1124: Would mentioning the attribute explicitly be more comprehensible?
i.e.

> […] However, if the postblit is `@disabled`, […]

jmdavis: I don't know why it would be more comprehensible. It would also arguably be wrong if you want to be really technical, because it's possible for a postblit constructor to be disabled for other reasons. The one that comes to mind at the moment would be if one or more of the member variables has had its postblit constructor disabled, and in such a case, the type wouldn't have necessarily declared any postblit or copy constructors (and it would need to declare one to be copyable).

jmdavis:
Ah, apparently, you need a `#` on the issue number for github to link the commit message appropriately. :|

Well, that's been fixed now.

spec/struct.dd
```diff
@@ -1106,6 +1106,26 @@ $(H3 $(LNAME2 field-init, Field initialization inside a constructor))
 
 $(H2 $(LEGACY_LNAME2 StructCopyConstructor, struct-copy-constructor, Struct Copy Constructors))
 
+    $(P $(RED Warning:) The plan is for
+    $(RELATIVE_LINK2 struct-copy-constructor, copy constructors) to replace
+    $(RELATIVE_LINK2 struct-postblit, postblit constructors), leaving postblit
+    constructors only in legacy code. However, because the compiler hooks in
+    druntime which deal with dynamic arrays and associative arrays have not yet
+    all been updated to support copy constructors properly
+    $(HTTPS github.com/dlang/dmd/issues/20970, issue #20970), any type that might
```
dkorpel, line 1115:
```suggestion
    ($(HTTPS github.com/dlang/dmd/issues/20970, issue #20970)), any type that might
```

I'd put parentheses around this, because it doesn't fit in the sentence the way it is worded now.

dkorpel (approved):

spec/struct.dd
jmdavis: Thanks for catching that. It's what I had intended, but I missed it because of the set of parens that's there for the ddoc macro.

Fixed.

spec/struct.dd
```diff
@@ -1106,6 +1106,26 @@ $(H3 $(LNAME2 field-init, Field initialization inside a constructor))
...
+    be called in all the cases where it should be for the elements of a dynamic
+    array or the keys or values in an associative array. Postblit constructors
+    do not have this problem.)
+
+    For backward compatibility reasons, a `struct` that explicitly defines both
+    a copy constructor and a postblit will only use the postblit for implicit
+    copying. However, if the postblit is disabled, the copy constructor will be
+    used. If a struct defines a copy constructor (user-defined or generated)
+    and has fields that define postblits, a deprecation will be issued,
+    informing that the postblit will have priority over the copy constructor.)
```
ntrel, line 1127: This closing `)` has no opening `$(P` tag. Ditto for the postblit 2nd paragraph.

ntrel: As the warnings are each 2 paragraphs, might be nice to wrap them in a `$(PANEL`.

jmdavis: Nice catch. I clearly screwed it up when I was rearranging the text and then failed to catch it. I really should just take the time to figure out how to build the documentation locally again to see what it actually looks like. It's been ages since I've done that and didn't want to bother, but I'm clearly making too many mistakes here which would be easily caught if I actually built the documentation.

And I'll have a look at what `$(PANEL` does. I don't recall seeing that particular macro before.

jmdavis: Okay. I finally figured out how to build the documentation and see the changes. It looks fine now. I don't know if using PANEL looks better or not, since it's grouping the two paragraphs, but it doesn't affect the count (though the count is kind of odd in a few places already anyway), so it looks a bit funny in that regard. But it probably is better to group them, so the changes now use it.

PR #4193, opened 2025-03-21
Add one line definition for betterC

dkorpel:
Inspired by Tsoding's recent D stream where he couldn't find a quick definition: https://youtu.be/Gj5310KnUTQ&t=286

### DIFF
--- a/spec/betterc.dd
+++ b/spec/betterc.dd
@@ -4,6 +4,9 @@ $(SPEC_S Better C,
 
 $(HEADERNAV_TOC)
 
+$(P BetterC is a subset of D that doesn't depend on the D runtime library, only the C runtime library.
+)
+
 $(H2 $(LNAME2 linking, Linking))
 
     $(P It is straightforward to link C functions and libraries into D programs.


spec/betterc.dd
```diff
@@ -4,6 +4,9 @@ $(SPEC_S Better C,
 
 $(HEADERNAV_TOC)
 
+$(P BetterC is a subset of D that doesn't depend on the D runtime library, only the C runtime library.
+)
```
TurkeyMan, line 8: I feel like it could be a little bit more detailed. Ie;

+++ This allows software to be written that can trivially interact with C, C++, or other native ecosystems without relying on any language-specific ecosystem support.
In particular, it is possible to produce libraries that can be used directly by C/C++ projects without having to become D-aware.

Maybe also include a reference to the automatic header-generator?

0xEAB: If it’s  supposed to provide a “quick definition”, then it should focus on doing that and only that. If it gets drowned in details, people will accidentally skip reading over it once again.

0xEAB: I’d rather visually enhance the whole thing; turn it into a colorful box or something. Maybe put further details below then.

ntrel (approved):

spec/betterc.dd
bangbangsheshotmedown: I's not a subset, it's D without the runtime library, basically D without GC and without Exception Handling, calling it a "subset" sounds bad

```
-betterC disables the GC and EH and everything else that depends on these two features, see the <link_to_the_doc>
It allows better integration in languages with C FFI, and allows targeting esoteric platforms such as WASM
```

dkorpel: I'm just going to merge this because this PR's scope is just a small quick fix that fits with the existing text, I don't have time to go over the rest of the page and rephrase / restructure / add other things right now.

0xEAB: @bangbangsheshotmedown 
> calling it a "subset"

Yet, isn’t it exactly that?

bangbangsheshotmedown: GC/EH are not part of the language, they are runtime library features, they'll not function without an external shared library

The runtime is not the language

- language = base
- runtime = superset

I'd argue -betterC is the base of the D language

PR #4199, opened 2025-03-27
Add example of associative array

schlupa:
Add another example of string lookup using associative array. It allows to introduce AA type that a pure C developer would not expect as a builtin in language.

### DIFF
--- a/articles/ctod.dd
+++ b/articles/ctod.dd
@@ -637,6 +637,21 @@ void dostring(string s)
        to generate a fast lookup scheme for it, eliminating the bugs
        and time required in hand-coding one.
 
+       Another possibility of associating a string to a number is to
+       use an associative array.
+
+----------------------------
+ulong[string] map = ["hello"    : 1616515,
+                     "goodbye"  : 42,
+                     "maybe"    : ulong.max
+                    ];     
+string s = "goodbye";
+
+writeln("number is ", map[s]);
+----------------------------
+
+       Any type can be used as key or as value. 
+
 <hr>$(COMMENT  ============================================ )
 $(H2 $(LNAME2 align, Setting struct member alignment))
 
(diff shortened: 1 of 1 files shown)


articles/ctod.dd
```diff
@@ -637,6 +637,21 @@ void dostring(string s)
        to generate a fast lookup scheme for it, eliminating the bugs
        and time required in hand-coding one.
 
+       Another possibility of associating a string to a number is to
+       use an associative array.
+
+----------------------------
```
thewilsonator, line 643: make this a runnable example?

thewilsonator (approved):
Otherwise looks good

thewilsonator:
```
./articles/ctod.dd:647:                    ];     
./articles/ctod.dd:653:       Any type can be used as key or as value. 
```

articles/ctod.dd
schlupa: Ok, I will look how to do a runnable example.

thewilsonator (approved):

articles/ctod.dd
```diff
@@ -637,6 +637,23 @@ void dostring(string s)
        to generate a fast lookup scheme for it, eliminating the bugs
        and time required in hand-coding one.
 
+       Another possibility of associating a string to a number is to
+       use an associative array.
+
+$(SPEC_RUNNABLE_EXAMPLE_COMPILE
```
thewilsonator, line 643: Complains of undefined macro, maybe something is missing is the build script for it?
```
/tmpfs/dautotest-work/repo/dlang.org/web/articles/ctod.html:749: [UNDEFINED MACRO] <!--UNDEFINED MACRO: "SPEC_RUNNABLE_EXAMPLE_COMPILE"-->
```
Anyway I don't mind if you remove the `SPEC_RUNNABLE_EXAMPLE_COMPILE`.

thewilsonator (approved):

articles/ctod.dd
ntrel: That macro only works in the `spec` subdirectory. Instead:
```suggestion
$(RUNNABLE_EXAMPLE_COMPILE
```

PR #4201, opened 2025-03-28
Add warning for -betterC as somethings that may be expected to work will not

rikkimax:


thewilsonator (approved):

spec/betterc.dd
```diff
@@ -7,6 +7,11 @@ $(HEADERNAV_TOC)
 $(P BetterC is a subset of D that doesn't depend on the D runtime library, only the C runtime library.
 )
 
+$(P $(RED Warning:) There may be hidden assumptions built into a D compiler and its matching runtime.
+As an advanced compiler feature, there may be aspects to it that have not been accounted for i.e. bindings in druntime that have C macros translated as non-templated free-functions.
```
dkorpel, line 11: Well I fixed that specific thing :slightly_smiling_face:  https://github.com/dlang/dmd/pull/21108

I think the phrasing 'there may be hidden assumptions' is too abstract for a warning targeted to new D users. I'd phrase it as "While the D runtime library and standard library are not linked in, it is still possible to import several `core.*` and `std.*` packages and use certain templates, types, and extern definitions from them. However, these libraries were not written with `-betterC` in mind so you are on your own with figuring out which functions you can and can't use. Expect cryptic compiler or linker errors when trying this out."

dkorpel (changes requested):

spec/betterc.dd
rikkimax: I slightly rewrote it.

Oh and good job with that PR.

dkorpel (approved):

spec/betterc.dd
```diff
@@ -7,6 +7,12 @@ $(HEADERNAV_TOC)
 $(P BetterC is a subset of D that doesn't depend on the D runtime library, only the C runtime library.
 )
 
+$(P $(RED Warning:) While the D runtime library and standard library are not linked in,
+it is still possible to import several packages (``core.*`` and ``std.*``) and use some templates, types and function declarations without bodies.
+However, these libraries were not written for BetterC, so you will need to determine which symbols that may be used.
```
dkorpel, line 12:
```suggestion
However, these libraries were not written for BetterC, so you will need to determine which symbols may be used.
```

spec/betterc.dd
dkorpel: Thanks. This warning is also a nice addition.

spec/betterc.dd
```diff
@@ -7,6 +7,12 @@ $(HEADERNAV_TOC)
 $(P BetterC is a subset of D that doesn't depend on the D runtime library, only the C runtime library.
 )
 
+$(P $(RED Warning:) While the D runtime library and standard library are not linked in,
+it is still possible to import several packages (``core.*`` and ``std.*``) and use some templates, types and function declarations without bodies.
+However, these libraries were not written for BetterC, so you will need to determine which symbols may be used.
+Expect cryptic compiler or linker errors whilst figuring this out.
```
mdparker, line 13: Please cut the "whilst figuring this out".

mdparker (changes requested):

spec/betterc.dd
rikkimax: Done.

PR #4202, opened 2025-03-30
Make the runnable example actually work

schlupa:
The runnable example added yesterday unfortunately did not work. Changed the code so that it actually works. Added also another features of an interpolated string.  I noticed during Tsoding's stream and also from personal experience, that discovering features often happen by looking for something else.

### DIFF
--- a/articles/ctod.dd
+++ b/articles/ctod.dd
@@ -646,9 +646,12 @@ ulong[string] map = ["hello"    : 1616515,
                      "goodbye"  : 42,
                      "maybe"    : ulong.max
                     ];
-string s = "goodbye";
-
-writeln("number is ", map[s]);
+void main()
+{
+	writeln("number is ", map["goodbye"]);
+	writeln(i"map contains $(str.length) elements");
+	writeln("map is ", map);
+}
 ----------------------------
 )
 
--- a/articles/ctod.dd
+++ b/articles/ctod.dd
@@ -649,9 +649,9 @@ ulong[string] map = ["hello"    : 1616515,
 unittest
 {
     import std.stdio;
-	writeln("number is ", map["goodbye"]);
-	writeln(i"map contains $(str.length) elements");
-	writeln("map is ", map);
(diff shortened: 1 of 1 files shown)


thewilsonator (approved):

thewilsonator:
```
searching for tabs
./articles/ctod.dd:652:	writeln("number is ", map["goodbye"]);
./articles/ctod.dd:653:	writeln(i"map contains $(str.length) elements");
./articles/ctod.dd:654:	writeln("map is ", map);
make: *** [posix.mak:822: test] Error 1
```

articles/ctod.dd
```diff
@@ -646,9 +646,13 @@ ulong[string] map = ["hello"    : 1616515,
                      "goodbye"  : 42,
                      "maybe"    : ulong.max
                     ];
-string s = "goodbye";
-
-writeln("number is ", map[s]);
+unittest
```
ntrel, line 649: I don't know that unittest blocks are compiled using the RUNNABLE_EXAMPLE_ macros.

articles/ctod.dd
```diff
@@ -646,9 +646,13 @@ ulong[string] map = ["hello"    : 1616515,
...
+{
+    import std.stdio;
+    writeln("number is ", map["goodbye"]);
+    writeln(i"map contains $(str.length) elements");
```
ntrel, line 653: `str` doesn't exist.

articles/ctod.dd
ntrel: It seems the Compile button does compile them, but DAutoTest does not.

ntrel: `tools/dspec_tester.d` doesn't look at `articles/`. I'm looking at fixing that but running into other issues too. Hopefully I can fix it.

articles/ctod.dd
ntrel: Fixed in #4203

PR #4213, opened 2025-04-02
Fix examples in DDOX documentation for std.algorithm.mutation.remove

### LINKED ISSUE #4007: Documentation for std.algorithm.mutation.remove: some code examples look wrong
### CC List

* Richard (Rikki) Andrew Cattermole
* Nick Treleaven (@ntrel)

### Description

````markdown
Created attachment 1924
wrong-looking code example

Hello,

right now on this page:

https://dlang.org/library/std/algorithm/mutation/remove.html

some code examples look wrong, see attachment screenshot.
````

**!!!There are attachements in the bugzilla issue that have not been copied over!!!**

Prthmsh7:
In the issue #4007 where examples in the DDOX-generated documentation for `std.algorithm.mutation.remove` were reported as incorrect.

The examples have several issues:
- Missing `std.stdio` import for examples using `writeln` and not showing that reassignment is necessary to modify the original array which lead to potential confusion in examples

**Solution**
I've implemented a client-side fix that:
- Adds warning notices above problematic examples
- Enhances examples to show proper variable declaration and usage
- Clarifies the need for reassignment when using `remove`

Fixes #4007

### DIFF
--- a/DDOX_FIXES.md
+++ b/DDOX_FIXES.md
@@ -1,0 +1,33 @@
+# DDOX Documentation Fixes
+
+This branch adds fixes for various documentation issues in the DDOX-generated documentation.
+
+## Fix for `std.algorithm.mutation.remove` examples
+
+GitHub Issue: [#4007](https://github.com/dlang/dlang.org/issues/4007)
+
+### Problem
+
+The examples in the DDOX-generated documentation for `std.algorithm.mutation.remove` had several issues:
+
+1. Missing `std.stdio` import for the examples that use `writeln`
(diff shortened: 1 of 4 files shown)


CyberShadow (changes requested):
Hi, as I think you noticed, this repository does not host the documentation for the standard library. Instead, it is recorded alongside the implementation in [the Phobos source code](https://github.com/dlang/phobos).

By the way, did you use an LLM to create this patch? If so, it would be nice if you disclosed this.

CyberShadow:
Closing with same rationale as https://github.com/dlang/dlang.org/pull/4214#issuecomment-2772117877.

PR #4225, opened 2025-04-11
[spec/expression] Document class comparison with `object.__cmp`

### LINKED ISSUE #4224: [spec/expression] wrongly says *CmpExpression* doesn't handle `null` for classes
https://dlang.org/spec/expression.html#class-comparisons
https://dlang.org/spec/operatoroverloading.html#compare

Class comparisons are actually lowered to `object.__cmp`, which checks references for identity and handles null.

PR incoming.

ntrel:
Fixes #4224.

Also briefly describe struct comparison.
Update example - comparing null is not a runtime error.


ntrel:
Note: DAutoTest failed for an unrelated error, so I force pushed. Now both tests are failing with unrelated errors.

thewilsonator (approved):

PR #4230, opened 2025-04-18
[spec/template] Improve lvalue sequence element docs

ntrel:
First state that an element can be modifiable.
Clarify that the elements only need to be compatible (not equal types) in a binary expression. 
Update example.


ntrel:
Ping?

thewilsonator (approved):

PR #4236, opened 2025-05-03
[spec/expression] Improve AssignExpression docs for arrays

ntrel:
Differentiate between a slice expression and assigning to an array. 
Rename 'Array Setting' section to 'Array Filling'.


spec/arrays.dd
```diff
@@ -418,12 +418,12 @@ assert(s == [2, 3, 3, 4]);
...
 
-$(H2 $(LNAME2 array-setting, Array Setting))
+$(H2 $(LNAME2 array-setting, Array Filling))
 
         $(P If a slice operator appears as the left-hand side of an assignment
         expression, and the type of the right-hand side is the same as the
-        element type of the left-hand side, then the array contents of the
-        left-hand side are set to the right-hand side.
+        element type of the left-hand side, then each element of the slice
+        expression will be assigned to the value on the right-hand side.
```
ntrel, line 426: BTW, 'slice expression' is much easier to talk about than a `SliceOperation` (https://github.com/dlang/dlang.org/pull/3680), as the latter is awkward and not an expression. I prefer it before when we had `SliceExpression`, which is more natural and consistent with the other docs.

spec/expression.dd
```diff
@@ -384,9 +384,13 @@ $(H3 $(LNAME2 simple_assignment_expressions, Simple Assignment Expression))
...
 
-    $(LI If the lvalue is a static array or a slice, the behavior is as
+    $(LI If the left operand is a slice expression, the behavior is as
     described in $(DDSUBLINK spec/arrays, array-copying, Array Copying) and
-    $(DDSUBLINK spec/arrays, array-setting, Array Setting).
+    $(DDSUBLINK spec/arrays, array-setting, Array Filling).
+    )
+
+    $(LI If the lvalue is an array, the behavior is as described in
+    $(DDSUBLINK spec/arrays, assignment, Array Assignment).
```
ntrel, line 393: Note that link covers static array copying.

dkorpel (approved):

PR #4239, opened 2025-05-04
[spec/function] Specify null dereference behavior for `@safe` code

### LINKED ISSUE #4240: [spec] `@safe` code generated must not assume null dereference won't occur
Walter from https://forum.dlang.org/post/vv6s8p$lfh$1@digitalmars.com:

> D relies on the null dereference not being "optimized away". Maybe there's a switch for that on ldc, there should be if there isn't one.

ntrel:
Forbid optimizations which assume a null dereference will not occur. `ldc2` does use those optimizations with `-O2` - see:
https://forum.dlang.org/post/vv6o31$ac9$1@digitalmars.com. Fixes #4240.

Cc @tgehr @WalterBright.

Specify that the code generated must detect null dereferences if the system (by default) does not.
Specify that the code generated must detect when any expression causes a null pointer to be indexed outside the protected first page. Include warning that `dmd` does not implement this yet - see https://github.com/dlang/dmd/issues/17776.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -4069,15 +4069,15 @@ $(H3 $(LNAME2 safe-values, Safe Values))
 
         $(P A pointer is a safe value when it is one of:)
         $(OL
-            $(LI `null`)
+            $(LI `null` - $(RELATIVE_LINK2 null-dereferences, see below))
             $(LI it points to a memory object that is live and
             the pointed to value in that memory object is safe.)
         )
         $(P Examples:)
         $(SPEC_RUNNABLE_EXAMPLE_RUN
         ---
-        int* n = null; /* n is safe because dereferencing null is a well-defined
-            crash. */
+        int* n = null; /* n is safe because dereferencing null must either crash
+            or abort. */
         int* x = cast(int*) 0xDEADBEEF; /* x is (most likely) unsafe because it
             is not a valid pointer and cannot be dereferenced. */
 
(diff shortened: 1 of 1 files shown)


ntrel:
Walter from https://forum.dlang.org/post/vv6s8p$lfh$1@digitalmars.com:
> D relies on the null dereference not being "optimized away". Maybe there's a switch for that on ldc, there should be if there isn't one.

That needs to be in the spec, hence this pull.

rikkimax:
This is going to need @kinke and @ibuclaw also, as they support targets that the codegen doesn't currently offer read barriers.

ibuclaw:
It's probably no big deal to "bounds check" pointer and class deferences in safe code only. Pointer slicing and indexing is already disallowed, and array slicing/indexing should rightly assume the underlying structure of the array isn't corrupted.

ntrel:
Timon & Derek Fawcus on the forum found this for LLVM:
> If "null-pointer-is-valid" is set to "true", then null address in address-space 0 is considered to be a valid address for memory loads and stores. Any analysis or optimization should not treat dereferencing a pointer to null as undefined behavior in this function

From under https://releases.llvm.org/10.0.0/docs/LangRef.html#function-attributes.

kinke:
Oh I see, it's only clang that 'optimizes' the artificial tiny testcase, not gcc.

Issue #4245, opened 2025-06-02
Associative Arrays Class Key Type


fieldvoid:
The spec says that for class types to be used as keys in associative arrays, two member functions must be overridden - opEquals and opHash. However the compiler accepts code where the class has not overridden these two member functions, providing some default implementation.
I think that the spec might be outdated - if so it needs to be updated to specify that classes *may* provide custom implementations of opHash and opEquals, and to clarify whether having one member function overriden necessitates also overriding the other.

Please see example below:

```
import std.stdio;

class Foo {
}

class Bar : Foo {
}

void main () {
    bool[Foo] set;
    
    Foo a = new Foo;
    Foo b = new Bar;
    
    set[a] = true;
    set[b] = true;
    
    if (a in set) {
        "%s is in set".writefln(a);
    }
    
    if (b in set) {
        "%s is in set".writefln(b);
    }
    
    set.remove(b);
    
    if (b !in set) {
        "%s is not in set.".writefln(b);
    }
}
```

fieldvoid:
If we are trying to keep track of class instances in a recursive type such that `class Obj { bool[Obj] branches; }`, or with a struct example - `struct Obj { bool[Obj*] branches; }`, wouldn't the `is` behavior of `opEquals` be desirable? I think that checking by address is an expectation (at least it was for me) and seeing the spec claim otherwise was quite a surprise.

Consider a GUI system with widgets that keep track of subwidgets. I would want to keep track of the addresses of the instances. I think that comparing the objects themselves (a sort of deep comparison) is all that common of a use-case.

jmdavis:
> I wouldn't say that overriding `opEquals` but leaving `toHash` is inherently buggy, no more than overloading an operator like `+` but having it do something that is not math related is buggy - it is just unexpected. [Edit]: I see that the spec does say that it is undefined behavior for overriding one member function but not the other (I think that's fine).

It is most definitely a bug if you don't provide a `toHash` which matches the behavior of that type's `opEquals`. This is because the AAs (and really much of anything doing hashing) rely on two equal values having the same hash, and if you don't provide a `toHash` that matches `opEquals`, that's not going to happen.

PR #4246, opened 2025-06-04
Update hash-map.dd

fieldvoid:
This update is in reference to <https://github.com/dlang/dlang.org/issues/4245>.

The spec currently says that for classes to be used as keys in associative arrays, they *must* override toHash and opEquals. This is untrue because the default implementations use the address of the class instance for computing the hash and for comparisons.


fieldvoid:
I've edited it, thanks

PR #4249, opened 2025-06-19
put predefined version identifiers into categories

WalterBright:
Adding a bit more structure helps with understanding them.


ntrel:
The trailing comma after each THEAD invocation is breaking the ddoc output:
https://dtest.dlang.io/artifact/website-07af77ef2c58ccf7eec5577ac1dbe7b6f115834f-a7c8b57019b09fdc2bc5219ab3a59987/web/spec/version.html#predefined-versions

I've fixed that in a PR to this PR:
https://github.com/WalterBright/d-programming-language.org/pull/2

It also changes the formatting to not use `THEAD`, which distinguishes the categories from the column headers.

PR #4255, opened 2025-06-27
[trivial] Remove spaces

Bolpat:
Just removes 4 spaces that don’t look good.


0xEAB (approved):
Also makes things consistent with <https://dlang.org/spec/statement.html#scope-guard-statement>.

PR #4256, opened 2025-07-03
[spec/class] Improve constructor docs

ntrel:
Describe a constructor.
~~Change~~ Extend example to show member variable is in scope.
`return;` can be used.

### DIFF
--- a/spec/class.dd
+++ b/spec/class.dd
@@ -430,8 +430,8 @@ $(H2 $(LNAME2 class-instantiation, Class Instantiation))
 
 $(H3 $(LNAME2 constructors, Constructors))
 
-        $(P A constructor is a special method which is normally
-        $(RELATIVE_LINK2 instantiation-process, invoked by the compiler)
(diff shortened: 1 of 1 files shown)


spec/class.dd
```diff
@@ -430,25 +430,29 @@ $(H2 $(LNAME2 class-instantiation, Class Instantiation))
...
         $(SPEC_RUNNABLE_EXAMPLE_RUN
         ------
         class A
         {
             int i;
 
-            this(int i) // constructor taking an int argument
+            this(int val) // constructor taking an int argument
             {
-                this.i = i; // initialize field `i` from parameter `i`
```
dkorpel, line 451: Is there a still example of `this.i = i` somewhere else? That's a common pattern also worth noting as officially supported.

dkorpel (approved):

spec/class.dd
ntrel: I don't think so, not nearby anyway. Updated so that example is kept and added a constructor overload to demonstrate member variables being in scope.

PR #4273, opened 2025-07-28
 [ddoc_preprocessor] Skip GRAMMAR sections with suffix 

ntrel:
Suffixed GRAMMAR_ macros should not be included in `spec/grammar.html`.
Follow up to https://github.com/dlang/dlang.org/pull/3860.

The only change in `grammar.html` diff should be that the *Named Character Entities* section is skipped, because it uses GRAMMAR_LEX.


ntrel:
You can see the diff was as expected:
https://dtest.dlang.io/results/6dcceef5daa609dc8ef342fe07109c052866fbf2/7dc1dcc3f277e8650dbade0334b37525c83520e6/

thewilsonator (approved):

PR #4275, opened 2025-07-30
spec/pragma: Fix example for pragma(mangle)

ibuclaw:
The .stringof property attempts to mangle the type as the string literal "MyClassA!(int)", which isn't what the example expects.

### DIFF
--- a/spec/pragma.dd
+++ b/spec/pragma.dd
@@ -302,7 +302,7 @@ $(H3 $(LNAME2 mangle, $(D pragma mangle)))
         extern(C++) struct _function {}
         template ScopeClass(C)
         {
-            pragma(mangle, C.stringof, C)
+            pragma(mangle, C)
             struct ScopeClass { align(__traits(classInstanceAlignment, C)) void[__traits(classInstanceSize, C)] buffer; }
         }
         extern(C++)


ibuclaw:
@thewilsonator ping.

spec/pragma.dd
```diff
@@ -302,7 +302,7 @@ $(H3 $(LNAME2 mangle, $(D pragma mangle)))
         extern(C++) struct _function {}
         template ScopeClass(C)
         {
-            pragma(mangle, C.stringof, C)
+            pragma(mangle, C)
```
thewilsonator, line 305: those args should be the other way around, not removed, see `compilable/issue21203.d`
```d
template ScopeClass(C , string name = C.stringof)
//if (is(C == class) && __traits(getLinkage, C) == "C++")
{
    //enum name = C.stringof;
    enum ns = __traits(getCppNamespaces,C);
    extern(C++, class)
    {
        extern(C++,(ns))
        {
            pragma(mangle, C, name)
            struct ScopeClass
            {
                char[__traits(classInstanceSize, C)] buffer;
                //... all the things ...
            }
        }
    }
}
```

PR #4279, opened 2025-08-01
[spec/statement] Define alias and enum on sequence foreach

### LINKED ISSUE #4278: Spec doesn't define `alias` and `enum` on sequence `foreach`
https://dlang.org/spec/statement.html#foreach_over_tuples

Note: They are mentioned for _StaticForeach_.

PR incoming.

ntrel:
Fixes #4278.

### DIFF
--- a/spec/statement.dd
+++ b/spec/statement.dd
@@ -1178,6 +1178,29 @@ hi has type string
 [2, 5] has type int[]
 )
 
+        $(P Each $(GLINK ForeachType) declaration will be a runtime variable,
+        unless `alias` or `enum` are specified:)
+
+        * The `alias` storage class will create a symbolic alias
+        * The `enum` storage class will create a compile-time constant
+
+$(SPEC_RUNNABLE_EXAMPLE_RUN
+---
+void main()
+{
+    import std.meta : AliasSeq;
+
+    int i = 1;
+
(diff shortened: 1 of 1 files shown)


spec/statement.dd
```diff
@@ -1176,6 +1176,29 @@ $(CONSOLE
...
+{
+    import std.meta : AliasSeq;
+
+    int i = 1;
+
+    foreach (alias a; AliasSeq!(i, i))
+    {
+        a++;
+    }
+    assert(i == 3);
```
Bolpat, line 1199: The example is weird. I’d rather not use the same variable twice in the sequence.
```suggestion
    int i = 1, j = 2;

    foreach (alias a; AliasSeq!(i, j))
    {
        a++;
    }
    assert(i == 2 && j == 3);
```

spec/statement.dd
```diff
@@ -1176,6 +1176,29 @@ $(CONSOLE
 7.4 has type double
 hi has type string
 [2, 5] has type int[]
+)
+
+        $(P Each $(GLINK ForeachType) declaration will be a runtime variable,
+        unless `alias` or `enum` are specified:)
```
ntrel, line 1182: I think it's only a runtime variable if that would compile.

thewilsonator (approved):

PR #4290, opened 2025-08-17
module: Add implementation defined note for imports

ibuclaw:
A similar note exists for module declarations:

    Implementation Defined:
    The mapping of package and module identifiers to directory and file names.

This adds an equivalent note about import name resolution.

---

The behaviour itself should be documented in dmd/cli.d as part of the `-I` option.


ibuclaw:
@thewilsonator please check the dautotest generated pages in case there was a mistake in the balancing of `(` / `)`.

0xEAB:
At a first glance, the source code looks off by one.

Also the generated page has this odd `<li>` with no corresponding `<ul>`.
image

PR #4312, opened 2025-09-17
Adds a link to the source code of the current module

denizzzka:
Adds a link to the source, without editing enabled. I am found that it is very convenient to have link here.

(Yes, I know there's usually a link to the source at the bottom of the page)


denizzzka:
Ok (I'm too lazy to deal with this))

PR #4313, opened 2025-09-17
[spec/function] Improve safe variadic docs

### LINKED ISSUE #4311: [Functions] d-style variadic basic type parameter poorly explained
```d
import std;
struct foo(T){
    void bar(T a...){
        a.writeln;
}}
unittest{
    foo!int().bar(1);
    foo!(int[])().bar(1,2,3);//all works
    foo!(int[3])().bar([1,2,3]);
    foo!(int[3])().bar(1,2,3);
}
```
https://dlang.org/spec/function.html#typesafe_variadic_functions paragraph 6.
https://github.com/dlang/dmd/issues/21874

Spec implies this but doesn't explain usage in the slightest; I suggest its called an "optional d-style vartatic"

ntrel:
Reword first paragraph (partly as class variadics are deprecated). 
Explain array variadics, make it clearer an array can be passed. 
Make examples runnable; simplify & use `assert`.
Explain basic type variadic use. Fixes #4311.

### DIFF
--- a/spec/function.dd
+++ b/spec/function.dd
@@ -2407,14 +2407,16 @@ Bar
 $(H4 $(LNAME2 typesafe_variadic_functions, Typesafe Variadic Functions))
 
         $(P A typesafe variadic function has D linkage and a variadic
-        parameter declared as either an array or a class.
-        The array or class is constructed from the arguments, and
-        is passed as an array or class object.
+        parameter, which is typically an array. When passing component
+        arguments (e.g. array elements), the variadic parameter is
+        implicitly constructed from the given arguments.
         )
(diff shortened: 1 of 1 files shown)


spec/function.dd
```diff
@@ -2532,6 +2529,7 @@ $(H4 $(LNAME2 typesafe_variadic_functions, Typesafe Variadic Functions))
 
         void buzz()
         {
+            neil();     // error, missing argument
```
ntrel, line 2532: Note: This was an ICE, fixed in https://github.com/dlang/dmd/pull/21876.

PR #4316, opened 2025-09-20
[spec/function] Improve variadic function docs

ntrel:
Move `core.stdc.stdarg` sentence to where it's actually used. 
Make examples runnable, add `writeln` calls.
Add links.
Fix link to typesafe variadics.
Remove `void*` from `_argptr`. Fixes https://github.com/dlang/dmd/issues/21884. 
`_argptr` is a pointer, not a reference.
Print the class member variables in last d-style example (the old output wasn't the class reference address anyway).
D-style variadics can be `@safe` if they don't call `va_arg` (though not very useful).
A typesafe variadic parameter must be last.


thewilsonator (approved):

thewilsonator:
Already done.

PR #4353, opened 2025-12-22
Redefine `__traits(getPointerBitmap, S)` for types without pointers to return size only

JohanEngelen:
This greatly reduces compile time for large types without pointers, because no need to create a large array literal with zeros.

See DMD and druntime implementation here: https://github.com/dlang/dmd/pull/22289


JohanEngelen:
these changes should be made part of https://github.com/dlang/dmd/pull/22289 (because of the repo merge)

PR #4368, opened 2026-01-12
Change bugzilla bug count of 0 to Github bug count of 127

schveiguy:
See https://forum.dlang.org/post/ycgdukvlllitvjwcgnib@forum.dlang.org


ibuclaw:
@schveiguy this file is autogenerated. So need to fix the cause too.

ibuclaw:
https://github.com/dlang/tools/blob/master/changed.d#L857-L871

PR #4370, opened 2026-01-18
Refactor installation instructions and update paths

duantihua:
1. In a Linux environment, libraries are installed by default to /usr/local/lib64
2. Fix the inconsistency between /usr/local/include in the documentation and /usr/include in the scripts
3. Fixed the copy error in the include library


adamdruppe:
opend instructs users to keep the downloaded package together and just put a forwarder in path. Everything else then works relative to the thisExePath. Advantages include:

* easier uninstall. just nuke the one directory and (optionally) delete the forwarder, not hunting across different paths.
* easier updates, since you just replace the one directory, without having to uninstall the old to avoid leaving files behind that break (anyone remember when std.datetime got broken up into packages and users flooded the forum with problems because the old datetime.d and the new datetime folder lived side by side?)
* easier parallel versions, you can have as many directories as you want without overwriting each other (the forwarder can even specify version in its name, or you use a direct path when you want a specific version)
* configuration is separated per-version 
* does not require `root` access, users can do it entirely contained in their personal accounts by putting it in ~ instead of /opt.

Disadvantages include:

* configurations living with the package means user customizations may have to be redone after updates. (somewhat common to have the package one just include, if exists, some user config extension file so they can be separate). note that updates may also change config meaning, requiring user to fix up their custom things anyway, but this virtually never happens with dmd 
* you might want several forwarder programs / symlinks for dmd, dub, ddemangle, dustmite, etc. opend consolidates all under one thing but you don't so it might take more. Alternatively, you could add /opt/dmd/bin to PATH but then you lose some of the parallel version advantages.
* some pedantic types might be like "man 7 hier" but you can reply to them "/opt should contain add-on packages that contain static files so booyah".

I have never "installed" a D compiler in all the time i used it. The advantages are compelling to me. (Actually fun fact, that's why I used Digital Mars C++ back in the day too: I could install it without admin access on the computer. I burned it to a CD-ROM and used it right off that disc, a truly portable install lol.)

AMDphreak:
> Part of the problem is that which set of lib folders a distro has is distro-specific. AFAIK, there is no standard that they all follow. 

Welcome to Linux stupidity, but more than that the stupidity of hierarchical namespaces. Hierarchies were the wrong choice for file organization.

PR #4392, opened 2026-02-11
[spec/type] Explain derived data types

ntrel:
const3.dd: qualifiers apply to component types of derived data types, and fields of an aggregate type.
expression.dd: be more precise about pattern matching capabilities.

type.dd:
Include `TypeCtor(T)`.
List function pointers, not function types (which are not part of the type grammar). 
Rename 'subtype' to 'component type' and list them.
Link to IsExpression pattern matching.

*Edit:* I've commented out the component types for function pointers and delegates for now - see commit.

### DIFF
--- a/spec/type.dd
+++ b/spec/type.dd
@@ -188,9 +188,10 @@ $(H3 $(LNAME2 component-types, Component Types))
     $(THEAD Type, Component Types)
     $(TROW *TypeCtor*`(T)`, `T`)
     $(TROW `T*`, `T`)
-    $(TROW `R function(Params)`,
+    $(COMMENT TROW `R function(Params)`,
         $(ARGS `R(Params)` - see $(RELATIVE_LINK2 functions, Function Types)))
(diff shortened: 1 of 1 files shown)


spec/type.dd
```diff
@@ -160,11 +170,40 @@ int[] da; // dynamic array/slice
...
+
+    $(P A derived data type has at least one *component type*:)
+
+    $(TABLE
+    $(THEAD Type, Component Types)
+    $(TROW *TypeCtor*`(T)`, `T`)
+    $(TROW `T*`, `T`)
+    $(COMMENT TROW `R function(Params)`,
+        $(ARGS `R(Params)` - see $(RELATIVE_LINK2 functions, Function Types)))
+    $(COMMENT It's unclear if delegates should have component types)
```
dkorpel: They should in my opinion, but there are implementation issues so it's fine to defer this for now

PR #4406, opened 2026-02-20
Add `typeof(null)` and `noreturn` to Basic Data Types

Bolpat:


thewilsonator (approved):

dkorpel:
`typeof(null)` is supposed to be `noreturn*` according to the bottom type DIP, but at this point that's for a future edition

PR #4412, opened 2026-03-02
List `core.sys` modules with proper docs in library sidebar

ntrel:
This uses a shell script to find modules with more than one doc-comment. The first doc-comment is the module one which probably isn't interesting on its own.

Fixes https://github.com/dlang/dmd/issues/22375.

- Hopefully find, xargs, grep, awk and tr are acceptable for the CI.
- I couldn't get the shell commands to work inside the makefile with `$(shell ...)` so I put them in a separate file.
- Maybe DRUNTIME_LATEST_DIR should be used for some doc targets instead of DRUNTIME_DIR.

~~*Update:* I haven't got dmd to define other platform versions from the host system, so for this PR it's just linux and posix.~~

### DIFF
--- a/tools/core_sys_docs.sh
+++ b/tools/core_sys_docs.sh
@@ -1,5 +1,9 @@
 #!/bin/sh
-# list all modules in $1 which have more than one doc-comment
+# Usage: ./core_sys_docs.sh DIR
+#
+# List all modules in DIR which have more than one doc-comment.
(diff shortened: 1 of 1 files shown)


ntrel:
I could undo the previous commit as https://github.com/dlang/dmd/pull/22691 doc tests are now passing.

Issue #4421, opened 2026-03-10
ddoc preprocessor misses every <word> after first from compiler flags HTML


ntrel:
Source:
```
-edition[=<NNNN>[<filename>]]
-mv=<package.module>=<filespec>
-target=<arch>-[<vendor>-]<os>[-<cenv>[-<cppenv>]]
```
The last one above is a change from dlang/dmd#22713.

HTML in browser shows:
```
-edition[=NNNN [ ]]
-mv=package.module =
-target=arch -[- ][-[-
```
https://dlang.org/dmd-linux.html#switch-edition[

`<filename>` etc is interpreted by the browser as non-existent HTML tags.

I have a PR almost ready.

PR #4422, opened 2026-03-11
[ddoc preprocessor] Highlight multiple `<word>` and nested `[...]` in CLI switches

ntrel:
Fixes https://github.com/dlang/dlang.org/issues/4421.

Before only the first <> and first [] was matched in the flag string.
Also inner [ chars were [wrongly italicised](https://dlang.org/dmd-linux.html#switch-HC[).

Before a space was added before the ] presumably because a tall italic character before runs into it. The trouble with that is it can look weird in other places, adding spaces where there shouldn't be any (e.g. `<arch>-<os>`, so I removed that.

*Update:* Also highlight any `<word>` in helpText even if not in flag string.

### DIFF
--- a/ddoc/source/preprocessor.d
+++ b/ddoc/source/preprocessor.d
@@ -479,10 +479,7 @@ private void highlightSpecialWords(ref string flag, ref string helpText)
             const specialWord = caps[1];
             helpText = helpText
                 .splitter(" ")
-                .map!((w) {
-                    auto wPlain = w.filter!(c => !c.among('<', '>', '`', '\'')).to!string;
-                    return wPlain == specialWord ? text("$(I ", wPlain, ")") : w;
-                })
+                .map!(w => w == specialWord ? text("$(I ", w, ")") : w)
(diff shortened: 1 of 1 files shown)


ddoc/source/preprocessor.d
```diff
@@ -447,46 +447,112 @@ auto genSwitches(string fileText)
...
+    }
+}
 
-        // highlight individual words in the description
-        helpText = helpText
-            .splitter(" ")
-            .map!((w){
-                auto wPlain = w.filter!(c => !c.among('<', '>', '`', '\'')).to!string;
-                return wPlain == specialWord ? wPlain.italic: w;
-            })
```
ntrel: BTW removing the filtering is fine because:
- Italicizing `<foo>` is not dependent on using `<foo>` in the flag string.
- `'foo'` is not actually used for the HTML docs - only for the man page (even though short help is copied sometimes)
- \` is not needed as matching `<foo>` is now by regex, not by space splitting

PR #4436, opened 2026-03-20
feat: Implement systematic format logic documentation injection

AMDphreak:
This PR introduces a centralized 'Format Logic' page and updates the Ddoc preprocessor to automatically inject formatting guidance into relevant Phobos function documentation. See the new [format.dd](https://github.com/dlang-supplemental/dlang.org/blob/feature/fmt-logic-docs/format.dd) for the centralized guide.


CyberShadow:
Hi, so... is there a rationale for why we want to do this?

The new page and the `injectFmtLogic` approach are a bit awkward, and I'm not sure it belongs in this repository since it's a bit of a Phobos concern.

PR #4449, opened 2026-05-21
Upgrade jquery 1.7.2 -> 3.7.1

dkorpel:
So I asked Claude to fix the dependabot alerts from using an old version of jquery, and this is the result.

> jQuery 1.7.2 released March 21, 2012 — ~14 years old

Yeah that seems about right

---

## Summary
- Replaces vulnerable jquery 1.7.2 with 3.7.1 (CDN URL + local fallback)
- Patches dependabot alerts: CVE-2015-9251, CVE-2012-6708, CVE-2019-11358, CVE-2020-7656
- Removes `$.browser.msie` branches in `js/run.js` (`$.browser` was removed in jQuery 1.9); the fallback `o.text(...)` path is kept

## Test plan
- [ ] Verify run-button output rendering on a code example page
- [ ] Verify subnav expand/collapse (uses `js/dlang.js`)
- [ ] Verify pages still load with CDN blocked (local fallback path)


0xEAB (approved):
I’d advocate for getting rid of it all together, but otherwise lgtm

PR #4461, opened 2026-08-07
improve exception safety article title

WalterBright:
otherwise no one will find it

### DIFF
--- a/articles/index.dd
+++ b/articles/index.dd
@@ -69,7 +69,7 @@ $(D_S Articles,
         )
         $(DIVC row,
             $(DIVC item,
-                $(H4 $(LINK2 $(ROOT_DIR)articles/exception-safe.html, Exception Safety))
+                $(H4 $(LINK2 $(ROOT_DIR)articles/exception-safe.html, Exception Safety - Transactions - Scope Guard ))
                 $(P This article is about exceptions and how RAII and scope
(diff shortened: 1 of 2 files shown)


articles/index.dd
```diff
@@ -69,7 +69,7 @@ $(D_S Articles,
         )
         $(DIVC row,
             $(DIVC item,
-                $(H4 $(LINK2 $(ROOT_DIR)articles/exception-safe.html, Exception Safety))
+                $(H4 $(LINK2 $(ROOT_DIR)articles/exception-safe.html, Exception Safety - Transactions - Scope Guard ))
```
0xEAB, line 72:
```suggestion
                $(H4 $(LINK2 $(ROOT_DIR)articles/exception-safe.html, Exception Safety - Transactions - Scope Guard Programming))
```

0xEAB, line 72:
```suggestion
                $(H4 $(LINK2 $(ROOT_DIR)articles/exception-safe.html, Exception Safety - Transactions - Scope Guards))
```

WalterBright: Isn't programming implicit in a web site on programming?

0xEAB: sure ;)
But it’s for consistency with the other instance where the title is found.

PR #4471, opened 2026-09-16
Update download links for d-apt sources and keyrings

jordisayol:


dkorpel:
Both links seem to work for me, what's the difference?

jordisayol:
Yes, they both works, but the last ones are the Sourceforge Standard for download files, and is not server mirror dependent.
I've done this pr to avoid possible future problems. SF are doing many changes on download rules.

dkorpel (approved):

dkorpel:
Got it, thanks!

